PR #24, opened 2015-04-24
Fix interpolation

julienguy:
Improved (clarified) interpolation 
 - integration in output bins of piece-wise linear function of input
 - consider the flux density=0 outside of the edges of the input bins
 - passes test of conservation of flux density, total integrated flux, inverse variance scaling with bin size,
   total inverse variance

### DIFF
--- a/py/desispec/interpolation.py
+++ b/py/desispec/interpolation.py
@@ -5,7 +5,7 @@ Utility functions for interpolation of spectra over different wavelength grid
 import numpy as np
 import sys
 from desispec.log import get_logger
-
+import pylab
 #import time # for debugging
 
 def bin_bounds(x) :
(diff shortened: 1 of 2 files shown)


sbailey:
This looks good, though there is a leftover "import pylab" probably from debugging.  Let's remove that and figure out how to update the pull request then I'll merge it.

More general comments about fits I/O:
  * When we do need fits I/O, astropy.io.fits is preferred (that is what pyfits became)
  * fitsio is OK if performance is absolutely crucial
  * file reading/writing should be isolated to the io package, so algorithm code should never need to import fits/pyfits/fitsio anyway

julienguy:
done (see the commit :-)

PR #25, opened 2015-04-24
Improve logger

julienguy:
add DESI_LOGLEVEL environment variable, add doc, define WARNING... in module
add unit test/example of log

### DIFF
--- a/py/desispec/log.py
+++ b/py/desispec/log.py
@@ -5,18 +5,57 @@ We can have something specific for DESI in the future but for now we use the sta
 
 import sys
 import logging
-
+import os
 
 desi_logger = None
 
-def get_logger(level=logging.DEBUG) :
+# just for convenience to avoid importing logging
+# we duplicate the logging levels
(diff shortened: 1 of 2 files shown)


sbailey:
Several comments :)

The intention was that $DESI_LOGLEVEL would override the level passed into get_logger().  i.e. if tag x.y.z had specified a certain loglevel for some piece of code, you could temporarily override that by using $DESI_LOGLEVEL without having to edit the code from that tag.  Order of priority
  * $DESI_LOGLEVEL highest
  * then whatever was passed in for level
  * then default to logging.INFO if level wasn't specified

the dico of valid options should include CRITICAL and FATAL (which are even worse than ERROR).  

Consider using desi_level.upper() just in case someone tries "info" instead of "INFO".

julienguy:
I think we should force to have at least the ERROR reported in the logs.
that's why I didn't include it in the list (this 'dico' is the list of
levels that are dumped to the output).
I am doing the change of priority with $DESI_LOGLEVEL

j

Issue #19, opened 2015-04-22
resolution module shouldn't hardcode the number of diagonals


sbailey:
The desispec.resolution module shouldn't hardcode the number of diagonals stored for the resolution matrix when it is constructed from the sparse diagonals data.  The fits file format keeps the same number of upper and lower diagonals and the number of total diagonals should be odd.

When constructing from a dense matrix, it does need to pick something for the number of diagonals to keep.  Consider making this an optional argument of the constructor.

sbailey:
fixed with with https://github.com/desihub/desispec/commit/556ded6f899a481520058c9adbc0c647d306972f

PR #26, opened 2015-04-26
Allow various number of diagonals when creating Resolution matrix

sbailey:
This fixes issue #19 to allow the Resolution object to be created with an odd but otherwise arbitrary number of diagonals.  This is needed for an upcoming specter bug fix that will auto-tune the number of diagonals based upon the PSF size and wavelength step of extractions.

David K: can you review this to make sure I've covered the various cases for Resolution created with sparse input?  I added a few more tests and I think this is correct, but your original code had already covered some cases that I hadn't originally thought about so another pair of eyes on this edit would be appreciated.

(sorry for all the git pushes that probably generated emails leading up to this...)

### DIFF
--- a/py/desispec/resolution.py
+++ b/py/desispec/resolution.py
@@ -110,6 +110,16 @@ def run_unit_tests(n = 100):
     for ndiag in [3,5,11]:
         R6 = Resolution(np.ones((ndiag, n)))
         assert len(R6.offsets) == ndiag, 'Constructor broken for ndiag={}'.format(ndiag)
+        
+    #- An even number if diagonals is not allowed
+    try:
+        ndiag = 10
+        R7 = Resolution(np.ones((ndiag, n)))
+        raise ValueError('an even number of diagonals is not supposed to be allowed')
+    except ValueError, err:
+        #- it correctly raised an error, so pass
+        pass
+        
 
 if __name__ == '__main__':
     run_unit_tests()
--- a/py/desispec/resolution.py
+++ b/py/desispec/resolution.py
@@ -40,7 +40,10 @@ class Resolution(scipy.sparse.dia_matrix):
             # We just need to put the diagonals in the correct order.
             diag_order = np.argsort(data.offsets)[::-1]
             ndiag = len(data.offsets)
+            assert ndiag%2 == 1, "Number of diagonals ({}) should be odd".format(ndiag)
             self.offsets = np.arange(ndiag//2,-(ndiag//2)-1,-1)
+            if not np.array_equal(data.offsets[diag_order], self.offsets):
+                raise ValueError('Offsets of input matrix are non-contiguous or non-symmetric')
             scipy.sparse.dia_matrix.__init__(self,(data.data[diag_order],self.offsets),data.shape)
 
         elif isinstance(data,np.ndarray) and len(data.shape) == 2:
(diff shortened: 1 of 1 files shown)


tskisner:
So are you planning on just changing the correlation length along the wavelength direction, or are you trying to support nearest neighbor fiber terms as well?

sbailey:
This is only the minimal change necessary to support reading a resolution matrix that was written with something other than 21 diagonal elements.  It does not expand the format to include fiber-to-fiber terms; it just removes the hardcoded 21.

Supporting nearest neighbor fiber-to-fiber terms could be included in the future but that will be a separate ticket that will need to be motivated by a specific study showing that it is necessary.

tskisner:
ok, sounds good.

py/desispec/resolution.py
```diff
@@ -33,18 +34,27 @@ class Resolution(scipy.sparse.dia_matrix):
...
     def __init__(self,data):
 
-        if scipy.sparse.isspmatrix_dia(data) and np.array_equal(np.sort(data.offsets)[::-1],self.offsets):
+        ### if scipy.sparse.isspmatrix_dia(data) and np.array_equal(np.sort(data.offsets)[::-1],self.offsets):
+        if scipy.sparse.isspmatrix_dia(data):
             # Input is already in DIA format with the required diagonals.
             # We just need to put the diagonals in the correct order.
             diag_order = np.argsort(data.offsets)[::-1]
+            ndiag = len(data.offsets)
+            assert ndiag%2 == 1, "Number of diagonals ({}) should be odd".format(ndiag)
```
dkirkby: Why is this an assertion rather than a ValueError?

py/desispec/resolution.py
```diff
@@ -33,18 +34,27 @@ class Resolution(scipy.sparse.dia_matrix):
...
+            self.offsets = np.arange(ndiag//2,-(ndiag//2)-1,-1)
+            if not np.array_equal(data.offsets[diag_order], self.offsets):
+                raise ValueError('Offsets of input matrix are non-contiguous or non-symmetric')
             scipy.sparse.dia_matrix.__init__(self,(data.data[diag_order],self.offsets),data.shape)
 
         elif isinstance(data,np.ndarray) and len(data.shape) == 2:
             n1,n2 = data.shape
-            if n1 == num_diagonals and n2 > n1:
+            if n2 > n1:
+                ndiag = data.shape[0]
```
dkirkby: `ndiag = n1`

py/desispec/resolution.py
```diff
@@ -33,18 +34,27 @@ class Resolution(scipy.sparse.dia_matrix):
...
+                self.offsets = np.arange(ndiag//2,-(ndiag//2)-1,-1)
```
dkirkby: Check that `ndiag` is odd here?

py/desispec/resolution.py
```diff
@@ -60,7 +70,7 @@ def to_fits_array(self):
         Convert to an array of sparse diagonal values.
 
         This is the format used to store resolution matrices in FITS files. Note that some
-        values in the returned rectangular array do correspond to actual matrix elements
+        values in the returned rectangular array do not correspond to actual matrix elements
```
dkirkby: Good catch!

dkirkby:
Sorry I missed this earlier.  Other users generally don't get an automatic notification for all pull requests, and this one did not connect to me in a way that github could detect.  If you replace "David K" with "@dkirkby" I am more likely to see it (see https://github.com/blog/821).

py/desispec/resolution.py
sbailey: Well, well.  A bug in my unit test. :)

    #- An even number if diagonals is not allowed
    try:
        ndiag = 10
        R7 = Resolution(np.ones((ndiag, n)))
        raise ValueError('Incorrectly created Resolution with even number of diagonals')
    except ValueError, err:
        #- it correctly raised an error, so pass
        pass

The code was accidentally correctly failing to create a Resolution matrix with an even number of diagonals because the number of offsets ended up as odd anyway and dia_matrix.__init__ complained that the number of diagonals didn't match the number of offsets.  But even if it had (incorrectly) succeeded my test made matters worse by raising a ValueError -- the same type of error it was looking to catch so it would have passed anyway.

So thanks for the code review catching that I wasn't checking for ndiag odd here.  I'll update and submit a new pull request since I already merged this one (oops).

PR #27, opened 2015-05-01
Allow various rsize

sbailey:
@dkirkby please take a look at this.  Several fixes:

  1. Explicitly check for even number of diagonals (bad) when initializing from a dense rectangular ndarray (thanks for comment on previous pull request)
  2. Fixed unit test which wouldn't have caught that anyway
  3. Fix check for odd ndiag when initializing from sparse array
  4. raise ValueError instead of assertions or RuntimeErrors
  5. Reformat comments to fit within 80 columns (admittedly unrelated to the bugs, but I'm old-school on 80 columns...)


dkirkby:
This looks good to me. Perhaps we should write down some 'best practices' for which exceptions to raise in desihub library code and when.  I like the use of ValueError here, rather than the more generic RuntimeError or some custom DesiSpecError.  This could be accompanied by a main() template for reporting/logging exceptions.

Issue #22, opened 2015-04-22
Write a fast raw to redshifts functional test


sbailey:
Write a fast raw to redshifts functional test, e.g. like the [dogwood](https://desi.lbl.gov/trac/wiki/Pipeline/DataChallenges/2015-01) production on 3-5 fibers.

Wrap that in a cronjob script that updates a "master" branch installation of all code, fetches input test data, runs the script, and report if anything is broken.  This should run every night.

Note: this could be integrated into a [Travis](https://travis-ci.org) test, but a basic version running as a cronjob at NERSC will be simpler to start with.  Travis might be more appropriate for our lower level unit tests.

sbailey:
See pull request #28 for a first version of this.

sbailey:
Closing issue; basic functional test exists in desispec.test.integration_test / etc/cron_dailytest.sh.  This can be expanded for additional code in the future.

PR #28, opened 2015-05-05
Dailytest

sbailey:
This pull request brings in a raw data through redshifts integration test of 5 fibers for issue #22 .  It simulates 5 fibers of raw data (2 SKY, 1 STD, 1 LRG, 1 ELG) plus arcs and flats and then runs extractions, fiber flats, sky subtraction, flux calibration, bricks, and redshift fitting.

  * etc/cron_dailytest.sh is the cronjob script to run on edison at NERSC (currently running under
    Stephen's account at 4am)
  * py/desispec/test/integration_test.py is the primary test orchestration code
  * py/desispec/pipeline/core.py has a new runcmd() function for running a pipeline command, with
    optional ability to check inputs and outputs.
  * util.py night2ymd() and ymd2night() commands came along for the ride to convert between
    YEARMMDD string and (year, month, day) tuple of integers though in the end these aren't used (yet).

This integration test doesn't yet cover:
  * specex PSF estimation (it uses the desimodel true PSFs).  The intention is to make this an optional test in a separate update — if specex is installed, use it, otherwise fall back to the desimodel PSFs and keep going)
  * coadds (too slow for quick test)

Example output is at /project/projectdirs/desi/spectro/redux/dailytest/dailytest.log (though I'm considering switching the test to use scratch instead).

### DIFF
--- a/etc/cron_dailytest.sh
+++ b/etc/cron_dailytest.sh
@@ -1,11 +1,8 @@
 #!/bin/bash
 #- Cronjob to run daily integration tests on edison.nersc.gov
 
-#- Recreate a normal login environment
-# export NERSC_HOST=edison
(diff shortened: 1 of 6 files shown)


moustakas:
This looks good.  The cron job looks for the templates in the "data challenge 2" directory which is probably good enough but we may also want to build a set of fixed templates specifically for this test.

PR #35, opened 2015-05-06
Redmonster doc

### LINKED ISSUE #32: Importing redmonster during documentation builds
Refactor the code in desispec.zfind.redmonster so that documentation can be built successfully even if redmonster is not in `$PYTHONPATH`.

### LINKED ISSUE #33: Backward-compatible Sphinx configuration
In PR #31, the conf.py file has been changed to assume that the Sphinx Napoleon extension is installed.  This is a safe assumption for Sphinx >= 1.3.  Can we assume that everyone in DESI who wants to build documentation is already using 1.3?  For example, MacPorts has Sphinx 1.3.1.

The alternative is to include a copy of the napoleon code in desispec itself.

weaverba137:
Fixes #32. Fixes #33.

Moving the import of redmonster code inside the `__init__()` function allows the redmonster.py module to compile even if redmonster is not installed.

Also adds a test for the presence of the Sphinx extension 'napoleon' in the doc/conf.py file.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -77,9 +77,8 @@ The desispec API
 .. automodule:: desispec.zfind
     :members:
 
-.. This can't be imported right now unless redmonster is in PYTHONPATH
-.. .. automodule:: desispec.zfind.redmonster
-..     :members:
+.. automodule:: desispec.zfind.redmonster
+    :members:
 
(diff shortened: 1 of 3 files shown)


weaverba137:
I'm working off of my own fork of desispec.

PR #36, opened 2015-05-07
Add Travis CI functionality.

weaverba137:
This PR adds a .travis.yml file to support basic tests.  Right now `python setup.py test` is *not* run, because I wasn't sure how ready we were for that.  Only basic tests of package integrity & documentation are performed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ matrix:
         # Check for sphinx doc build warnings - we do this first because it
         # runs for a long time
         - python: 2.7
-          env: SETUP_CMD='build_sphinx' # -w'  # OPTIONAL_DEPS=true
+          env: SETUP_CMD='build_sphinx OPTIONAL_DEPS=true
           # OPTIONAL_DEPS needed because the plot_directive in sphinx needs them
           # -w is an astropy extension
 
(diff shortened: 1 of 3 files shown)


sbailey:
Thanks!  Right now the basic tests can be run with 

    python -c "from desispec.test import runtests; runtests()"

Travis could be configured to call that (right?) or setup.py could be updated to do that, or we could alter that command to be whatever those expect it to be.  Travis would need numpy, astropy, scipy, and perhaps yaml.  I don't think those tests require desimodel, but future higher level tests might.

I'll merge this now, so these comments are for the next steps.

weaverba137:
The convention is to have the tests run through 'python setup.py test', & I know how to do that.

The .travis.yml file already loads numpy, scipy, & astropy so that's no issue.

Loading redmonster or desimodel might be trickier.  It should be possible to create tests that skip themselves if those are unavailable.

sbailey:
Under the hood, what is 'python setup.py test' expecting to run?  Can be that completely arbitrary?  i.e. can you make it call "from desispec.test import runtests; runtests()" ?

Currently redmonster and desimodel only enter the tests as part of the higher level daily integration test; they are not required for the (current) desispec unit tests.

weaverba137:
The traditional way to do this is to pass the 'test_suite' keyword to the setup() function in setup.py.  The value of 'test_suite' is a function that returns a unittest.TestSuite instance.  Making it some arbitrary script would require a bit more hacking.

PR #37, opened 2015-05-07
Test sky

julienguy:
* Use scipy.linalg.cho_factor and scipy.linalg.cho_solve instead of lapack calls which was depending on scipy version, add docstring, add cholesky_invert.
* add test/test_linalg.py
* fix bugs in test/test_sky.py (invar didn't have same dimension as flux) , add test that fitted sky is ok
* Update wrong docstring in sky.compute_sky

### DIFF
--- a/py/desispec/linalg.py
+++ b/py/desispec/linalg.py
@@ -9,51 +9,57 @@ import scipy,scipy.linalg,scipy.interpolate
 from desispec.log import get_logger
 
 def cholesky_solve(A,B,overwrite=False,lower=False):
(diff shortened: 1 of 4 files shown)


sbailey:
@weaverba137 can you take at look at the Travis failures?  It looks like a Travis config problem, not a problem with this pull request itself.

weaverba137:
the test_sky branch doesn't contain the .travis.yml file.  It was created a while ago.

weaverba137:
Any reason we haven't deleted test_sky yet?

PR #41, opened 2015-05-08
Test refactor

weaverba137:
Based on my tests, all of the following methods now work:

```
python -c "from desispec.test.desispec_test_suite import runtests; runtests()"
```

```
python -m "desispec.test.desispec_test_suite"
```

```
python setup.py test
```

So there are three different ways to run tests, all of which do the same thing.


weaverba137:
PS, my version of the test-refactor branch (weaverba137:test-refactor) is no longer in sync with the test-refactor branch in desihub/desispec.  Once this PR is done, we'll want to delete both.

PR #45, opened 2015-05-15
add ability to create Resolution from 1D array of Gaussian sigmas

sbailey:
This PR adds the ability to generate a sparse Resolution matrix from a 1D array of Gaussian sigmas (in addition to the other input options of dense arrays, dense data from sparse arrays, or sparse matrices).  I developed this for quickly generating Resolution matrices for test cases, but it could also be useful for quickgen.

@dkirkby could you take a look?

### DIFF
--- a/py/desispec/resolution.py
+++ b/py/desispec/resolution.py
@@ -11,10 +11,38 @@ from __future__ import division, absolute_import
 
 import numpy as np
 import scipy.sparse
+from scipy.special import erf
 
 # The total number of diagonals that we keep in the sparse formats when
 # converting from a dense matrix
 default_ndiag = 21
+
+def _gauss_pix(x, mean=0.0, sigma=1.0):
+    """
+    Utility function to integrate Gaussian density within pixels
+    
+    Args:
+        x (1D array): pixel centers
+        mean (float): mean of Gaussian
+        sigma (float): sigma of Gaussian
+        
+    Returns:
+        array of integals of the Gaussian density in the pixels.
+        
+    Note:
+        All pixels must be the same size
+    """
+    x = (np.asarray(x, dtype=float) - mean) / sigma
+    dx = x[1]-x[0]
+    if not np.allclose(np.diff(x), dx):
+        raise ValueError('all pixels must have the same size')
+        
+    edges = np.concatenate([x-dx/2, x[-1:]+dx/2])
+    assert len(edges) == len(x)+1
+    
+    y = erf(edges)
+    return (y[1:] - y[0:-1])/2
+    
 
 class Resolution(scipy.sparse.dia_matrix):
     """Canonical representation of a resolution matrix.
@@ -56,7 +84,7 @@ class Resolution(scipy.sparse.dia_matrix):
                 raise ValueError('Offsets of input matrix are non-contiguous or non-symmetric')
             scipy.sparse.dia_matrix.__init__(self,(data.data[diag_order],self.offsets),data.shape)
 
-        elif isinstance(data,np.ndarray) and len(data.shape) == 2:
+        elif isinstance(data,np.ndarray) and data.ndim == 2:
             n1,n2 = data.shape
             if n2 > n1:
                 ndiag = n1  #- rename for clarity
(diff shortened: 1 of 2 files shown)


py/desispec/resolution.py
```diff
@@ -94,4 +105,32 @@ def to_fits_array(self):
         """
         return self.data
 
-#- (Unit tests moved to desispec.test.test_resolution)
\ No newline at end of file
+def _gauss_pix(x, mean=0.0, sigma=1.0):
```
dkirkby: Since this is an implementation support routine and there is no obvious rationale for mean != 0 in a resolution context, I would probably drop the mean arg here.  If you want to keep it, then you should add a unit test for mean != 0.

py/desispec/resolution.py
```diff
@@ -94,4 +105,32 @@ def to_fits_array(self):
...
+    """
+    Utility function to integrate Gaussian density within pixels
+
+    Args:
+        x (1D array): pixel centers
+        mean (float): mean of Gaussian
+        sigma (float): sigma of Gaussian
```
dkirkby: Passing sigma as an ndarray would also work and allow you to eliminate the loop when you call this function above.

py/desispec/resolution.py
```diff
@@ -94,4 +105,32 @@ def to_fits_array(self):
...
         return self.data
 
-#- (Unit tests moved to desispec.test.test_resolution)
\ No newline at end of file
+def _gauss_pix(x, mean=0.0, sigma=1.0):
+    """
+    Utility function to integrate Gaussian density within pixels
+
+    Args:
+        x (1D array): pixel centers
```
dkirkby: You mention here that the centers must be on a grid but, even better, would be to remove this restriction so, for example, this code would work with BOSS spectra where the new constructor would actually be quite useful for interfacing with the DESI pipeline.  Your error function calculation is already general enough, you just need to pass in the bin egdes here instead of bin centers.

py/desispec/resolution.py
```diff
@@ -94,4 +105,32 @@ def to_fits_array(self):
...
+    x = (np.asarray(x, dtype=float) - mean) / sigma
+    dx = x[1]-x[0]
+    if not np.allclose(np.diff(x), dx):
+        raise ValueError('all pixels must have the same size')
+
+    edges = np.concatenate([x-dx/2, x[-1:]+dx/2])
+    assert len(edges) == len(x)+1
+
+    y = erf(edges)
+    return (y[1:] - y[0:-1])/2
```
dkirkby: return np.diff(y)/2 ?

py/desispec/resolution.py
```diff
@@ -94,4 +105,32 @@ def to_fits_array(self):
...
+    """
+    x = (np.asarray(x, dtype=float) - mean) / sigma
+    dx = x[1]-x[0]
+    if not np.allclose(np.diff(x), dx):
+        raise ValueError('all pixels must have the same size')
+
+    edges = np.concatenate([x-dx/2, x[-1:]+dx/2])
+    assert len(edges) == len(x)+1
+
+    y = erf(edges)
```
dkirkby: I had to go hunting to find which erf you are using.  Since you are just using it once, a fully qualified name might be easier to read here.

py/desispec/test/test_resolution.py
```diff
@@ -54,3 +54,11 @@ def test_resolution(self, n = 100):
         except ValueError:
             #- correctly raised an error, so pass
             pass
+            
+        #- Test creation with sigmas - it should conserve flux
+        R9 = Resolution(np.linspace(1.0, 2.0, n))
+        self.assertTrue(np.allclose(np.sum(R9.data, axis=0), 1.0))
+
```
dkirkby: Add a test for non-equally spaced x and mean != 0 ?

py/desispec/resolution.py
sbailey: Good idea, though this would require adding an optional argument to the Resolution constructor to specify what those bin centers/edges are.  If we do this, I might favor switching the sigma units to be Angstroms (currently pixels) and specify the wavelength bins in Angstroms.  I'll think about how to express this cleanly without complicating the interface.  "if `data` is 1D it is interpreted as sigmas and then the `wavelength` option is required, otherwise it is ignored" sounds a bit messy.  But it would be good to support this somehow.

py/desispec/resolution.py
sbailey: This would take more than just passing in an array for sigma, since the bin rescaling line `x = np.asarray(x, dtype=float) / sigma` would then need to become a matrix operation with other matrix operations downstream.  This is only worth it if the loop is a slow limiting factor.

weaverba137:
This has been on hold for four months.  Is anything going to happen with this PR?

sbailey:
The intention is yes, but it is pretty low priority.  The only holdup is minor bookkeeping hassles related to the good suggestions make it more general.  If I don't get to those by the end of September, I may just merge the existing functionality and leave the rest for future updates if we need them.

sbailey:
The extensions suggested in this PR review are all good ideas, but since they aren't features we currently need they have been pretty low priority to implement and this has been dangling for awhile.  I'm going to merge in what we have for now and close this out.  Those extensions (e.g. non-uniform bins, non-0 mean) can be implemented later if/when we need them.

Issue #16, opened 2015-04-22
streamline io.read_* outputs


sbailey:
Most of the desispec.io.read_* functions return an inconveniently long tuple of outputs.  Using it interactively almost always requires looking up the arguments to figure out which arguments are which.  e.g.

    flux, ivar, wave, resolution_data, header = io.read_frame(filename)

Update these to return lightweight classes instead that contain the equivalent member variables, e.g.

    spectra = io.read_frame(filename)
    plot(spectra.wave, spectra.flux[0])

My style preference is for these to be super-lightweight classes that are basically syntactic sugar compared to just returning dictionary with keys of "wave", "flux", "ivar", etc.  They could provide some minimal convenience functions, but the intention is that these are *not* containers for core algorithms. They should expose the raw numpy arrays just like the current io.read_* routines return raw numpy arrays.

This change will break a lot of code, so it might be best to wait for basic functional unit tests to be in place first.  After making the change, if it breaks code that wasn't caught by a test, that should be added as a test too.

weaverba137:
Looks like you want an instance returned by [namedtuple()](https://docs.python.org/2/library/collections.html#collections.namedtuple).

sbailey:
I'll submit a pull request for this soon.  In the end I didn't use namedtuple() because I wanted to have a small amount of functionality in these classes, e.g. checking shape consistency between flux and ivar.  namedtuple() does match the original issue request (and is nifty, thanks for pointing it out), but the scope slightly shifted during implementation.

sbailey:
Addressed by PR #47 .  Closing issue.

PR #47, opened 2015-07-28
Refactor io objects

sbailey:
This PR addresses issue #16 to refactor I/O to return lightweight objects, e.g.

    frame = io.read_frame(filename)

instead of

    wave, flux, ivar, mask, resolution, header = io.read_frame(filename)

This has convenience and greater future flexibility since additional parameters can be added to the frame object without having to change any of the code using it.  It has the downside of somewhat obfuscating what is returned by the I/O functions — you have to go look it up in an object rather than directly seeing the 6 elements.

This touched a lot of files, since those lightweight objects also get passed around for the various algorithm code.  Multiple tests were added as part of this PR.  It passes both the unit tests and the integration test.


sbailey:
Update: the tests pass on my laptop but the Travis tests failed.  I'll look into that and fix it and update the pull request.  The basic changes can still be reviewed in the meantime.

weaverba137:
There are a few problems here.

1. The Python package needs to internally set up and know the paths to these scripts at compile/test time.
1. Once *executed*, the scripts need to be able to import the desispec package.
1. One can't assume that the same version of Python will execute the scripts, *i.e.* if any non-standard Python needs to be used to execute these scripts, that executable needs to be written into the test.

These are solvable problems, but it would be much easier if the tests simply tested the *components* of the scripts, or imported the `main()` function and executed that.

weaverba137:
I dealt with the internal package paths and environment variables, but in a very fragile way.  I advocate moving away from running scripts and just testing the internal components.

sbailey:
Previously all unit tests were just for components, but I found problems where scripts were failing because component interfaces and their tests had changed but the scripts hadn't been updated, and thus the problems weren't found until later integration tests.  The test_binscripts unit test was an attempt to verify that the scripts themselves worked.  This test works when you already have the git clone configured via PATH and PYTHONPATH, but it was failing the Travis tests because of difficulty bootstrapping itself.

Our previous script guidance was that it was ok for a script to:
  * parse command line arguments
  * perform I/O using desispec.io.* functions
  * call algorithms using the data received from the I/O

but scripts should not include algorithms themselves.

Wrappering all that into a single main() slightly obfuscates the script structure, but could be ok if that would cleanly solve the Travis configuration bootstrapping.

weaverba137:
Those guidelines are fair enough, but when I actually look inside `desi_compute_fiberflat.py` and `desi_compute_sky.py`, I see those are *already* wrappered in `main()` functions.  So why not just put those main functions in the package hierarchy, where they can be easily found by the package tests, and as a bonus, those functions will be found by automated documentation processing?

sbailey:
Forgot to add in the merge comments: this addresses issue #16 .  Apologies for the extra email, but I do want to maintain PR vs. issue linkages in the comments.

PR #50, opened 2015-08-05
Support for file download

### LINKED ISSUE #48: Create `io.download()` function.
We need a `io.download()` function to handle downloads of DESI data from the main repository.  It should:

* Be aware of the structure of the DESI repository.
* Be aware of files that have already been downloaded (*i.e.* have a cache).
* Accept a single file or list of files.

Example function signature:

```python
def download(filenames,baseurl='http://portal.nersc.gov/project/desi'):
    """Download files from DESI repository.
    """
    if isinstance(filename,str):
        filenames = [filenames]
    # ...
```

### LINKED ISSUE #49: Create `io.findfile()` function
We need a `io.findfile()` function that will:

* Be aware of the DESI file hierarchy.
* Work at NERSC & on someone's laptop.
* Download files as needed, using `io.download` (see #48).
* Optionally, accept a file *type*, with parameters, rather than a full file name.

weaverba137:
This PR:

* adds download support via `desispec.io.download()` (closes #48)
* adds download keyword option to `desispec.io.findfile()` (closes #49)
* Includes new tests for `findfile()`
* Includes tests of downloads, however, tests are
  - skipped if no ~/.netrc file is present
  - only performed on one file, which automatically disables multiprocessing
* Incidentally fixes some documentation bugs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,7 +90,7 @@ install:
     - export CONDA_INSTALL="conda install -c astropy-ci-extras --yes python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION"
 
     # CORE DEPENDENCIES
-    - if [[ $SETUP_CMD != egg_info ]]; then $CONDA_INSTALL numpy=$NUMPY_VERSION scipy=$SCIPY_VERSION pyyaml pip Cython jinja2; fi
+    - if [[ $SETUP_CMD != egg_info ]]; then $CONDA_INSTALL numpy=$NUMPY_VERSION scipy=$SCIPY_VERSION pyyaml pip Cython jinja2 requests; fi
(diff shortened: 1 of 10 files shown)


weaverba137:
One more item to do before merging: clean up any files left over from download tests.

sbailey:
This PR brings in a dependency upon the external requests package.  That is included in the anaconda python distribution by default and is easy to "pip install requests" otherwise, but it isn't part of hpcports AFAICT.  This will temporarily break our installations at NERSC until Ted is back and can add requests to HPCports (or we install requests directly into the desi software area, but HPCports would be better).

For the record, this PR also includes some docstring edits and absolute vs. relative import cleanup that came along for the ride.

weaverba137:
Good point.  I have already created a PR on hpcports to add requests.

The last commit on this branch includes the clean-up of files/directories downloaded by tests.

PR #53, opened 2015-09-02
select sky fibers

profxj:
Use SPECMIN,SPECMAX to select sky fibers in frame
in desispec.sky


sbailey:
For context, there was a bug (probably introduced in the I/O refactor) where the sky subtraction was trying to use sky fibers from other wedges and was crashing.  This PR ensures that it only uses fibers within its own wedge.

Issue #52, opened 2015-08-12
combine_ivar doesn't work as advertised


weaverba137:
The test in `desispec.test.test_util.TestNight.test_combine_ivar()` performs this test:

```python
        ivar = util.combine_ivar(1, 2)
```

`desispec.util.combine_ivar` tries to convert the scalar values to an array, using

```python
    iv1 = np.asarray(ivar1)  #- handle list, tuple, ndarray, and scalar input
    iv2 = np.asarray(ivar2)
```

However, for scalar input, `np.asarray()` returns a numpy *scalar*, which cannot be indexed.  When the combined inverse variance is computed, 

```python
    ivar[ii] = 1.0 / (1.0/iv1[ii] + 1.0/iv2[ii])
```

the indexing of the scalar values throws an exception.

It is very strange that this test *ever* passed.

weaverba137:
This may be due to the numpy version.  In 1.9.x, this actually seems to work. However, in 1.8.x it does not.

sbailey:
Good to support numpy 1.8.x, but it would probably be better to handle this using np.atleast_1d() instead of the original np.asarray(), rather than the explicit numpy version checking in commit d4f7735 .

weaverba137:
Ah, yet another useful function that numpy thought of already!

weaverba137:
This should have been automatically closed by #56.  However, I think you have to reference the issue in the *body* of a PR in order to have a PR automatically close it.

PR #56, opened 2015-09-04
Fix combine ivar for numpy 1.8

sbailey:
This updates util.combine_ivar to support old numpy 1.8 by using `np.atleast_1d` instead of `np.asarray` when casting the input python scalars into numpy arrays.  It also handles the case of the inputs being numpy 0-dim arrays.  This has been tested on both numpy 1.8.2 and numpy 1.9.2.


sbailey:
This fixes issue #52.

weaverba137:
The code looks fine so far, although it duplicates some work that I've done on the metadata-db branch.  Let me make sure there won't be any conflicts.

weaverba137:
Let's go ahead & merge this.  If there are any conflicts with my branch they will be very minor.

Issue #51, opened 2015-08-06
Load file metadata into database


weaverba137:
Note: the branch 'metadata-db' has been created to address this issue.

We want a metadata database to track the relationships among nights, exposures, bricks and files.  The schema of such a database is described in the file etc/file_db.sql.  We still need code to:

* Register files at file creation time.
* Load a set of files after creation (*i.e.* do the initial population of the database).
* Encapsulate commonly-used queries.

Based on the schema in the file mentioned above, we know how to fill in everything except these columns:

* exposure.telra, exposure.teldec.  These are supposed to be in the pix*.fits files but appear to be missing.
* exposure.tileid.  This should also be in a header somewhere.

weaverba137:
Is there a separate issue already created to address the issues with some missing data related to exposures mentioned above?

sbailey:
I opened [desisim issue 19](https://github.com/desihub/desisim/issues/19) to fix this in the pixel-level simulation output.  That ticket can also be used for any other missing header keywords that we want to fix in the short term.

sbailey:
The discussion of PR #61 raised the issue of how spectro pipeline outputs trace their input provenance.  Currently they don't.  In the future:

* the spectro pipeline planning script could record the dependency chain while putting together the commands to run (this script doesn't yet exist)
* the output files could keep in their header the names of all of their inputs (similar to tracking code versions, also still to be done)
  * this should include the location of the raw data inputs, which crosses out of the production output base directory and into a differently-located raw data input directory.
* both

Discuss.  Let's pick a path and then open a separate issue for implementing that.  My preliminary vote is for "both" even though this does replicate information — each file is standalone for knowing its inputs, while the pipeline planning full chain output could be useful for the big picture.

In the meantime, it seems that the metadata loading script should use `$DESI_SPECTRO_REDUX/$PRODNAME` as the default placeholder for where to find processed data, and `$DESI_SPECTRO_DATA` as the default for where to find the raw data.  Both should be override-able with command line options.  Output should go into `$DESI_SPECTRO_REDUX/$PRODNAME/etc/metadata.db`, even if we're only loading raw data for now.  If we want to have a raw-data-only metadata loading script, let's keep that as a separate thing.

weaverba137:
It's getting toward the end of the day here, so I'll just quickly say two things:

* I also would vote for both, unless it becomes too difficult to manage.
* Which `PRODNAME` should I use for testing purposes for the next round of metadata development.

[...]

PR #61, opened 2015-09-09
Metadata database

weaverba137:
This PR implements a SQLite metadata database.  This branch is getting quite a bit ahead of master, so we might want to consider doing this PR to get the bulk of the code integrated, then create new branches for smaller tasks.  Specific functionality *not* yet implemented:

* Mapping of files to bricks:  This is non-trivial due to potential size of the mapping.
* Mapping of files to files: Where does the database loader get the information on what files depend on what other files.  From the datamodel?  From metadata in the files themselves?

Please comment!


sbailey:
Yes, let's merge this now to prevent master from getting too far out of sync.  You can decide if you want to keep making changes on this branch or if you want to delete it and open new branches for the remaining features.

Prior to merging:

`bin/load_database` is too generically named for something that will end up in our path.  We'll have multiple database loading scripts (targets, spec pipeline metadb, spec pipeline output, maybe LSS catalog...).  It's a mouthful, but maybe something like load_spec_metadb?

`desispec.io.crc` is impressive, but could we use sha1 or md5 available in the python standard library `hashlib` instead?  I see that sha1sum and md5sum are *not* included by default on MacOS as executables, but do we specifically need to support the ability to calculate a hash from both python and a pre-installed executable?

desispec.io.database.main() appears to write `$DESISPEC/etc/metadata.db`, i.e. writing into a *code* directory rather than into into the production *data* directory.  It also has hardcoded alpha-5, a brick filename, and a specific night.  It should take a production directory as input and write the output into an etc/ directory of that input production, with options for writing to a separate location.  It could default to using `$DESI_SPECTRO_REDUX/$PRODNAME` for input and writing to `$DESI_SPECTRO_REDUX/$PRODNAME/etc/metadata.db` as output.

weaverba137:
Good comments.  Yes, there's some hard-coded stuff that definitely needs to be removed.

* `load_database`: sure, that name can change.
* `desispec.io.crc`: Turns out there are some subtleties there that I still need to work out.  The fundamental problem there is what to use as a unique ID for a file.  The checksum (of some variety) is perfect, but it comes down (again!) to whether one prefers integer or text primary keys.  The md5/sha1 varieties are harder to represent as a conventional, human-readable integer.
* Absolutely, that database should be written elsewhere.  Those are good suggestions.

weaverba137:
`bin/load_database` -> `bin/desi_load_metadata`

weaverba137:
Hard-coded paths are now removed, and the database file is written to the correct directory.  However, the *default* brick filename is still hard-coded, because it's not obvious how else to specify it.  There is a command-line option to specify it though.

weaverba137:
Now uses SHA1 as the file id.  We'll see if that works, or is too cumbersome.

The latest version of the database is in desi/spectro/sim/alpha-5/etc/metadata.db.  Please take a look.

sbailey:
Using sha1 (160 bits of information) is safe to avoid random hash collisions.  Any 64 bit integer hash risks collisions by the time we scale up to full DESI (e.g. see discussion at http://stackoverflow.com/questions/754657/ok-to-use-only-64-bits-of-sha1-hash-as-an-id).

Do you actually use the feature that the file primary key is the hash, vs. just using `INTEGER PRIMARY KEY` to get a guaranteed unique integer?  Using a hash seems nifty, but I'm not quite sure what the use case is.  If sha1 turns out to be too cumbersome, is there a reason not to just use the built-in unique integer facilities of the db rather than trying to derive our own?

The output is going to `$DESI_SPECTRO_SIM/$PRODNAME/etc/metadata.db`, but it should go to `$DESI_SPECTRO_REDUX/$PRODNAME/etc/metadata.db`.  i.e. it does into the reduction directory along with the output files, not into the input raw data / sim directory (which multiple reductions could use as a starting point).

i.e. sha1 is good for now; let's try it.  The output location should be changed before merging.  Otherwise looks good (but I did want to reach agreement on output location prior to merging rather than updating it myself by fiat and merging out from under you).

weaverba137:
You have identified a chicken-and-egg problem.  `INTEGER PRIMARY KEY AUTOINCREMENT` creates the key *after* the data is inserted, but because of all the connections to other tables, it makes the code much simpler and reduces load on the database if one knows the key beforehand.  Note that `AUTOINCREMENT` is required to enable to database to compute the key.  While there are database functions that allow one to obtain the next value, or reset a value, again, this adds extra database overhead, makes error checking more complicated, etc.

Since we use .sha1sum files for data integrity anyway, I have added the feature that these files will be read to obtain the checksums, rather than recomputing them on the fly for every file.  This can result in substantial speed-up.

In terms of the output directory, I need to make sure that we're not overloading `$PRODNAME`, because that has different meanings depending on whether one is dealing with simulated raw data, real raw data or reductions.  Right now I'm *only* loading raw data, since I don't know what reduced data is connected to the `alpha-5` simulation, and more importantly, what metadata is relevant to reduced files versus raw files.  (This is more me thinking "out loud"...)

weaverba137:
I read the SQLite documentation much more carefully and:

1. `AUTOINCREMENT` is not actually necessary to create an auto-generated integer primary key.
2. SQLite nevertheless does not provide the same level of access to auto-generated integer sequences as Postgresql does.  While it is possible to query a SQLite database with a simple function to obtain the *last* integer generated, it is not necessarily so easy to predict the *next* value.
3. This gets into an area where SQLite-flavoured SQL starts to differ significantly from other flavors, especially Postgresql, meaning that converting between the two would not be straightforward.
4. It is not written anywhere in the SQLite documentation that I could see that the auto-generated values are perfectly sequential, only that they are 64-bit integers.
5. Imagine one is attempting to `INSERT` a set of rows, and that one has attempted to predict the primary keys beforehand to use as foreign keys in another table.  Now imagine one ore more of the `INSERT`s fails.  What happens to the sequence of primary keys predicted, versus what actually got created in the database?

sbailey:
Clarifying environment variables for picking output directories:

Simulated raw data go to `$DESI_SPECTRO_SIM/$PIXPROD`.  Grouping by $PIXPROD allow us to keep different simulations isolated from each other.

The spectro pipeline reads data from `$DESI_SPECTRO_DATA`.  When running on simulated data, we set `$DESI_SPECTRO_DATA = $DESI_SPECTRO_SIM/$PIXPROD`, but the spectro pipeline itself doesn't know that it is simulated data, i.e. it should never use `$DESI_SPECTRO_SIM` or `$PIXPROD` since those are meaningless for the final real data.

Spectro redux goes to `$DESI_SPECTRO_REDUX/$PRODNAME` .  I'm considering renaming `$PRODNAME` -> `$SPECPROD` to give it better symmetry with `$PIXPROD` and remove the ambiguity of pixel simulation vs. spectro pipeline productions.  Note that `$DESI_SPECTRO_SIM/$PRODNAME` is mixing two things and shouldn't be used; perhaps the rename would help avoid this confusion.

Note: the current spectro redux don't seem to have kept the link back to which raw data dir was used.  This is an oversight.  For now, assume inputs come from `$DESI_SPECTRO_DATA` by default, and we'll just have to set that.

I had been thinking about the metadata db as being primarily about the spectro pipeline outputs, and it just happens to include the raw data files too as the starting point.  i.e. the metadata.db should be under the spectro pipeline output.  But I can see the usefulness of having a metadata description for just the raw data, independent of what productions use it.  But then it becomes tempting to include other raw data metadata such as exposure times, flavor (arc/flat/science/bias/dark...), etc.

If we separate out databases for raw data vs. processed data, we should keep that distinction clear.  If the current script is intended to eventually include processed data, then its output should go to `$DESI_SPECTRO_REDUX/$PRODNAME/etc/` by default.  If loading spectro pipeline outputs is intended to be a different script, we should scope out what the looks like and how it relates to this script prior to committing to this one.

weaverba137:
OK, I get all that, & it's consistent with what we discussed in person back in August.

The problem is, the *only* data that I have so far been able to figure out how to load is the data in `$DESI_SPECTRO_SIM/alpha-5`.  For example, all the metadata in the exposure table is obtained from the files in that directory.  Also the brick file in there.  In some sense, I'm working forwards from the raw data instead of backwards from the final reduced data.

So in order for `$DESI_SPECTRO_REDUX/$SPECPROD` (I like that suggestion, by the way) to be the *starting* point for the database loading, I need to know which `$SPECPROD` to run the tests on, and which metadata to use to find all the pointers back to the exposure metadata.  We never really established that in August.

Also, we still need to work out how to establish the file dependency tree, & the above problem is related to that.

Finally, in terms of raw data metadata, all the stuff you were thinking about is already supported!  The exposure time is already in the exposure table.  The flavor is just another entry in the exposureflavor table, etc.  I feel like you're looking at the inputs to the load, but not looking at all at the output, *i.e.* the database itself.

sbailey:
I'll merge this now, but let's change the metadata.db output location in a future PR after we've sorted out how to relate the spectro pipeline outputs to each other and to their input raw data.  Let's move that discussion back to issue #51 rather than in the thread of this PR.  @weaverba137 can decide if he wants to close this branch or leave it open for continued development and future pull requests.

PR #62, opened 2015-09-11
Image class

sbailey:
This pull request implements a lightweight Image class, similar to the lightweight Frame, SkyModel, etc. classes.  It bundles a 2D image with inverse variance and a mask, to be used for preprocessing, cosmic ray finding, and pixel-simulation code.  This PR includes io.read_image(), io.write_image(), and unit tests.

The current class treats readnoise as a single float.  A future extension should make this (optionally?) a 2D array, e.g. to keep track of the different quadrants and the effect of pixel flats.

Assigned to @weaverba137 for review, but @profxj and others are welcome to comment too (github, unlike bitbucket, only allows a single reviewer to be assigned).


weaverba137:
Looks OK so far, but I'll have to take a closer look on Monday.

weaverba137:
I think this is OK to merge.

rainwoodman:
Minor comments:

- I like meta is a dictionary, not as keyward arguments. 

- I am a bit concerned about 'rawdata' vs 'data'; aka, in cases function name suggests 'data' but the actual string is 'rawdata'. What is the difference? Is it due to API back-ward compatibility function names need to be preserved?

sbailey:
`rawdata` vs. `data`: thanks for the comment.  I converted io.findfile to use `rawdatadir` internally rather than just `data` since that was ambiguous about raw data vs. processed data, which live in different directory trees.  But I missed the `data_root()` function, which remained ambiguous.  I just renamed this to `rawdata_root()` to be clearer.  `specprod_root()` is the partner function for finding the root of the current spectroscopic data production directory.

Tests pass on my laptop; after confirming that Travis agrees I will merge this.

rainwoodman:
Thanks for clarifying.

PR #68, opened 2015-09-17
Allow Image slicing

sbailey:
Enables slicing of Image objects, e.g.

    img1 = Image(pix, ivar, mask)
    img2 = img1[0:100, 0:200]
    img3 = img1[0:100]
    img4 = img1[:, 0:200]

If NAXIS1 and NAXIS2 are in the meta header, those are updated with the new slice dimensions.  Tests are included.

### DIFF
--- a/py/desispec/image.py
+++ b/py/desispec/image.py
@@ -1,6 +1,7 @@
 '''
 Lightweight wrapper class for preprocessed image data
 '''
+import copy
 
 class Image(object):
     def __init__(self, pix, ivar, mask=None, readnoise=0.0, camera='unknown',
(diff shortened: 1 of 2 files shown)


weaverba137:
Looks good to me.

weaverba137:
Actually, maybe I'm being paranoid, but you don't do any input validation in the initial stages of slicing the `Image`'s attributes.  For example, `pix = self.pix[xyslice]` is evaluated without verifying that `xyslice` is actually a slice.  In the unlikely event that someone tries to slice with something other than a slice (maybe they think Image is like a dictionary and try to do keyword access), the underlying attributes *might* catch the error, but they might not, since it is possible to do keyword access on certain subclasses of NumPy arrays.

sbailey:
Good point; I had already merged and deleted the branch before this comment came in, but I just added some input validation, raising a ValueError if it isn't a proper slice.  I'll merge this update via another PR.

PR #69, opened 2015-09-17
test Image slicing inputs

sbailey:
Adds input validation when slicing Image objects.  Includes tests for bad slicing.


weaverba137:
Looks good.  Very thorough!

sbailey:
And I added two more tests to be super extra thorough.
Some CCD bit mask rearranging came along for the ride.

PR #70, opened 2015-09-17
Image io workaround for astropy.io.fits bug on uint16 output

sbailey:
It appears that astropy.io.fits has a bug when writing compressed images of uint16.  It gets the TFIELD and BITPIX header keywords in the wrong order, which can choke other fits io libraries like cfitsio.  This PR works around that by storing the Image.mask as a int16 rather than uint16, and then casting back to uint16 upon reading.  I was unable to find a workable verify option to get astropy.io.fits to write it correctly in the first place.

If you want to chase the astropy.io.fits bug to see if there is a more direct workaround, here is example bad fits output:

    from astropy.io import fits
    import numpy as np
    hx = fits.HDUList()
    hx.append( fits.PrimaryHDU(np.zeros(10)) )
    hx.append( fits.CompImageHDU(np.zeros(10).astype(np.uint16)) )
    hx.writeto('blat.fits', clobber=True)

The resulting blat.fits file fails [fitsverify](https://heasarc.gsfc.nasa.gov/docs/software/ftools/fitsverify/).

### DIFF
--- a/py/desispec/image.py
+++ b/py/desispec/image.py
@@ -2,6 +2,7 @@
 Lightweight wrapper class for preprocessed image data
 '''
 import copy
+import numpy as np
 
 class Image(object):
     def __init__(self, pix, ivar, mask=None, readnoise=0.0, camera='unknown',
@@ -31,7 +32,10 @@ class Image(object):
             
         self.pix = pix
         self.ivar = ivar
-        self._mask = mask
+        if mask is not None:
+            self._mask = mask.astype(np.uint16)
+        else:
+            self._mask = None
         self.meta = meta
         
         #- Optional parameters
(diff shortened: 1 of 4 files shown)


weaverba137:
Is this bug known to the astropy developers?

sbailey:
It doesn't seem to be in their issue list.  I'll file a report.  As an update this does work:

    hx.append( fits.CompImageHDU(np.zeros(10).astype(np.uint16), uint=True) )

But these do not:

    hx.append( fits.CompImageHDU(np.zeros(10).astype(np.uint16)) )
    hx.append( fits.CompImageHDU(np.zeros(10).astype(np.uint16), uint=True, name='BLAT') )

i.e. apparently you are supposed to tell astropy.io.fits that your data includes uint; I don't understand why that is required instead of just auto-discovering it.  But using uint=True is broken when also trying to set the EXTNAME via name=...

sbailey:
Hmm.  Correction.  astropy.io.fits is doing some sort of state caching such that things that the above "working" example only works if you have previously failed and remain in the same ipython session and try to create a new HDUList and write a new file.  If you start over again from scratch it also fails.  So I haven't found any options that work with uint, with or without setting the extname.

rainwoodman:
Just a rant. "state caching" smells like really poor design decisions.

PR #72, opened 2015-09-26
QA Exposure

profxj:
Generated a QA Class for pipeline assessment on
a given exposure.  

Includes simple QA on Skysub algorithm.

Includes simple I/O tests.
Includes an iPython Notebook for usage/tests.

### DIFF
--- a/py/desispec/qa/qa_exposure.py
+++ b/py/desispec/qa/qa_exposure.py
@@ -1,21 +1,23 @@
 """
-Class to organize and execute QA for a DESI exposure
+Classes to organize and execute QA for a DESI exposure
 """
 
 from __future__ import print_function, absolute_import, division, unicode_literals
 
 import numpy as np
 
-class QA_Exposure(object):
-    def __init__(self, exptype='none', expid=-1, in_data=None):
-        """
-        Class to organize and execute QA for a DESI exposure
+from desispec.sky import qa_skysub
 
-        x.expid, x.exptype
+class QA_Frame(object):
+    def __init__(self, flavor='none', camera='none', in_data=None):
+        """
+        Class to organize and execute QA for a DESI frame
+
+        x.flavor, x.data, x.camera
         
         Args:
-            exptype: str, optional exposure type (e.g. flat, arc, science)
-            expid: int, optional exposure number
+            flavor: str, optional exposure type (e.g. flat, arc, science)
+            camera: str, optional camera (e.g. 'b0')
             in_data: dict, optional -- Input data 
               Mainly for reading from disk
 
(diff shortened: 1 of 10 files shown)


sbailey:
Several usage/style questions:

The intension is that there will be one qa file per exposure, and each step will read that in as a `QA_Exposure` object, update the contents, and then write it back out?  The "planned in my head but not yet implemented or even written down" pipeline design was going to allow each frame to be processed independently and in parallel.  This single qa-{expid}.yaml file creates a race condition point that would need to be coordinated.  Options:
  * deal with it at the pipelining level.  Not quite sure how to handle this while still allowing parallel frame processing.  File locking is a bit fragile (in particular, `fcntl.flock` doesn't work from Edison batch nodes to the /project file system...)
  * one qa file per frame (expid,camera) instead of per exposure.  Optionally these could be merged at the end of processing in a separate step.
  * perform QA as a separate after-burner step, doing each frame in a row.  This requires extra I/O to load all the data files back in instead of doing the QA as you go when the data are already in memory.
  * something else?

I was surprised to see a `QA_Exposure` object as an *input* parameter to `qa_skysub`, rather than the returned output.  It looks like this object is modified by the function call.  Normally that pattern is avoided in python unless there is a specific memory or performance reason to prefer an in-place modification.  At minimum it should be clearly stated in the docstring that the `qaexp` object will be modified by this call.  Design alternatives to consider:

  * `qa_skysub(frame, fibermap, skymodel, qaexp=None)` : if qaexp is None, create one, fill it in, and return it.  Otherwise, update the one that was passed in and return that.
    * only mildly different from existing implementation, but limits boilerplate code for creating new QA_Exposure object outside of qa_{xyz} code.
  * `qa_skysub(frame, fibermap, skymodel)` always returns a new `QA_exposure` object, which is explicitly merged into the existing qaexp structure with code that looks something like:
    * `qaexp.append( qa_skysub(frame, fibermap, skymodel) )`
  * let `write_qa_exposure(outfile, qaexp)` handle the merging: if outfile already exists, then read it in, merge in the info from qaexp, then write it back out.  We would probably want to use QA_Exposure.append() like the previous option rather than having the write code know the details of how to merge.
    * This saves each pipeline step from having to do the boilerplate of checking if the file already exists, reading it in, and passing the qaexp object forward.  Perhaps the function should be called `append_qa_exposure` in this case?

All algorithms need to define what should happen in the case of a collision: if an algorithm is trying to update a qa field that already exists, should it overwrite (silently or with warning?), raise an exception, raise an exception unless force=True, ...?

Is `qaexp.init_skysub(frame.meta['camera'])` necessary before calling `qa_skysub`, or should `qa_skysub` just do that internally?

`qa_skysub` is directly accessing the ._data member of the input QA_Exposure object.  If you really intend individual algorithms to directly manipulate this structure, don't hide it with an underscore.  Consider providing access via something like `QA_Exposure.add_qa(camera, algorithm, qadict)` instead ('algorithm' would be 'skysub', 'fluxcalib', etc., and could maybe use a better name).

py/desispec/qa/qa_exposure.py
```diff
@@ -0,0 +1,76 @@
+"""
+Class to organize and execute QA for a DESI exposure
+"""
+
+from __future__ import print_function, absolute_import, division, unicode_literals
+
+import numpy as np
+
+class QA_Exposure(object):
+    def __init__(self, exptype='none', expid=-1, in_data=None):
```
sbailey: Following SDSS/BOSS, we've been using "flavor" instead of "exptype" elsewhere within desi code and files.

profxj:
Made a number of modifications, following SB suggestions
and my own inspirations:

1. Generated QA_Frame to deal with parallelization.  I think a QA_Exposure
class should still exist (eventually) to collate the various frames and turn the various QA files
into a single one (by camera).

2. Made aspects of the QA_Frame class more modular

3. Made the call to QA calculations a method in the Class, e.g.
  qaframe.run_qa('SKYSUB', (frame, fibermap, skymodel))

4. Separated QA parameters (which are initialized within the Class) from
 QA outputs (which are generally generated elsewhere).

5. Updated the Notebook accordingly

6. Ran unit tests and a test case on my laptop (didn't repeat on Edison)

PR #73, opened 2015-09-28
fixes new location of star templates

sbailey:
stellar_templates directory at NERSC had been renamed to "star_templates", which broke the end-to-end pipeline tests (flux calib uses those files).  This one liner fixes that.


moustakas:
I inadvertently broke that.  I wanted the top-level directory to better match the corresponding spectral type (STAR).

sbailey:
The new directory name makes sense, and the integration tests did their job of catching the issue before we ran a bunch of jobs with the wrong config.  The only delay was me waiting until today to see why they had started failing.

Issue #14, opened 2015-04-22
test_fiberflat needs more tests


sbailey:
py/desispec/test/test_fiberflat.py includes two tests flagged as expected failures since they haven't been written yet.  Fill these in.  The "test_resolution" test that does exist can serve as a template for a non-trivial test (that found a non-trival bug that has since been fixed).

    #- Tests to implement.  Remove the "expectedFailure" line when ready.
    @unittest.expectedFailure
    def test_throughput(self):
        """
        Test that spectra with different throughputs but the same resolution
        result in the expected variations in fiberflat.
        """
        raise NotImplementedError

    @unittest.expectedFailure
    def test_throughput_resolution(self):
        """
        Test that spectra with different throughputs and different resolutions
        result in fiberflat variations that are only due to throughput.
        """
        raise NotImplementedError

sbailey:
More unit tests added with PR #74.  Closing issue.

PR #74, opened 2015-09-28
Unittests

sbailey:
This PR adds more fiberflat tests to address issue #14, and addresses several fiberflat problems found in the process:

  * fixes incorrect error propagation (it was using spec.flux after modifying it by the flat, when the formula was intended to use the original input flux; beware of altering inputs...)
  * sets specmask.BADFIBERFLAT where fiberflat=0, fiberflat_ivar=0, or fiberflat_mask==0
  * doesn't set flux=0 if fiberflat=0; let the mask handle that case instead

It also implements two previous placeholder algorithmic tests of the fiberflat calculation.  There are edge effects in the fiber flats in the case of varying resolution and varying throughput; this test avoid those by not testing the fiberflat within 2 sigma of the edge of the spectrum.

Changes that came along for the ride:
  * maskbits.spmask renamed to maskbits.specmask
  * misc comments and docstrings
  * more FiberFlat input validation when creating the object


weaverba137:
* Double-check the _version.py file, it's a leftover from before we cleaned up the 1.2 tag.
* GitHub says there are some conflicts. You might need to merge with trunk.

sbailey:
Short version: I think the conflict is resolved, with version == 0.2.dev399 .

Longer version:  I merged with master and got a conflict in _version.py, which I resolved by hand by accepting the master version (0.2.dev387).  I committed that, and then reran "python setup.py version" to try to bump the dev version, but it took me back to 0.1.1.dev396.  I changed that by hand to 0.2.dev, committed that, checked out master (looked around but didn't do anything), checked out unittests again, and reran "python setup.py version", upon which it took me to 0.2.dev399.

So I'm still getting the hang of the versioning workflow, but I think things are ok now for a merge.  I'd still appreciate a cross check from Julien on the error propagation changes.

PR #87, opened 2015-10-31
Pip install

weaverba137:
Please double-check this one!

* pip install support
* Travis tests now use their docker infrastructure, so they're faster.
* Read the Docs support
* Test coverage support.


sbailey:
Lots of cleanup changes to comments and spaces; I don't know of a reasonable way to test the pip install stuff without actually merging and trying it out.  That part doesn't touch the core algorithms so it shouldn't otherwise break anything, even if there are remaining things to do on the pip install / config side.  Merging.

dkirkby:
@sbailey You can test pip uploading and installing from an un-merged branch using https://testpypi.python.org/pypi instead of the usual pypi.python.org

PR #98, opened 2015-11-19
Add --objtype option to desi_zfind

sbailey:
This PR adds a --objtype option to desi_zfind.py to restrict which templates are considered.  This is especially useful for leaving out the slow QSO templates while debugging ELGs and LRGs.  Redmonster doesn't distinguish between ELG and LRG so those are mapped to GALAXY.  Default is to fit GALAXY,STAR,QSO.

Examples:
```
desi_zfind.py -o blat.fits -b {brickname} --objtype ELG
desi_zfind.py -o blat.fits -b {brickname} --objtype GALAXY,STAR
desi_zfind.py -o blat.fits -b {brickname}
```

### DIFF
--- a/py/desispec/zfind/redmonster.py
+++ b/py/desispec/zfind/redmonster.py
@@ -17,20 +17,21 @@ from desispec.log import get_logger
 class RedMonsterZfind(ZfindBase):
     """Class documentation goes here.
     """
-    def __init__(self, wave, flux, ivar, R=None, dloglam=1e-4):
+    def __init__(self, wave, flux, ivar, R=None, dloglam=1e-4, objtype=None):
         """Uses Redmonster to classify and find redshifts.
 
         See :class:`desispec.zfind.zfind.ZfindBase` class for inputs/outputs.
 
+        optional:
+            objtype : list or string of template object types to try
+                [ELG, LRG, QSO, GALAXY, STAR]
+
         TODO: document redmonster specific output variables
         """
-        try:
-            from redmonster.physics.zfinder import ZFinder
-            from redmonster.physics.zfitter import ZFitter
-            from redmonster.physics.zpicker2 import ZPicker
-        except ImportError:
-            get_logger().error("You are attempting to use RedMonster, but it is not available for import!")
-            raise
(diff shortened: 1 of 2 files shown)


py/desispec/zfind/redmonster.py
```diff
@@ -58,14 +59,35 @@ def __init__(self, wave, flux, ivar, R=None, dloglam=1e-4):
...
-        ]
+        self.templates = list()
+        for x in templatetypes:
+            if x == 'GALAXY':
+                self.templates.append(('ndArch-ssp_em_galaxy-v000.fits', 0.6, 1.6))
+            elif x == 'STAR':
+                self.templates.append(('ndArch-spEigenStar-55734.fits', -0.005, 0.005))
+            elif x == 'QSO':
+                self.templates.append(('ndArch-QSO-V003.fits', 0.0, 3.5))
+            else:
```
moustakas: Are these values setting the redshift ranges over which redmonster will search for a solution?  In other words, if I set --objtype ELG will redmonster only search over the redshift range [0.6,1.6]?  Hopefully not!  On the hand, we *do* want setting OBJTYPE to trigger a narrowing of the redshift search range, for example, [0,2] for ELG, [0,1.5] for LRG, etc.

sbailey: Whoop!  Good catch.  Redmonster doesn't distinguish between ELGs and LRGs for its templates, but this code is restricting the GALAXY=(ELG or LRG) search to a ELG-ish redshift range of 0.6 to 1.6.  At minimum we should include a range appropriate to both and a superset of what might be in the targets rather than what we want to be in the targets.  However, if a future redmonster version (or alternate redshift fitter) had ELG-optimized templates, I'd be suspicious of answers with z>1.67 since [OII] has walked off the edge of the spectrograph and there wouldn't be much real signal to latch onto.  How about [0,1.67] for this range?  Or should it really be out to z=2?

moustakas: I think we should go to z=2 by default.  There's a MgII 2800 (narrow) emission line that may occasionally yield a redshift.

PR #99, opened 2015-11-26
enable read_frame to read brick files

sbailey:
brick files are essentially frame files that mix spectra from different exposures plus additional information like the fibermap  Both have FLUX, IVAR, WAVELENGTH, RESOLUTION, ... .  This PR enables desispec.io.read_frame to read a brick file as if it was a Frame.  The only change needed was to set SPECMIN to a default of 0 since that is missing (and not so meaningful) in the brick file header.

This is motivated by the current Brick object being entangled with the underlying file format, while the Frame objects are completely decoupled from their I/O (which I prefer, though I think I understand why the Brick objects are organized as they are for convenience while constructing the bricks from multiple input exposures).  Refactoring this to isolate data objects from I/O and maximize the symmetry between frame and brick files might be the better long-term solution.

i.e. this PR might be a bad idea, so please comment.  I'm assigning to @profxj since he was the one who originally required that the fiber array be set.  @dkirkby and others are also welcome to comment.

The memmap=False was added because astropy.io.fits will secretly and evilly keep your file open after you "close" it unless you tell it memmap=False.


sbailey:
Documenting a conversation with X:

* At minimum we should log a warning that SPECMIN wasn't in the header
* We should add a keyword to *every* file output defining what type of file it is (i.e. not relying only on the filename for that info).  If that had existed, then this could have used "if hdr['FILETYPE'] == 'brick'" instead of "if 'SPECMIN' not in hdr", since SPECMIN should be in frames but not bricks.

PR #101, opened 2015-11-27
Test

gdhungana:
This PR adds few tests for flux calibration.


sbailey:
Thanks for adding these tests.  They are failing the automated Travis tests integrated with github because the flux calib code still requires desisim to find the filter curves; see [desisim issue #8](https://github.com/desihub/desisim/issues/8) about moving the filter curves.  Travis doesn't know about this dependency and thus fails to find envvar $DESISIM .

Rather than hack Travis to add a dependency we don't want just to fix the tests, let's keep this PR on hold and fix the filter curve location within the next ~2 weeks.  I'll add this as an agenda item for the DESI data call this week (as part of a broader discussion about our dependency tree).

weaverba137:
What is the status of this PR?  It is now almost four months old.

sbailey:
As a reminder: desispec has a crazy dependency on desisim because that is where our filter/mag calculation code used to be.  This PR made this especially obvious because it would need Travis to install desisim to work.  We kept changing where we were going to put the filter/mag calculations, but it seems to be converging on speclite.filters.  Let's switch the desisim flux calibration code to use that for the magnitude calculations — and these tests will help confirm that we didn't break anything in the meantime.  Then merge.

@gdhungana I know you have multiple things going on right now.  Please give an update for when/if you'll be able to do this change or if we should find someone else to do it (like @rstaten maybe?)

gdhungana:
@sbailey Thanks for the speclite information. I just talked with Bob also and Ryan is out of town as well this week. This should be easy fix so I will transfer this to speclite.filters tomorrow if I get a chance, otherwise the day after. I hope that should be okay.

sbailey:
this PR is refactored into PR #116 .  closing.

PR #112, opened 2016-01-30
Speedup fiber gauss

julienguy:
* faster routine for fiber_gaus (gain of a factor 6), now total desi_bootcalib is 3 min per frame
 * legendre degree as option
 * some cosmetics for logging


julienguy:
The code is failing a test on the gaussian fit (sigma != 1.142540313858601), but it does not seem that the profile of test image https://portal.nersc.gov/project/desi/data/spectest/pix-sub_b0-00000001.fits.gz is actually gaussian so I am not sure the test is valid. (if the profile is not gaussian, one gets different sigmas depending on how the data are weighted).

figure_1

PR #115, opened 2016-02-17
speclite.filters for flux calibration

gdhungana:
This PR replaces the original filter responses in desisim/data by speclite.filters for flux calibration.

### DIFF
--- a/bin/desi_fit_stdstars.py
+++ b/bin/desi_fit_stdstars.py
@@ -79,7 +79,6 @@ def main() :
     filters=fibers["FILTER"]
     if 'DESISIM' not in os.environ:
         raise RuntimeError('Set environment DESISIM. Can not find filter response files')
-    basepath=DESISIM+"/data/"
 
     #now load all the skyfiles, framefiles, fiberflatfiles etc
     # all three channels files are simultaneously treated for model fitting
@@ -199,7 +198,7 @@ def main() :
 
         log.info('Star Fiber: {0}; Best Model Fiber: {1}; TemplateID: {2}; Chisq/dof: {3}'.format(l[0],bestModelIndex[k],templateid[bestModelIndex[k]],chi2dof[k]))
         # Normalize the best model using reported magnitude
-        modelwave,normalizedflux=normalize_templates(stdwave,stdflux[bestModelIndex[k]],mags,filters,basepath)
+        modelwave,normalizedflux=normalize_templates(stdwave,stdflux[bestModelIndex[k]],mags,filters)
         normflux.append(normalizedflux)
 
     # Now write the normalized flux for all best models to a file
--- a/bin/desi_fit_stdstars.py
+++ b/bin/desi_fit_stdstars.py
@@ -14,7 +14,7 @@ desi_fit_stdstars.py
 """
 
 #- TODO: refactor algorithmic code into a separate module/function
-#- TODO: move filters out of desisim and remove $DESISIM dependency
+
 
 from desispec import io
 from desispec.fluxcalibration import match_templates,normalize_templates
@@ -44,10 +44,6 @@ def main() :
 
     DESI_SPECTRO_REDUX=os.environ['DESI_SPECTRO_REDUX']
     PRODNAME=os.environ['PRODNAME']
-    if 'DESISIM' not in os.environ:
-        raise RuntimeError('Set environment DESISIM. It will be neede to read the filter transmission files for calibration')
-
-    DESISIM=os.environ['DESISIM']   # to read the filter transmission files
 
     if args.fibermap is None or args.models is None or \
        args.spectrograph is None or args.outfile is None or \
@@ -77,8 +73,6 @@ def main() :
     NIGHT=fiber_header['NIGHT']
     EXPID=fiber_header['EXPID']
     filters=fibers["FILTER"]
-    if 'DESISIM' not in os.environ:
-        raise RuntimeError('Set environment DESISIM. Can not find filter response files')
 
     #now load all the skyfiles, framefiles, fiberflatfiles etc
     # all three channels files are simultaneously treated for model fitting
@@ -153,12 +147,13 @@ def main() :
 
     stdwave,stdflux,templateid=io.read_stdstar_templates(args.models)
 
+    #- Withdrew trimming below to address filters with long tails
     #- Trim standard star wavelengths to just the range we need
-    minwave = min([min(w) for w in frameWave.values()])
-    maxwave = max([max(w) for w in frameWave.values()])
-    ii = (minwave-10 < stdwave) & (stdwave < maxwave+10)
-    stdwave = stdwave[ii]
-    stdflux = stdflux[:, ii]
+    # minwave = min([min(w) for w in frameWave.values()])
+    # maxwave = max([max(w) for w in frameWave.values()])
+    # ii = (minwave-10 < stdwave) & (stdwave < maxwave+10)
+    # stdwave = stdwave[ii]
+    # stdflux = stdflux[:, ii]
 
     log.info('Number of Standard Stars in this frame: {0:d}'.format(len(stars)))
     if len(stars) == 0:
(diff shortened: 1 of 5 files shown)


py/desispec/fluxcalibration.py
```diff
@@ -175,7 +175,7 @@ def findappMag(flux,wave,filt):
         #Normalizing using only SDSS_R band magnitude
         if v.upper() == 'SDSS_R' or v.upper() =='DECAM_R' :
             refmag=mags[i]
-            filter_response=read_filter_response(v,basepath) # outputs wavelength,qe
+            filter_response=load_filter_response(v) # outputs wavelength,qe
             rebinned_model_flux=rebinSpectra(stdflux,stdwave,filter_response[0])
             apMag=findappMag(rebinned_model_flux,filter_response[0],filter_response[1])
```
dkirkby: Is there a good reason not to use the convolution method implemented by the returned `filter_response` object here?  In other words, replace these two lines with:
```
apMag = filter_response.get_ab_magnitude(stdflux, stdwave)
```

bin/desi_fit_stdstars.py
```diff
@@ -79,7 +79,6 @@ def main() :
     filters=fibers["FILTER"]
     if 'DESISIM' not in os.environ:
         raise RuntimeError('Set environment DESISIM. Can not find filter response files')
```
dkirkby: Does DESISIM still need to be defined?  If so, this exception message should be updated.

py/desispec/io/filters.py
```diff
@@ -8,6 +8,7 @@ def read_filter_response(given_filter,basepath):
     Read requested filter (SDSS_r, DECAM_g, ...) from files in basepath dir
     
     Returns tuple of wavelength, throughput arrays
+    This is for reading files from desisim/data etc.
```
dkirkby: Are the docs here about reading from basepath still valid?

py/desispec/io/filters.py
```diff
@@ -23,3 +24,34 @@ def read_filter_response(given_filter,basepath):
     ## filter_response=(wave, throughput, tck)
     ## return filter_response
     return wave, throughput
+
+def load_filter_response(given_filter):
+    """
```
dkirkby: Does it make sense to use the names already defined by `speclite.filters` so we can take advantage of their docs directly (http://speclite.readthedocs.org/en/latest/filters.html) instead of documenting these translation rules.

gdhungana:
@dkirkby Thanks for the comments. I don't think there is any desisim requirement. I didn't delete the earlier function to read the filter response, so just left it. Also thanks for the naming suggestion. Also I agree that filter_response.get_ab_magnitude looks more straightforward. That may also get rid of the convoluton function. Will change now!

gdhungana:
@dkirkby It seems the decam filters have long tails. This spans outside the desi coverage and can't get ab magnitude. Below is an example. It is okay for SDSS_R as my given wavelength goes beyond the filter coverage, but DECAM_R has long tail and returns a ValueError. Other DECAM filters also have the same issue. Can't calibrate as it stands.
 
In [1]: import numpy as np

In [2]: import speclite
In [3]: wave=np.linspace(3500.,9800.,10000.)
In [4]: flux=(np.random.normal(5.0,0.2)+np.ones(wave.shape[0]))*1.0e-17
In [5]: filter_response=speclite.filters.load_filter('sdss2010-r')
In [6]: filter_response.get_ab_magnitude(flux,wave)
Out[6]: 19.139648204867722
In [7]: filter_response=speclite.filters.load_filter('decam2014-r')
In [8]: filter_response.get_ab_magnitude(flux,wave)
ValueError                                Traceback (most recent call last)
<ipython-input-8-5f11a137e47e> in <module>()
----> 1 filter_response.get_ab_magnitude(flux,wave)

/usr/local/lib/python2.7/dist-packages/speclite/filters.pyc in get_ab_magnitude(self, spectrum, wavelength, axis)
    988         float or array
    989         """
--> 990         maggies = self.get_ab_maggies(spectrum, wavelength, axis)
    991         return -2.5 * np.log10(maggies)
    992 

/usr/local/lib/python2.7/dist-packages/speclite/filters.pyc in get_ab_maggies(self, spectrum, wavelength, axis)
    958             convolution = self.convolve_with_array(
    959                 wavelength, spectrum, photon_weighted=True,
--> 960                 interpolate=True, axis=axis, units=default_flux_unit)
    961         try:
    962             return convolution.value / self.ab_zeropoint.value

/usr/local/lib/python2.7/dist-packages/speclite/filters.pyc in convolve_with_array(self, wavelength, values, photon_weighted, interpolate, axis, units, method)
    911         """
    912         convolution = FilterConvolution(
--> 913             self, wavelength, photon_weighted, interpolate, units)
    914         return convolution(values, axis, method)
    915 

/usr/local/lib/python2.7/dist-packages/speclite/filters.pyc in __init__(self, response, wavelength, photon_weighted, interpolate, units)
   1193                 'Wavelengths do not cover {0} response {1:.1f}-{2:.1f} {3}.'
   1194                 .format(self.response.name, self.response._wavelength[0],
-> 1195                         self.response._wavelength[-1], default_wavelength_unit))
   1196 
   1197         # Find the smallest slice that covers the non-zero range of the

ValueError: Wavelengths do not cover decam2014-r response 3330.0-10990.0 Angstrom.

moustakas:
FYI -- I ran into the same issue here
  https://github.com/dkirkby/speclite/issues/12
In desisim at least I'm dealing with this by padding the templates, but I'm a little nervous about padding actual data during actual flux-calibration...

dkirkby:
How are the DECam magnitudes actually used in the flux calibration?

gdhungana:
I agree that templates go beyond these range and currently the template spectrum has been trimmed to little outside desi coverage in the flux calibration, which probably shouldn't be done, and is likely causing this. But I was more curious here.

sbailey:
Please merge from current master to get updated versions of numpy/scipy/astropy into the travis tests, which are currently failing due to old dependency versions.

sbailey:
@gdhungana please add your flux calib tests from #101 to this PR and close #101.

Side note: desispec.io.filters currently has some logic to translate between filter names like 'DECAM_R' and the names that speclite expects like 'decam2014-r' .  In the long term we can/should fix the fibermap files to use more consistent filter names, but we can leave the current name translation logic in place for now.

gdhungana:
@sbailey I have been merging the desispec:master (upstream master) for me locally, but Travis test is still failing. All I have been doing is "git pull upstream master" from my local branch after making commits and pushing them to the origin master (gdhungana:master) from where the PR is made. Pulling upstream master says "everything up-to-date." This had been working earlier, even yesterday when I first generated the PR. What should I try?

sbailey:
It looks like gdhungana:desispec has the correct numpy/scipy/astropy versions.  The tests are now failing due to travis missing speclite.

@dkirkby or maybe @weaverba137 :  How should speclite be installed?  e.g. maybe

```
in .travis.yml:
SPECLITE_VERSION=0.4

in etc/travis_env_common.sh:
$PIP_INSTALL speclite $SPECLITE_VERSION
```

@gdhungana please also push your version to a *branch* of desihub:desispec — that will give the rest of us write access to it so that we can contribute quick fixes like this too.

gdhungana:
@sbailey Isn't that strange? If speclite was not installed, how could it pass the PR the first time? I had used speclite then at first. Only after I made other commits and pushed, it failed! Anyway, I have pushed a new branch to desihub:desispec. Hope that helps to track this down.

dkirkby:
For specsim, I added speclite to my setup.py:
```
setup(name=PACKAGENAME,
   ...
   install_requires=['astropy', 'scipy', 'pyyaml', 'speclite'],
   ...
```
I also added a line to `.travis.yml` under the `install` target:
```
- if [[ $SETUP_CMD != egg_info ]]; then $PIP_INSTALL pytest-xdist speclite; fi
```

sbailey:
Tests now pass on the desihub/desispec:fluxcalib branch.  Closing this PR and re-opening a PR from that branch.

PR #120, opened 2016-02-26
fix flux units for mag calc

sbailey:
This PR corrects the flux units in the magnitude calculations in desispec.fluxcalibration.normalize_templates() .  Internally the fluxes are in 1e-17 erg/s/cm2/A but speclite filters have a default without the 1e-17, resulting in crazy magnitude calculations.  This adds an explicit unit.  The calculation is only ~100 us slower per stdstar.

It also upgrades the tests of desispec.fluxcalibration.normalize_templates() to check that the returned spectrum has the requested normalization.


gdhungana:
@sbailey Thanks Stephen for noticing the units inconsistency. I was suspecting that could happen as we saw that before as well but hadn't checked yet. I will also test on few more and report if I find anything else.

dkirkby:
In the (unlikely) event that 100us / star is an issue, you should be able to achieve 100us total overhead for N stars by creating a single filter convolution object that you re-use for all stars.

PR #128, opened 2016-03-01
Bootcalib: Small fix for z camera

profxj:
Fixes issue wrecking Ted's run
Hopefully improves robustness too


sbailey:
Hmm.  A one line change to the line_guess default, but line_guess isn't documented... :)

Could you comment on what "line_guess" is in the docstrings of load_gdarc_lines (where it is an output) and id_arc_lines (where it is an input)?

profxj:
Good point.  
docs now added

sbailey:
I'm merge this now, but please slip a change directly into master that also documents what "wmark" is in the load_gdarc_lines output.  You currently document 3/4 of the return values, and one of those (line_guess) is documented in terms of the one that is missing (wmark).  Not to be too persnickety...

PR #132, opened 2016-03-03
Fibermap format

### LINKED ISSUE #123: update fibermap data model
Review fibermap datamodel compared to upstream steps.  For any variables that have the same meaning, make them also have the same name and datatype.  e.g. TARGETMASK0 -> DESI_TARGET, and add BGS_TARGET and MWS_TARGET.

### LINKED ISSUE #124: include fibermap in frame and cframe files
The original fibermap is included in the raw data directory $DESI_SPECTRO_DATA/{night}/fibermap-{expid}.fits .  Spectro pipeline outputs are written to a different directory structure under $DESI_SPECTRO_REDUX , make it somewhat of a pain to be crossing directory hierarchies to get back to the fibermap file.

To simplify this, include the fibermap as a table in the extracted frame files, and also propagate into the cframe files.  This replicates data, but is similar to what BOSS does and what we have done for the brick files for convenience.  This makes the c/frame files more stand-alone useful, and makes the redux directory stand-alone useful after the spectral extraction step.

Probably include only the fibermap entries for the fibers in each frame, e.g. frame-b0* would have fibers 0-499, while frame-b1* would have 500-999, etc.

Include the fibermap in b\*, r\*, z\* so that each can be handled independently, and to provide datamodel robustness for the case when one of the channels is unavailable.

Currently we use the desi-agnostic 'exspec' script in specter for the extractions.  This should be wrapped by desispec to provide the desi-specific functionality of propagating the fibermap.  This will require some script -> module refactoring of exspec.

sbailey:
this PR fixes #123 and fixes #124 via
* updates the fibermap format to match upstream data, in particular DESI_TARGET, MWS_TARGET, and BGS_TARGET
* updates the Frame object to include the fibermap as part of its data, both in memory and as an HDU in the file.
* adds scripts bin/desi_extract_spectra.py and bin/desi_merge_bundles.py which are DESI-specfic versions of specter bin/exspec and bin/merge_bundles.  This requires specter>=0.4.  These DESI scripts propagate the fibermap into the frame outputs.

github says this can't be automatically merged into master, but starting this PR to trigger tests and review.  I'll separately track down the merge conflict.

A companion desisim PR will require this version of desispec to get the updated fibermap format.


sbailey:
Had an analog chat with Ted to coordinate work with automate branch.  This PR shouldn't conflict with the current automate branch.  Merging now.

Issue #133, opened 2016-03-03
Data files not installed / usable


tskisner:
When doing "python setup.py install --prefix=<blah>", the files in the data subdirectory are not installed.  At run time, the path to data is formed with

desispec_path = imp.find_module('desispec')[1]+'/../../'

Do we have a standard way of installing data files with setup.py, and obtaining the path to their installed location?

weaverba137:
@tskisner, where exactly are you trying to install desispec?  Which version?  Any particular reason you're not using pip?  Can you use NERSC as a workaround?

tskisner:
This was installing from a git checkout on edison, where I was manually installing harp, specex, desiutil, desisim, specter, and desispec.  So I guess the usual "python setup.py install --prefix=" command is not intended to be used like this?

weaverba137:
Ah, OK.  All software DESI on NERSC should *only* be installed by desiInstall.  Can you explain further why you needed to manually install it?  This might be a use case that we've missed.

tskisner:
ok, I guess this is just user error on my part.  I was not aware that I had to use desiutil / desiInstall to install a working software stack, and thought I could manually install development versions directly from the git checkouts.

Is it possible to do a "python setup.py develop --prefix=<blah>" to install links back to the live git checkouts?  Basically I'm trying to find the best workflow for "active hacking" on the packages.  If develop mode is supported, then I can just use that for this kind of workflow, and use desiInstall when installing static versions of things.

weaverba137:
OK, again, just trying to establish the use case here.  At NERSC we already have git checkouts of desispec on cori and edison.  These can be accessed with `module load desispec/master`.  You should double-check, but I don't think anyone is actively using them, so you could probably do a git pull or git checkout in those.

However, more broadly, I think you may have stumbled on a dirty little secret, which is that neither pip nor `python setup.py install` will install directories *outside* of the package it is installing.  This is why all Python developers recommend putting data files *inside* the package directory.  In desispec's case it would involve moving the data directory to py/desispec/data.  The data can then be accessed using the `pkg_resources` module.

The reason that this is a dirty secret is that even though we have supported pip installs of DESI packages for work on laptops & other systems (*i.e.* anywhere that is not NERSC) for some time, I strongly suspect that nobody has ever actually used this feature very much.  Instead, most people, myself included, run code out of git checkouts, & with *those* the data directory is right there.

[...]

PR #134, opened 2016-03-04
prefer the DESISPEC environment variable

weaverba137:
This PR allows using the `DESISPEC` environment variable to set the path to the data directory, instead of relying on a semi-hard-coded path.  This addresses some of the concerns in #133.

### DIFF
--- a/py/desispec/bootcalib.py
+++ b/py/desispec/bootcalib.py
@@ -19,6 +19,7 @@ import yaml
 import glob
 import math
 import time
+import os
 
 from astropy.modeling import models, fitting
 from astropy.stats import sigma_clip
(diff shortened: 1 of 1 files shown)


weaverba137:
PS, this PR is only intended as a temporary solution.   The real solution is to move the data files into the package directory itself, *i.e.* `py/desispec/data`, and then to access the files using the `pkg_resources` module.

moustakas:
```desisim.templates.EMSpectrum``` relies on the data files in desisim/data and I get to them using the ```pkg_resources.resource_filename``` method:
  https://github.com/desihub/desisim/blob/master/py/desisim/templates.py#L133-L138
I had never used this package before but hopefully it's the right idea (i.e., a stable solution).

PR #139, opened 2016-03-14
Pipeline automation

tskisner:
This work is the first implementation of automating the pipeline.  There are MPI wrappers used for the most expensive steps, and NERSC slurm scripts are generated for nearly all steps.  There are still some small issues to work on, as well as expanding unit tests and documentation.  However it would be good to get this large work back into master before making smaller branches to address specific issues.


sbailey:
For the record, this PR includes changes to a lot more files than one would naively expect because we were also fixing bugs that we discovered while automating the pipeline, in particular to how fibermap information was propagated and used for spectrograph > 0 (lots of bugs relating to whether spectrograph 1 fiber indices begin at 0 or 500).  Merging this large PR now.

PR #140, opened 2016-03-15
Fiberflat

sbailey:
This PR from Julien Guy improves the robustness and performance of the fiberflat code in the presence of cosmics.  Small tests work; getting this PR going while queueing up larger batch tests for robustness.


julienguy:
validated on dc3a-lite sims

Issue #144, opened 2016-03-26
parallelize redmonster wrapper


sbailey:
Background: redmonster isn't parallelized and as a result some test bricks can't finish within the 30 minute NERSC debug queue time limit, which makes it a pain to debug the pipeline end-to-end.

To do: update desispec.zfind.redmonster.RedMonsterZfind to wrap parallelism around redmonster using the python subprocess module.  This involves splitting the inputs into N subsets of spectra, processing each independently through redmonster, and then stitching the results back together.

Alternative: fork redmonster, parallelize it, and submit a pull request to redmonster itself on bitbucket.  Check with @timahutchinson and Adam Bolton first to make sure they would accept such a PR.

moustakas:
Done in #150.

tskisner:
Actually, in the current pipeline, we use MPI to parallelize the calls to redmonster.  See the main function here:

https://github.com/desihub/desispec/blob/master/py/desispec/scripts/zfind.py

For many productions (depends on the number of bricks) we can fit zfind into the debug queue.

moustakas:
Thanks for the added info -- I was just cleaning up stale issues!

PR #150, opened 2016-04-07
Parallelize desi zfind

cballand:
This PR modifies desi_zfind.py to allow for multiprocessing. This follows from issue #144 by @sbailey. It uses Pull.map() from the python multiprocessing package. The number of cores to be used is passed as an option in the command line. 
Test: fitting 150 ELG spectra bricks using desi_zfind.py with option '--objtype elg' on cori@nersc:
ncores = 1: 27 min.
ncores = 32: 4 min.

### DIFF
--- a/bin/desi_zfind.py
+++ b/bin/desi_zfind.py
@@ -8,12 +8,13 @@ Fit redshifts and classifications on DESI bricks
 import sys
 import os
 import numpy as np
+import multiprocessing
 
 from desispec import io
 from desispec.interpolation import resample_flux
 from desispec.log import get_logger
 from desispec.zfind.redmonster import RedMonsterZfind
-from desispec.io.qa import load_qa_brick, write_qa_brick
+from desispec.zfind import ZfindBase
 
 import optparse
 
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks!  I am super excited to see this PR since this was the last step that didn't fit within the debug queue 30 minute limit for a typical bricksize.  I committed an update that rebundles the targets for the Pool.map() so that redmonster will be called only once per cpu rather than once per target, so that it doesn't keep reloading the large template files over and over.

However, I just realized that I didn't update the repacking of the redmonster results into the final output files.  Don't merge this until I fix that (tonight or tomorrow morning).

sbailey:
I fixed the packing of the results, which I had broken in the previous push.  I also made the default to use parallelism and renamed --ncpu to --nproc to be a more accurate option name: we are controlling the number of parallel processes, which could be more or less than the number of physical cores used (which is also different than the number of CPUs, each of which has multiple cores).  Full disclosure: desisim/bin/pixsim-desi has a --ncpu option (not --nproc) that I probably introduced.  Mea culpa.

Please take a look and let me know if you have any complaints or suggestions before the final merge.  Merging this will also mean that we need to update the pipeline wrapper script, but that is worth delaying the pipeline tag.

PR #155, opened 2016-04-08
Run desi_zfind with one process per node.

tskisner:
Now that it uses multiprocessing, run one desi_zfind per node when generating slurm scripts.


sbailey:
Oops.  I added a call to find_bricks(proddir) to count the number of bricks, forgetting that this is in desi_pipe and those brick directories don't exist yet.  Fortunately I had also written some code for parsing bricknames from the fibermap files; one moment while I switch to that...

sbailey:
Verbally discussed with @tskisner who approved this change.  merging.

PR #156, opened 2016-04-09
Preproc

sbailey:
work in progress — feel free to review and comment but don't merge yet.

This introduces basic preprocessing code

    image = desispec.preproc.preproc(rawimage, header, bias=False, pixflat=False, mask=False)

following the raw data format described in [DESI-1229](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=1229).

This includes the following steps:
  * bias image subtraction
  * overscan subtraction (from BIASSEC* keyword defined regions)
  * readnoise estimation (from BIASSEC* keyword defined regions)
  * comparison to expected readnoise in RDNOISE* keywords
  * gain correction (from GAIN* keywords)
  * pixel flat correction
  * cosmic ray masking
  * propagation of input known bad pixel mask
  * inverse variance estimation

i.e. image = ((rawimage-bias-overscan)*gain)/pixflat

For bias, pixflat, and mask, it currently accepts input arrays or input filenames, but doesn't yet have the ability to lookup default calibration files for each frame for a given night.  If any of these are False it will skip that step.

To do before merging:
  * Check coverage, add more tests
  * Add desispec.io.read_raw(filename, camera)

Creating a repo for default calibration files and updating this to look those up will be a future PR.


py/desispec/preproc.py
```diff
@@ -0,0 +1,196 @@
...
+        - pixel flat correction
+        - cosmic ray masking
+        - propagation of input known bad pixel mask
+        - inverse variance estimation
+        
+    Notes:
+    
+    The bias image is subtracted before any other calculation to remove any
+    non-uniformities in the overscan regions prior to calculating overscan
+    levels and readnoise.
```
moustakas: Is this really the right thing to do? I've typically always modeled the overscan region (e.g., median average across rows or columns), subtracted from everything, including the bias frame, and then bias-subtracted.  Here you're taking the overscan level to be a scalar.

py/desispec/preproc.py
```diff
@@ -0,0 +1,196 @@
...
+    if pixflat:
+        if pixflat is True:
+            pixflat = desispec.io.preproc.read_pixflat(camera=camera, dateobs=dateobs)
+        elif isinstance(pixflat, (str, unicode)):
+            pixflat = desispec.io.preproc.read_pixflat(filename=pixflat)
+
+        if pixflat and pixflat.shape != image.shape:
+            raise ValueError('shape mismatch pixflat {} != image {}'.format(pixflat.shape, image.shape))
+
+        image /= pixflat
```
moustakas: Do you need to protect against divide-by-zero?

py/desispec/preproc.py
sbailey: I based this on BOSS / idlspec2d, which models the 2D shape of the bias per amplifier and subtracts that prior to calculating the overscan as a remaining constant.  If the bias shape is constant in time this allows a higher statistics estimate of the remaining overscan offset vs. the noisier per-row per-exposure estimate.  See idlspec2d [bolton_biasgen.pro](http://www.sdss3.org/svn/repo/idlspec2d/trunk/pro/spec2d/bolton_biasgen.pro) and [bolton_biassub.pro](http://www.sdss3.org/svn/repo/idlspec2d/trunk/pro/spec2d/bolton_biassub.pro) (NB: these are sdss3.org links which someday may go away; I haven't found the equivalent public sdss.org link where idlspec2d now lives).  Example bias images are at http://www.sdss3.org/svn/repo/specflat/trunk/biases/ .  This method of subtracting the bias was implemented by Adam Bolton so we can ping him for details about the motivation behind this; David Schlegel @djschlegel probably knows too.

We did occasionally have electronics problems that caused time-variable biases.  These were handled as special cases for amplfiers and date ranges in the preprocessing code, falling back to row-by-row overscan subtraction as you describe.  For BOSS this was usually discovered by the quicklook equivalent as apparently large readnoise (taking the RMS of a non-flat 2D distribution).  If we use this method for DESI, we should add specific tests for non-flatness of the bias-subtracted overscan region.  For BOSS we chose *not* to automatically fall back to row-by-row overscan subtraction, since this was a symptom of actually needing humans to do something with the electronics and we wanted it to mess things up and make us notice.

dkirkby:
Is there any way to turn off email notifications from coveralls on pull requests?

weaverba137:
I think I've got them turned off now.

sbailey:
I believe this PR is ready for merging now.  @moustakas could you take a final look?  @julienguy is on vacation this week.  Items to do in a future PR:
* have desisim / pixsim-desi optionally write the multi-HDU raw data format using the desispec.io.write_raw function included here
* add placeholder bias, mask, and pixflat files to svn (this PR has hooks to use them, but not actual example files)

After we have real teststand data, we can evaluate the stability of the overscan regions and whether the BOSS-like pre-overscan bias images can be used or whether we need to do the more classic row-by-row overscan subtraction.  DESI uses the same CCDs as BOSS for r and z, though the electronics will be completely different.

py/desispec/preproc.py
```diff
@@ -0,0 +1,275 @@
...
+
+        if pixflat.shape != image.shape:
+            raise ValueError('shape mismatch pixflat {} != image {}'.format(pixflat.shape, image.shape))
+
+        if np.all(pixflat != 0.0):
+            image /= pixflat
+            readnoise /= pixflat
+        else:
+            good = (pixflat != 0.0)
+            image[good] /= pixflat[good]
```
moustakas: I haven't tracked down the piece of code that builds the pixel flats, but will that code guarantee that the pixel values in the flat are never negative?  That's the only case not covered here.

py/desispec/preproc.py
```diff
@@ -0,0 +1,275 @@
...
+            readnoise[good] /= pixflat[good]
+            mask[~good] |= ccdmask.PIXFLATZERO
+            
+        lowpixflat = (0 < pixflat) & (pixflat < 0.1)
+        if np.any(lowpixflat):
+            mask[lowpixflat] |= ccdmask.PIXFLATLOW
+
+    #- Inverse variance, estimated directly from the data (BEWARE: biased!)
+    var = image.clip(0) + readnoise**2
+    ivar = 1.0 / var
```
moustakas: This will result in a divide-by-zero for negative pixel values in the image.  One alternative I've used before in lieu of ```np.clip()``` is to do something like ```var = np.sqrt(image**2) + rednoise**2```.

moustakas:
I obviously can't check this any further without being able to run it on data, but from reading the code everything looks good (modulo my two previous comments). @sbailey you should feel free to merge once you've decided what to do about my comments, if anything.

py/desispec/preproc.py
sbailey: the pixflat generation code doesn't yet exist for DESI (it exists for BOSS, but with somewhat specific BOSS tuning).  That code should enforce that pixflat values are positive, but the belt-and-suspenders approach here could be replacing `good = (pixflat != 0.0)` with
```
Does that seem reasonable to you?
if np.any(pixflat < 0.0):
    log.warn('negative pixflat values')
good = (pixflat > 0.0)
```

py/desispec/preproc.py
sbailey: I don't think it is enforced, but `readnoise` should always be > 0 such that `var` should have a minimum value of `readnoise**2` -- i.e. I don't think we are in danger of a divide-by-zero.  Note: the current line is `var = image.clip(0) + readnoise**2`, not `var = (image + readnoise**2).clip(0)` which could result in zeros if the image had large negative pixels.

I'm thinking about the statistical question about whether negative pixel values should contribute to the variance model or not.  We're in questionable statistical territory either way...

moustakas: Yes, of course, I misread the code.

PR #159, opened 2016-04-18
Quicklook Update

gdhungana:
Still working on the pipeline but sufficient update for this PR. This PR updates the boxcar to be using the bootcalib psf, as well as specter.psf.PSF. A similar psf interface is added. This only works for CCD x/y centroids and the wavelength solution. No resolution accounted yet.

PR #160, opened 2016-04-19
quicklook with bootcalib psf

sbailey:
I don't understand why the original PR didn't work, but this is a replacement for PR #159.  GitHub claimed there was a conflict between ghungana:quicklook and desihub:quicklook that had to be resolved at the command line, but fetching both of those and merging didn't show any conflicts.  This PR contains the changes in gdhungana:quicklook that are not yet in desihub:quicklook.  I will add some additional git-fu comments to #159 and close that out, and review this one for merging.

The original PR comment:

> Still working on the pipeline but sufficient update for this PR. This PR updates the boxcar to be using the bootcalib psf, as well as specter.psf.PSF. A similar psf interface is added. This only works for CCD x/y centroids and the wavelength solution. No resolution accounted yet.

### DIFF
--- a/bin/desi_quicklook
+++ b/bin/desi_quicklook
@@ -13,14 +13,14 @@ from __future__ import absolute_import, division, print_function
 
 from desispec.quicklook import quicklook,qllogger
 import os,sys
-import yaml,pickle
+import yaml
 
 import optparse as op
 
 def ql_main():
 
     p = op.OptionParser(usage = "%")
-    p.add_option("-c", "--config_file", type=str, help="yaml/pickle file containing config dictionary",dest="config")
+    p.add_option("-c", "--config_file", type=str, help="yaml file containing config dictionary",dest="config")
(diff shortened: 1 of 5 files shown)


sbailey:
I'll merge this now, but some future cleanup requests:

* desispec/bin/desi_bootcalib.py should be refactored into a lightweight
  script wrappering an algorithm, such that quicklook.BootCalibration.do_bootcalib can call that
  algorithm instead of having to replicate much of that script
* `BoxcarExtraction.run(*args, **kwargs)` obscures what keyword names are actually required.
  Better to explicitly have `run(psffile, boxwidth, dw, nspec)` (with optional defaults).
* Test coverage went down on this PR

gdhungana:
@sbailey On your first request, where should the wrapper algorithm go? I am assuming inside desispec.bootcalib, right? The desispec/bin/desi_bootcalib or quicklook.BootCalibration.do_bootcalib can call that.

PR #162, opened 2016-04-20
QA: ztruth

profxj:
Adds a script that runs on a given night to perform QA on
redshifts using the 'truth' in simpix files.

No tests.

Will integrate QA figures from z1dc soon.

### DIFF
--- a/bin/desi_night_ztruth.py
+++ b/bin/desi_night_ztruth.py
@@ -70,8 +70,8 @@ def main():
     meta = dict(SIMSPECV='9.999', PRODNAME=os.getenv('PRODNAME'))
 
     # Run stats
+    summ_dict = dsqa_z.summ_stats(simz_tab)
     if args.qafile is not None:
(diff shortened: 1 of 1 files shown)


sbailey:
Somewhat distracted by a review today, but my initial reaction is that this should go in desisim instead of desispec to keep a sharp boundary between truth/sims vs. real data processing, and also to avoid having desispec and desisim both depend upon each other (generally a dependency no-no).  i.e. desisim depends upon desispec, but desispec shouldn't depend upon desisim.

For convenience, we probably do want an option to the pipeline to optionally run the final "compare against truth for QA" step, but that should be limited to running a script or function in desisim, with all the real truth-comparing code in desisim itself.

This does have the disadvantage of splitting the QA code apart, but I think "QA that can be done using only real data" and "QA that can be done using truth information" is a fairly natural split.

profxj:
Ok, am closing this PR and pushing this script to desisim.

PR #163, opened 2016-04-21
use mask in ivar of images

julienguy:
One line change in desi_extract_spectra.py to have images ivar=0 for mask!=0. This reduces considerably the cosmics spikes in the frames. Tested on expid 6 , spectro 1 of dc3c. It would be interesting to rerun the whole dc3c with this fix.


sbailey:
d'oh!  Of course.  I thought I was already doing that but missed it.  Thanks for this fix.  merging now.

PR #165, opened 2016-04-25
Changed desi_zfind.py to accept less filters

ngbusca:
I've minimally modified desi_zfind.py to use as many filters as given as arguments, instead of requiring  "b", "r", and "z". This allows the user to run on the bricks produced from BOSS data, which have only "b" and "r".

### DIFF
--- a/bin/desi_zfind.py
+++ b/bin/desi_zfind.py
@@ -8,22 +8,33 @@ Fit redshifts and classifications on DESI bricks
 import sys
 import os
 import numpy as np
+import multiprocessing
 
 from desispec import io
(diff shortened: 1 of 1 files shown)


sbailey:
A lot more than just filters changed in this pull request.  Parallelism and QA code disappeared.  We should not merge this as-is.  @ngbusca please ping me or a local git expert if this was the result of a git merge gone wrong, and either update this PR or open a new one with just the minimal changes needed to auto-discover the available filters.  Thanks.

https://github.com/desihub/desispec/pull/165/files

ngbusca:
@sbailey wooooops. It should be fixed now...

Issue #65, opened 2015-09-16
scripts shouldn't end in .py


sbailey:
This proposed change is embarrassingly minor, but it could create confusion for anyone trying to use pipeline code, so advertising it via an issue prior to implementation:

Our coding guidelines say that command line scripts should *not* end with .py (code you import ends in .py; scripts you run do not).  I'd like to fix this for all our scripts prior to documenting the next major pipeline production and prior to writing the pipeline wrapper code that will be calling all these scripts.

weaverba137:
Since we're also thinking about how to increase the test coverage of scripts, now is a good time to think about moving all the actual code into the desispec package, and only having the thinnest possible wrapper as the script itself.

dkirkby:
One way to encourage / enforce this is to use the entry_points mechanism in setup.py, or are you already doing this?

weaverba137:
I strongly endorse the entry_points mechanism, but we haven't implemented that yet.

sbailey:
For the most part the algorithms and I/O are already factored out into
packages and testable and reusable independent of the script.  The
difficulty with testing the scripts isn't whether they live in bin/ or
whether they live in a package with a main() function that starts with
argument parsing.  The tricky part is setting up appropriate input files
that are light weight enough for testing but still realistic.  Moving the
code into a package or using entry_points doesn't help with that issue at
all.

I find that entry points tend to obfuscate which scripts exist and what
those scripts do.    I advocate for scripts like

[parse args]
data = io.read_something(args.input)
results = some_algorithm(data)
io.write_outputs(args.output, results)

the algorithms and I/O are still independently usable and testable, the
script is lightweight but still shows the basic concept of what it does.

Regards,

Stephen

moustakas:
+1

weaverba137:
Responding to: 

> The difficulty with testing the scripts isn't whether they live in bin/ or whether they live in a package with a main() function that starts with argument parsing.  The tricky part is setting up appropriate input files that are light weight enough for testing but still realistic.  Moving the code into a package ...

That is actually very easy to do if you move *both* the scripts *and* the test files into the package.  There are many examples of this in the astropy packages, and you can see simple example in the desiutil and desidatamodel packages as well.

sbailey:
Agreed.  The only issue is that the naive version of some of the input files are several GB of data which we don't want to keep in the repo.  So there is work to be done to trim those into tiny files that are viable to keep in the repo but are also realistic enough to test the codes.

[...]

PR #166, opened 2016-04-26
Reorganization of code to move main() for all scripts into the package.

tskisner:
This pull request is for moving the main() functions for scripts into the package, as discussed in #65 .  Later, when running the pipeline, options for each step will be parsed from yaml and turned into a list of strings, which can then be passed to main().  This way we have all the option parsing in one place.


weaverba137:
I thought the reason for splitting out command-line parsing into a separate function was so we could run unit tests on it.  Is there a reason you haven't done that here?

tskisner:
we want to call main() from 3 places:  a tiny wrapper script getting options from sys.argv, a pipeline running function reading options from yaml, and in the unittests with either fixed options or also read from some default yaml.  That was the motivation behind putting the main into the package.  So that we don't have to use subprocess to call it.  Do you think it will be too painful to make a list of strings (fake sys.argv) from within the unit test when calling main()?

tskisner:
ah- I think you meant "why haven't I implemented the unit test here".  I could do that, but did not want to go that far until @sbailey was happy with the general plan...

weaverba137:
In fact there is are examples of unit testing an `argparse` wrapper function in desiutil.  See https://github.com/desihub/desiutil/blob/master/py/desiutil/test/test_install.py

tskisner:
sure, so just to compare the 2 approaches:

1. (this code)
-  main takes array of strings (sys.argv or other), parses args, runs code.
-  script calls main with sys.argv
-  unittest creates fake list of strings, calls main.
-  pipeline runner parses yaml, creates list of strings, calls main.

2. (desiutil approach)
-  main class has method to parse args and return an argparse Namespace (this could obviously also be 2 separate functions, one to parse, and a main which takes a Namespace).
-  script calls parse function with sys.argv, passes the returned Namespace to main()
-  unittest calls parse function with custom strings, passes Namespace to main()
-  pipeline runner calls parse function with custom strings from yaml, then calls main()

I think these are basically equivalent, right?  Every use case is calling parse and then main, so we just move the parsing into main.  And the "thing" which we would have passed to the parse function is a list of strings, which we now pass to main...

Is there a use case where passing around a Namespace object is preferable?

weaverba137:
You will always be creating a Namespace object anyway, so you might as well use it.  It makes writing tests very easy, because all the options have a name!

tskisner:
ok, that is a valid point.  will update this pull request with that change.

tskisner:
ok, updated to pass around Namespace objects.  I also added a unit test for bootcalib.main().

sbailey:
Thanks for humoring me with a tangible example.  Every case here starts with a list of strings and then uses `parse` and `main` as a pair, e.g.
```
argstr = ['-a', '1', '-b', '2']
args = parse(argstr)
main(args)
```
My first reaction was to favor how Ted originally coded it, with the list of option strings passed directly into `main`, but I'm not sure that is such a big deal.  Ben's desiutil test code shows one case where having an intermediate object is useful so that it can be manipulated to change the value of an option before calling `main` again.

As a cautionary note, we are in danger of wandering off into cool-coding land and alienating typical astrocoders who expect minimal fuss between their algorithms and their scripts.  Once you understand it, this refactor does make it cleaner to keep algorithm code, command line scripts scripts, parallelism wrappers, and tests all in sync, but I'm not sure it achieves the "minimal fuss" metric.  Let's proceed with this refactor, but I'm not sure we're out of the woods yet.  At minimum we need some code guidelines documentation about the underlying motivations for this method.

moustakas:
From my perspective as a relatively new Python coder---and recognizing the need for the algorithms to be accessed several different ways---I think I like how this has been refactored.  (Although one point of confusion is that a few comments back, @tskisner, you said that you switched the code to use the Namespace object explicitly, but I'm not seeing it in the code that has been checked in.)

My primary concern is that because the parse() and main() functions have been moved out of the executable script it's harder to (quickly) track down what that script *does*, particularly with respect to the command-line options.  We can mitigate this issue by 1) having good documentation in the bin/ script itself as well as when the script is called without any arguments; and 2) moving the parse() and main() functions to the *top* of the file in the relevant module (in this case desispec.bootcalib).  Although if we satisfy 1) we need to be careful to not have duplicated documentation (both in desispec.parse and in bin/desi_bootcalib) that ends up getting out of sync.

My last comment is that we should only do this refactor for scripts that are on the critical path.  For example, it's still convenient to have easily readable top-level scripts that simply parse command-line arguments and do their main thing: https://github.com/desihub/desisim/blob/master/bin/simulate-templates

tskisner:
Hi @moustakas, the Namespace class is the thing returned by the "parse_args" function inside the bootcalib.parse function.  It is the thing that is passed to main.  I think it is fine to put top-level docstrings in the script.  However, the "help" strings for each command line argument have to go inside the parse function when we add each option to the parser.

Also, I think we could limit this refactor to the functions in desispec that need to be called by the pipeline.  No need to do this to the desisim repo for example.  Although it does help to run the "script" as part of a unit test.

profxj:
I'm rather fond of this refactor.  The desi modules are getting longer, but I'd prefer that to having to maintain multiple modules and scripts.

Unlike John, I prefer the parse and main to remain at the end of the module.  It is where I expect to find main() and it is good to have parse right there with it.  I should add that you've probably convinced me to refactor a few of my own scripts to follow this (i.e. strip the parser out of main).

weaverba137:
@moustakas brings up a valid point, but there are two standard solutions to this problem, *both* of which should be used.

1. The command-line help functions provided by `argparse.ArgumentParser` need to be employed to their fullest, so that, *e.g.* `desi_bootcalib --help` is as helpful as it can be.
2. Write up more extensive algorithm and interface-level documentation in the doc directory.  For example, the file `doc/bootcalib.rst` (which doesn't exist yet) should describe in great detail what `desi_bootcalib` does, with references to the individual functions called.

I also want to address @sbailey's point about "cool-coding".  We're not doing anything "cool" here, we're following a standard practice, which you can see in astropy and astropy-affiliated packages.  As the manager in charge of code oversight, it is my responsibility to exhort us to follow established standards, and I am doing so right now.

sbailey:
Thanks for the comments, everyone.  Full steam ahead with the refactor.  We'll need to keep an eye on the documentation issue that @moustakas raised.  `desi_bootcalib --help` still works for documenting the options to the script at the command line, but we've lost the docstring/introspection (e.g. from pydoc or ipython) way of seeing what the options to `main()` should be -- the interface to the `main()` function is now effectively in the code of the `parse()` function.

I agree with Ben's point about including more doc/*.rst documentation, but that is in addition to the lower-level API docstrings, not in replacement of them.

tskisner:
I've renamed this pull request to not be temporary.  I'll continue work on this branch, updating all the packages.  One note on the recent commit- to avoid the "algorithm" and "I/O" packages from having to import each other (circular reference), I've adopted the astropy-ish technique of putting the main and parse functions into a "scripts" subpackage.  Hopefully this is not offensive to anyone.  I'll keep moving forward in this direction.

tskisner:
Stephen suggests that we merge this work so far, so that new code can be added in the same spirit.  I'll continue work on this branch converting other scripts.

sbailey:
I fixed some bugs found with the integration test.  Merging now, but leaving the branch open for Ted to continue the refactoring work.

PR #167, opened 2016-04-29
Finish reorganization of scripts

### LINKED ISSUE #65: scripts shouldn't end in .py

tskisner:
This is the remainder of the changes.  In hindsight, we should have closed this branch after the previous pull request merge and opened a new one.  The history will be somewhat messy now.  This also closes #65


sbailey:
There are enough changes that I haven't reviewed every file in detail, but tests pass including the integration test, so I'll merge now, close this branch, and we can make further updates in other pull requests from branches off of the new master.

PR #169, opened 2016-05-01
Pipeline work in progress

tskisner:
This pull request adds some new functions, but does not break anything or enable anything new.  This is the first of several pull requests to merge pipeline changes in an incremental way, in order to keep development better in-sync with master.


sbailey:
We'll eventually need a cleanup pass that adds docstrings and other high-level documentation/examples so that others can understand how to use this, but I won't let that stand in the way of incremental PR and merges for now as long as tests continue to pass.

tskisner:
Sure, I will make sure documentation is in place before the code is turned on / enabled.

PR #175, opened 2016-05-04
Merging modifications of bootcalib developed in teststand branch

julienguy:
The changes are 

 * additional options to desi_bootcalib   (default is not changed)

  --lamps LAMPS         comma-separated used lamp elements, ex: HgI,NeI,ArI,CdI,KrI
(choice of lamp lines different from the ones used by default in desisim)
  --triplet-matching    use triplet matching method for line identification (slower but expected more robust)
  --ntrack NTRACK       Number of solutions to be tracked (only used with triplet-matching, more is safer but slower)

* additional of several NIST spectra to the data/arc_lines/ directory.

I checked this version runs on both standard sim and teststand dedicated sims (all of this with wavelength in vacuum).


sbailey:
Test failed with:
```
FAIL: test_cholesky_invert (desispec.test.test_linalg.TestLinalg)
  File "/home/travis/build/desihub/desispec/py/desispec/test/test_linalg.py", line 74, in test_cholesky_invert
    self.assertAlmostEqual(Id[0,1],0.)
AssertionError: 5.9604644775390625e-08 != 0.0 within 7 places
```
This involves code that wasn't changed with this PR and it has passed many many times before, so it is probably just bad luck with the random number matrix being nearly singular.  I reran the Travis tests and is passes.  If this comes up again we can investigate more.

sbailey:
I'm a little nervous about all the hardcoded wavelengths in the code itself; we should consider refactoring for more flexibility in the future.  OK to merge for now if both Julien and Xavier are statisfied with the current state.

julienguy:
I agree with you. That's the next thing I wanted to change but there are other improvements needed so I suggest we merge and restart a new branch (as discussed in telecon).

julienguy:
desi_bootcalib --triplet-matching is successful for all fibers, b,r,z, in 10 spectros except for one in z due to a spurious bright column (due to some yet unidentified error in the pixel sims). We may merge this.

weaverba137:
Is there a reason the teststand branch still exists?

julienguy:
I delete this branch for now. Might open a new one with open issues when then come

PR #177, opened 2016-05-06
desi_preproc

sbailey:
Adds desi_preproc script and bug fixes/features needed to support raw data preprocessing.
```
usage: desi_preproc [-h] --infile INFILE [--outdir OUTDIR] [--pixfile PIXFILE]
                    [--cameras CAMERAS] [--bias BIAS] [--pixflat PIXFLAT]
                    [--mask MASK]

Preprocess DESI raw data

optional arguments:
  -h, --help         show this help message and exit
  --infile INFILE    path of DESI raw data file
  --outdir OUTDIR    output directory
  --pixfile PIXFILE  output preprocessed pixfile
  --cameras CAMERAS  comma separated list of cameras
  --bias BIAS        bias image calibration file
  --pixflat PIXFLAT  pixflat image calibration file
  --mask MASK        mask image calibration file

By default, all HDUs of the input file are processed and written to pix*.fits
files in the current directory; use --outdir to override the output directory
location. --pixfile PIXFILE can be used to override a single output filename
but if only a single camera is given with --cameras. --bias/--pixflat/--mask
can specify the calibration files to use, but also only if a single camera is
specified.
```

Updates header keywords to match the [DESI-1964](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=1964) description which supersedes [DESI-1229](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=1964).  Amps are now named 1-4 instead of A-D.  Fixing DocDB descriptions to be consistent is in the works.  I'm sure that we'll need to update again after we actually have real raw data.

The changes in this PR are needed for a pending companion PR from desisim that will actually simulate the fully raw data.

### DIFF
--- a/bin/desi_preproc
+++ b/bin/desi_preproc
@@ -1,0 +1,8 @@
+#!/usr/bin/env python
+
+'''
+Preprocess a DESI raw data exposure, outputing individual pix files
+'''
+
+from desispec.scripts import preproc
+preproc.main(preproc.parse())
--- a/py/desispec/image.py
+++ b/py/desispec/image.py
@@ -13,9 +13,9 @@ class Image(object):
         
         Args:
             pix : 2D numpy.ndarray of image pixels
-            ivar : inverse variance of pix, same shape as pix
             
         Optional:
+            ivar : inverse variance of pix, same shape as pix
             mask : 0 is good, non-0 is bad; default is (ivar==0)
             readnoise : CCD readout noise in electrons/pixel (float)
             camera : e.g. 'b0', 'r1', 'z9'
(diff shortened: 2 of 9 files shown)


julienguy:
One minor comment :  preproc._overscan(pix, sigma=5, niter=3)
   I find the parameter name 'sigma' counterintuitive, it's generally naming the standard deviation whereas here it means the number of standard deviations. I'd rather call it nsigma or nsig.

For the rest, we cannot really check this without the sims but you say this should be merged before the sims so let's merge this and look at the preprocessed sims when ready.

PR #194, opened 2016-05-18
Update travis environment to include harp, specex, and specter.

tskisner:
The merge of new pipeline code (and improved tests) broke the travis build, since those tests were more realistic and required the real pipeline dependencies (harp, specex, and specter).  I caused this breakage- mea culpa.  This pull request appends the needed dependencies to the travis environment, and also switches to using the newer ubuntu 14.04 with a newer gcc that better supports C++11.  The build of HARP is quite slow.  I can work on making an ubuntu package to speed up the linux travis build, but the OS X install is more challenging to package.

The travis tests now take approximately 20 minutes to complete.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ os:
     # - osx
 
 # Setting sudo to false opts in to Travis-CI container-based builds.
-sudo: false
+sudo: required
+dist: trusty
 
 # The apt packages below are needed for sphinx builds, which can no longer
 # be installed with sudo apt-get.
@@ -27,6 +28,10 @@ addons:
             - graphviz
             - texlive-latex-extra
             - dvipng
+            - libblas-dev
+            - liblapack-dev
+            - libboost-all-dev
+            - libcfitsio3-dev
 # python:
     # - 2.6
     # - 2.7
@@ -44,6 +49,9 @@ env:
         - SPHINX_VERSION=1.3
         - DESIUTIL_VERSION=1.3.4
         - SPECLITE_VERSION=0.4
+        - SPECTER_VERSION=0.4.1
+        - HARP_VERSION=1.0.1
+        - SPECEX_VERSION=0.3.9
         - CONDA_INSTALL='conda install -c astropy-ci-extras --yes'
         - PIP_INSTALL='pip install'
         - OPTIONAL_DEPS=false
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,13 @@ os:
     # - osx
 
 # Setting sudo to false opts in to Travis-CI container-based builds.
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -17,7 +17,8 @@ os:
     # - osx
 
 # Setting sudo to false opts in to Travis-CI container-based builds.
-sudo: false
+sudo: required
```
weaverba137: Is `sudo: required` required in order to choose a different distribution?

tskisner:
Yes, in order to get ubuntu trusty (14.04).  I got that from the figure here:  https://docs.travis-ci.com/user/ci-environment/

weaverba137:
OK, thanx for that table.  Could you add a timestamp to the end of the install process and to the end of the test process?  This will allow us to see how much of that ~20 minutes is compiling versus actually running the test.

tskisner:
I added the date stamps, but travis logs already show the time spent running travis_env_[OS].sh.  See for example:

https://travis-ci.org/desihub/desispec/jobs/131408973

where you can see that the install script took 1090 seconds and the test command took 62 seconds.

sbailey:
I am confused about this test dependency.  I don't have HARP or specex installed on my laptop, yet I am able to run the desispec unit tests for this branch (`python setup.py test`).  Why are HARP and specex needed for the travis tests?

weaverba137:
@sbailey that's a very good point!  The only new dependency that the new unit tests *definitely* introduced was on specter, but that's a pip-installable Python package.

tskisner:
Ok, fair enough.  However, as soon as we introduce unit tests that call desispec.scripts.specex.run_frame, then we will hit this.  If you like, I could comment out those install commands for now, and we can un-comment them when we need them.  How does that sound?

weaverba137:
Yes, please go ahead.

When I refactor the Travis tests to use astropy/ci-helpers, I'll be sure to leave the install commands in, but deactivated.

sbailey:
Please comment out for now since we don't actually need HARP or specex.

Longer term, I would argue that we should continue to avoid that dependency for the unit tests.  20 minutes is too long to wait for Travis tests to finish, and I don't want to place any other barriers to developers running the unit tests as a matter of habit on their laptops (sure, you only have to install HARP + specex once, but I suspect that Ted and Julien are the only developers who have done that...).

It would be good to have specex tested as part of the integration test, but even then it would be useful to have it gracefully step over that if specex isn't installed.  e.g. a dummy specex executable could be added to the path as part of the test stetup; it would create a link to the desimodel PSFs instead of actually fitting a new PSF.

tskisner:
Ok, should be good to merge now.

Issue #131, opened 2016-03-03
sky model file name does not match datamodel


dkirkby:
The following line in `io.meta.findfile()`:
```
sky = '{specprod_dir}/exposures/{night}/{expid:08d}/sky-{camera}-{expid:08d}.fits',
```
does not match the naming scheme in the datamodel:

https://github.com/desihub/desidatamodel/blob/master/doc/DESI_SPECTRO_REDUX/PRODNAME/exposures/NIGHT/EXPID/skymodel-CAMERA-EXPID.rst

Probably the datamodel should be updated.

weaverba137:
This is fixed in the latest data model.

PR #201, opened 2016-05-27
Rawpixsim

sbailey:
A few changes related to raw pixel simulations:
  * `io.write_raw`: raise ValueError if camera already exists in the output raw data file
  * don't force the CAMERA keyword to uppercase
  * include --preproc in integration test pixsim

Also see [desisim PR #131](https://github.com/desihub/desisim/pull/131)


sbailey:
continuum.io has been having load problems since last night, so tests are failing since they can't download conda.

sbailey:
continuum.io is back; tests pass; merging

PR #202, opened 2016-05-30
Quicklook update

gdhungana:
Several things still on progress but have enough for this PR. This PR updates 

- QAs based on the updated data model
- adds an extraction QA to counts the spectral bins/fiber above from threshold
- adds try-catch block to make a request using HTTP to post to QLF server (not available now)
- adds quicklook.meta which will be useful in setting up the configuration etc. Next update to incorporate this.
- fixes a arcimage.ivar argument bug in bootcalib.bootcalib()


sbailey:
I disagree with the design of how running the quicklook algorithm automatically tries to update results on a quicklook framework server, even if that server is a placeholder that doesn't exist yet.  e.g. if I'm running quicklook on some simulated test dataset on my laptop, it shouldn't try to automatically connect to some master QLF server and upload results.  The algorithms should run and generate local outputs and quit.  If I want those on some server, I should run a separate command to upload the local results to that server.

Has security been considered in the design of the QLF REST interface?  How does it know that I'm a legitimate source of data?

desispec/quicklook/meta.py : this looks like a copy of desispec/io/meta.py

gdhungana:
@sbailey Thanks and I understand what you are saying. May be a configurable boolean will do that. Will configure accordingly. Also I'll bring up the issue of security with the QL-QLF team.
> desispec/quicklook/meta.py : this looks like a copy of desispec/io/meta.py
This is import from desispec/io/meta.py as you said with some modification where I defined the default to "QUICKLOOK_SPECTRO_DATA" instead of "DESI_QUICKLOOK_DATA" etc.. for the directory hierarchy, not to conflict the two but suggestions welcome. But also add bias and dark although those names may not be final. I also don't see the dark frame in the preprocess.  For the psf file, desispec.io.meta seems to have an expid too. Generation of a psf file using bootcalib involves arc and flat image, which are separate expid. What expid should be there?

sbailey:
I'd still prefer the "load results to the server" to be a completely separate step from running the QA algorithm.  If it is an option to the QA algorithm it should be the server URL itself, not just a boolean that can write to one and only one master server.  We certainly don't want the observers to accidentally see our results from running a test at NERSC and think it was real data coming in...

gdhungana:
The idea was to update the QA output to the QLF server once that QA is executed and output is available, without the need to complete the QAs downstream. So the pipeline has to make this request after each QA execution. It is done from run_qa() that calls the relavant QA algorithm, so url or a boolean is not an argument to QA algorithm itself but to the run_qa method. Although I admit that there are few short qa algorithms inside run_qa method as well, which I think can be moved outside and then run_qa can call those QA algorithms from there. So instead of passing the boolean from the configuration, will something like 'url=None' be a default argument for run_qa and it can then try to post the qa output to QLF server if url is not None. Is this something close to what you are suggesting or something different?

gdhungana:
The idea was to update the QA output to the QLF server once that QA is executed and output is available, without the need to complete the QAs downstream. So the pipeline has to make this request after each QA execution. It is done from run_qa() that calls the relavant QA algorithm, so url or a boolean is not an argument to QA algorithm itself but to the run_qa method. Although I admit that there are few short qa algorithms inside run_qa method as well, which I think can be moved outside and then run_qa can call those QA algorithms from there. So instead of passing the boolean from the configuration, will something like 'url=None' be a default argument for run_qa and it can then try to post the qa output to QLF server if url is not None. Is this something close to what you are suggesting or something different?

sbailey:
Yes, url=None is a good default, with internal run_qa code like:
```
if url is not None:
    try: 
        import requests
        response=requests.get(url)
        #- Check if the api has json
        api=response.json()
        #- proceed with post
        response=requests.post(url,json=retval) #- no need of json.dumps as the api has it
    except:
        log.info("Skipping HTTP post...")            
```

Please also remove `quicklook/meta.py`, or strip it down to just what is needed for quicklook itself.  Right now it is mostly a copy of `io/meta.py` .

PR #203, opened 2016-05-31
Desispec pecvel

rstaten:
I have implemented standard star peculiar velocities by allowing a canonical f-type model to shift from -600 km/s to 600 km/s, then performing chi square minimization. It takes longer to run than desirable and there is an offset when comparing fit redshift to true redshift, but I am working to fix this. The attached plot compares the fit redshifts to those from simspec files for 12 different spectra.
[fit_shift.pdf](https://github.com/desihub/desispec/files/292015/fit_shift.pdf)

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -14,7 +14,6 @@ from .io.filters import load_filter
 import scipy, scipy.sparse, scipy.ndimage
 import sys
 from astropy import units
-import speclite.redshift
 
 #rebin spectra into new wavebins. This should be equivalent to desispec.interpolation.resample_flux. So may not be needed here
 #But should move from here anyway.
@@ -65,7 +64,6 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
     def applySmoothingFilter(flux):
         return scipy.ndimage.filters.median_filter(flux,200) # bin range has to be optimized
 
-
     rnorm=flux["r"]/applySmoothingFilter(flux["r"])
     bnorm=flux["b"]/applySmoothingFilter(flux["b"])
     znorm=flux["z"]/applySmoothingFilter(flux["z"])
@@ -75,22 +73,6 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
     bivar=ivar["b"]*(applySmoothingFilter(flux["b"]))**2
     rivar=ivar["r"]*(applySmoothingFilter(flux["r"]))**2
     zivar=ivar["z"]*(applySmoothingFilter(flux["z"]))**2
-
-    Chisq=1e100
-    bestId=-1
-    bchisq=0
-    rchisq=0
-    zchisq=0
-
-    bmodels={}
-    rmodels={}
-    zmodels={}
-    for i,v in enumerate(stdflux):
-        bmodels[i]=rebinSpectra(v,stdwave,wave["b"])
-        rmodels[i]=rebinSpectra(v,stdwave,wave["r"])
-        zmodels[i]=rebinSpectra(v,stdwave,wave["z"])
-
-    Models={"b":bmodels,"r":rmodels,"z":zmodels}
 
     def convolveModel(wave,resolution,flux):
 
@@ -102,6 +84,10 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
 
         return convolved
 
+    Chisq=1e100
+    bchisq=0
+    rchisq=0
+    zchisq=0
     nstd=stdflux.shape[0]
     nstdwave=stdwave.shape[0]
     maxDelta=1e100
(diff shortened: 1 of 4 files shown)


sbailey:
A 1 km/s scan from -700 to +700 km/s is unnecessarily fine a resolution given that you are fitting a chi2 scan minimum.
```
min_index=np.argmin(chisq[100:-100])
fit_x=pec_vel[min_index:min_index+200]
fit_y=chisq[min_index:min_index+200]
```
This only fits one side of the minimum, instead of symmetric about the minimum, e.g.  I think you meant min_index-100:min_index+101.  At the same time, fitting a parabola with ~200 points is overkill, and probably the minimum is much much sharper than that.  A much coarser scan and fitting 3 or 5 points about the minimum is probably sufficient.  Please take a look at chisq vs. pec_vel and inspect the minima to get a sense of what resolution is needed and what range around the minimum is needed.

The Travis tests are failing because it isn't installing/finding specter.  I'm not sure how that happened if this was branched off of a clean version of master, but we can fix that up separately from the peculiar velocity implementation.

tskisner:
WARNING:  looking at the network graph, this branch diverged from master *before* the giant cleanup / mess we had to do manually recovering things.  @sbailey, we should at least be really careful here, and should consider rebasing this branch off current master and force-pushing that before merging.  Obviously we need to communicate that if we are going that route.

weaverba137:
I was about to say the same thing!

moustakas:
This is a great development!  Just a high-level comment:  algorithmically, it might be easier to pull out the same functionality from redmonster, assuming the license allows it, or to just call the low-level routines directly.  The cross-correlation and parabolic chi^2 minimization (of multiple possible minima) is pretty optimal. All you would need is a small number of continuum-normalized basis templates or archetypes.

rstaten:
@sbailey, I changed the scanning to every 10 km/s rather than 1 km/s. This doesn't change the fitted redshift significantly. However, I have kept the fit to the same distance around the minimum (now 20 points rather than 200 due to the change in resolution). I had it at 200 initially based on looking at chi square vs. peculiar velocity, as there are occasionally oscillations around the minimum that cause the fit to be skewed if not enough points are included. Also, the reason for having min_index+200 is because of the line before (min_index=np.argmin(chisq[100:-100])). I find the minimum chi square from -600 km/s to 600 km/s, but in order to be able to apply a fit to velocities approaching this limit, I extend the overall chi square region to -700 km/s to 700 km/s. Thus, min_index is found for the smaller chi square array, so when applied to the full chi square array, it must be [min_index:min_index+200] rather than [min_index-100:min_index+100] in order to keep the fit symmetric about the minimum.

rstaten:
Also, if it will be beneficial to start a new branch I can certainly do that and just make changes there.

sbailey:
Thanks for clarifying the indexing.  That makes sense.  Please post a few plots of example chi2 vs. pecvel with the fit over the range of the points used.

Given the state of the git branch merging mess, please start a new branch off of current master and re-apply your updates there and open a new PR (with the plots there).  Thanks.

tskisner:
No need to start a new branch!

If you (and maybe Stephen) are the only ones with a local checkout of this, you could do from your local checkout:

git fetch
git checkout master
git rebase origin/master
git checkout desispec_pecvel
git rebase master
(resolve any conflicts along the way- you may have to edit files, "git add" them, and then continue the rebase).
git push -f origin desispec_pecvel

rstaten:
Thanks @tskisner, I followed these steps and everything should be up to date now.

The attached plots show chi square vs. peculiar velocity with a minimum chi square fit for three different spectra.
[min_fit_20160521.pdf](https://github.com/desihub/desispec/files/294020/min_fit_20160521.pdf)
[min_fit_20160517.pdf](https://github.com/desihub/desispec/files/294022/min_fit_20160517.pdf)
[min_fit_20160421.pdf](https://github.com/desihub/desispec/files/294023/min_fit_20160421.pdf)

rstaten:
This is coming from the fact that I had to change the read_stdstar_templates function in the io directory so that the correct temperature, surface gravity, and metallicity can be found to compare to the std spectra. Also, based on this change I had to edit stdstars.py in the scripts directory.

sbailey:
If the tests need to be updated to match the new interface, that is ok too.

rstaten:
Ok, I have changed the tests accordingly, but haven't been able to check yet as I am getting the following error when running python setup.py test:

  File "setup.py", line 18, in <module>
    from desiutil.setup import DesiTest, DesiVersion, get_version
  File "/Users/ryanstaten10/desi/desiutil/py/desiutil/setup.py", line 29, in <module>
    from .modules import configure_module, process_module, default_module
  File "/Users/ryanstaten10/desi/desiutil/py/desiutil/modules.py", line 129
    exec(out, globals(), locals())
SyntaxError: unqualified exec is not allowed in function 'desiutil_module' it is a nested function

julienguy:
I obtain a bad matching of the std stars on a simulation run with this code/branch.
Two obvious problems I see 

- seems to a bug at line 152 of stdstars.py 
```
stars.append((i,{"b":[frameFlux["b"][i]/fiberFlat["b"][i]-convolvedsky["b"][i],frameWave["b"]],
                         "r":[frameFlux["r"][i]/fiberFlat["r"][i]-convolvedsky["r"][i],frameWave["r"]],
                         "z":[frameFlux["z"][i]/fiberFlat["z"][i]-convolvedsky["z"][i],frameWave]},fibers["MAG"][i]))
```
should be ```frameWave["z"]``` ?

- in fluxcalibration.match_templates , the redshift is fitted, but this information seems to be lost because the routine returns 
```
return bestId,stdwave,stdflux[bestId],red_Chisq
```

julienguy:
Potential improvements (after the merge) 
- The code is quite slow, we could probably make it faster and also can easily parallelize it. 
- Also ```desi_fit_stdstars``` has peculiar inputs and needs two environments variables to run, we could improve this in a future version.

julienguy:
```
   def applySmoothingFilter(flux):
        return scipy.ndimage.filters.median_filter(flux,200) # bin range has to be optimized
```
is problematic on the edge of the flux array 

I don't know if it's  a problem or not once the same procedure is applied to the data and model star,
but it might be safer to set ivar=0 on the edges.

sbailey:
@rstaten do you need to upgrade desiutil on your local machine?  Travis is not reporting those desiutil related errors, but now it has these errors:
```
  File "/home/travis/build/desihub/desispec/py/desispec/test/test_flux_calibration.py", line 214, in test_outliers
    frame = get_frame_data()
  File "/home/travis/build/desihub/desispec/py/desispec/test/test_flux_calibration.py", line 53, in get_frame_data
    wave, model_flux = get_models(nspec, nwave, wavemin=0, wavemax=10)
ValueError: too many values to unpack
```

py/desispec/fluxcalibration.py
```diff
@@ -102,36 +84,100 @@ def convolveModel(wave,resolution,flux):
...
     bestId=-1
     red_Chisq=-1.
 
-    for i in range(nstd):
+    # search stellar template metadata for canonical f-type model
+
+    tcm=np.where(teff==6000.0)[0]
+    lcm=np.where(logg==4.0)[0]
+    fcm=np.where(feh==-1.5)[0]
+    can_mod=list(set(tcm).intersection(set(lcm).intersection(fcm)))[0]
```
sbailey: This appears to be picking only one point in the standard star phase space rather than considering multiple standard star options and finding the one that fits best.

rstaten:
@julienguy I was also wondering what to do about edge effects. I have tried shortening the arrays used to compute chi square values by various lengths and it doesn't change chi square significantly, but it can cause a different model to be chosen. What should I define as "edge," so that I can amend ivar accordingly?

rstaten:
@sbailey desiutil is up to date on my machine, but this error still occurs. I made a change based on the error message you sent, but it appears that the tests are still failing.

Also, I only chose one point in standard star space to mimic the BOSS code, where I chose the same teff, logg, and feh values. I tried looping through each model, but it takes a very long time. Should I still try to find the best chi square for a range of these values?

sbailey:
@rstaten thanks for clarifying — you fit the redshift using a single canonical model, and then using that redshift you check the other models for which one provides the best fit.  Given that this approach is vetted by BOSS, that is ok for now.

@weaverba137 do you have any ideas for why Ryan would be getting desiutil related test errors?  I don't get this on my machine (python 2.7.11), nor does Travis.

@rstaten what version of python are you using? (python --version)

rstaten:
I'm using 2.7.8, so this could be the issue.

rstaten:
Yeah that was the problem. I have 2.7.11 now and python setup.py test now works, thanks.

weaverba137:
Indeed, you need a more recent Python version.  2.7.11 is good.

sbailey:
@rstaten: now that you aren't getting desiutil errors, are you getting the "too many values to unpack" errors?  It looks like that still needs to be fixed.

rstaten:
No, I have fixed all of the paces where those occur. I'm now getting:

LinAlgError: Array must not contain infs or NaNs

which is generated by:

model_shift=fit.deriv().r[0]/c

This is where the root of the chi square fit is found and converted to redshift.

sbailey:
Julien has provided a new method of fitting stdstars to the models, submitted in PR #206 .  It has much better accuracy at selecting the correct model and includes peculiar velocities in the fit.  That supercedes this PR so I'm going to close this one now.

Issue #153, opened 2016-04-07
refactor spectral extraction parallelism


sbailey:
The current spectral extraction parallelism involves an MPI wrapper that spawns single process python scripts desi_extract_spectra.py .  We're getting hit with python startup for both the MPI wrapper and for the individual extractions, and it is especially bad since desi_extract_spectra.py isn't parallelized and requires one script per bundle instead of one script per frame (or spectrograph or exposure...).  This causes a large and stochastic overhead to the most CPU expensive step of our processing.

Redo the parallelism of the spectral extractions to streamline this.

At the same time, maintain the ability to easily rerun any individual bundle, which is the only advantage of how the current parallelism is implemented.  It should be cut-and-paste level of easy to rerun a failed bundle, and it shouldn't require maintaining two separate API paths for how to get from a command line to an extraction.

sbailey:
Ted refactored the parallelism to use MPI to process each of the bundles and merge them together.  Startup times are still stochasitic, but at least now we only have the initial MPI wrapper startup and not the second spawned python process startup too.  Example startup times from multiple runs of madrone (2 spectrographs x 6 exposures to extract) and maple (10 spectrographs x 12 exposures to extract):

* madrone 6:10, 1:28, 2:13, 3:36 (minutes:seconds)
* maple 3:40, 3:00, 6:15, 8:17

Runtimes should be algorithmically the same, but they also vary presumably due to overheads in reading the input data, reading the PSFs, and writing the output.  Runtimes vary from 6 to 18 minutes, with the longest runtime occurring for the job with the longest startup time too.

As of tag 0.7.0, desi_pipe is allocating 2 cores per extraction process even though the extractions were only using 1 core.  I've verified that we can go to 1 core per extraction process and still typically finish in under 15 minutes while using half as many cores per job.  I'll including that in an upcoming PR later today, after which this ticket can be closed.

PR #205, opened 2016-06-02
pipeline refactor

### LINKED ISSUE #151: enable QA metrics in the pipelined scripts
Turn on the --qafile and --qafig for every pipeline step that has one.

Add QA hooks for any pipeline step that doesn't have one.

### LINKED ISSUE #152: refactor how pipeline planning is done
Ted is organizing a summary of this, but getting a ticket in place.

Refactor the pipeline planning to make it easier to:
* see what will run and what files will be generated ahead of time
* track what has run / succeeded / failed and why
* (re)run subsets of the processing

The current pipeline doesn't plan a step until it can see what was produced by the previous step.  Refactor that to pre-organize the relationships between exposures (e.g. which arcs/flats flow into which psfs which should be used by which exposures).  These can expanded on the fly to know what should be run and what files should exist.

### LINKED ISSUE #158: fix incorrect desi_fit_stdstars pipeline logic
The current `desispec.pipeline` logic incorrectly treats `desi_fit_stdstars.py` as working on an individual frame, like all of the other calibration steps.  However, `desi_fit_stdstars.py` simultaneously fits all 3 cameras of a spectrograph so that it finds the best template for each standard star.  It only needs to be run once per spectrograph, not once per frame.  Inspecting test output from the current pipeline shows that `stdstars-[brz]{n}-{expid}.fits` are identical.

The pipeline logic is being refactored, but this throws a new wrinkle in the dependency tree since `desi_fit_stdstars.py` currently depends upon 10 input files over 3 frames:
  * fibermap for that exposure
  * 3 frame-[brz]{n}-{expid}.fits files
  * 3 sky-[brz]{n}-{expid}.fits files
  * 3 fiberflat-[brz]{n}-{expid}.fits files
where {n} is the spectrograph ID 0-9 and {expid} is the exposure ID.

i.e. `desi_fit_stdstars.py` needs to wait for all 3 different frames to finish with sky subtraction before it can proceed.  I think the currently pipeline may accidentally enforce that by at least running all sky steps before moving on to the stdstar steps.

`desi_fit_stdstars.py` should also gracefully fall back to using 2 channels if a 3rd channel is missing.  At first I thought that that the pipeline should check those inputs and not bother running unless all 3 exist, but that creates a fragility if one CCD fails.  It might be better for the pipeline *not* to enforce that, but just require 2/3 to exist and then proceed, letting `desi_fit_stdstars.py` itself handle a missing channel.

sbailey:
Opening this PR on behalf of Ted who did most of the coding in it.  This is the continuation of a big refactor for how the batch job pipelining is done.  I've independently tested it and gotten far enough that it seems that the *pipelining* portion works and current problems are in the algorithms themselves.  We'd like to merge this back into master now so that they don't get too far out of sync, and then proceed with new branches for fixing remaining bugs / features.

Fixes #158 (stdstar pipeline logic)
Fixes #152 (refactor pipeline planning)

placeholders for #151 (enable QA in pipeline scripts)

partially addresses #153 (refactor extraction parallelism) but it seems we need to study this more.


tskisner:
The QA stuff is integrated but commented out, since I had problems with thousands of processes calling matplotlib on the compute nodes and cluttering the logs with warnings.  We should leave that issue open until we can revisit it and maybe do something better.

sbailey:
Hmm; I don't want commented out QA to hold up this merge, but automated QA is one of the "must have" items for our L2 milestone, so this is high priority to fix.

tskisner:
I suggest we open a separate issue to deal with stupid matplotlib stuff.

PR #206, opened 2016-06-03
new stdstars fit

sbailey:
This PR provides a new standard star template matching algorithm by @julienguy that has much better accuracy than the current code which usually selects the wrong template.  This plot shows the (model-truth)/truth for the old (top) and new (bottom) algorithms.  Note the 100x difference in y scale.
fluxcalib

I updated the pipeline code to work with the new `desi_fit_stdstars` command line options.  It is also now multiprocessing parallelized.  I'll need some help from @tskisner for how that should really be done; it currently involves a hardcoded 24 processes.  But tests by hand on a batch node worked; real batch jobs are running now for final confirmation.  It may need some tweaking for number of nodes and threads and multiprocessing to get it to all fit within 30 minutes still.

This PR supercedes #203 .

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -202,7 +202,7 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
     best_chi2=model_chi2[best_model_id]
     #log.info("model star#%d chi2/ndf=%f best chi2/ndf=%f"%(star,chi2/ndata,best_chi2/ndata))
     
-    return best_model_id,z,best_chi2/ndata, z
+    return best_model_id,z,best_chi2/ndata
     
 
 def normalize_templates(stdwave, stdflux, mags, filters):
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -19,9 +19,12 @@ from astropy import units
 #But should move from here anyway.
 
 def rebinSpectra(spectra,oldWaveBins,newWaveBins):
-    tck=scipy.interpolate.splrep(oldWaveBins,spectra,s=0,k=5)
+    tck=scipy.interpolate.splrep(oldWaveBins,spectra,s=0,k=1)
     specnew=scipy.interpolate.splev(newWaveBins,tck,der=0)
     return specnew
+
+def applySmoothingFilter(flux):
+    return scipy.ndimage.filters.median_filter(flux,200) 
 
 #import some global constants
 import scipy.constants as const
@@ -32,7 +35,9 @@ c=const.c
 erg=const.erg
 hc= h/erg*c*1.e10 #(in units of ergsA)
 
-def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux):
+
+
+def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, logg, feh):
     """For each input spectrum, identify which standard star template is the closest
     match, factoring out broadband throughput/calibration differences.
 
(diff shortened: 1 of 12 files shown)


py/desispec/fluxcalibration.py
```diff
@@ -61,76 +86,124 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux):
...
+    
+    log.info("starting multiprocessing with %d cpus"%ncpu)
+    pool = multiprocessing.Pool(ncpu)
+    model_chi2 =  pool.map(_func, func_args)
+    best_model_id=np.argmin(np.array(model_chi2))
+    best_chi2=model_chi2[best_model_id]
+    #log.info("model star#%d chi2/ndf=%f best chi2/ndf=%f"%(star,chi2/ndata,best_chi2/ndata))
+    
+    return best_model_id,z,best_chi2/ndata, z
+
```
gdhungana: Looks like returning 'z' twice here.

py/desispec/scripts/stdstars.py
```diff
@@ -17,200 +17,232 @@
...
+            band=camera[0]
+            wave[band]=frames[camera].wave
+            flux[band]=frames[camera].flux[star]
+            ivar[band]=frames[camera].ivar[star]
+            resolution_data[band]=frames[camera].resolution_data[star]
+        
+        
+        bestModelIndex[star],redshift[star],chi2dof[star],z=match_templates(wave,flux,ivar,resolution_data,stdwave,stdflux, teff, logg, feh,ncpu=args.ncpu)
+        
+        log.info('Star Fiber: {0}; TemplateID: {1}; Redshift: {2}; Chisq/dof: {3}'.format(starfibers[star],bestModelIndex[star],redshift[star],chi2dof[star]))
```
gdhungana: Here, there are now two different variables. Are redshift[star] and z different?

sbailey:
Good catch @gdhungana.  Thanks for looking at this.  When fixing up the tests to work with the new code I incorrectly added an extra redshift to match the test rather than noticing that multiple return values were different.  I just pushed a new version without the duplicate return redshifts.

sbailey:
The batch job the does fiberflat, sky subtraction, std star fitting, and flux calibration no longer fits within the 30 minute debug queue.  Let's fix that before merging.  Options:
  1. split these into two chained jobs
  2. parallelize one of the other steps
  3. `match_templates` currently compares to *all* stars in the templates file, even though many of them are not F-type flux standards.  The templates could be trimmed down (based on Teff, logg, FeH or colors) to only the templates that might actually be standard stars.

I think I prefer option (3), then (2), then (1).  @julienguy, @tskisner, @others?  Let's try to close this on Monday.

tskisner:
Yes, there are lots of things I want to work on, including splitting some
of the calibration steps.  I can't deal with this until tomorrow though.

Ted

julienguy:
Hi,

 I pushed a modification to the code that consists in preselecting models based on a measured imaging color of the target. 

 - The color (like 'G-R') can be optionally modified. The color argument, e.g. 'G-R', is automatically converted into mag["DECAM_G"]-mag["DECAM_R"] or any other filter set found in the fibermap ending with G and R (in the default setup).
-  The max allowed color difference between imaging data and spectroscopic models is also an optional argument (0.1 by default). 
- 57 models out of 616 pass the |delta G-R|<0.1 cut.
- This naturally makes the final spectroscopic selection approximately 10 times faster.

Two important caveats 
 - There are NO MAGNITUDE ERRORS in the fibermap. We have to change this in the sims and use this info in the code.
- There is NO GALACTIC EXTINCTION in the fibermap. We also have to change this.

sbailey:
With the updates from @julienguy and some tweaking to the number of processes used for the parallelism, fiberflat through flux calibration now easily fits within the debug queue time limit.  I will re-verify this during daylight hours before merging this, but things are looking good.

tskisner:
Thanks @sbailey for implementing the global default nproc calculation to use in all multiprocessing codes.  I am fine if you merge this, and then I'll work on matplotlib things and re-enabling the QA outputs.

sbailey:
Re-test jobs ran and all steps combined ran in 3.5 minutes (fiberflat, sky, stdstars, fluxcalib, procexp).  Merging.  Now we just need to get redmonster and extractions to be this fast.

PR #211, opened 2016-06-08
fix default_nproc circular import dep

sbailey:
This PR fixes a circular import dependency by moving `desispec.pipeline.utils.default_nproc` into `desispec.util.default_nproc`.  It also fixes the integration test to match @julienguy's new `desi_fit_stdstar` interface.

Prior to this PR, here is the circular dependency that I was hitting:
```
In [2]: import desispec.scripts.stdstars
AttributeError                            Traceback (most recent call last)
<ipython-input-2-7483a0ce6f0a> in <module>()
----> 1 import desispec.scripts.stdstars

/Users/sbailey/temp/desispec/py/desispec/scripts/stdstars.py in <module>()
     25 from desispec.interpolation import resample_flux
     26 from desispec.log import get_logger
---> 27 from desispec.pipeline.utils import default_nproc
     28 from desispec.io.filters import load_filter
     29 

/Users/sbailey/temp/desispec/py/desispec/pipeline/__init__.py in <module>()
     20     create_prod, select_nights, graph_read_prod)
     21 
---> 22 from .run import (finish_task, is_finished, run_task, run_step, retry_task,
     23     step_file_types, run_step_types, run_steps, prod_state, file_types_step,
     24     pid_exists, shell_job, nersc_job, qa_path)

/Users/sbailey/temp/desispec/py/desispec/pipeline/run.py in <module>()
     36 import desispec.scripts.fiberflat as fiberflat
     37 import desispec.scripts.sky as skypkg
---> 38 import desispec.scripts.stdstars as stdstars
     39 import desispec.scripts.fluxcalibration as fluxcal
     40 import desispec.scripts.procexp as procexp

AttributeError: 'module' object has no attribute 'stdstars'
```
However, Ted informs me that `desispec.pipeline.utils` was created in the first place to avoid a circular import very similar to this.  It is quite possible that a different refactor of imports or location of `default_nproc` would address this better.

BTW, I finally figured out that command-click on the assignees list allows you to assign multiple people to a ticket, which I previously thought was impossible.  I'm assigning this to both @tskisner and @weaverba137 .  Some combination of them and/or others please take a look and comment since I don't have a deep understanding of python import logic.

### DIFF
--- a/py/desispec/pipeline/utils.py
+++ b/py/desispec/pipeline/utils.py
@@ -34,9 +34,3 @@ def option_list(opts):
                 optlist.append("{}".format(val))
     return optlist
 
-#- Default number of processes to use for multiprocessing
-if 'SLURM_CPUS_PER_TASK' in os.environ.keys():
-    default_nproc = int(os.environ['SLURM_CPUS_PER_TASK'])
-else:
-    import multiprocessing as _mp
-    default_nproc = max(1, _mp.cpu_count() // 2)
--- a/py/desispec/scripts/stdstars.py
+++ b/py/desispec/scripts/stdstars.py
@@ -24,9 +24,8 @@ from desispec import io
 from desispec.fluxcalibration import match_templates,normalize_templates
 from desispec.interpolation import resample_flux
 from desispec.log import get_logger
-from desispec.pipeline.utils import default_nproc
+from desispec.util import default_nproc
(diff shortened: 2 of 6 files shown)


tskisner:
Already looked through this and it is fine- I actually think it is better to put this in the top level util file anyway.  I will merge as soon as tests pass.

Issue #20, opened 2015-04-22
add checksums and code versions to output files


sbailey:
All io.write* functions should use the checksum=True option when writing FITS files to automatically add DATASUM and CHECKSUM keywords to the headers.

Add code versions to the headers.  This may require some creative naming to keep within 8 characters.
Make this semi-optional.  e.g. if desimodel is configured, include a VDSIMODEL keyword with the desimodel version that is set up.  If it isn't set up, don't crash (e.g. writing a frame file doesn't actually require desimodel).  Codes that might be considered for caching their version number:
  * desispec
  * desimodel
  * desisim
  * redmonster
  * specex
  * harp

Note: some (all?) of those may not actually provide a version.  Fix that too if needed.

rainwoodman:
I wonder if the first 8 chars of git sha hash can serve the purpose?

sbailey:
Yes.  Here is a snippet of code from [desiUtil](https://desi.lbl.gov/trac/browser/code/tools/desiUtil/trunk/py/desiUtil/install/git_version.py) that returns the results of `git describe --tags --dirty --always` .  If you have a tag you get a nice clean tag number; if you have a checkout of some random commit you get the hash.

    def git_version(git='git'):
        """Use ``git describe`` to generate a version string.
        Parameters
        ----------
        git : str, optional
            Path to the git executable, if not in :envvar:`PATH`.
        Returns
        -------
        git_version : str
            A version string.
        """
        from subprocess import Popen, PIPE
        myversion = '0.0.1.dev'
        try:
            p = Popen([git, "describe", "--tags", "--dirty", "--always"], stdout=PIPE)
        except EnvironmentError:
            return myversion
        out = p.communicate()[0]
        if p.returncode != 0:
            return myversion
        return out.rstrip()

desiUtil also has similar code for parsing svn versions.  The original intention was for code to use desiUtil for standard version parsing, but that got deprecated at some point where there were problems installing desiUtil and it seemed silly to break code over a dependency just for getting a version number.

weaverba137:
In terms of FITS header keyword names, we came up with a solution in the DECam Legacy Survey pipeline.  It is similar to how table column names and types are defined.  For example:

```
DEPNAM01= 'astropy'
DEPVER01= '1.0.4'
DEPNAM02= 'astrometry_net'
DEPVER02= '3.1.4.1.5.9'
```

sbailey:
That is great.  Let's use that.

sbailey:
[...]

PR #212, opened 2016-06-08
I/O cleanup

### LINKED ISSUE #85: spectra as 32-bit instead of 64-bit
Update io.write_frame() to write 32-bit floats, and update io.read_frame() to cast back to 64-bit double for internal calculations.  This is a very simple and easily reproducible form of lossy data compression.
Related: We had a long discussion on the desi-data mailing list about the pros and cons of this vs. FPACK tile compression.  FPACK can do somewhat better in terms of compression vs. precision lost, but the basic float vs. double compression seems "good enough" for now.

This is similar to what is already done for io.write_image() and io.read_image().

sbailey:
This PR addresses a number of I/O cleanup issues:
  * fixes #85 by storing data as 32-bit float on disk while casting back to 64-bit when reading in.  This is essentially a factor-of-2 lossy compression technique.
  * stores masks as compressed data HDUs, which oddly requires the masks to be 32 bit not 64 bit since the compressed integer fits convention doesn't support 64-bit integers.  If we really need 64 bit masks we can undo this on a case by case basis; in the meantime that HDU will be much much smaller.
  * uses `memmap=False` for most fits I/O, since otherwise `astropy.io.fits` leaves the file open even after you call `close()` (grr)
  * Optimizes the fits I/O by doing more "open the file, read everything you need, close the file" instead of a bunch of calls to `fits.getdata()` and `fits.getheader()`, each of which re-opens the file and parses it anew.
  * Initially writes output files to {filename.fits}.tmp.  After succeeding in writing the entire file it moves it to the final {filename.fits} name.  This avoids corrupted files with the "right" name if the processes is killed in the middle of I/O.  The pipeline was already doing something like this (which I haven't undone yet, maybe that should be added here), but it seems better to directly do it as part of the desispec.io module by default.
  * adds header and data checksums to fits file output to address half of issue #20 .

### DIFF
--- a/py/desispec/fiberflat.py
+++ b/py/desispec/fiberflat.py
@@ -432,7 +432,7 @@ class FiberFlat(object):
             ivar: 2D[nspec, nwave] inverse variance of fiberflat
             
         Optional inputs:
-            mask: 2D[nspec, nwave] mask where 0=good; default ivar==0
+            mask: 2D[nspec, nwave] mask where 0=good; default ivar==0; 32-bit
             meanspec: (optional) 1D[nwave] mean deconvolved average flat lamp spectrum
             chi2pdf: (optional) Normalized chi^2 for fit to mean spectrum
             header: (optional) FITS header from HDU0
@@ -475,7 +475,7 @@ class FiberFlat(object):
         self.wave = wave
         self.fiberflat = fiberflat
         self.ivar = ivar
-        self.mask = mask
+        self.mask = mask.astype(np.uint32)
         self.meanspec = meanspec
         self.chi2pdf = chi2pdf
 
--- a/py/desispec/fiberflat.py
+++ b/py/desispec/fiberflat.py
@@ -13,6 +13,7 @@ from desispec.linalg import cholesky_solve
 from desispec.linalg import cholesky_solve_and_invert
 from desispec.linalg import spline_fit
 from desispec.maskbits import specmask
+from desispec import util
 import scipy,scipy.sparse
 import sys
 from desispec.log import get_logger
@@ -475,7 +476,7 @@ class FiberFlat(object):
         self.wave = wave
         self.fiberflat = fiberflat
         self.ivar = ivar
-        self.mask = mask.astype(np.uint32)
+        self.mask = util.mask32(mask)
         self.meanspec = meanspec
         self.chi2pdf = chi2pdf
 
--- a/py/desispec/frame.py
+++ b/py/desispec/frame.py
@@ -6,9 +6,11 @@ from __future__ import absolute_import, division
 
 import numpy as np
 
+from desispec import util
 from desispec.resolution import Resolution
 from desispec.coaddition import Spectrum
 from desispec.log import get_logger
+from desispec import util
 log = get_logger()
 
 # class Spectrum(object):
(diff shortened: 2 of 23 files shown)


py/desispec/fiberflat.py
```diff
@@ -475,7 +475,7 @@ def __init__(self, wave, fiberflat, ivar, mask=None, meanspec=None,
         self.wave = wave
         self.fiberflat = fiberflat
         self.ivar = ivar
-        self.mask = mask
+        self.mask = mask.astype(np.uint32)
```
weaverba137: What insurance do you have that the input value of mask will always be a type that can be cast to `np.uint32`?  It says 32 bit in the documentation, but if someone puts in something 64-bit anyway, numpy will throw a (possibly obsure) exception.

py/desispec/frame.py
```diff
@@ -89,7 +91,9 @@ def __init__(self, wave, flux, ivar, mask=None, resolution_data=None,
         if mask is None:
             self.mask = np.zeros(flux.shape, dtype=np.uint32)
         else:
-            self.mask = mask
+            if np.any(np.abs(np.asarray(mask)) > 2**32-1):
+                log.error('mask contains elements > 2**32-1; downcasting to 32-bit')
+            self.mask = np.asarray(mask, dtype=np.uint32)
```
weaverba137: This looks like a different method to set the mask than in fiberflat.  Is there a reason you use `np.asarray()` here and not `mask.astype(np.uint32)`?

weaverba137:
Is it possible to include a unit test where you deliberately introduce 64-bit masks, and then handle any exceptions that arise gracefully, *e.g.* still raise an exception, but with a helpful error message that says, "You tried to input a 64-bit mask, when in fact it has to be a 32-bit mask."

sbailey:
Thanks for the comments.  Mask code was written at different times for different objects so there is some inconsistency in how they finally cast it to the variable that is kept in the object.  I'll clean that up.

Given that 64-bit is the default integer size, I don't want to automatically raise an exception if someone passes in a 64-bit mask array if they are only using the first 32-bits.  If they are using more than 32 bits, that should be an error.  In some places the current code silently moves on, in other places it logs an error but keeps going.  This case should probably be an exception with a useful message and not just a logged error.

Detail: numpy type casts involve memory copies, so I want to make sure that these checks are efficient without needless copies.

sbailey:
Based on @weaverba137's comments, I updated the masks such that all of them go through a new util.mask32() code that takes care of up/downcasting, endianness, and 32-bit overflows while standardizing the output to unsigned 32-bit integers.  An exception is raised only if the input mask can't be converted into a uint32 or if downcasting it would result in the loss of a bit that it set.

weaverba137:
I'm satisfied with this PR, but @tskisner should still take a look.

tskisner:
It looks fine to me, merge away.

sbailey:
For the record, the use of CompImageHDU introduced here has a problem when using astropy <1.2.1.  Those versions of astropy wrote the header keywords in the wrong order such that the files could still be read by astropy.io.fits but not by cfitsio-based libraries.  This was fixed in astropy 1.2.1 via astropy/astropy#5125 (thanks to John Parejko wearing his LSST hat).

Issue #127, opened 2016-02-28
rename envnvar $PRODNAME to $SPECPROD


sbailey:
rename $PRODNAME to $SPECPROD to have better symmetry with $PIXPROD.

Pixel simulation output to $DESI_SPECTRO_SIM/$PIXPROD = $DESI_SPECTRO_DATA.

Spectro pipeline outputs to $DESI_SPECTRO_REDUX/$SPECPROD.

The idea is that $DESI_SPECTRO_REDUX is set once by the DESI environment configuration to a canonical location, and then different production runs (with different pipeline versions, different simulation inputs, different parameters, etc.) set $SPECPROD to keep them separated.

sbailey:
I have this working on my laptop, but will wait until the pipeline infrastructure is robustly working before rearranging deck chairs and opening a PR for this.

PR #213, opened 2016-06-09
Matplotlib and Integration Test Fixes

tskisner:
These commits:

- Ensure that the matplotlib backend is set before any matplotlib use, and that this backend is set only once.
- Re-enable QA output in the pipeline steps which support it.
- Make the new integration test the default.  This calls the actual desi_pipe command to create a production and then runs the generated bash scripts in sequence.

Unit tests and integration test pass on my laptop, on an edison interactive session, and in an edison batch job.  More testing by other people is encouraged.


sbailey:
Unit tests work on my laptop; running integration test now.  I disabled the xvfb for the Travis tests, which should test if all the matplotlib stuff is working without needing a real $DISPLAY.

sbailey:
Integration test on my laptop fails with
```
$ python py/desispec/test/integration_test.py
:
:
Updating production /data/desi/spectro/redux/test
  File "/Users/sbailey/desi/git/desispec/bin/desi_pipe", line 349, in <module>
    main()
  File "/Users/sbailey/desi/git/desispec/bin/desi_pipe", line 183, in main
    with open(args.env, 'r') as f:
IOError: [Errno 2] No such file or directory: 'env.txt'
  File "py/desispec/test/integration_test.py", line 244, in <module>
    integration_test()
  File "py/desispec/test/integration_test.py", line 140, in integration_test
    sp.check_call(com, shell=True)
  File "/Users/sbailey/anaconda/lib/python2.7/subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command 'desi_pipe --env env.txt --spectrographs 0 --fakeboot --fakepsf' returned non-zero exit status 1
```
Where is `env.txt` supposed to come from?  I think that we could/should assume that at the point you are running the integration test you already have an working environment configured.  i.e. can we drop the `--env env.txt` option or have the test generate a dummy blank one (with a different name, and cleanup after itself...)?

tskisner:
Sure- sorry I left that in, feel free to commit that change to remove it
from the system call to desi_pipe.

sbailey:
I pushed two changes:
  * don't use env.txt
  * use $SPECPROD (future, see #127) or $PRODNAME (current) if `desi_pipe --prod PROD` isn't set)

Now it runs, but is failing somewhere along the way:
```
Generating scripts
Running extraction script...
/Users/sbailey/desi/git/desispec/bin/desi_pipe_run

real	5m33.252s
user	5m9.759s
sys	0m15.963s

real	4m32.173s
user	4m12.883s
sys	0m12.814s
Running calibration script...
/Users/sbailey/desi/git/desispec/bin/desi_pipe_run

real	2m10.947s
user	2m1.463s
sys	0m8.382s
Running makebricks script...
/Users/sbailey/desi/git/desispec/bin/desi_make_bricks

real	0m2.219s
user	0m1.279s
sys	0m0.240s
/Users/sbailey/desi/git/desispec/bin/desi_make_bricks

real	0m1.555s
user	0m1.287s
sys	0m0.240s
Running zfind script...
/Users/sbailey/desi/git/desispec/bin/desi_pipe_run

real	0m2.375s
user	0m2.005s
sys	0m0.307s

--------------------------------------------------
Brick     True  z        ->  Class  z        zwarn
--------------------------------------------------
```
But it isn't actually making it all the way through, and isn't saying where it failed.  We should have it print more error messages along the way if it fails and stop there.

tskisner:
Hi @sbailey, if you are running this on your laptop, you will have to edit the options.yaml file to point to the stdstar templates.  I don't think there is an environment variable for that.

weaverba137:
Ah, that's true.  You still have to set up the environment variable for the basis templates, but I had to figure out by trial-and-error what other templates directories are necessary for the integration test.

sbailey:
I updated the new integration test to look for star templates under `$DESI_ROOT//spectro/templates/star_templates/v1.1/star_templates_v1.1.fits` as the previous integration test did.  i.e. there is still hard coding, but at least it doesn't start with `/project/projectdirs`.

I also added some additional logging to help figure out what went wrong when things fail:
  * integration test output gives full path to scripts it is running, not just "Running extraction script..."
  * scripts print what log file they are piping to
  * when a step fails, the log now includes the path to the traceback and yaml dump, and includes the traceback directly in the log too.

Tests work on my laptop but Travis tests just failed... investigating...

sbailey:
The new integration test now runs on my laptop and Travis.  When something does go wrong mid-stream I find it harder to track down than before because the top level integration test output doesn't give any indication when one of its intermediate steps fail.  The final redshifts just don't appear and I have to go tracking back through various logs and failure dumps looking for the first step that failed and why.  I'll try to work with this for a few days and see if this is just a learning curve reaction, but I may want to revisit the usability of this.  Merging for now.

PR #217, opened 2016-06-14
Quicklook update

gdhungana:
This PR

- Adds some example QA plots
- Initiates the development of Online specific fiberflat and sky compute and their apply methods
- Updates boxcar output now that readnoise changed to full CCD pixel size array from a scalar in the pix file. Currently it is still taking a median but should be propagated from top. Work in progress.
- Adds readability of psfboot files for all channels from inside desispec.psf. Would be nice to have this info in the header itself.
- Updates new example config file for both offline and online QL run in desispec/data/quicklook directory.
- Other minor cleanups, fixes and updates.

With this version, the incorrect result seen on the last QL run at Nersc, has been corrected.


sbailey:
Could you please move the `quickfiberflat.py`, `quicksky.py`, and `procalgs.py` into py/desispec/quicklook to keep all the quicklook-custom code together in the same place?  Otherwise looks fine for merging.

rkehoe:
@sbailey:  Thanks.  We will re-examine code re-use with offline algorithms 
after Durham.

--------------------------------------------------------------------------
Robert Kehoe					      (214)768-1793 (off.)
Professor of Physics
Department of Physics				       (214)768-4095 (fax)
Southern Methodist University			     kehoe@physics.smu.edu
Dallas, TX 75275		          http://www.physics.smu.edu/kehoe

PR #220, opened 2016-06-15
Dependencies

### LINKED ISSUE #20: add checksums and code versions to output files

sbailey:
This PR adds code version dependencies to the output fits files, using desiutil.depend.add_dependencies() to add keywords of the form
```
DEPNAM00= 'python  '
DEPVER00= '2.7.11 |Anaconda 2.0.1 (x86_64)| (default, Dec  6 2015, 18:57:58)  &'
CONTINUE  '[GCC 4.2.1 (Apple Inc. build 5577)]'
DEPNAM01= 'numpy   '
DEPVER01= '1.10.4  '
DEPNAM02= 'scipy   '
```
This requires desiutil master, but otherwise should be a transparent change for all other work.

This fixes #20 (part of the pipeline 2016b milestone)

Assigning to @tskisner so that he can decide if this should be merged before or after his current pipeline debugging round.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,9 @@ env:
         # - SCIPY_VERSION=0.16
         - ASTROPY_VERSION=1.1.1
         # - SPHINX_VERSION=1.3
-        - DESIUTIL_VERSION=1.4.1
+        # Need desiutil master until >1.5.0 tag to get add_dependencies
+        # - DESIUTIL_VERSION=1.4.1
+        - DESIUTIL_VERSION=master
         - SPECLITE_VERSION=0.4
         - SPECTER_VERSION=0.5.0
         # - HARP_VERSION=1.0.1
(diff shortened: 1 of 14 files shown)


weaverba137:
I think I recall seeing that the compression aspect mentioned in #20 has already been implemented (#212 perhaps)?  Can you confirm?

tskisner:
Might as well merge this first, since my pull request is not even open.  I'll rebase my local branch before pushing.

sbailey:
Yes, the compression was handled in #212.  This PR addresses the code dependencies part.  Merging now.

PR #222, opened 2016-06-16
Fix many, many documentation problems.

weaverba137:
On some systems, but especially ReadTheDocs, documentation may be built in an environment that does not include many of the dependencies of desispec.  In order for automated API documentation to work, all desispec modules must be *importable*, which means not throwing `ImportError` when a module is missing.

Sphinx has a mechanism for dealing with this specific problem.  It can simulate missing modules.  See http://www.sphinx-doc.org/en/stable/ext/autodoc.html#confval-autodoc_mock_imports for more details.  Be sure to take a look at the changes to doc/conf.py to see how many modules need to be mocked up.

In addition, there should be no *calculations* performed at module level, because if one module-level calculation fails, it can prevent not only the module itself from being imported, but any *other* modules that use the module.  `desispec.fluxcalibration` has this property.

Finally, there were a bunch of warnings from, let's be honest here, crappy docstrings that are not following standards.


weaverba137:
PS, please don't merge this until I am certain that ReadTheDocs documentation is building without errors.  You can follow the progress on http://desispec.readthedocs.io/en/rtd-api/api.html

weaverba137:
The documentation is building without errors on ReadTheDocs.  Again, take a look: http://desispec.readthedocs.io/en/rtd-api/api.html.  This is a lot of API.

This PR is safe to merge now.

sbailey:
Thanks for the tour-de-force cleanup of the docstrings and related documentation building.  I'll merge this now, but some questions/comments that may spawn other tickets:
* I was surprised that the entire API ended up as a single long page rather than being hierarchical by submodule.  Was this intended, necessary, an accidental side-effect, other?
* It looks like the "design and use" part is very out of date; this is the placeholder where @tskisner can document the new pipeline structure, but currently it isn't accurate.
* Many of us write docstrings at the level that they are useful for `pydoc` and ipython help, without
being aware of the impact of non-standard formatting unless it is so bad that it causes the Travis sphinx build to fail.  Is there a better way for us to tell as part of a test or pre-merge check whether the docstrings are good?

Regarding module level calculations: I agree that we shouldn't do them arbitrarily, but if we have good reason for a module-level calculation (coding convenience can be a good reason), the sort of try/except wrappers that you added to fluxcalibration seem to be an acceptable workaround.

weaverba137:
We can certainly split up the API page (api.rst) into sub-pages.  Nothing requires it to be on one page.

You can check for sphinx build errors two ways:

1. run `python setup.py build_sphinx` yourself.  Sphinx will report the number of warnings.
2. Look at the Travis logs for the build_sphinx tests.  If *that* doesn't report any warnings, you're in good shape.

PR #227, opened 2016-06-23
Make substantial changes to the way that logging is handled.  

tskisner:
The logs for each pipeline task are now routed to a per-task file by remapping stdout to a file.  Rewrite desi_pipe_status to allow exploring the state of the full pipeline, a pipeline step, or an individual task.  By redirecting stdout for each task to a file, previous fragility is reduced.  This is likely due to reducing the amount of data that must be buffered and merged by slurm.

### DIFF
--- a/bin/desi_pipe_status
+++ b/bin/desi_pipe_status
@@ -25,6 +25,22 @@ import desispec.io as io
 import desispec.pipeline as pipe
 
 
+class clr:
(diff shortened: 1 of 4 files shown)


tskisner:
Also note that after this is merged, you will have to re-run desi_pipe to create the new log and failure directory structure.

PR #230, opened 2016-06-24
fix peak finder algorithm with clustering

julienguy:
Change algorithm to find peaks in slice of CCD image in bootcalib.py , find_fiber_peaks. The original algorithm was failing in rare cases where two pixels in the median slice had the exact same value (possible with ADC counts). Now use first a clustering algorithm.


tskisner:
Note that this fix is in response to some failures on Stephen's "maple" simulation production.  I am merging this now in order to continue testing the pipeline at scale.

PR #232, opened 2016-06-28
Redmonster fix

sbailey:
This branch is work of @julienguy @ngbusca @cballand, though I'm opening a PR to see the changes that need to be synchronized the the redmonster updates in desihub/redmonster#1 .


sbailey:
This requires desihub/redmonster master, not the bitbucket or timhutchinson/redmonster versions.  I confirmed that it works with either the templates shipped with redmonster or the lightweight ones in /project/projectdirs/desi/spectro/templates/redmonster/v0.1 .  Merging now.

moustakas:
Minor comment: we should consider changing ```TYPE``` and ```SUBTYPE``` in the output data model to ```CLASS``` and ```SUBCLASS```, respectively, to not conflict with the ```TYPE``` morphological classification in the Tractor photometric catalogs.

PR #234, opened 2016-06-29
Raw data meta database

### LINKED ISSUE #51: Load file metadata into database

weaverba137:
This PR

- Closes #51 
- Add support for loading or simulating metadata from raw data files.

Please inspect, but let me do the merging, just in case.


tskisner:
Hi Ben, this schema looks good, and maps very well to the dependency graph that I construct in memory (and currently write to yaml) during pipeline production set-up.

I know this PR is focused on the raw data, but hopefully we could eventually make production-local copies of the raw data database, and add derived data products (and their states) to that db.  The pipeline currently loads the initial dependencies and states from per-day yaml files and from querying the filesystem.  Then a snapshot of the whole production metadata and file states are dumped by the pipeline to yaml files after each step.

Once we have signed off on the big milestone in a couple weeks, I think we should convert the pipeline to use your database for this state tracking.

weaverba137:
In c9b5a4f, I added hooks to grab alt, az from the headers if available.  This assumes that the headers are actually called `ALT`, `AZ`.  Please confirm that these really are the official names for these headers before merging, or otherwise tell me what these headers will be called.

sbailey:
Some of the discussion for this PR has been happening in the comments of issue #51.  There is an open question about tile and tile2brick there, but that doesn't need to block this.  Go ahead and merge @weaverba137 when you are ready.

weaverba137:
OK, I'm going to merge.  We can open other issues for specific changes.

PR #235, opened 2016-06-30
make stdstar multiprocessing optional

sbailey:
This PR:
  * disables multiprocessing in fluxcalibration.match_templates if ncpu=1 
  * uses ncpu=1 when wrapping match_templates in the batch pipeline MPI wrapper; this works around a hang when combining MPI+multiprocessing
  * adds some log.debug messages
  * runs redmonster in the regular queue for 1 hour using 12 cores per brick.  This may also have the mpi4py+multiprocessing hang problem too, but the previous "debug queue for 30 min + 2 cores per brick" wasn't going to finish in time anyway, even with the trimmed down QSO templates

### DIFF
--- a/bin/desi_pipe
+++ b/bin/desi_pipe
@@ -21,7 +21,11 @@ import desispec.io as io
 import desispec.pipeline as pipe
 
 
-def compute_step(setupfile, rawdir, proddir, envcom, first, last, specs, night, ntask, taskproc, shell_mpi_run, shell_maxcores, shell_threads, nersc_maxnodes, nersc_nodecores, nersc_threads, nersc_mp, nersc_queue_thresh):
+def compute_step(setupfile, rawdir, proddir, envcom, first, last, specs, night, 
+    ntask, taskproc, shell_mpi_run, shell_maxcores, shell_threads, nersc_maxnodes,
+    nersc_nodecores, nersc_threads, nersc_mp, nersc_queue_thresh,
+    queue='debug', minutes=30,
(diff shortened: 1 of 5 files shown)


tskisner:
I see you are calling sorted() on a dictionary items() iterator, which is basically a list of tuples.  I have never tried that, but I guess it works if this code does not give an error :-)  One of the travis tests died with a timeout downloading stuff from continuum.  Any way to re-run the tests without pushing to this branch?

tskisner:
I confirm this branch works for me on the madrone simulation that previously failed.  The travis failure looks like a network issue on the test server.  @weaverba137, do you mind if I merge this?  Alternatively, how can I re-trigger the travis tests?

moustakas:
On the right hand side of the Travis page there's an icon to redo the
tests.  It's annoyingly subtle but it's there!

tskisner:
I have seen the same information on the web, including screenshots of where the "rebuild" button should be.  Sadly, I see no such button when viewing the travis build page in chrome on Linux...

sbailey:
`sorted(grph.items())` -- that works under python 2.7 since .items() actually returns a list that can be sorted.  But it won't work under python 3 where it returns an iterator...  So I preemptively updated it to `sorted(list(grph.items()))` to be ready for python 3.  And pushing that triggered rerunning the tests too.

julienguy:
Looked at the code, travis tests ok, tested by Ted, so I'm merging

PR #236, opened 2016-07-01
check and fix bootcalib coefficient outliers

sbailey:
This PR improves the robustness of bootcalib by checking the XCOEFF and YCOEFF values vs. fiber number.  Assuming that the coefficients should vary smoothly with fiber number, it identifies outliers and replaces them with the interpolated value.  The attached plot shows an example from maple z2 where one fiber has a very bad fit but gets fixed.

I also added EXTNAMEs to the FITS headers while I was at it.

Note: bootcalib is doing calculations within its output function `bootcalib.write_psf()`.  Gack.  I didn't try to clean that up; I just added an additional check+fix that the fitted values were reasonable before writing them out.

bootcalib

### DIFF
--- a/py/desispec/bootcalib.py
+++ b/py/desispec/bootcalib.py
@@ -1690,6 +1690,49 @@ def trace_fweight(fimage, xinit, ycen=None, invvar=None, radius=2., debug=False)
     # Return
     return xnew, xerr
 
+def fix_coeff_outliers(coeff, deg=5, sigma_clip=5):
+    '''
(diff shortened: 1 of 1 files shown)


tskisner:
This looks good to me.  Leaving it open until @julienguy has a chance to look at it.

julienguy:
We need to add warnings in log file and a fibermask HDU in output file for this.
We can do it later though. I am posting an enhancement issue on this and merge.

PR #238, opened 2016-07-03
MPI zfind

tskisner:
This work adds an alternate MPI parallelization built into desispec.scripts.zfind.  If a communicator is passed to main(), then it forces multiprocessing nproc --> 1 and distributes the targets across the MPI processes.  The results from each process are then merged to the root process and written out as usual.  The commandline desi_zfind script always calls main without MPI.  The new wrapper desi_mpi_zfind calls main using MPI_COMM_WORLD.  And of course the pipeline calls this main with a communicator containing some number of processes (current default is 24).  I have tested on several bricks that using MPI (even with multiple nodes) produces the same output redshifts as using multiprocessing on up to 24 procs.

This PR also integrates this new feature into the pipeline.  On a small simulation (40 bricks, with up to 160 targets per brick), the zfind step completes in ~15 minutes in the debug queue.

### DIFF
--- a/bin/desi_mpi_zfind
+++ b/bin/desi_mpi_zfind
@@ -1,0 +1,21 @@
+#!/usr/bin/env python
+# See top-level LICENSE.rst file for Copyright information
+
+import desispec.scripts.zfind as zfind
+
+comm = None
+rank = 0
+nproc = 1
+
+try:
+    from mpi4py import MPI
+    comm = MPI.COMM_WORLD
+    rank = comm.rank
+    nproc = comm.size
+except ImportError:
+    print("mpi4py not found, using only one process")
+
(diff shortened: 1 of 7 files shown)


tskisner:
Actually I'm going to push one more small thing to this pull request.  Don't merge quite yet...

tskisner:
This last commit generates a master script when running desi_pipe.  This script (run/scripts/run_slurm_all.sh) will submit the jobs to the queue with dependencies.

sbailey:
Ted tested twice in maple with results in /scratch1/scratchdirs/kisner/desi/pipeline/maple/repro_1
 and /scratch1/scratchdirs/kisner/desi/pipeline/maple/repro_2.  Merging.

PR #239, opened 2016-07-07
Bootcalib

julienguy:
Important refactoring of bootcalib
 - fix the algorithm to detect traces (original one was not working in some cases for sparse fiber images)
 - fix a bug in triplet matching algorithm that was leading to duplicate matches. 
 - do a two pass identification of lines after looping on all fibers, taking advantage of the fact that all fibers should see the same arc spectrum.
 - redo the fix of outlier trace solutions using CCD coordinate instead of fiber id for interpolation.
 - the list of good lines is now an ascii file, with an option to use an external one. the code systematically checks for the adequacy of those lines with the full list of lines from NIST.
 - save information in PSF fits file header (CAMERA, EXPID of flatfield and arc exposures)
 - add an option to write the list of lines found, this can then be used as input to specex.
 - add an option to down-select the number of detected lines if too numerous, selecting them by flux while keeping the lines on the edges of each camera wavelength range.
 - unit tests have been modified to match the changes in the function arguments and output.
 - still need to work on the documentation


julienguy:
The fit takes more time but should still fit in debug queue for 500 fiber exposures.
b channel = 6 minutes (single core)
r channel = 13 min
z channel = 17 min

sbailey:
Tested on 3 arcs + 3 flats x 10 spectrographs of maple test production.  Merging.

PR #241, opened 2016-07-11
Cleanup for 2016b release

### LINKED ISSUE #127: rename envnvar $PRODNAME to $SPECPROD

### LINKED ISSUE #137: fibermap propagation
Now that the fibermap is kept as part of the c/frame files, some scripts and functions can be simplified since they can get the fibermap directly from the frame instead of requiring a separate input file.

This has been cleaned up for most scripts and functions in the fibermap-propagation propagation branch, except for desi_fit_stdstars.py and desi_make_bricks.py, both of which use the fibermap file more deeply in their logic — they start with the fibermap file and use its contents to derive what other files they need.  It is still desirable to refactor them to not need the original fibermap file in the raw data directory so that they only depend upon files in the redux directory structure, but that will take a bit more work.

### LINKED ISSUE #233: rename TYPE -> CLASS
Problem: The imaging surveys use "TYPE" to categorize whether a photometric object is PSF-like, extended with an exponential or de vaucouleurs profile, etc.  The spectroscopic pipeline has been using "TYPE" for the redshift fitter output object type, e.g. GALAXY or QSO. 

Proposal: rename TYPE to CLASS (and SUBTYPE to SUBCLASS) in the redshift finder output.  This also has the advantage of going back to the naming used by SDSS/BOSS.  I'm not sure why I switched in the first place.

This would require updates to redshift fitter I/O wrappers, redshift fitting QA code, and quickcat.

Name changes are a pain, but better sooner than later.  Register your objections, support, or alternative suggestions now...

sbailey:
This PR addresses several small cleanup issues for the 2016b spectro pipeline release:

* finishes cleaning up the fibermap propagation, such that code uses the fibermap directly from the frame files instead of needing it to be passed in as an extra option.  Closes #137.
* renames the redshift output "TYPE" to "SPECTYPE" to avoid the ambiguity of the target catalogs / imaging teams also using "TYPE" to describe the Tractor fit object shapes (other considerations were ZTYPE, ZCLASS, or TARGET_TYPE; "CLASS" had the problem of being a python reserved keyword when spelled lowercase...)  Closes #233.
* renames $PRODNAME to $SPECPROD for less ambiguity vs. $PIXPROD, Tractor productions, etc.  Closes #127.
* Adds a `--nompi` option to `desi_pipe_run` to disable MPI and uses that option for the non-batch scripts generated by `desi_pipe`.  This was needed for the integration test to be able to run on edison login nodes where MPI is disabled (but mpi4py still exists and can be imported but if you try to use it it will crash so hard that a try/except can't catch it).

I have tested this with unittests and an integration test, but not a full pipeline run.


sbailey:
I tested this with a 2 spectrograph madrone production and it ran end-to-end without any failures.  After we merge this I'll submit a separate desisim pull request to update the QA analysis to use the new SPECPROD / SPECTYPE names.  I'll wait on merging until after the DESI data telecon tomorrow.

Issue #174, opened 2016-05-03
bootcalib function organization


sbailey:
Context:
* Originally a fair amount of logic was in the `bin/desi_bootcalib` script itself
* quicklook needs to call bootcalib as a function instead of a command line script and so factored this into `desispec.bootcalib.bootcalib()` (originally named `wrap_bootcalib`...)
* the full spectro pipeline also needs to call bootcalib, preferably by calling a python function instead of spawning an external script.  This was refactored into `desispec.scripts.bootcalib.main()`

My confusion is that `desispec.scripts.bootcalib.main()` and `desispec.bootcalib.bootcalib()` do not look particularly similar.  @profxj and @gdhungana please coordinate on a top level bootcalib function so that we aren't replicating features in both `main()` and `bootcalib()`.  I think we want this in the end:

* `desispec.bootcalib.bootcalib()` function that takes data as input and returns data as output, but doesn't directly do any I/O.  The existing `bootcalib` function is like that, but it isn't clear that is has all of the original functionality of the script.
* `desispec.scripts.bootcalib.main() + .parse()` parses command line options, performs I/O, and calls `desispec.bootcalib.bootcalib()` to do the main work.
* both `bin/desi_bootcalib` and the automated pipeline call `desispec.scripts.bootcalib.main()`

quicklook should be able to call `bootcalib()` with data structures rather than having to create lists of strings to pass into `scripts.bootcalib.main()` and then parse files to get the output back.  The automated pipeline goes via `main()` to get to `bootcalib()` so that it calls the code by the same route as one would do when calling `desi_bootcalib` from the command line.

gdhungana:
I was outputting the minimal requirement for desispec.bootcalib.write_psf(), so the I could output them and write to the file inside quicklook. It may be that some functionality were lost doing so or I was somehow ignoring them for an initial test, compared to original desispec/bin/desi_bootcalib.py or now in  desispec.scripts.bootcalib.py. 

I'll be updating QL with the new change. Thanks!

sbailey:
I went ahead and merged Julien's bootcalib updates in PR #239 since they were needed for teststand preparation, but that might have actually made the situation described in this PR even worse.  Too much algorithm is in `desispec.scripts.bootcalib.main()`, and that is where @julienguy updated the algorithm, causing it to get further out of sync with `desispec.bootcalib.bootcalib()` .  We need to clean this up so that the bootcalib algorithms are in one and only one place, to be used by both quicklook and the main pipeline script.

[...]

PR #243, opened 2016-07-12
Improve test coverage

sbailey:
This PR scratches an itch by improving the desispec test coverage (though it is still at a sad 51%).

* `bootcalib.bootcalib()` is broken; see issue #174.  This PR adds an "expectedFailure" test that will start passing once bootcalib.bootcalib() is fixed again.  Note: the command line script and the pipeline don't use this function, but quicklook does.  #174 describes some desired re-organization to remove this redundancy.
* updated `runcmd()` to accept either a function to run or a command line string to run, and use that to test that the bin scripts and the `scripts.*.main()` functions actually do the same thing.
* re-enabled coverage tests of `main()` functions -- we still have algorithms in some of them and we shouldn't get a free pass on lack of coverage there.
* Added keywords to bootcalib output files to be more like other PSF files (NPIX_X, NPIX_Y, PSFTYPE) and updated the PSF class to use header keywords instead of hardcoded npix_x/y and parsing camera name from the filename.
* improved coverage of `pipeline.plan` and `bootcalib.qa*`

Other big items missing unittest coverage:
* io.database (Ben is working on that, perhaps via a separate PR)
* quicklook
* several of the pipeline scripts, especially those that call external packages to work (specex, specter, redmonster); these are tested via the integration test but not the unit tests
* pipeline.run -- lots of MPI stuff in there


weaverba137:
Not sure why you commented out the ignore script hooks.  That will decrease coverage with no obvious benefit.

sbailey:
Ironically, this PR to improve test coverage is failing tests on Travis...  :)  However, this is due to a failure to download the bootcalib test files from NERSC, which is having big /project disk problems today.  After NERSC has recovered, I'll reset these tests to run again.

sbailey:
> Not sure why you commented out the ignore script hooks. That will decrease coverage with no obvious benefit.

I re-enabled coverage tests of main() functions because some of them still have substantial amounts of code and we shouldn't get a free pass on coverage just because they are in a function called main().  e.g. `scripts.bootcalib.main()`, `scripts.makebricks.main()`, `scripts.extract.main()`.  It would be preferable to further clean those up, but in the meantime we should know that we're not getting good coverage on them.

And on some of them, test.test_binscripts does cover them, which is good!

weaverba137:
I am working on a fix to these bootcalib errors on a separate branch.

There are a number of serious problems with test_bootcalib.py:

1. Uses urllib2 instead of requests.
2. Doesn't even try to catch http errors.
3. Runs setUp instead of setUpClass which means the download is retried for every method in class TestBoot.
4. Even if the download does succeed, it does not define a particular directory to write the files.
5. Doesn't clean up after itself properly.

sbailey:
> Uses urllib2 instead of requests.

This seems ok to me; urllib2 is built into python while requests is an external dependency, and the usage of urllib2 here is simple and already written.

> Doesn't even try to catch http errors.

That is a problem

> Runs setUp instead of setUpClass which means the download is retried for every method in class TestBoot.

setUpClass would be better, though at least the current code does check for the existence of the file so it doesn't actually redo the download every time.  `os.path.isfile()` would be a clearer check.

> Even if the download does succeed, it does not define a particular directory to write the files.
> Doesn't clean up after itself properly.

Leaving them in the top-level directory is a bit irritating, but the lack of cleanup is sort of a feature since it means that it doesn't have to redownload them the next time that the test is run.  It would be nice if they would be downloaded to some other more hidden directory (under etc? data? test? temp?) but still leave them behind for the next time that the test is run (e.g. on an airplane, or when NERSC is down...)

weaverba137:
Thanx for the comments, but I'm already way past that point, see #244.

sbailey:
#244 made good improvements but might have generated a merge conflict on the bootcalib tests.  I will rebase this branch off of master and re-push to integrate those changes.

PR #244, opened 2016-07-13
Integration test merge

### LINKED ISSUE #224: Merge various integration test shell scripts
In the process of working on pipeline install instructions, #170, I've created some shell scripts similar to cron_dailytest.sh.  I need to merge these into cron_dailytest.sh or some more general-purpose wrapper script.

### LINKED ISSUE #231: test_bootcalib is trying and failing to download remote data.
In a recent Travis test, tests failed because test_bootcalib.py tried to download files from portal.nersc.gov.  See https://travis-ci.org/desihub/desispec/jobs/140607436

It is not clear why the http access timed out, although it is entirely possible that after many, many tests, NERSC has decided that the Travis IP range is abusive.

Also, we really should be using requests, not urllib2.

weaverba137:
This PR

* Closes #231 by skipping tests that would fail if NERSC data is unavailable
* Closes #224 
* Removes unused `desispec.io.crc` module.
* Adds *framework* for testing `desispec.io.database`.

I'd like to merge this PR *before* #243, so that #243 can take advantage of the fixes to test_bootcalib.py.  When both this PR & #243 are merged, I'll work on additional tests of `desispec.io.database`.


sbailey:
Most changes look good.  Repeating some comment on another ticket to get them associated with this PR:

I'd suggest leaving the bootcalib test files behind (although not in the top level directory) so that they don't have to be re-downloaded every time the test is run.  I'm not sure if there is a standard location for such a thing.

I don't like replacing the couple lines of urllib2 code with requests code.  The requests code is easier to read, but previously the external requests library was an optional dependency that was only needed for quicklook and io.download; one could use the rest of the desispec code without it.  This effectively promotes requests to a required dependency since the tests won't run without it.  urllib2 is built-in to python, and this usage is simple enough (and already written); I'd advocate leaving it in place with a wrapper for catching URLError and HTTPError.  I'm a fan of requests for more complicated http stuff; I don't think it is needed here.

I like the graceful fall back of skipping the bootcalib tests if the input files can't be downloaded.

weaverba137:
I think the latest push addresses these concerns.

sbailey:
Looks good; thanks.  Elegant use of try/except/else.

Issue #242, opened 2016-07-12
desispec.pipeline.plan.graph_path* shouldn't replicate io.findfile functionality


sbailey:
`desispec.pipeline.plan.graph_path*` functions have a lot of code like:

```
    dir = os.path.join(proddir, 'exposures', night, expid) 
    path = os.path.join(dir, "frame-{}-{}.fits".format(cam, expid)) 
```

This replicates the functionality in `io.findfile()`, which is supposed to be the canonical way of mapping a file type plus camera, expid, night, etc. into a location on disk.

Update `desispec.pipeline.plan.graph_path*` to use `io.findfile()` instead.  If `io.findfile` has missing functionality, add it there.

sbailey:
desispec.pipeline.run.qa_path() is another area that should tie into desispec.io.findfile to get the canonical name and location.

tskisner:
![](https://github.trello.services/images/mini-trello-icon.png) [Pipeline Improvements](https://trello.com/c/sCuBELo3/4-pipeline-improvements)

tskisner:
Rather than fix this code and then change that same code as part of #292, I am going to leave this open until that is finished.

tskisner:
Closed by 1468b777f3a4eb2ad565cb10fb6fc1d3044ceaf4

PR #247, opened 2016-07-15
fix QA file output

### LINKED ISSUE #246: pipeline writes QA yaml and *.pdf
Should be easy to fix, but opening a ticket to remind us to do this.  The pipeline is mixing up the --qafile vs. --qafig arguments to the processing steps such that files like "sky-r9-00000007_QA.pdf" are actually yaml files, and the real pdfs don't exist...

sbailey:
fixes #246 so that the pipeline actually outputs QA plots (previously it was outputting the QA yaml files as *.pdf files but not actually making plots...)

Tested via the integration test that runs a mini pipeline, and visually inspecting the the pdf files are actually pdf files.


py/desispec/pipeline/run.py
```diff
@@ -95,12 +95,26 @@
 ]
 
 
-def qa_path(datafile, suffix="_QA"):
+def qa_path(datafile, suffix="_QA", figformat='pdf'):
+    '''
```
profxj: I suggest you add a qaformat keyword as well.  Not sure
we'll stick with YAML for the rest of time.  
Even consider hooking into desispec.io.qa 
to get the format from there..

profxj:
Consider my one comment and then I suggest you merge.

sbailey:
I added the qaformat option with a default of "yaml", but I'm going to punt the hook into desispec.io.qa / desispec.io.findfile to #242, which is dealing with other areas where the pipeline is deriving where files should be located.

PR #248, opened 2016-07-16
Remake QA

profxj:
Introduce QA_Prod class
   Includes method to remake QA on a production

Refactors fluxcalib QA

Edits to io.meta to handle Ted's naming convention
  Failing test_findfile as a result.  Not sure how to fix...


sbailey:
Looks good, except that the __init__ docstring for the QA_Prod class doesn't match the function arguments.  Please make sure that the documentation for how to use this is correct, and include a snippet of code in this PR for what someone should run to get outputs.  Perhaps attach some example plots as an advertisement...

Will there be an accompanying desispec/bin script to run this, or do you intend people to run this only via the python prompt?

I fixed the test to match the new stdstar naming convention, so everything passes again.

i.e. ok to merge other than updating instructions / docs for how someone other than yourself should use this.  Perhaps the ipython notebooks are intended to be that but it isn't necessarily the first place someone would look, so linking to that from here could be useful if you think those are sufficient.  Thanks.

profxj:
Added a script.  

Remake QA with a command like:
`desi_qa_prod --specprod_dir /Users/xavier/DESI/DESI_SCRATCH/redux/madrone/ --remake_frame 3`

From within Python, one may do:

`from desispec.qa import QA_Prod
qa_prod = QA_Prod(specprod_dir)
qa_prod.remake_frame_qa(remake_plots=True)`

Also see the Notebook QA_Prod in docs/nb

PR #249, opened 2016-07-19
Slurp QA

profxj:
Slurp QA (yaml files only so far) in a Production run.
Tested on mardrone

Depends on PR#39 in desiutils, although it won't
fail tests.

Notebook revised.  Will add notes on usage on
github too.


profxj:
Here is usage from within Python:

`from desispec.qa import qa_prod as dqap`
`qa_prod = QA_Prod(specprod_dir)`
`qa_prod.slurp(remake=True, remove=True)`

And as a script:

`desi_qa_prod --specprod_dir /Users/xavier/DESI/DESI_SCRATCH/redux/madrone --remake_frame 3`

py/desispec/qa/qa_exposure.py
```diff
@@ -85,7 +85,11 @@ def fluxcalib(self, outfil):
 
     def load_qa_data(self, remove=True):
         """ Load the QA data files for a given exposure (currently yaml)
+        Args:
+            remove: bool, optional
+              Remove frame files
```
sbailey: Presumably this means remove the QA frame files, not the frame files themselves.

py/desispec/scripts/qa_prod.py
```diff
@@ -14,8 +14,12 @@ def parse(options=None):
                         help = 'Path containing the exposures/directory to use')
     parser.add_argument('--remake_frame', type = int, default = 0,
                         help = 'Bitwise flag to control remaking the QA files (1) and figures (2) for each frame in the production')
-    #parser.add_argument('--outfile', type = str, default = None, required=True,
-    #                    help = 'path of DESI sky fits file')
+    parser.add_argument('-slurp', default = False, action='store_true',
```
sbailey: double dash for long-style options?  e.g. `--slurp`, `--remove`, `--clobber`

sbailey:
Looks fine, with minor comments inline.  Let's wait for desihub/desiutil#39 to be merged first to maintain consistency between the master copies of desispec and desiutil.

py/desispec/qa/qa_exposure.py
profxj: Yup, updated.

py/desispec/scripts/qa_prod.py
profxj: Ok.  I was trying to distinguish between switches and inputs, but 
there is no real need. 
Updated

profxj:
Waiting..

PR #250, opened 2016-07-19
QA_Prod figs

profxj:
Code to generate simple channel histograms for a production.

Here is a script command:

`desi_qa_prod --specprod_dir /Users/xavier/DESI/DESI_SCRATCH/redux/madrone/ --channel_hist=default`

See the Notebook for inline commands.


sbailey:
This uses `desiutil.io.combineDicts`, but we ended up with `desiutil.io.combine_dicts` instead.  Other than the re-rename, looks fine to merge.

profxj:
Excellent catch.

Fixed and merging.

PR #254, opened 2016-07-24
Fluxcalib fixes

profxj:
As titled.

Simple refactoring to eliminate fiber calculations.

Should fix integration test, but I was not successful
on my laptop (other errors appeared).


sbailey:
I confirmed that the integration test worked for me using this version.  In a separate email thread, I suspect that @profxj's problem was that the version number had been updated after a "python setup.py develop" installation.  From what I can tell, if the version number of the code changes you have to re-install develop.

PR #255, opened 2016-07-26
added template groups

timahutchinson:
Added template groups into redmonster wrapper.  Default is to give each template class (galaxy, star, qso) a separate group.

I somehow lost the ability to give this a test run on my laptop, so this needs to be tested before being merged.


timahutchinson:
I somehow lost the ability to give this a test run on my laptop, so this needs to be tested before being merged.

sbailey:
Tested with integration test on my laptop using recently updated redmonster/master .  Merging.

PR #260, opened 2016-08-04
Quicklook

rstaten:
This PR adds one quicklook QA and updates three additional QAs.

-  The new integrate spectrum QA calculates the integral of a spectrum for a given wavelength range.

- The pixel RMS QA now includes the RMS for the overscan region and overscan region per amp.

- The SNR QA now includes SNR vs. magnitude for object types ELG, LRG, QSO, and STAR. It also now includes fiber ID for each of these object types.

- The count spectral bins QA now returns fibers with at least one bin with more than 500 counts.

- These QAs now use the ISO format timestamp rather than MJD.


rstaten:
Two more QAs have been added to this PR.

- The sky peaks QA calculates the number of counts found at sky peak wavelengths, including RMS of these counts over all fibers, sky fibers, and sky fibers per amplifier.

- The XW sigma QA calculates sigma in both the x and w directions of pixels located at sky peak wavelengths. Note: due to the way in which scipy finds optimal parameters, the absolute value of the parameterized sigmas are taken.

rstaten:
The quicklook configuration file has been updated to reflect the changes in this PR.

Note: for snr vs. mag we are using only decam-r magnitude when available.

PR #261, opened 2016-08-15
Documentation updates

tskisner:
This makes progress towards documenting the use of the pipeline.  There is still work to be done, but this is a start.


doc/_static/desi_pipe_status_all.inc
```diff
@@ -0,0 +1,3 @@
+.. code-block:: console
+
+    blah blah blah
```
weaverba137: Are you going to expand on this?

tskisner:
This documentation is a work in progress.  It is much better than the incorrect and mostly non-existent documentation that is currently in master.

weaverba137:
That's fine.  I missed the 'work in progess' in the description.

Issue #219, opened 2016-06-15
How can one determine which template files are needed?


weaverba137:
During a recent integration test on my home system, the test crashed because I had templates/star_templates/v1.0, not templates/star_templates/v1.1.

This has been a recurring problem.  Unless one has expert knowledge, one must actually run the pipeline, find where it crashed, then figure out what files were missing.

So, where are the template versions recorded, & how can someone verify that they have the correct versions installed on their home system (without just downloading everything)?

weaverba137:
PS, I've been running with these templates installed `basis_templates/`, `calib/` and `star_templates/`.  I don't know if others are required, *e.g.* `cosmics/`.

Also, I need to send another round of `fix_permissions.sh` requests, because not all the template directories are visible at http://portal.nersc.gov/project/desi/collab/spectro/templates/.

tskisner:
Currently the lamplines for specex and the star templates are the two data files that cannot be determined by environment variables.  As such, they are written to the master config file for the pipeline (prod/run/options.yaml).  Although I find environment variables annoying, they do have the benefit of allowing data files to be portably found on any system.  If we could obtain these paths from the environment, then the remainder of the global features in options.yaml are portable.

weaverba137:
In #207 @sbailey mentioned basis_templates/v2.2, but in most of the desisim module files I've looked at on NERSC, `$DESI_BASIS_TEMPLATES` is still set to  `<= v2.0`.  The only place I can see where it is v2.2. is the master module file on edison.  So hypothetically, if I were looking for which version to use, & only looked at cori, I'd think v2.0.

sbailey:
1. I agree that the current situation is a problem
2. I welcome ideas for what would be easy and clean to do instead

Addeing @moustakas to the assignee list for this thoughts on how to manage where the templates are and what versions are required.

moustakas:
The templates in ```$DESI_BASIS_TEMPLATES``` are really designed for simulations; they're far too bulky for the pipeline. I propose we select a smallish set of templates centered around F-star temperatures and put those in the data directory of desispec.  (```idlspec2d``` does this -- the Kurucz stellar models are in the etc directory and take up about 2MB.)  

There are ~40 (out of ~930) stellar models with Teff around ~6000 K, which would take up around 20MB. We can probably cull this down further by being more selective with metallicity and surface gravity, although without u-band we're going to inevitably select F-stars a broader range of metallicities.  (Gaia spectrophotometry may help with this.)

[...]

PR #263, opened 2016-08-19
WISE mags for stdstars

sbailey:
We should have run the integration test prior to merging John's big template refactor desihub/desisim#132 .  This PR fixes at least one issue:  standard stars now have WISE magnitudes included, but our default stdstar template set /project/projectdirs/desi/spectro/templates/star_templates/v1.1/star_templates_v1.1.fits doesn't have sufficient wavelength coverage to calculate WISE mags to compare those templates with the data mags.  Perhaps we need to update the template set?

This PR has a work around of simply not calculating the WISE magnitudes for these stdstar templates since we don't normalize to them anyway.

I'm still getting other random redshift failures, but it might be because the new templates are generating fainter ELGs that sometimes should fail.

i.e. don't merge yet, but I wanted to get @moustakas looking at this to see if we should update the templates that we use for stdstar comparisons, or do what I did here, or something else.

### DIFF
--- a/py/desispec/scripts/stdstars.py
+++ b/py/desispec/scripts/stdstars.py
@@ -214,7 +214,12 @@ def main(args) :
     log.info("computing model mags %s"%model_filters)
     model_mags = np.zeros((stdflux.shape[0],len(model_filters)))
     fluxunits = 1e-17 * units.erg / units.s / units.cm**2 / units.Angstrom
+    
     for index in range(len(model_filters)) :
+        if model_filters[index].startswith('WISE'):
+            log.warning('not computing stdstar {} mags'.format(model_filters[index]))
+            continue
+
         filter_response=load_filter(model_filters[index])
         for m in range(stdflux.shape[0]) :
             model_mags[m,index]=filter_response.get_ab_magnitude(stdflux[m]*fluxunits,stdwave)
--- a/py/desispec/test/integration_test.py
+++ b/py/desispec/test/integration_test.py
@@ -217,18 +217,24 @@ def integration_test(night=None, nspec=5, clobber=False):
 
             j = np.where(fibermap['TARGETID'] == zbest.targetid[i])[0][0]
             truetype = siminfo['OBJTYPE'][j]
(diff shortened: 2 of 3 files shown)


moustakas:
Just to remind everybody, we discussed some possibilities in https://github.com/desihub/desispec/issues/219.  

I think we should revisit this question because I *don't* think we should be considering all the stellar templates as potential standard stars -- when we start generating datasets with more realistic systematic errors I think this is going to get our flux-calibration into trouble...  The point is that we should have a semi-static, restricted set of templates that we consider potential standard-star SEDs.

But to the question on this thread, I think we should just update the ```starmodels``` option in ```desispec.pipeline.plan``` to point to the latest stellar models in ```${DESI_ROOT}/spectro/templates/basis_templates/v2.2/star_templates_v2.1.fits```.   Is that acceptable?

Also, is the ```STD_TEMPLATES``` environment variable used anymore?  It's mentioned in the documentation:
* https://desi.lbl.gov/trac/wiki/Pipeline/Install#Data
* https://github.com/desihub/desispec/search?utf8=%E2%9C%93&q=star_templates

but isn't used in any code as far as I could tell.

sbailey:
Still a WIP, but two more updates so that the final integration test checks don't report false failures:
  * consider ELG [OII] flux when assigning ok/OOPS in the final integration test check
  * allow either STD or FSTD to be classified as a STAR

I also turned off the desi_qa_zfind in old_integration_test pending the fix to issue desihub/desisim#149

I have not yet tried using `${DESI_ROOT}/spectro/templates/basis_templates/v2.2/star_templates_v2.1.fits` but that seems reasonable.  The stdstar matching code already trims the templates down to those whose colors match the target (with some buffer to account for measurement error) so we aren't comparing to every template.  I'm not sure we need to do this switch prior to merging this though; the remaining integration test failures turned out to be a combination of the above bookkeeping false failures and a fibermap MAG issue in desisim for which I will submit a PR shortly.

moustakas:
Merging this now so we can proceed in other fronts, but I would like to revisit how the standard star templates are incorporated into the pipeline.

PR #265, opened 2016-08-24
2to3

sbailey:
This PR implements changes recommended by `2to3`.  It has been tested against python 2.7.11 but I have not yet tried running the unit tests on python 3.5.11 (that will be a separate PR after this one is reviewed and merged).  The remaining `2to3` recommendations that I have *not* implemented are all of the form:
```
-    for name, nd in grph.items():
+    for name, nd in list(grph.items()):
```
where wrapping it in a list is unnecessary and just visual clutter.

The only place where I used "from __future__ import unicode_literals" is in doc/conf.py where there were some strings that were explicitly unicode; this make them all unicode.

### DIFF
--- a/py/desispec/io/download.py
+++ b/py/desispec/io/download.py
@@ -73,7 +73,7 @@ def download(filenames,single_thread=False,workers=None):
         if workers is None:
             workers = cpu_count()
         p = Pool(workers)
-        downloaded_list = p.map(_map_download,list(zip(file_list,http_list,[a]*len(file_list))))
+        downloaded_list = p.map(_map_download,zip(file_list,http_list,[a]*len(file_list)))
     return downloaded_list
 
 def _map_download(map_tuple):
--- a/py/desispec/io/download.py
+++ b/py/desispec/io/download.py
@@ -73,7 +73,7 @@ def download(filenames,single_thread=False,workers=None):
         if workers is None:
             workers = cpu_count()
         p = Pool(workers)
-        downloaded_list = p.map(_map_download,zip(file_list,http_list,[a]*len(file_list)))
+        downloaded_list = p.map(_map_download,list(zip(file_list,http_list,[a]*len(file_list))))
     return downloaded_list
 
 def _map_download(map_tuple):
--- a/py/desispec/io/meta.py
+++ b/py/desispec/io/meta.py
@@ -162,7 +162,7 @@ def get_files(filetype, night, expid, specprod_dir=None):
             guaranteed to match the regular expression [brz][0-9].
     """
     glob_pattern = findfile(filetype, night, expid, camera='*', specprod_dir=specprod_dir)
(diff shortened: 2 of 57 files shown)


weaverba137:
Reviewing.....

py/desispec/io/download.py
```diff
@@ -73,7 +73,7 @@ def download(filenames,single_thread=False,workers=None):
         if workers is None:
             workers = cpu_count()
         p = Pool(workers)
-        downloaded_list = p.map(_map_download,zip(file_list,http_list,[a]*len(file_list)))
+        downloaded_list = p.map(_map_download,list(zip(file_list,http_list,[a]*len(file_list))))
```
weaverba137: Minor point: I'm not sure if `list(zip())` is necessary here.  `map()` might be able to accept an iterable.

py/desispec/io/meta.py
```diff
@@ -130,7 +130,7 @@ def get_raw_files(filetype, night, expid, rawdata_dir=None):
             guaranteed to match the regular expression [brz][0-9].
     """
     glob_pattern = findfile(filetype, night, expid, camera='*', rawdata_dir=rawdata_dir)
-    literals = map(re.escape,glob_pattern.split('*'))
+    literals = list(map(re.escape,glob_pattern.split('*')))
```
weaverba137: Minor point: This could also be written as a list comprehension: `[re.escape(f) for f in glob_pattern.split('*')]`.

py/desispec/qa/qa_plots.py
```diff
@@ -441,7 +441,7 @@ def frame_fiberflat(outfil, qaframe, frame, fiberflat):
     i0 = outfil.rfind('/')
     ax2.text(xlbl, ylbl, outfil[i0+1:], color='black', transform=ax2.transAxes, ha='left')
     yoff=0.10
-    for key in sorted(qaframe.data['FIBERFLAT']['QA'].keys()):
+    for key in sorted(list(qaframe.data['FIBERFLAT']['QA'].keys())):
```
weaverba137: Minor point: I think `sorted()` can accept an iterable.

py/desispec/io/meta.py
sbailey: Good point.  I implemented the 2to3 recommendation without checking, but you are right that the list() wrapper isn't necessary in this case.  `multiprocessing.Pool.map` and vanilla `map` accepts an iterable for the function arguments and doesn't require it to be a list.  I'll wait for other comments and update them in bulk.

weaverba137:
Aside from the minor comments, I think this looks good.

sbailey:
I removed some unnecessary list() wrappers as noted by Ben.  After confirming that Travis tests still pass (they do on my laptop...) I'll merge this and proceed with testing with python 3.5.

PR #266, opened 2016-08-31
Py3 (almost)

sbailey:
This PR adds more changes motivated by the py3 transition, though it doesn't pass tests under py3 due to the issues itemized in astropy/astropy#5280.  We're close, but I'd rather get astropy to fix their bugs than work around them in user code.  Changes here:
* `log.warn` -> `log.warning` ("warn" works, but generates a deprecation message)
* fix bug in fiberflat qadict comparison logic that accidentally works in py2.7 but fails in py3.5 because comparing a tuple with a number is no longer allowed
* fix indexing bug in flux calibration QA, also discovered by more restrictive py3.5 rules
* remove `np.asarray()` wrappers for `FITS_rec` objects because that corrupts unsigned integers (sigh)
* optimistic: use dtypes like `('x', (str, 5))` instead of `('x', 'S5')` or `('x', 'U5')` so that the numpy array dtype will use bytes or unicode to match the native string type for the python version.  Works for py2.7, generates the correct array for py3.5 but then `astropy.io.fits` falls on its face anyway.
* Misc little stuff:
  * `imp.reload()` instead of `reload`
  * when downloading a fits file via requests, write with mode `'wb'` instead of `'w'`
  * `isinstance(blat, numbers.Integral)` instead of `isinstance(blat, int)` (required in py3.5 for numpy integers)
  * update binscript location auto detection for py2.7 and py3.5
  * cache pipeline planning `graph.items()` before traversing, because the graph is updated during the planning traversal and that breaks the iterator in py3.5.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -2,10 +2,15 @@
 desispec change log
 ===================
 
-0.9.1 (unreleased)
+0.10.0 (unreleased)
 ------------------
 
-* No changes yet
+PR #266 update for python 3.5:
+
+* Many little updates to work for both python 2.7 and 3.5
+* internally fibermap is now an astropy Table instead of FITS_rec table
+* Bug fix for flux calibration QA
+* requires desiutil >= 1.8.0
 
 0.9.0 (2016-08-18)
 ------------------
--- a/py/desispec/bootcalib.py
+++ b/py/desispec/bootcalib.py
@@ -567,7 +567,7 @@ def parse_nist(ion, vacuum=True):
     nist_tbl.remove_column('Rel.')
     nist_tbl.remove_column('Ritz')
     nist_tbl.add_column(Column(agdrel,name='RelInt'))
-    nist_tbl.add_column(Column([ion]*len(nist_tbl), name='Ion', dtype='S5'))
+    nist_tbl.add_column(Column([ion]*len(nist_tbl), name='Ion', dtype=(str, 5)))
     nist_tbl.rename_column('Observed','wave')
     # Return
     return nist_tbl
--- a/py/desispec/bootcalib.py
+++ b/py/desispec/bootcalib.py
@@ -948,7 +948,7 @@ def fiber_gauss_old(flat, xtrc, xerr, box_radius=2, max_iter=5, debug=False, ver
         list of Gaussian sigma
     """
     log=get_logger()
-    log.warn("fiber_gauss uses astropy.modeling.  Consider an alternative")
+    log.warning("fiber_gauss uses astropy.modeling.  Consider an alternative")
     # Init
     nfiber = xtrc.shape[1]
     ny = xtrc.shape[0]
@@ -1121,7 +1121,7 @@ def find_fiber_peaks(flat, ypos=None, nwidth=5, debug=False) :
 
     # Book-keeping and some error checking
     if len(xpk) != Nbundle*Nfiber:
-        log.warn('Found the wrong number of total fibers: {:d}'.format(len(xpk)))
+        log.warning('Found the wrong number of total fibers: {:d}'.format(len(xpk)))
     else:
         log.info('Found {:d} fibers'.format(len(xpk)))
     # Find bundles
@@ -1129,13 +1129,13 @@ def find_fiber_peaks(flat, ypos=None, nwidth=5, debug=False) :
     medsep = np.median(xsep)
     bundle_ends = np.where(np.abs(xsep-medsep) > 0.5*medsep)[0]
     if len(bundle_ends) != Nbundle:
-        log.warn('Found the wrong number of bundles: {:d}'.format(len(bundle_ends)))
+        log.warning('Found the wrong number of bundles: {:d}'.format(len(bundle_ends)))
     else:
         log.info('Found {:d} bundles'.format(len(bundle_ends)))
     # Confirm correct number of fibers per bundle
     bad = ((bundle_ends+1) % Nfiber) != 0
     if np.sum(bad) > 0:
-        log.warn('Wrong number of fibers in a bundle')
+        log.warning('Wrong number of fibers in a bundle')
         #raise ValueError('Wrong number of fibers in a bundle')
 
     # Return
--- a/py/desispec/fiberflat.py
+++ b/py/desispec/fiberflat.py
@@ -544,7 +544,7 @@ def qa_fiberflat(param, frame, fiberflat):
     # Check amplitude of the meanspectrum
     qadict['MAX_MEANSPEC'] = float(np.max(fiberflat.meanspec))
     if qadict['MAX_MEANSPEC'] < 100000:
-        log.warn("Low counts in meanspec = {:g}".format(qadict['MAX_MEANSPEC']))
+        log.warning("Low counts in meanspec = {:g}".format(qadict['MAX_MEANSPEC']))
 
     # Record chi2pdf
     try:
@@ -564,7 +564,7 @@ def qa_fiberflat(param, frame, fiberflat):
     MAX_SCALE_OFF = float(np.max(np.abs(scale-1.)))
     fiber = int(np.argmax(np.abs(scale-1.)))
     qadict['MAX_SCALE_OFF'] = [MAX_SCALE_OFF, fiber]
-    if qadict['MAX_SCALE_OFF'] > param['MAX_SCALE_OFF']:
+    if qadict['MAX_SCALE_OFF'][0] > param['MAX_SCALE_OFF']:
         log.warn("Discrepant flux in fiberflat: {:g}, {:d}".format(
                 qadict['MAX_SCALE_OFF'][0], qadict['MAX_SCALE_OFF'][1]))
 
@@ -577,7 +577,7 @@ def qa_fiberflat(param, frame, fiberflat):
     mean = np.mean(fiberflat.fiberflat*gdp,axis=1)
     fiber = int(np.argmax(np.abs(mean-1.)))
     qadict['MAX_MEAN_OFF'] = [float(np.max(np.abs(mean-1.))), fiber]
-    if qadict['MAX_MEAN_OFF'] > param['MAX_MEAN_OFF']:
+    if qadict['MAX_MEAN_OFF'][0] > param['MAX_MEAN_OFF']:
         log.warn("Discrepant mean in fiberflat: {:g}, {:d}".format(
                 qadict['MAX_MEAN_OFF'][0], qadict['MAX_MEAN_OFF'][1]))
 
@@ -586,7 +586,7 @@ def qa_fiberflat(param, frame, fiberflat):
                       np.outer(mean, np.ones(fiberflat.nwave))),axis=1)
     fiber = int(np.argmax(rms))
     qadict['MAX_RMS'] = [float(np.max(rms)), fiber]
-    if qadict['MAX_RMS'] > param['MAX_RMS']:
+    if qadict['MAX_RMS'][0] > param['MAX_RMS']:
         log.warn("Large RMS in fiberflat: {:g}, {:d}".format(
                 qadict['MAX_RMS'][0], qadict['MAX_RMS'][1]))
 
--- a/py/desispec/fiberflat.py
+++ b/py/desispec/fiberflat.py
@@ -555,7 +555,7 @@ def qa_fiberflat(param, frame, fiberflat):
     # N mask
     qadict['N_MASK'] = int(np.sum(fiberflat.mask > 0))
     if qadict['N_MASK'] > param['MAX_N_MASK']:  # Arbitrary
-        log.warn("High rejection rate: {:d}".format(qadict['N_MASK']))
+        log.warning("High rejection rate: {:d}".format(qadict['N_MASK']))
 
     # Scale (search for low/high throughput)
     gdp = fiberflat.mask == 0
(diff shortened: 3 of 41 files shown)


weaverba137:
How much effort would it be to convert to using the astropy-recommended practice for writing FITS files?

sbailey:
For those who haven't been involved with the astropy/astropy#5280 discussion like @weaverba137 and myself:  the astropy-recommended way is to explicitly create each column individually and then call `BinTableHDU.from_columns()` instead of letting `BinTableHDU` or `io.fits.writeto` auto-derive the columns from the numpy structured array.

For the case of the fibermap with 22 columns it wouldn't be too bad, but it does get tricky in the brick making code that is both adding columns (NIGHT, EXPID) and appending to tables in pre-existing brick files.  Right now that just needs to know about what it is adding while being agnostic about what is already there.  Having to explicitly create the columns increases the number of places that have to keep in sync:  the code that creates the numpy array in memory the first time, the code that writes it out, the brick making code, and the datamodel documentation.  This gets more fragile for larger tables like targeting and the redshift catalog, which could end up with dozens of columns.  Auto deriving the FITS column type from the numpy column type sounds like the right way to go, and that is the gist of what the astropy io libraries *should* be doing in the first place.

weaverba137:
Can you point out the areas in the code where Python 3 tests are failing due to this specific problem?

sbailey:
A few examples:
```
ERROR: test_fibermap_rw (desispec.test.test_io.TestIO)
  File "/Users/sbailey/desi/git/desispec/build/lib/desispec/test/test_io.py", line 213, in test_fibermap_rw
    desispec.io.write_fibermap(self.testfile, fibermap)
  File "/Users/sbailey/desi/git/desispec/build/lib/desispec/io/fibermap.py", line 95, in write_fibermap
    extname="FIBERMAP", clobber=True)
  File "/Users/sbailey/desi/git/desispec/build/lib/desispec/io/util.py", line 144, in write_bintable
    fx.flush()
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/utils/decorators.py", line 636, in flush
    func = make_function_with_signature(func, name=name, **wrapped_args)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/util.py", line 250, in wrapped
    func(*args, **kwargs)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/hdulist.py", line 614, in flush
    self._flush_update()
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/hdulist.py", line 936, in _flush_update
    hdu._prewriteto(checksum=hdu._output_checksum, inplace=True)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/table.py", line 532, in _prewriteto
    return super(_TableBaseHDU, self)._prewriteto(checksum, inplace)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/base.py", line 580, in _prewriteto
    self._update_checksum(checksum)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/base.py", line 641, in _update_checksum
    datasum_keyword=datasum_keyword)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/base.py", line 1419, in add_checksum
    datasum_keyword=datasum_keyword)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/base.py", line 1369, in add_datasum
    cs = self._calculate_datasum(blocking)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/table.py", line 849, in _calculate_datasum
    return super(BinTableHDU, self)._calculate_datasum(blocking)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/base.py", line 1558, in _calculate_datasum
    self._data_offset)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/base.py", line 568, in _get_raw_data
    return self._file.readarray(offset=offset, dtype=code, shape=shape)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/file.py", line 283, in readarray
    buffer=self._mmap)
TypeError: buffer is too small for requested array
```
and in bricks, though this might be a different bug:
```
ERROR: test_brick (desispec.test.test_io.TestIO)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/utils/decorators.py", line 496, in __get__
    return obj.__dict__[self._key]
KeyError: 'data'

During handling of the above exception, another exception occurred:

  File "/Users/sbailey/desi/git/desispec/build/lib/desispec/test/test_io.py", line 288, in test_brick
    self.assertEqual(bx.get_num_targets(), nspec)
  File "/Users/sbailey/desi/git/desispec/build/lib/desispec/io/brick.py", line 193, in get_num_targets
    return len(np.unique(self.hdu_list[4].data['TARGETID']))
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/utils/decorators.py", line 498, in __get__
    val = self.fget(obj)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/table.py", line 404, in data
    data = self._get_tbdata()
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/table.py", line 171, in _get_tbdata
    self._data_offset)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/base.py", line 568, in _get_raw_data
    return self._file.readarray(offset=offset, dtype=code, shape=shape)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/file.py", line 283, in readarray
    buffer=self._mmap)
TypeError: buffer is too small for requested array
```
this might be a different bug:
```
ERROR: test_frame_rw (desispec.test.test_io.TestIO)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/utils/decorators.py", line 496, in __get__
    return obj.__dict__[self._key]
KeyError: 'data'

During handling of the above exception, another exception occurred:

  File "/Users/sbailey/desi/git/desispec/build/lib/desispec/test/test_io.py", line 144, in test_frame_rw
    frame = desispec.io.read_frame(self.testfile)
  File "/Users/sbailey/desi/git/desispec/build/lib/desispec/io/frame.py", line 109, in read_frame
    fibermap = fx['FIBERMAP'].data
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/utils/decorators.py", line 498, in __get__
    val = self.fget(obj)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/table.py", line 404, in data
    data = self._get_tbdata()
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/table.py", line 171, in _get_tbdata
    self._data_offset)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/hdu/base.py", line 568, in _get_raw_data
    return self._file.readarray(offset=offset, dtype=code, shape=shape)
  File "/Users/sbailey/anaconda/envs/py3/lib/python3.5/site-packages/astropy/io/fits/file.py", line 289, in readarray
    data.shape = shape
ValueError: total size of new array must be unchanged
```
Pipeline tests also fail because they can't write a valid fibermap to test running the pipeline, so other bugs might be lurking in there.

At this point I've probably spent more type submitting and replying to astropy issues and related emails than it would have taken me to work around the specific fibermap case, but this seems a slippery slope of code maintenance if we have to create the columns by hand every time we want to write a table...

weaverba137:
I'm getting my test infrastructure up-to-date, and I can see some of those errors now.

Here's some question to think about:

* Are we using unicode in FITS files because we actually need it?  In other words, do we have to support FITS table columns that contain non-ASCII data?
* Are we using unicode because we are not being as careful as we need to in handling ASCII data?
* Are we using unicode *only* because astropy.io.fits is "being nice" and trying to convert to unicode automatically (implying that we're doing everything else correctly)?

sbailey:
> Are we using unicode in FITS files because we actually need it? In other words, do we have to support FITS table columns that contain non-ASCII data?

No.  The only reason for using unicode is because that is the native string type in python 3.

> Are we using unicode because we are not being as careful as we need to in handling ASCII data?

Perhaps.

> Are we using unicode only because astropy.io.fits is "being nice" and trying to convert to unicode automatically (implying that we're doing everything else correctly)?

Sort of.  Specifically, I'm hoping to achieve a situation where we can have code like
```
def blat(x):
    if x == 'foo':
        ...
```
that "just works" without having to know whether `x` was created on the fly and passed in or whether it originated on disk or was extracted from some row of some column of some Table/array.

It appears that the best practices for this is the "bytes sandwich"
1. bytes for I/O
2. unicode internally
3. be explicit about the encoding used to convert between (1) and (2)

FITS doesn't have a standard for specifying (3), but that could be implicitly handled if the I/O library (like astropy.io.fits) would use a self-consistent encoding/decoding.  The alternative is that we handle it ourselves before passing objects to/from astropy.io.fits, fitsio, or h5py.

geordie666:
I'm intrigued by this:

The alternative is that we handle it ourselves before passing objects
to/from astropy.io.fits, fitsio, or h5py.

Could we just write an I/O wrapper that checks that everything read in is
converted to unicode, converts what isn't, and then passes the input back
to the code as our "standard"? Are there difficult use cases that prevent
that?

Ad.

weaverba137:
@geordie666, I'm thinking along similar lines.  If we're only ever going to have ASCII characters in string data, then we could just always explicitly declare that as bytes, *e.g.* `b'ELG'` or `b'1805p345'` instead of `'ELG'`, `'1805p345'`.

sbailey:
@weaverba137 (and others): are you otherwise ok with these changes as an incremental step toward py3, even if they don't yet get us all the way there?

For the record: I briefly had desispec in a state similar to what @weaverba137 described:
* All numpy arrays used dtype `'Sn'` (bytes) for strings rather than `(str, n)` (native string type for python version).
* All code had to know whether a given variable was coming from a numpy array / FITS file to know whether to compare to `b'BLAH'` or vanilla `'BLAH'`.  This mostly impacted comparisons to `fibermap['OBJTYPE']` to extract the standard star and sky fibers; and comparisons for pulling out targets from a particular `BRICKNAME`.

This worked, but seemed messy, so I tried going back to the general recommendation of using the native string type internally and dealing with bytes and its conversion at the I/O layer in the naive hope that astropy.io.fits (or fitsio or hdf5 or something) would be able to handle this generically...

@geordie666 commented:
> Could we just write an I/O wrapper that checks that everything read in is
converted to unicode, converts what isn't, and then passes the input back
to the code as our "standard"? Are there difficult use cases that prevent
that?

I think this is relatively easy for astropy Tables since they store the data per column and have convenience functions for swapping out one column with another.  It is more of a pain for native numpy structured arrays since (I think) you have to copy the entire array just to reformat one column.  But possible.

weaverba137:
As I just commented in astropy/astropy#5280, I think the bug can be avoided by using bytes and 'Sn' notation rather than native string types, so I think we should go back to that.

A particularly subtle part of this problem mentioned above is that (I think) it only gets triggered when a string column has items with different lengths (`['a', 'bb', 'ccc']`) or where the length of all items is shorter than the width of the field (`['foo', 'bar']` with 'S5').

I think this problem would be easy to fix on the astropy side--all one has to do is pad the data with the correct number of null bytes, but I don't know how quickly that fix could get into a release.

sbailey:
OK, for now I will heed @weaverba137 's suggestion and go back to just using bytes / 'Sn' and live with the warts of not using the native string type.  I haven't given up on the idea of handling this at the I/O layer (either by astropy or by us), but I'll try to get *something* working under py3 first.  I also admit that using pure bytes / 'Sn' would also make it easier to switch to fitsio or h5py if we wish.

sbailey:
Alas, using native bytes everywhere doesn't work, because astropy Tables and `astropy.io.fits` try to auto-convert to unicode for you anyway under python 3.  You can avoid that by using `np.asarray()`, but that munges unsigned integer columns!

BUT - I can smell success!  I'm almost done with a encoding/decoding wrapper for Table I/O that keeps bytes on disk while letting you use native string types internal to the code.  See desihub/desiutil#41 .  I've adapted desispec to use that and now I'm chasing other py3 bugs before updating this PR.

sbailey:
Tests on this branch now pass on both py 2.7 and py 3.5 on my laptop; this also requires desiutil master.

@weaverba137 could you update `.travis.yml` to test against both versions of python now?  It looks like the `.travis.yml` file could use some cleanup and I don't want to make it worse...

Prior to merging:
  * [ ] add py3.5 to travis tests
  * [ ] also test against desisim/py3 and integration tests to see if there are other pieces that need to be updated here
  * [ ] pipeline tests are leaving temporary files behind; clean that up
  * [ ] maybe: convert fibermap usage to astropy Tables throughout instead of FITS_rec which try to trick you into thinking they are numpy ndarrays but they aren't really.  Nowadays Tables are getting a lot more astropy love than FITS_rec and seem to behave more consistently.

weaverba137:
Are we ready for a new tag of desiutil?

sbailey:
I'm fine with a new tag of desiutil if you want to use that instead of master for the desispec travis tests.  Alternatively, we could wait until the full combination of all our products pass py3 tests and then update the tags for everyone.

weaverba137:
Good point, I'll wait.

weaverba137:
Acknowledged.  I'm going to tag desiutil momentarily, then update the .travis.yml file.

weaverba137:
Travis tests are passing under 3.5.  While working on this, I found some sphinx errors, but those can be fixed in a separate PR.

Also, tests are running under desiutil/master due to a potential merge conflict.  After the merge, we should switch to the latest tag of desiutil (1.8.0).

Finally, we should update the change log.  That should happen before the next tag, but doesn't necessarily have to be part of this PR.

sbailey:
Updated changes.rst; merging now and proceeding with updating the desiutil version in travis and tagging desispec 0.10.0.

Issue #272, opened 2016-09-16
resample_flux problem when two inputs have the same wavelength


sbailey:
Consider the following:
```python
x  = np.array([1, 2, 3, 3, 4, 5])
y1 = np.array([1, 2, 3, 4, 5, 6])
y2 = np.array([1, 2, 4, 3, 5, 6])
xx = np.array([1, 2.5, 3.3, 4.5])
z1 = resample_flux(xx, x, y1)
z2 = resample_flux(xx, x, y2)
```
Should `z1 == z2`?  Note that `x` has a repeated value, and the only difference between `y1` and `y2` is which order the values appear for that repeated `x` value.

The current resample code produces different resamplings for `z1` and `z2`, which came up as problematic in the redmonster wrapper that was coadding + loglamda resampling the different channels into a single spectrum.  The order in which the channels are added to the coadd matter, which doesn't smell right.

sbailey:
In a separate email, @julienguy pointed out that this example violates the assumption about the inputs documented in the code:
> The input is interpreted as the nodes positions and node values of
    a piece-wise linear function

i.e. using this code as a *resampler* as originally intended is valid.  Using it as a cheap *coadder* breaks the input assumptions about the data representing a piecewise linear function.  I'll close this issue and open a different one about the need for a fast and simple coadder.

PR #273, opened 2016-09-16
update redmonster wrapper for 2.7 vs. 3.5 reproducibility

### LINKED ISSUE #271: Consider using json.dumps() instead of repr() for serializing redmonster subtype data.
In `desispec.zfind.redmonster`, line 132:

```python
        self.subtype = np.asarray([repr(self.zpicker.subtype[i][0]) for i in range(nspec)])
```

I am pretty sure that `repr()` is not a *standard* way to serialize dictionary data.  It is not even particularly safe.  Supposing that one of the values has a representation that is not easy to convert back into an object of that type (a lot of objects representations do have nice representations, but a lot don't; this is suggested but not enforced by the Python standard). Converting this back into a dictionary would also require an `eval()` call which is also potentially dangerous (or `ast.literal_eval()` which looks safer).

Encoding the dictionary as JSON would not only be a safer way to serialize the data, it would also be compatible with other languages!

sbailey:
This PR updates the desi_zfind wrapper of redmonster so that the results are reproducible between python 2.7 and 3.5.  Previously they weren't even reproducible from back-to-back runs under python 3.5 (!).  The changes are:
* always add the brick files in the same order; previously they were added in whatever order came out of `filters=list(brick.keys())`, which apparently under python 3.5 can vary from run to run.  The order matters due to the `resample_flux` issue raised in issue #272 .  A commented out test in `test_resample` could be used for studying #272, but the minimal change here at least makes the code reproducible.
* convenience: `Brick.get_target_ids()` will return them in the order in which they first appeared in the original input file.
* fixes #271 by storing the redmonster subtype with `json.dumps` instead of `repr`.

I verified this with runs on zdc1 brick-?-elg-10.fits for multiple runs under python 2.7 and 3.5 .  I see that the Travis tests just failed so I'll submit this PR and then go chase that down.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ env:
         # - SPECEX_VERSION=0.3.9
         - MAIN_CMD='python setup.py'
         # These packages will always be installed.
-        - CONDA_DEPENDENCIES=""
+        - CONDA_DEPENDENCIES="qt=4"
         # These packages will only be installed if we really need them.
         - CONDA_ALL_DEPENDENCIES="scipy matplotlib coverage==3.7.1 pyyaml"
         # These packages will always be installed.
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,8 @@ desispec change log
 0.10.1 (unreleased)
 -------------------
 
-* no changes yet
+* update redmonster wrapper for reproducibility
+* Brick.get_target_ids() returns them in the order they appear in input file
 
 0.10.0 (2016-09-10)
 -------------------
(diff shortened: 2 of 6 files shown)


sbailey:
With some help from Ben to work around a Travis QT version 4 vs. 5 problem, tests now pass on Travis too.  Please review.  Thanks.

tskisner:
This looks good to me.  Note that my small pull request #275 also had the travis failure due to Qt4, so this will fix that as well.

weaverba137:
This looks good to me.

sbailey:
I also fixed the qt problem here; tests now pass again.  Merging.

PR #275, opened 2016-09-20
Convert strings to bytes when passing to ctypes create_string_buffer.…

tskisner:
…  Tested that this works on a small production using both the 2.7 and 3.5 desi-conda stacks.


tskisner:
ok to merge this?

PR #278, opened 2016-09-29
Small fixes for python3

tskisner:
This cleans up several small things:  format strings, dictionary iteration, and disabling the (inherently broken) writing of serialized subtypes to a fixed-width ascii column of a binary FITS table. 

Tested this version from bootcalib to redshifts under python3.5 and python2.7, although some zfind tasks ran out of time.

### DIFF
--- a/py/desispec/pipeline/plan.py
+++ b/py/desispec/pipeline/plan.py
@@ -317,18 +317,6 @@ def graph_night(rawdir, rawnight):
         fmdata = io.read_fibermap(fibermap)
         flavor = fmdata.meta['FLAVOR']
         fmbricks = {}
-        for fmb in fmdata['BRICKNAME']:
-            fmb = str(fmb)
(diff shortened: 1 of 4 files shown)


py/desispec/pipeline/plan.py
```diff
@@ -398,7 +398,8 @@ def graph_night(rawdir, rawnight):
             node['out'] = []
             grph[name] = node
 
-    for name, nd in grph.items():
+    current_items = list(grph.items())
```
weaverba137: Why do you need to convert to a list first?  dict.items() is still an iterable in Python 3.

tskisner:
This is just making a copy before modifying the original.  For consistency, I used the same temporary variable name that Stephen used in other parts of the code.

weaverba137:
Aha, if you are modifying the original, then I get it.

PR #280, opened 2016-09-30
Pipeline logging and reliability

tskisner:
This work ensures that all outputs from all processes are captured by writing to per-process log files and then merging those into the corresponding per-task logfile.  It also ensures that all processes working on a single task raise an exception if any one process raises.  This prevents "hanging" at barriers at the end of each task.  At each pipeline step, if all tasks fail for some reason, then the pipeline stops.


sbailey:
Please add a docstring to `collective_log` to explain arguments and how it should be used.  Otherwise looks good; go ahead and merge when the docstring is added.

PR #284, opened 2016-10-10
updated BUNIT to proper units

### LINKED ISSUE #283: cframe files should set BUNIT keyword
Our cframe I/O uses fluxes in units of 1e-17 erg/s/cm2/A, but they don't set the BUNIT keyword to tell anyone that.  They should.

rstaten:
This PR:

- fixes #283 

- adds 1e-17 to the BUNIT keyword

### DIFF
--- a/py/desispec/io/brick.py
+++ b/py/desispec/io/brick.py
@@ -78,14 +78,13 @@ class BrickBase(object):
             # Create empty HDUs. It would be good to refactor io.frame to avoid any duplication here.
             hdr = desispec.io.util.fitsheader(header)
             add_dependencies(hdr)
-            hdr['EXTNAME'] = ('FLUX', 'no dimension')
-            hdu0 = astropy.io.fits.PrimaryHDU(header = hdr)
-            hdr['EXTNAME'] = ('IVAR', 'no dimension')
-            hdu1 = astropy.io.fits.ImageHDU(header = hdr)
-            hdr['EXTNAME'] = ('WAVELENGTH', '[Angstroms]')
-            hdu2 = astropy.io.fits.ImageHDU(header = hdr)
-            hdr['EXTNAME'] = ('RESOLUTION', 'no dimension')
-            hdu3 = astropy.io.fits.ImageHDU(header = hdr)
+            hdr['EXTNAME'] = ('FLUX', '1e-17 erg/(s cm2 Angstrom)')
+            hdr['BUNIT'] = '1e-17 erg/(s cm2 Angstrom)'
+            hdu0 = astropy.io.fits.PrimaryHDU(header=hdr)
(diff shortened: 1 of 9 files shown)


sbailey:
Thanks!  This PR fixes the BUNIT for fluxcalib output.  Could you also add BUNIT for cframe?  In that case I think it is completely missing.

rstaten:
It is completely missing, but the cframe files seem to be generated using the same write_frame function as the frame files. Since these two files do not have the same output flux units, this may be why the BUNIT keyword is not included.  If desired, I could simply write a new function or change the current one to distinguish between the outputs.

Also, I'm assuming the stdstars output should also include the factor of 1e-17 as well? If so, this is a simple fix.

sbailey:
> It is completely missing, but the cframe files seem to be generated using the same write_frame function as the frame files. Since these two files do not have the same output flux units, this may be why the BUNIT keyword is not included. If desired, I could simply write a new function or change the current one to distinguish between the outputs.

Good point.  Let's add an optional `units` keyword to `desispec.io.write_frame`, with something like
```python
if units is not None and 'BUNIT' not in hdr:
    hdr['BUNIT'] = str(units)
```
That should work for `units` as a string and if `units` is an `astropy.units.Units` object.  To check: are there any python 3 unicode gotchas?  I hope not...

And then update the code that calls `desispec.io.write_frame` with the proper units for frame vs. cframe.

> Also, I'm assuming the stdstars output should also include the factor of 1e-17 as well? If so, this is a simple fix.

Yep.

rstaten:
I have set the BUNIT for frame files to photons/bin and cframe files to 1e-17 erg/s/cm^2/A.

sbailey:
We had a misunderstanding about my suggestion for how to propagate units to the frame files.  Implementing it was easier than re-describing it, so I went ahead and changed it.  Since we do propagate header keywords from inputs to outputs, I would like to run an integration test on this to check if we end up with any intermediate steps with incorrect BUNITs.  i.e. don't merge yet; I'll fix or report any problems, then merge.

sbailey:
I made a number of small changes:
* Added more BUNIT throughout
* Made BUNIT strings that can be parsed by `astropy.unit.Unit` (even if we aren't doing that in the main code).  e.g. "Angstrom" instead of "A" and "electron" instead of "electrons".
* fixed fiberflat single- vs. double-precision when reading in (unrelated to original purpose of this PR)
* fixed fiberflat EXTNAME (now "FIBERFLAT" not "FLUX"; unrelated to original purpose of this PR)

I checked this with an integration test to make sure that BUNITs weren't incorrectly propagated from one file type to another.  After the Travis tests pass I think this is ready to merge.  Thanks @rstaten for getting this started.

PR #285, opened 2016-10-14
Quicklook QA

rkehoe:
This update is part 1 in bringing QuickLook QA and offline QA closer into consistency.  It includes:

- merge metric calculations from QuickLook sky residuals into offline QA (@gdhungana)
- update all QuickLook QAs to include FLAVOR, NIGHT key/values, and change from ARM, SPECTROGRAPH --> CAMERA
- some clean up

@profxj Could you check qa_skysub to make sure as you expect?

Next on list: decide w/ @profxj  and implement a common convention for handling QA metadata, parameters and the hierarchy of QA dicts.


sbailey:
OK with me.  @profxj can do final review and merge.

py/desispec/qa/qa_quicklook.py
```diff
@@ -1405,41 +1462,31 @@ def run(self,*args,**kwargs):
...
-        residuals_ivar = util.combine_ivar(frame.ivar[skyfibers], skymodel.ivar[skyfibers])
-
-        #- Now get the median residual for each fiber
-        med_resid_fiber=np.median(residuals,axis=1)
-        med_resid_wave=np.median(residuals,axis=0)
-        wavelength=frame.wave
+        qadict=qa_skysub(param,frame,skymodel,quick_look=True)
 
-        retval["VALUE"]={"MED_RESID":qadict["MED_RESID"], "NREJ": qadict["NREJ"], "NSKY_FIB": qadict["NSKY_FIB"], "RESID_PER": qadict["RESID_PER"], "NBAD_PCHI": qadict["NBAD_PCHI"], "MED_RESID_FIBER": med_resid_fiber, "MED_RESID_WAVE": med_resid_wave,"WAVELENGTH": wavelength}
+        retval["VALUE"]={"MED_RESID":qadict["MED_RESID"], "NREJ": qadict["NREJ"], "NSKY_FIB": qadict["NSKY_FIB"], "RESID_PER": qadict["RESID_PER"], "NBAD_PCHI": qadict["NBAD_PCHI"], "MED_RESID_FIBER": qadict["MED_RESID_FIBER"], "MED_RESID_WAVE": qadict["MED_RESID_WAVE"],"WAVELENGTH": qadict["WAVELENGTH"]}
```
profxj: If the intent here is to duplicate the key names (as it appears, how about:

retval["VALUE"] = {}
for key in qadict.keys():
   retval["VALUE"][key] = qadict[key]

and if there are specific keys you want to skip you can do that too.

profxj:
I left one minor comment.  Consider it and then merge.

py/desispec/qa/qa_quicklook.py
gdhungana: @profxj Agreed. I think we want whole dictionary here. I just pushed the change. Merging after the tests pass.

gdhungana:
@profxj Agreed. I think we want the whole dict here. I updated the changes and now merging as the tests also have passed.

PR #288, opened 2016-10-19
enable asymmetric resolution matrix

### LINKED ISSUE #286: Review resolution matrix requirements
Resolution matrix data formats and utilities are organized in the `resolution.Resolution` class, which internally uses the scipy.sparse DIA format but knows how to convert this to and from a dense matrix or the arrays stored in a brick file.

The resolution matrix of simulated bricks is obtained (in quickbrick and quickgen) using code like:
```
R = Resolution(camera.get_output_resolution_matrix())
```
The recent PR desihub/specsim#50 changes the return value of `get_output_resolution_matrix` from a dense matrix to a DIA-format matrix.  The DIA format is supported by the Resolution constructor, but it makes additional requirements that will not generally be met (e.g., having an odd-number of off diagonals).

This issue is to review what the Resolution requirements should be, making changes if necessary, and update the Resolution constructor to accept any valid DIA matrix, padding or trimming as necessary.

sbailey:
This PR fixes #286 (Review resolution matrix requirements) to enable creating resolution matrices with a different number of upper/lower diagonals.  In particular, it enables desisim quickbrick and quickgen to work with the latest specsim, which now can return resolution matrices with an even number of diagonals (see desihub/specsim#50).

Tests are expanded to have 100% desispec.resolution code coverage.  I think algorithmic input options coverage is pretty good too, but let me know if you think there are important missing cases to test.

I implemented this by padding the underlying sparse resolution matrix to have (2\*n+1) diagonals (the main diagonal plus n upper and n lower diagonals).  An alternative would be to update the Frame file format to explicitly track the diagonals, but since this would involve backwards compatibility issues and touch more pieces of code, I'm not pursuing that now.  In practice the real extractions will have (2\*n+1) meaningful diagonals, and the simulations will have at most 1 zero padded diagonal, so I don't think this padding approach is problematic.

### DIFF
--- a/py/desispec/resolution.py
+++ b/py/desispec/resolution.py
@@ -42,29 +42,38 @@ class Resolution(scipy.sparse.dia_matrix):
     The last format is the one used to store resolution matrices in FITS files.
 
     """
-    def __init__(self,data):
+    def __init__(self, data, offsets=None):
 
         ### if scipy.sparse.isspmatrix_dia(data) and np.array_equal(np.sort(data.offsets)[::-1],self.offsets):
         if scipy.sparse.isspmatrix_dia(data):
             # Input is already in DIA format with the required diagonals.
             # We just need to put the diagonals in the correct order.
             diag_order = np.argsort(data.offsets)[::-1]
-            ndiag = len(data.offsets)
-            if ndiag%2 == 0:
-                raise ValueError("Number of diagonals ({}) should be odd".format(ndiag))
-            self.offsets = np.arange(ndiag//2,-(ndiag//2)-1,-1)
-            if not np.array_equal(data.offsets[diag_order], self.offsets):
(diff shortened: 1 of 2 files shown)


dkirkby:
The changes look good to me (modulo pep8).

Is there already a use case for passing an offsets array with a dense matrix?

Does (or should) the data model place any restrictions on the min/max number of diagonals?

sbailey:
> Is there already a use case for passing an offsets array with a dense matrix?

I'm not sure we need that, but it was easy enough to add.
* non-square dense matrix + offsets is interpreted as sparse data to create a scipy.sparse.dia_matrix
* square dense matrix + offsets is interpreted as a dense resolution matrix where you only want to
   keep the specified diagonals

> Does (or should) the data model place any restrictions on the min/max number of diagonals?

I put some sanity checks on very small (<3) and very large (>4*default_ndiag) offset arrays.

PR #291, opened 2016-10-24
work around non-ASCII characters in NIST line lists

### LINKED ISSUE #289: NIST arc files have non-ascii characters
(Opening after offline communication with Stephen)

The files that bootcalib uses cause a read error with the latest python3.5 version of the desi-conda stack at NERSC.  Stephen noted that this is due to the non-ascii Angstrom symbol.  To reproduce:

>>> from desispec import bootcalib                                        
>>> bootcalib.load_arcline_list('b0', vacuum=True)
      
INFO:bootcalib.py:547:parse_nist: reading NIST file /scratch2/scratchdirs/kisner/software/desi-dev/desispec/pipeout-01321ba/lib/python3.5/site-packages/desispec-0.11.0.dev1033-py3.5.egg/desispec/data/arc_lines/CdI_vacuum.ascii                                                                                                  
  File "<stdin>", line 1, in <module>
  File "/scratch2/scratchdirs/kisner/software/desi-dev/desispec/pipeout-01321ba/lib/python3.5/site-packages/desispec-0.11.0.dev1033-py3.5.egg/desispec/bootcalib.py", line 624, in load_arcline_list
    tbl = parse_nist(iline, vacuum=vacuum)
  File "/scratch2/scratchdirs/kisner/software/desi-dev/desispec/pipeout-01321ba/lib/python3.5/site-packages/desispec-0.11.0.dev1033-py3.5.egg/desispec/bootcalib.py", line 548, in parse_nist
    nist_tbl = Table.read(nist_file, format='ascii.fixed_width')
  File "/global/common/edison/contrib/desi/conda/conda_3.5-20160913/lib/python3.5/site-packages/astropy/table/table.py", line 2332, in read
    return io_registry.read(cls, *args, **kwargs)
  File "/global/common/edison/contrib/desi/conda/conda_3.5-20160913/lib/python3.5/site-packages/astropy/io/registry.py", line 351, in read
    data = reader(*args, **kwargs)
  File "/global/common/edison/contrib/desi/conda/conda_3.5-20160913/lib/python3.5/site-packages/astropy/io/ascii/connect.py", line 37, in io_read
    return read(filename, format=format, **kwargs)
  File "/global/common/edison/contrib/desi/conda/conda_3.5-20160913/lib/python3.5/site-packages/astropy/io/ascii/ui.py", line 289, in read
    table = fileobj.read()
  File "/global/common/edison/contrib/desi/conda/conda_3.5-20160913/lib/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 613: ordinal not in range(128)

sbailey:
This PR fixes #289 where bootcalib was failing with python 3.5 at NERSC due to non-ASCII characters in NIST line lists and NERSC not setting $LANG.

Thanks to @weaverba137 for pinging the astropy list about this, and Derek Homeier for providing the useful solution of `locale.setlocale(locale.LC_ALL, str('en_US.utf8'))` .

@weaverba137 also asked whether this should be moved into the `desispec.io` module.  As long as these files are used only for bootcalib and they remain a part of the package data, I think it is ok to keep their I/O as a detail inside bootcalib.  If we need to read these files for other reasons (e.g. a python-based PSF or wavelength fitter), then their I/O wrapper should move out into `desispec.io`.  For now, I left the location as-is for the minimal change to fix the bug.

I also confirmed that the unit tests failed at NERSC prior to this fix, and pass after this fix (they already passed on many systems including Mac and Travis since $LANG is set by default).

### DIFF
--- a/py/desispec/bootcalib.py
+++ b/py/desispec/bootcalib.py
@@ -547,7 +547,7 @@ def parse_nist(ion, vacuum=True):
     nist_file = resource_filename('desispec', srch_file)
     log.info("reading NIST file {:s}".format(nist_file))
     default_locale = locale.getlocale(locale.LC_CTYPE)
-    locale.setlocale(locale.LC_CTYPE, 'en_US.utf8')
+    locale.setlocale(locale.LC_CTYPE, 'en_US.UTF-8')
     nist_tbl = Table.read(nist_file, format='ascii.fixed_width')
     locale.setlocale(locale.LC_CTYPE, default_locale)
     gdrow = nist_tbl['Observed'] > 0.  # Eliminate dummy lines
--- a/py/desispec/bootcalib.py
+++ b/py/desispec/bootcalib.py
@@ -546,10 +546,10 @@ def parse_nist(ion, vacuum=True):
     # Read, while working around non-ASCII characters in NIST line lists
     nist_file = resource_filename('desispec', srch_file)
     log.info("reading NIST file {:s}".format(nist_file))
-    default_locale = locale.getlocale(locale.LC_ALL)
-    locale.setlocale(locale.LC_ALL, 'en_US.utf8')
+    default_locale = locale.getlocale(locale.LC_CTYPE)
+    locale.setlocale(locale.LC_CTYPE, 'en_US.utf8')
     nist_tbl = Table.read(nist_file, format='ascii.fixed_width')
-    locale.setlocale(locale.LC_ALL, default_locale)
+    locale.setlocale(locale.LC_CTYPE, default_locale)
     gdrow = nist_tbl['Observed'] > 0.  # Eliminate dummy lines
     nist_tbl = nist_tbl[gdrow]
     # Now unique values only (no duplicates)
--- a/py/desispec/bootcalib.py
+++ b/py/desispec/bootcalib.py
@@ -22,6 +22,7 @@ import time
 import os
 import sys
 import argparse
+import locale
 from pkg_resources import resource_exists, resource_filename
 
 from astropy.modeling import models, fitting
@@ -542,10 +543,13 @@ def parse_nist(ion, vacuum=True):
     if not resource_exists('desispec', srch_file):
         log.error("Cannot find NIST file {:s}".format(srch_file))
         raise Exception("Cannot find NIST file {:s}".format(srch_file))
-    # Read
+    # Read, while working around non-ASCII characters in NIST line lists
     nist_file = resource_filename('desispec', srch_file)
     log.info("reading NIST file {:s}".format(nist_file))
+    default_locale = locale.getlocale(locale.LC_ALL)
+    locale.setlocale(locale.LC_ALL, 'en_US.utf8')
     nist_tbl = Table.read(nist_file, format='ascii.fixed_width')
+    locale.setlocale(locale.LC_ALL, default_locale)
     gdrow = nist_tbl['Observed'] > 0.  # Eliminate dummy lines
     nist_tbl = nist_tbl[gdrow]
     # Now unique values only (no duplicates)


weaverba137:
When you say "they already passed on ... Travis", I'm not sure that's entirely correct.

I actually wanted to write (and am writing) a unit test on parse_nist(), to make sure we know exactly when that passes or fails, *then* implement the fix.

One weird problem is that the trick `del os.environ['LANG']` doesn't work in unit tests.  Python still "remembers" what LANG was, so you can't force an expected failure that way.

There are several cases to test here:

* Python 2 versus 3
* LANG set before python starts versus not
* LANG set or unset during execution of the test, *i.e.* force Python into a "blank" locale.
* locale workaround

sbailey:
Hmm!  My attempt to cache the previous state via `locale.getlocale(locale.LC_ALL)` works on edison (py3.5) and my Mac (py2.7 and py3.5), and under Travis py2.7 but not Travis py3.5:
```
  File "/home/travis/build/desihub/desispec/build/lib/desispec/test/test_bootcalib.py", line 131, in test_wavelengths
    llist = desiboot.load_arcline_list(camera)
  File "/home/travis/build/desihub/desispec/build/lib/desispec/bootcalib.py", line 628, in load_arcline_list
    tbl = parse_nist(iline, vacuum=vacuum)
  File "/home/travis/build/desihub/desispec/build/lib/desispec/bootcalib.py", line 549, in parse_nist
    default_locale = locale.getlocale(locale.LC_ALL)
  File "/home/travis/miniconda/envs/test/lib/python3.5/locale.py", line 576, in getlocale
    raise TypeError('category LC_ALL is not supported')
TypeError: category LC_ALL is not supported
```
I empirically verified that LC_CTYPE also works on edison, and that appears to be the default category for `locale.getlocale(category)`, so hopefully that will work for Travis too.  Pushing in a moment.

I agree that the Travis tests weren't testing the case of $LANG not being set ahead of time.  If you can do that without too much work, ok.  But if that turns out to be a pain, as long as we have a setup that works for Travis 2.7 and 3.5 (normal Linux), NERSC 3.5 (Crayz Linux), and Mac Anaconda (2.7 and 3.5) I'd call that good enough.

weaverba137:
An additional twist is that Travis sets *both* LANG and LC_ALL (see https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables). So that even if one is unset, it is possible in principle for the locale to be set from the other.

weaverba137:
While working on the unit test for this, I noticed that test_bootcalib.py is failing to clean up test_arc.fits.gz and test_flat.fits.gz.  I suppose that's meant to "cache" those files, but I really don't like extraneous files in my git checkout.

sbailey:
Feel free to move `test_arc.fits.gz` and `test_flat.fits.gz` to a more hidden location in desispec (perhaps `py/desispec/data/test` or under etc?), but please don't auto-delete them since they have to be redownloaded.  They are purposefully cached and they are already in .gitignore .

sbailey:
BTW, I just discovered that Mac OSX needs the 'en_US.UTF-8' instead of 'en_US.utf8'.  Either works at NERSC.  Hopefully 'en_US.UTF-8' will also work for Travis.  This might be the first case we've had of a Mac vs. Linux bug that didn't just turn out to be different versions of astropy/numpy/whatever installed.

weaverba137:
Interesting, and it may actually be more fundamental than that: BSD C library versus GNU C library.

weaverba137:
I can't get a consistent result from the unit tests I am writing, which is kind of scary, but I wouldn't have time to track it down today.  Merge if you want.

PR #295, opened 2016-11-01
update of QA dictionaries: PARAMS and METRICS in online and offline QAs

rkehoe:
Update of QAs to bring 'offline' and 'online' into closer agreement.

- for 'online' QAs (those written originally for quicklook):  Add 'PARAMS' subdictionary to include configurable thresholds and reference values to 3 QAs (sky residual, countbins, countpix).  Adopt data subdirectory label 'METRICS'  (i.e. 'VALUES' --> 'METRICS')

- for 'offline' QAs (those written initially for offline pipeline): relabel data and parameter subdictionaries as 'METRICS' and 'PARAMS', respectively.

PR #298, opened 2016-11-04
Quicklook configuration

gdhungana:
This PR focuses on configuration updates to quicklook:

- Adds a module for generating configuration to run QL
- Refactors the QL pipeline to bin area wrapper, and main pipeline code in scripts 
- replaces hardcoded QLF URL access parameters with dependency on QLF environment variables for QLF based QL run
- Add QA params in the configuration following recent changes in #295 
- use datamodel convention for naming QA output yaml and png files


sbailey:
I'm merge this now, but some comments for the future:

The string formatting syntax of `"blat foo %i" % n` still works but is officially deprecated.  In future updates, please switch to `"blat foo {:d}".format(n)` .  We still have a lot of the old style string formatting in our code that we'll need to update someday, but any new code should use the `.format()` syntax.  Full details (except for the reason for deprecating the old syntax) are at https://docs.python.org/3/library/string.html#formatstrings .

This uses environment variables for the QLF username and password.  OK for now, but prior to deployment we'll need to do a full security review for how passwords are propagated, https, firewalls, etc.  It would be handy to have read-only access from offsite, but we want to make sure that we don't accidentally exposure the ability for a script kiddie to post fake QL results to the real production QLF.

PR #311, opened 2016-11-13
Do not setup the software environment inside of slurm scripts.

tskisner:
Do not attempt to do environment setup inside the SLURM script.  This has proven unreliable since SLURM itself has bugs that prevent it from working cleanly.  Instead, the desired DESI software packages must be loaded at the time of job submission.  I personally disagree with this kind of workflow, but since this is a limitation of SLURM, we don't have much choice.

Also change the string representing the state of running jobs ("wait") to be more informative ("running").

### DIFF
--- a/bin/desi_pipe_run_mpi
+++ b/bin/desi_pipe_run_mpi
@@ -1,0 +1,18 @@
+#!/usr/bin/env python
+#
+# See top-level LICENSE.rst file for Copyright information
+#
+# -*- coding: utf-8 -*-
+
+"""
+Run one or more steps of the DESI pipeline using MPI.
+"""
+
+from mpi4py import MPI
+
+import desispec.scripts.pipe_run as pipe_run
+
+if __name__ == '__main__':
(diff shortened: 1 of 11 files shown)


sbailey:
I also disagree with this kind of workflow where the batch script itself is not self-describing for the environment setup necessary to run it.  Prior to merging, let's discuss this a bit in person and consider pushing more on NERSC about this.  The last answer I heard from them way basically "oh well, you should be using Shifter if you want reproducible job scripts anyway", which is not a very satisfying answer.

tskisner:
Based on the SLURM discussion forums, we are not the only users unsatisfied with this.  One thing we could do is more extensive logging at the start of the slurm scripts which dump the list of loaded modules, path to desi_pipe_run, etc.  At least then we have a record of the job submission environment that got propagated.

sbailey:
I forgot about this ticket.  Please add `module list` and `printenv` or equivalent to the job scripts to at least cache what the environment was, then ok to merge this.

Issue #312, opened 2016-11-18
corrupted brick file generation


sbailey:
This command generates a corrupted brick file:
```
quickbrick -b test2 --objtype BGS -n 100 \
--outdir temp \
--seed 784077896 --airmass 1.5 --verbose --zrange-bgs 0.1 0.2 \
--moon-phase 0.1 --moon-angle 30 --moon-zenith 20
```
Note: there is an open PR to merge quickbrick and quickgen to reduce identical code.  This was tested on desisim commit 620f28a344b58627e340e8603311eb80427cd97d.

Opening one of the `temp/brick-*-test2.fits` files has the warning:
```
WARNING: VerifyWarning: Error validating header for HDU #5 (note: Astropy uses zero-based indexing).
    'ascii' codec can't decode byte 0xa2 in position 82: ordinal not in range(128)
There may be extra bytes after the last HDU or the file is corrupted. [astropy.io.fits.hdu.hdulist]
```
The appears to be coming from the header reporting the size of the fibermap data (HDU4) to be different from what it actually is, causing it to start parting HDU5 at the wrong location.  I think this comes from `desispec.io.brick.Brick.add_objects` where the fibermap table and header are updated in-place, but I haven't gotten further than that.

gdhungana:
This seems same but adding here with std error output when running quickbrick. This occured with previous unmerged quickbrick and now merged in quickgen also.

 quickgen --brickname "moon09" --objtype "ELG" --nspec 1000 --seed 1234 --moon-phase 0.9 --moon-angle 80 --moon-zenith 60 > moon09.log 2>&1

INFO:quickgen.py:275:main: Initializing SpecSim with config desi
INFO:quickgen.py:329:main: Generating ELG template
INFO:io.py:601:read_basis_templates: Reading /home/govinda/Desi/spectro/templates/basis_templates/v2.2/elg_templates_v2.0.fits
Traceback (most recent call last):
  File "/home/govinda/Desi/softwares/desisim/bin/quickgen", line 11, in <module>
    sys.exit(quickgen.main(args))
  File "/home/govinda/Desi/softwares/desisim/py/desisim/scripts/quickgen.py", line 644, in main
    fx = fits.open(filepath, mode='append')
  File "/usr/local/lib/python2.7/dist-packages/astropy/io/fits/hdu/hdulist.py", line 139, in fitsopen
    return HDUList.fromfile(name, mode, memmap, save_backup, cache, **kwargs)
[...]

PR #314, opened 2016-11-30
fix hdu update in brick I/O, add relevant test

gdhungana:
This PR
- fix the HDU update in desispec.io.brick so that fits file written is not corrupted ( #312 )
- adds a unit test that fails in the previous case but passes for this update. The written file is reopened and appended to confirm if the file written is correct.


sbailey:
Thanks for this fix.  I updated doc/changes.rst in master post-merge.

PR #315, opened 2016-12-02
Major cleanup of pipeline documentation

tskisner:
Major cleanup of pipeline documentation:

 - Update install instructions with examples for Linux and OS X

 - Expand pipeline use section to include examples of desi_pipe_status

 - Add new section on pipeline internals which covers details about
   the processing and how to add new workers


sbailey:
Travis claims the tests pass, but the build_sphinx test actually blew up due to a missing yaml dependency.  Please add that to the travis install.

Documentation updates themselves look good.

tskisner:
Interesting.  So the sphinx_build test runs using the system python?  I see pyyaml is already in the conda dependencies.  I'll try adding python-yaml to the APT list then.

weaverba137:
You should not add python-yaml, because the Travis test system uses conda, not the system Python install.

tskisner:
ok, I see that it didn't help.  But pyyaml is already in the list of conda packages.  So why can't sphinx "import yaml" ?

weaverba137:
That remains to be seen.

tskisner:
Ok, I'll keep poking at this, apologies for lots of test pushes :-)

tskisner:
I think I just need to add CONDA_DEPENDENCIES to the build_sphinx entry in the matrix.

weaverba137:
See if astropy is reporting any problems today.

weaverba137:
Yes, that should help.

tskisner:
That worked, just cleaning up some warnings now.

sbailey:
Reading these in more detail in their pretty readthedocs form, I noticed that "Installing Dependencies on an OS X System" is based on homebrew whereas elsewhere we recommend using conda.  Would it be possible to use conda for the python side and homebrew for the MPI/boost side?

tskisner:
Conda's openmpi seems broken for me, and you can't force conda mpi4py to
use non-conda openmpi.  I know you love conda, but it has some drawbacks.

PR #316, opened 2016-12-12
Test that NIST data files are read correctly.

weaverba137:
This PR adds a test to catch problems related to #289.

Also note that this sets `use_2to3 = False` in the top-level setup.py file.  Since we *have* converted to Python 3, we don't need this run.  For details on what this does or doesn't do, see http://setuptools.readthedocs.io/en/latest/python3.html

This branch is a little old now, but it should merge fine.


sbailey:
#317 was merged to fix xrange vs. range, but py3 tests are still failing (py2 is ok):
```
FAIL: test_parse_nist (desispec.test.test_bootcalib.TestBoot)
Test parsing of NIST arc line files.
  File "/home/travis/build/desihub/desispec/py/desispec/test/test_bootcalib.py", line 118, in test_parse_nist
    tbl = desiboot.parse_nist('CdI')
AssertionError: UnicodeDecodeError not raised
```
Please investigate.

weaverba137:
I'm surprised that the push is succeeding but the merge is not.  Did the NIST files change since October?

weaverba137:
Ah, of course.  The function `parse_nist()` already fixes the locale in master.  So the test is actually kind of vestigial.

PR #317, opened 2016-12-12
replace xrange with range

weaverba137:
`xrange()` is a Python 2 function that should not be in our code.

Please merge this *before* #316.


sbailey:
Close notes:
* we have boxcar tests that check this code and on my laptop they properly fail under python 3 due to the `xrange` call
* this `xrange` code was brought in via PR #293, where travis said they pass
* I suspect but am not sure that this is due to `setup_keywords['use_2to3'] = True` automagically converting `xrange` to `range`, thus obscuring the problem
* PR #316 turns off `use_2to3` (good!)

Merging now.

PR #326, opened 2017-01-10
Pipeline improvements and tests

tskisner:
This commit updates several aspects of the pipeline:
 
 * Adds new unit tests for pipeline operations and
   removes stale tests.

 * Adds new helper functions and unit tests for
   graph path operations.

 * Adds functionality to desi_pipe to generate batch
   scripts that use shifter images.


tskisner:
Ok, this is ready for merge.  There will be some small fixes for shifter slurm scripts coming, but I want to get this merged before it gets too out of sync with master.

py/desispec/test/test_pipeline_run.py
```diff
@@ -4,145 +4,84 @@
...
-# from desispec.pipeline.run import *
-# import desispec.io as io
+import subprocess as sp
 
-#- TODO: override log level to quiet down error messages that are supposed
-#- to be there from these tests
+from desispec.pipeline.common import *
+from desispec.pipeline.graph import *
+from desispec.pipeline.plan import *
+from desispec.pipeline.run import *
```
sbailey: Stylistically let's avoid using "import *" in the future since it makes it harder for people less familiar with the code to know which of the packages a given function/class came from.  In this case it is within some test code so it isn't that big a deal and I won't block the PR for it.

sbailey:
For the record: this is another case where github defaulted to "squash and merge" rather than a vanilla merge commit.  Ted is the only contributor and several of the commits are bug fixes to get 2.7 tests running so they aren't particularly interesting by themselves, so I will proceed with the squash and merge.

PR #333, opened 2017-01-16
Qa refactor

rkehoe:
This pull request facilitates integration of QAs from quicklook at the offline spectroscopic pipeline.  It follows discussion with @profxj and implementation consistent with those discussions.  It 
includes:

- modification methods for signal-to-noise and sky residuals calculation, refactoring QA code to be reused readily in both quicklook and offline pipelines
- inclusion of signal-to-noise QA into offline QA for first time
- integration of data model of these two QAs at sky-subtraction step of pipeline
- integration of signal-to-noise into plotting for offline
- modification of quicklook to use new refactored code
- modification test_biscripts to accomdate changes
- addition of new unit tests for new refactored computation methods
- minor fixes for plotting

This will be used as a template for integrating further QAs across offline/quicklook pipelines, and implementing a common data model.

### DIFF
--- a/py/desispec/qa/qa_plots.py
+++ b/py/desispec/qa/qa_plots.py
@@ -99,20 +99,20 @@ def frame_skyres(outfil, frame, skymodel, qaframe):
     """
 
     # Sky fibers
-    skyfibers = np.where(frame.fibermap['OBJTYPE'] == 'SKY')[0]
-    assert np.max(skyfibers) < 500  #- indices, not fiber numbers
+    ##skyfibers = np.where(frame.fibermap['OBJTYPE'] == 'SKY')[0]
+    ##assert np.max(skyfibers) < 500  #- indices, not fiber numbers
 
     # Residuals
-    res = frame.flux[skyfibers] - skymodel.flux[skyfibers] # Residuals
-    res_ivar = util.combine_ivar(frame.ivar[skyfibers], skymodel.ivar[skyfibers])
-    med_res = np.median(res,0)
+    ##res = frame.flux[skyfibers] - skymodel.flux[skyfibers] # Residuals
+    ##res_ivar = util.combine_ivar(frame.ivar[skyfibers], skymodel.ivar[skyfibers])
+    ##med_res = np.median(res,0)
 
     # Deviates
-    gd_res = res_ivar > 0.
-    devs = res[gd_res] * np.sqrt(res_ivar[gd_res])
+    ##gd_res = res_ivar > 0.
+    ##devs = res[gd_res] * np.sqrt(res_ivar[gd_res])
 
     # Calculations
-    wavg_res = np.sum(res*res_ivar,0) / np.sum(res_ivar,0)
+    ##wavg_res = np.sum(res*res_ivar,0) / np.sum(res_ivar,0)
     '''
     wavg_ivar = np.sum(res_ivar,0)
     chi2_wavg = np.sum(wavg_res**2 * wavg_ivar)
@@ -121,6 +121,8 @@ def frame_skyres(outfil, frame, skymodel, qaframe):
     chi2_med = np.sum(med_res**2 * wavg_ivar)
     pchi2_med = scipy.stats.chisqprob(chi2_med, dof_wavg)
     '''
+    med_res = qaframe.qa_data['SKYSUB']["METRICS"]["MED_RESID_WAVE"]
+    wavg_res = qaframe.qa_data['SKYSUB']["METRICS"]["WAVG_RES_WAVE"]
 
     # Plot
     fig = plt.figure(figsize=(8, 10.0))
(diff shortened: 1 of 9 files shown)


py/desispec/sky.py
```diff
@@ -233,55 +231,30 @@ def qa_skysub(param, frame, skymodel, quick_look=False):
...
+        param = dict(
+                     PCHI_RESID=0.05, # P(Chi^2) limit for bad skyfiber model residuals
+                     PER_RESID=95.,   # Percentile for residual distribution
+                     )
+    #qadict_resid = qa.sky_resid(param, frame, skymodel, quick_look=quick_look)
+    #qadict=qadict_resid.copy()
+
+    #- QAs
+    #- first subtract sky to get the sky subtracted frame. This is only for QA. Pipeline does it separately. 
+    tempframe=copy.deepcopy(frame) #- make a copy so as to propagate frame as it is
```
profxj: Again, we should avoid deepcopy()

It would be better to add a copy() method to the Frame class.

py/desispec/sky.py
```diff
@@ -233,55 +231,30 @@ def qa_skysub(param, frame, skymodel, quick_look=False):
...
-    # Output dict
-    qadict = {}
-    qadict['NREJ'] = int(skymodel.nrej)
-
-    # Grab sky fibers on this frame
-    skyfibers = np.where(frame.fibermap['OBJTYPE'] == 'SKY')[0]
-    assert np.max(skyfibers) < 500  #- indices, not fiber numbers
-    nfibers=len(skyfibers)
-    qadict['NSKY_FIB'] = int(nfibers)
+    from desispec.qa import qa_quicklook as qa
```
profxj: Please name as ql_qa or qa_ql instead

py/desispec/sky.py
```diff
@@ -233,55 +231,30 @@ def qa_skysub(param, frame, skymodel, quick_look=False):
...
+    #qadict_resid = qa.sky_resid(param, frame, skymodel, quick_look=quick_look)
+    #qadict=qadict_resid.copy()
+
+    #- QAs
+    #- first subtract sky to get the sky subtracted frame. This is only for QA. Pipeline does it separately. 
+    tempframe=copy.deepcopy(frame) #- make a copy so as to propagate frame as it is
+    subtract_sky(tempframe,skymodel)
+
+    qadict_resid = qa.sky_resid(param, tempframe, skymodel, quick_look=quick_look)
+    qadict=qadict_resid.copy()
```
profxj: Rather odd to generate then copy for qadict.

py/desispec/sky.py
```diff
@@ -233,55 +231,30 @@ def qa_skysub(param, frame, skymodel, quick_look=False):
...
+
+    qadict_snr = qa.SignalVsNoise(tempframe,param)
```
profxj: I'm not convinced we want this offline but someone can argue otherwise.

If not, then wrap inside:

```
if quick_look:
```

py/desispec/qa/qa_quicklook.py
```diff
@@ -214,6 +217,84 @@ def integrate_spec(wave,flux):
     integral=np.trapz(flux,wave)
     return integral
 
+def sky_resid(param, frame, skymodel, quick_look=False):
```
profxj: Why move this method inside qa_quicklook?

If isn't specific to quicklook and I would argue that
having it within the desispec.sky module makes it
easier to understand what is being used.

profxj:
Have reviewed the code and offer suggested edits.

As a related aside, I note that the qa_quicklook.py module contains
a number of methods that could be more generally useful.  Someone
should consider whether any of these would go best in desiutils.

py/desispec/qa/qa_plots.py
```diff
@@ -156,30 +159,99 @@ def frame_skyres(outfil, frame, skymodel, qaframe):
...
+    ##binsz = 0.1
+    ##i0, i1 = int( np.min(devs) / binsz) - 1, int( np.max(devs) / binsz) + 1
+    ##rng = tuple( binsz*np.array([i0,i1]) )
+    ##nbin = i1-i0
     # Histogram
-    hist, edges = np.histogram(devs, range=rng, bins=nbin)
+    ##hist, edges = np.histogram(devs, range=rng, bins=nbin)
+
+    binsz = qaframe.qa_data['SKYSUB']["PARAMS"]["BIN_SZ"]
+    hist = np.asarray(qaframe.qa_data['SKYSUB']["METRICS"]["DEVS_1D"])
```
rkehoe: Initially what was there was a sky residual calculation repeated in
   the plotting routine itself, even though it was already done in the sky
   resid function.  In addition, the hist and edges quantities are metrics
   of the physical distribution of residuals like other QAs.  This
   calculation is also something we would want in quicklook, for instance.
   By refactoring, we omit the duplicate calculations, and modularize the
   code so that sharing offline and online is straightforwad, which is a
   main part of the purpose of this PR.  The offline plotting
   functionality of residuals is preserved.

py/desispec/qa/qa_plots.py
```diff
@@ -156,30 +159,99 @@ def frame_skyres(outfil, frame, skymodel, qaframe):
...
     ax1.plot(xppf, area*scipy.stats.norm.pdf(xppf), 'r-', alpha=1.0)
 
     ax1.set_xlabel(r'Res/$\sigma$')
     ax1.set_ylabel('N')
     ax1.set_xlim(xmin,xmax)
 
-
     # Meta text
+    #- limit the dictionary to residuals only for meta
+    qaresid=copy.deepcopy(qaframe)
```
rkehoe: We started to replace with a shallow copy but ran into the problem we
 had originally where this keeps the references to subsidiary  
 information in the original compound object.  This is what we are
    trying to avoid by doing the copy, so we have left it.  If a copy() is 
    added to the Frame class, we can certainly use that later, but for now
    we've reverted to the deepcopy usage.  There are other places in the
    pipeline where deepcopy is used.

py/desispec/qa/qa_plots.py
```diff
@@ -156,30 +159,99 @@ def frame_skyres(outfil, frame, skymodel, qaframe):
...
+    resid_keys=['NREJ','NSKY_FIB','NBAD_PCHI','MED_RESID','RESID_PER']
+    qaresid.qa_data['SKYSUB']['METRICS']={key:value for key,value in qaframe.qa_data['SKYSUB']
+                                         ['METRICS'].items() if key in resid_keys}
+
     ax2 = plt.subplot(gs[1,1])
     ax2.set_axis_off()
-    show_meta(ax2, qaframe, 'SKYSUB', outfil)
+    show_meta(ax2, qaresid, 'SKYSUB', outfil)
+    
+    #- SNR Plot
```
rkehoe: Yes, the plots worked when running the offline pipeline.  We tested.
   It only prints sky resid meta quantities.  In the future, there are
   likely some SNR quantities to report, but we leave that for subsequent
   update.

rkehoe:
This update accomplishes further modifications and tests to accomodated questions.  If this PR is signed off, please indicate it would be preferred to clean up commented out former code before merge.

py/desispec/qa/qa_plots.py
profxj: Ok, this is fine.

profxj:
These modifications look good to me.  I think the only lingering concern I have
is that the QA parameters for offline/online are set in different places and can
have different values.  I suppose that is ok and maybe even desirable but we
need to be aware of this.

rkehoe:
Agreed on the issue with PARAMs being set differently online and offline.  It would be better to work toward more commonality in a future PR.

Should I go ahead and remove the commented out code before merging?

profxj:
Yes, please remove and merge.

rkehoe:
Cleanup completed.

PR #341, opened 2017-02-09
allow bootcalib tests to run even if NERSC is down

sbailey:
This PR fixes a bug in test_bootcalib, which requires downloading some example data from NERSC (which is still unavailable today).  The test had logic for a failure to download, but it wasn't catching the case of the server returning a 403 error code and some HTML text instead of a fits file.

I would appreciate a cross check from someone like @tskisner or @weaverba137: this code considers an HTTP return code of 200 as good, and any other code as bad.  Are there any other return codes that should be considered a success (418 perhaps)?

Thanks to @gdhungana for reporting the problem.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -9,6 +9,7 @@ Desispec Change Log
 * close PSF file after initializing PSF object
 * fix graph_path usage in workers
 * update io.write_raw to enable writing simulated raw data with new headers
+* allow test_bootcalib to run even if NERSC portal is returning 403 errors
 
 0.12.0 (2016-11-09)
 -------------------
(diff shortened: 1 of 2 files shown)


weaverba137:
As a practical matter, only accepting 200 is probably fine, but if you want to be thorough, you might need to consider accepting codes in the 3xx range.  See https://en.wikipedia.org/wiki/List_of_HTTP_status_codes.

sbailey:
It looks like the 3xx status codes require further action (e.g. redirecting to a different URL), which isn't worth implementing for this case.  The other 2xx codes are successes but don't see to apply to the case of "give me this file".  I'll go ahead and merge this and if we see other non-200 codes pop up in practice, we can add those as needed.

Issue #310, opened 2016-11-10
move desispec.brick elsewhere


sbailey:
**Summary**: move `desispec.brick` to desimodel, desitarget, or desiutil to avoid the dependency loop introduced in desihub/desitarget#112 .

`desispec.brick` contrains some utility functions for calculating the mapping from RA,dec to brickname for various sized bricks, but it isn't actually used anywhere in desispec itself, which just propagates the input bricknames without questioning them.

It is used by desisim when generating fake input data, and recently it is also used by desitarget.mocks when generating mock input catalogs.  This creates a dependency loop since desispec depends upon desitarget (to get mask bit definitions) and now desitarget depends on desispec (to get brick info).  Ugh.

Candidate locations to move it would be desitarget (it is related to targeting), desimodel (analogous to keeping footprint and tiling info), or perhaps desiutil.

Related: At the same time we should consider refactoring it into using a brick definition file from the imaging surveys rather than recalculating the ra,dec vs. brickname from scratch.  The current implementation has some rounding differences with the IDL version used by the imaging surveys resulting in slightly different brick definitions.

Also related: the imaging surveys continue to use 0.25x0.25 deg2 bricks, while we have been optimistically assuming that they will eventually switch to 0.5x0.5 deg2 bricks that have a better match for number of spectra per file and inode limits.  With each imaging data release that looks less and less likely.  We may have to break our organizational dependency upon imaging survey bricks if they continue to use the tiny bricks.  We will have to face that at some point, but let's not let that issue become a blocking factor for moving this module to avoid the desitarget/desispec dependency loop.

moustakas:
```
Also related: the imaging surveys continue to use 0.25x0.25 deg2 bricks, while we have been 
optimistically assuming that they will eventually switch to 0.5x0.5 deg2 bricks that have a better 
match for number of spectra per file and inode limits. With each imaging data release that looks less 
and less likely. We may have to break our organizational dependency upon imaging survey bricks if 
they continue to use the tiny bricks. We will have to face that at some point, but let's not let that 
issue become a blocking factor for moving this module to avoid the desitarget/desispec 
dependency loop.
```
@dstndstn should comment, but for (the forthcoming, ~Jan 2017) DR4 we should be able to more easily "repackage" the Tractor catalogs into larger-sized bricks. 0.5x0.5 deg2 bricks may be back on the table!

djschlegel:
[...]

PR #343, opened 2017-02-13
speed up brick.Bricks.brickname

moustakas:
This speeds up ```brick.Bricks.brickname``` by looping on each unique brick row rather than looping on each object in the input array (see #310).

With 5M input objects the previous version took 6.81s vs 2.98s now.  Here's the snippet of code used for testing.

```
from time import time
import numpy as np
from desispec.brick import Bricks

B = Bricks()
rand = np.random.RandomState(444)

nobj = 5000000
ra = rand.uniform(0.1, 359.1, nobj)
dec = rand.uniform(-89.0, 89.0, nobj)

%time brickname = B.brickname(ra, dec)
```

### DIFF
--- a/py/desispec/brick.py
+++ b/py/desispec/brick.py
@@ -75,11 +75,11 @@ class Bricks(object):
 
     def brickname(self, ra, dec):
         """Return string name of brick that contains (ra, dec) [degrees]
-        
+
         Args:
             ra (float) : Right Ascension in degrees
             dec (float) : Declination in degrees
-            
+
         Returns:
             brick name string
         """
@@ -90,7 +90,7 @@ class Bricks(object):
 
         ncol = self._ncol_per_row[irow]
         jj = (ra/360.0 * ncol).astype(int)
-        names = np.empty(len(ra), dtype='S8')
+        names = np.empty(len(ra), dtype='U8')
         for thisrow in set(irow):
             these = np.where(thisrow == irow)[0]
             names[these] = np.array(self._brickname[thisrow])[jj[these]]
@@ -98,7 +98,7 @@ class Bricks(object):
         if np.isscalar(inra):
             return names[0]
         else:
-            return np.array(names)
+            return names
 
     def brick_radec(self, ra, dec):
         """Return center (ra,dec) of brick that contains input (ra, dec) [deg]
--- a/py/desispec/brick.py
+++ b/py/desispec/brick.py
@@ -94,7 +94,7 @@ class Bricks(object):
         for thisrow in set(irow):
             these = np.where(thisrow == irow)[0]
             names[these] = np.array(self._brickname[thisrow])[jj[these]]
-            
+
         if np.isscalar(inra):
             return names[0]
         else:
--- a/py/desispec/brick.py
+++ b/py/desispec/brick.py
@@ -87,12 +87,14 @@ class Bricks(object):
         dec = np.atleast_1d(dec)
         ra = np.atleast_1d(ra)
         irow = ((dec+90.0+self._bricksize/2)/self._bricksize).astype(int)
-        names = list()
-        for i in range(len(ra)):
-            ncol = self._ncol_per_row[irow[i]]
-            j = int(ra[i]/360 * ncol)
-            names.append(self._brickname[irow[i]][j])
 
+        ncol = self._ncol_per_row[irow]
+        jj = (ra/360.0 * ncol).astype(int)
+        names = np.empty(len(ra), dtype='S8')
+        for thisrow in set(irow):
+            these = np.where(thisrow == irow)[0]
+            names[these] = np.array(self._brickname[thisrow])[jj[these]]
+            
         if np.isscalar(inra):
             return names[0]
         else:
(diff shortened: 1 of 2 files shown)


moustakas:
Ugh, tests are broken because of a unicode issue I do not yet fully understand.  Specifically, with my proposed changes the bricknames end up with a "b" prefix, e.g., ```b'0434m165'``` which I know I should understand but don't.

weaverba137:
You've declared `names` as a numpy array (which contains bytes not str) instead of a list.  So the values in `names` are bytes.

weaverba137:
I'm trying to fix this up.

py/desispec/brick.py
```diff
@@ -75,28 +75,30 @@ def __init__(self, bricksize=0.5):
...
+        names = np.empty(len(ra), dtype='U8')
+        for thisrow in set(irow):
+            these = np.where(thisrow == irow)[0]
+            names[these] = np.array(self._brickname[thisrow])[jj[these]]
 
         if np.isscalar(inra):
             return names[0]
         else:
-            return np.array(names)
+            return names
```
weaverba137: Already fixed, but since `names` is already an array, calling `np.array()` on it again is useless.

py/desispec/brick.py
```diff
@@ -75,28 +75,30 @@ def __init__(self, bricksize=0.5):
...
         irow = ((dec+90.0+self._bricksize/2)/self._bricksize).astype(int)
-        names = list()
-        for i in range(len(ra)):
-            ncol = self._ncol_per_row[irow[i]]
-            j = int(ra[i]/360 * ncol)
-            names.append(self._brickname[irow[i]][j])
+
+        ncol = self._ncol_per_row[irow]
+        jj = (ra/360.0 * ncol).astype(int)
+        names = np.empty(len(ra), dtype='U8')
```
weaverba137: In other parts of this function, you take into account the brick size.  However, here you assume the brick name has exactly 8 characters, which is only true for bricks of a certain size.

py/desispec/test/test_brick.py
```diff
@@ -1,31 +1,53 @@
...
+        self.assertTrue((bricknames == self.names).all())
 
     def test3(self):
-        brickname = desispec.brick.brickname(np.array(self.ra), np.array(self.dec))
-        self.assertEqual(len(brickname), len(self.ra))
-        self.assertTrue(np.all(brickname == self.names))
-                
-if __name__ == '__main__':
-    unittest.main()
+        """This test is nonsensical, because it is a repeat of test2.
```
weaverba137: This is just a repeat of `test_brickname_array`.  Specifically, since `self.ra` and `self.dec` are *already* arrays, calling `np.array()` on them does nothing.

py/desispec/brick.py
sbailey: Commentary: In an earlier version of the code, `names` started out as a list that was appended and then was converted to an array here at the end.  In other cases where you are dealing with user input that might be a list or an array, a good trick to know is `np.asarray`: if the input was already an array it does a fast pass-through (faster than `np.array` which creates a new array), otherwise it does the slower operation of converting it to an array.  I don't think that applies in this particular case.

py/desispec/test/test_brick.py
sbailey: At the risk of bike shedding: it would be good to confirm that the code works with list inputs too, i.e. the removed `test3` should have been `test_brickname_list` with a cast to `list` instead of a redundant case to `array`.

weaverba137:
I've converted the redundant test into an explicit test of lists of floats to names.

py/desispec/test/test_brick.py
weaverba137: Fixed.

weaverba137:
I think the only outstanding review item is the one about the number of characters in a brick name, which depends on the brick size.

py/desispec/brick.py
moustakas: Thanks, @weaverba137.  For the record, I didn't write this function.  However, I think it's OK for ```brickname``` to be hard-coded to 8 characters, which gives four significant figures in RA and three significant figures in Declination (with a p/m to indicate the sign of the declination).  What's nice about this is that it works well with the 0.25x0.25 deg2 bricks currently in use by the imaging surveys, although it also works with the 0.5x0.5 deg2 bricks we're using in the mocks.  We'll never go smaller than 0.25 deg on a side and even with 1 deg bricks it means we'll be carrying around a few extra bytes -- no big deal.

py/desispec/brick.py
moustakas: Thanks, yeah, this bit of code was there previously when ```names``` was a list.  I just neglected to remove it.

py/desispec/brick.py
weaverba137: I don't understand how that can be possible, unless the 0.25 degree bricks actually drop some significant figures to fit into that space.  I was not aware of that "feature" and I don't particularly like it.  It implies that it is not possible, even in principle, to deduce the brick size from the brick name.

weaverba137:
I am willing to open a separate issue on the number of characters in a brick name versus brick size.

py/desispec/brick.py
moustakas: That's correct.  The documentation from DECaLS is [here](http://legacysurvey.org/dr3/files) and screenshot below.  

screen shot 2017-02-13 at 8 17 17 pm

Issue #339, opened 2017-02-08
Improve fiberflat


julienguy:
Tests of desi_compute_fiberflat on teststand data and BOSS data show the code needs to be improved.
Work in progress on branch desi_on_boss by Nicolas.

ngbusca:
Tests were failing due to small scale (~ 50A) features not being captured by the default modelling with a smoothing scale at 100 A. Lowering the default value to 20A solved the problem. 

Do we want the code to automatically adapt the smoothing scale? These small scale features are a property of the spectrograph and are unlikely to change over the course of the survey (unless the spectrograph is replaced). I think it might be an overkill for something relatively simple (i.e. the fiberflat).

sbailey:
What you did in PR #344 is good: the smoothing resolution is an optional parameter to `compute_fiberflat`, with a default that matches what DESI needs.  We don't need the code to auto-adapt the smoothing scale.

I'm running the integration test on the desi_on_boss branch as a final check and then will merge PR #344 and close this issue.  If there are other problems that need to be fixed that weren't addressed in PR #344, please reopen this with a specific description.

PR #344, opened 2017-02-17
Desi on boss

ngbusca:
I'm opening this pull request with the modifications needed to run the fiberflat through make bricks on boss data. 

The main changes are the introduction of the svd decomposition when cholesky fails. This handles small ivars. 

Other modifications include adapting default parameters, moving hardcoded values to parameter options and minor bug fixes. 

The code has been tested on all (6) exposures of BOSS plate 4200 and a flatfield file from the test-bench.


weaverba137:
Please briefly describe the changes in the `doc/changes.rst` file.

sbailey:
The changes seem ok except that the fiberflat code seems much much slower than it used to be.  The `desi_compute_fiberflat` test has been running for 15 minutes on my laptop and is still only on the 3rd iteration.  Travis tests are failing due to running for 10 minutes without output and then timing out.

Is this slowdown known and necessary, or are the tests probing some case that you haven't encountered with real data?

ngbusca:
The only difference that will increase the running time is that the deconvolution loop ("2nd pass") is forced to iterate at least once. Filling the resolution matrix (used for the deconvolution) in this loop dominates the running time if many iterations are needed.

I've run a benchmark on pine (30 files). Flatfielding all 30 files with the master code took 31min vs 42min with the desi_on_boss code. 

Which input file is the integration test using? I'd like to take a look to understand what's failing.

sbailey:
I was just running the standard unit tests with "python setup.py test" in
the top level desispec directory.  I think that creates a fake flat on the
fly to use. Does that work for you?   Glad to hear that it is still fast on
real data and full pixsim examples.

Stephen

ngbusca:
I've found the problem (and fixed it), it was due to the test being run on a frame with only 20 pixels.

ngbusca:
I still have to address @weaverba137 comment

py/desispec/scripts/fluxcalibration.py
```diff
@@ -77,9 +77,14 @@ def main(args) :
...
-    model_fibers = model_fibers%500
-    if np.any(fibermap['OBJTYPE'][model_fibers] != 'STD'):
-        for i in model_fibers:
+
+    ## get the fiber indices
+    w = np.array([objtype.strip() == 'STD' for objtype in fibermap['OBJTYPE'][:]])
+    fibers = fibermap['FIBER'][:][w]
+    if np.any(fibers != model_fibers):
+        i_model_fibers = list(fibers)
+        i_model_fibers = [model_fibers.index(i) for i in model_fibers]
```
sbailey: This line is causing using test errors with the message
```
  File "/Users/sbailey/desi/git/desispec/bin/desi_compute_fluxcalibration", line 16, in <module>
    fluxcal.main(args)
  File "/Users/sbailey/desi/git/desispec/py/desispec/scripts/fluxcalibration.py", line 86, in main
    i_model_fibers = [model_fibers.index(i) for i in model_fibers]
  File "/Users/sbailey/desi/git/desispec/py/desispec/scripts/fluxcalibration.py", line 86, in <listcomp>
    i_model_fibers = [model_fibers.index(i) for i in model_fibers]
AttributeError: 'numpy.ndarray' object has no attribute 'index'
```
This happens within test_binscripts.py resulting in the final test error message
```
FAIL: test_compute_fluxcalib (__main__.TestBinScripts)
  File "py/desispec/test/test_binscripts.py", line 203, in test_compute_fluxcalib
    self.assertEqual(err, 0)
AssertionError: 1 != 0
```
Please check this code.  Note that line 86 blows away the `i_model_fibers` defined in line 85.  You might need to use `np.where(model_fibers == i)[0]` instead of `model_fibers.index(i)`.

Please also run `python setup.py test` to confirm that the unit tests run locally.

ngbusca:
@weaverba137 should I list the changes under a new version number or add them under the current 0.12.1 (unreleased)?

weaverba137:
Add them under the current 0.12.1 (unreleased) please.

sbailey:
I also did a final check that this works with the integration test (it passes).  Merging now.  This fixes #339 (AFAIK).

PR #345, opened 2017-02-18
add optional bricksize input to brick.brickname

moustakas:
Minimal change to the `brick.brickname` function (*not* the `Bricks.brickname` method) to make the bricksize a variable input.  This input is needed for https://github.com/desihub/desitarget/issues/136.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -10,6 +10,7 @@ Desispec Change Log
 * fix graph_path usage in workers
 * update io.write_raw to enable writing simulated raw data with new headers
 * allow test_bootcalib to run even if NERSC portal is returning 403 errors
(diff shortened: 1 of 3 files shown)


sbailey:
Since the `_bricks` object is cached to avoid recreating it every time `brickname` is called, this code wouldn't have worked if the same program tried to calculate things for two different brick sizes.

I added a test to verify that :) and code to work around that so that you can call `brickname` with different brick sizes and have it work correctly.

@moustakas if the Travis tests pass and you agree with this change, go ahead and merge.

moustakas:
Thank you @sbailey I made this change on the airplane bleary-eyed and didn't think it through.

PR #351, opened 2017-02-25
Quicklook

rstaten:
This PR:

- Adds a function to use psfboot Xsigma as a zeroth order proxy of psf resolution. Subsequent PR will update better approximation for resolution in dispersion direction.
- Adds Wsigma plot to XWsigma quicklook QA.
- Removes SNR threshold, allowing negative values to be included in SNR calculation.
- Fixes histogram plotting issue for several quicklook QAs.


py/desispec/quicklook/procalgs.py
```diff
@@ -6,10 +6,40 @@
...
+    nwave=flux.shape[1]
+    resolution_data=np.zeros((nspec,21,nwave))
+    if usepsfboot:
+        if hasattr(psf,'xsigma_boot'): #- only use if xsigma comes from psfboot
+            log.info("Getting resolution matrix band diagonal elements from constant Gaussing Xsigma")
+            for ispec in range(nspec):
+                thissigma=psf.xsigma(ispec,wave) 
+                Rsig=Resolution(thissigma)
+                resolution_data[ispec]=Rsig.data
+    return resolution_data
```
sbailey: This function tests for the existence of `psf.xsigma_boot`, but then uses `psf.xsigma`.  Is that what was intended?

py/desispec/qa/qalib.py
```diff
@@ -301,8 +301,7 @@ def SN_ratio(flux,ivar):
     medsnr=np.zeros(flux.shape[0])
     #totsnr=np.zeros(flux.shape[0])
     for ii in range(flux.shape[0]):
-        signalmask=flux[ii,:]>0 #- mask negative values
-        snr=flux[ii,signalmask]*np.sqrt(ivar[ii,signalmask])
+        snr=flux[ii]*np.sqrt(ivar[ii])
         medsnr[ii]=np.median(snr)
```
sbailey: I'd like to check with Kyle and David about what we do for BOSS SoS (quicklook) in the presence of noisy negative signal when calculating S/N .  If you have other info that motivates this choice, please comment.

rstaten: @sbailey We decided to change this to reduce a biased S/N calculation, but input would be appreciated. The plots below show the SNR QA with all values (top) and no negative values (bottom), if that helps.

allvalues
no_negatives

py/desispec/quicklook/procalgs.py
gdhungana: Hi @sbailey Yes this is intended here. The reason is that xsigma is meant to be more general (may be calculated from different ways) but is equal to xsigma_boot when the psf used is coming from psfboot.

I have added a function to desispec.psf.PSF in the same commit, 

def xsigma(self,ispec,wave):
+        if hasattr(self,'xsigma_boot'):
+            if np.isscalar(wave):
+                return self.xsigma_boot[ispec]
+            else:
+                return np.full(len(wave),self.xsigma_boot[ispec]) #- constant xsigma for a given fiber

For now this is only taking the case where per fiber xsigma comes from psfboot and making copies when the wavelength is a vector, but this function is intended to be more general for case when there may/will be different approach to measure xsigma. This way the same function is sufficient given the content in psf object (coming from psfboot or otherwise). If this is conflicting, please let me know and I'll only have xsigma_boot explicitly for now and worry about xsigma when we really have a different approach for it.

sbailey:
For comparison, here is the equivalent S/N code from BOSS (idlspec2d/pro/apo2d/quickextract.pro lines 355-374)
```idl
   ;----------
   ; Compute average (but median-filtered) flux and signal-to-noise
  
   meanflux = fltarr(nfiber)
   meansn = fltarr(nfiber)
   for ifib=0, nfiber-1 do begin
      ; Select unmasked pixels in the wavelength range for this fiber
      iwave = where(logwave[*,ifib] GT alog10(wrange[0]) $
                AND logwave[*,ifib] LT alog10(wrange[1]) $
                AND objsubivar[*,ifib] GT 0, nwave)
      if (nwave GT 0) then begin
         meanfluxvec = djs_median( flux[iwave,ifib], width=filtsz<nwave, $
          boundary='reflect' )
         meansnvec = djs_median( objsub[iwave,ifib] $
          * sqrt(objsubivar[iwave,ifib]), width=filtsz<nwave, $
          boundary='reflect' )
         meanflux[ifib] = djs_mean(meanfluxvec)
         meansn[ifib] = djs_mean(meansnvec)
      endif
   endfor
```
i.e. the quoted "S/N" is the mean of a median filtered skysubtractedflux/err vector, without clipping negatives or taking absolute values.

PR #353, opened 2017-02-28
fixed bug in sky subtraction

ngbusca:
This PR fixes a bug in sky subtraction (sky.py) that was causing a crash whenever the final cholesky decomposition succeeded.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -11,6 +11,7 @@ Desispec Change Log
 * update io.write_raw to enable writing simulated raw data with new headers
 * allow test_bootcalib to run even if NERSC portal is returning 403 errors
 * add Brick.bricksize property; allow brick.brickname to specify bricksize
+* do svd inverses when cholesky decompositions fail in fiberflat, sky subtraction,   and flux calibration
 
 0.12.0 (2016-11-09)
 -------------------
--- a/py/desispec/sky.py
+++ b/py/desispec/sky.py
@@ -140,7 +140,7 @@ def compute_sky(frame, nsig_clipping=4.,max_iterations=100) :
     # solve once again to get deconvolved sky variance
     try :
         skyflux,skycovar=cholesky_solve_and_invert(A.todense(),B)
-    except np.linalg.linalg.LinAlgError,e :
+    except np.linalg.linalg.LinAlgError :
         log.warning("cholesky_solve_and_invert failed, switching to np.linalg.lstsq and np.linalg.pinv")
         skyflux = np.linalg.lstsq(A.todense(),B)[0]
         skycovar = np.linalg.pinv(A.todense())
--- a/py/desispec/sky.py
+++ b/py/desispec/sky.py
@@ -136,10 +136,14 @@ def compute_sky(frame, nsig_clipping=4.,max_iterations=100) :
 
     log.info("nout tot=%d"%nout_tot)
 
-
+    
     # solve once again to get deconvolved sky variance
-    skyflux = np.linalg.lstsq(A.todense(),B)[0]
-    skycovar = np.linalg.pinv(A.todense())
+    try :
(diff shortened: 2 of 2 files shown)


sbailey:
@ngbusca is it known why the Cholesky decomposition is failing?  Is this rare or common on BOSS data?  Should this final attempt be wrapped in try/accept like lines 88-92 which try Cholesky first (faster) and then fall back to SVD only if needed?  I'm concerned that we might be hiding an ill conditioned matrix that might now be silently giving us wrong answers with SVD instead of crashing Cholesky.

@julienguy could you take a look at this and comment?

julienguy:
np.linalg.lstsq , np.linalg.pinv was called only once, but it's true it's 5 times slower than cholesky_solve_and_invert. So with Nicolas' agreement I commited a modification, first trying cholesky_solve_and_invert then the other if a linalg exception is raised.
I tested this modif. on pine sims with/without ivar=0 in a wavelength range to test the failure.

julienguy:
travis says "There was an error while trying to fetch the log" for one python 2.7 config. 
I give up and let this PR open for a travis guru.

sbailey:
The previous Travis failure Julien mentioned was a hiccup with the coveralls.io service.  I restarted that, but now we're getting python 2.7 (but not python 3.5) Travis errors related to QA plots.  I don't immediately see why this is triggered now, and I haven't been able to reproduce it on my laptop yet, but cutting and pasting the error here before restarting Travis again to see if this is random:
```
  File "/home/travis/build/desihub/desispec/bin/desi_compute_fluxcalibration", line 16, in <module>
    fluxcal.main(args)
  File "/home/travis/build/desihub/desispec/py/desispec/scripts/fluxcalibration.py", line 106, in main
    qa_plots.frame_fluxcalib(args.qafig, qaframe, frame, fluxcalib)
  File "/home/travis/build/desihub/desispec/py/desispec/qa/qa_plots.py", line 408, in frame_fluxcalib
    icalib = fluxcalib.calib[stdfibers[ii]][gdp]
IndexError: in the future, 0-d boolean arrays will be interpreted as a valid boolean index
```
and
```
ERROR: testSkyPeaks (desispec.test.test_ql_qa.TestQL)
  File "/home/travis/build/desihub/desispec/py/desispec/test/test_ql_qa.py", line 362, in testSkyPeaks
    resl=qa(inp,**qargs)
  File "/home/travis/build/desihub/desispec/py/desispec/quicklook/qas.py", line 16, in __call__
    return self.run(*args,**kwargs)
  File "/home/travis/build/desihub/desispec/py/desispec/qa/qa_quicklook.py", line 521, in run
    return self.run_qa(input_frame,paname=paname,amps=amps,psf=psf, qafile=qafile, qafig=qafig, qlf=qlf)
  File "/home/travis/build/desihub/desispec/py/desispec/qa/qa_quicklook.py", line 575, in run_qa
    peak1_flux=np.trapz(frame.flux[i,iwave1-dw:iwave1+dw+1])
TypeError: slice indices must be integers or None or have an __index__ method
```

sbailey:
Travis test was transient; I'll go ahead and merge this now but we'll need to watch out if this happens again in the future.

ngbusca:
@sbailey Note that the problem will always be ill conditioned whenever the number of un-masked data pixels is less than the number of model pixels or there are model pixels that are mapped to masked regions (i.e. further than the full width of the resolution). In these cases, linalg.lstsq will set the model to zero wherever is can't be determined from the data (and also the model covariance via the pseudo inverse linalg.pinv). I think this is the right behaviour. 

On boss data it happened a lot in my first tries where I was using the entire wavelength range. Now that I've reduced it to 3500-1000A the calls to lstsq are almost gone (but not completely, I'll take a look at the cases where it is called to try and understand why it's needed).

Issue #349, opened 2017-02-21
move desispec.log into desiutil.log


sbailey:
logging is general enough that we should move it into desiutil instead of desispec.  e.g. this would let desitarget and desimodel use the logging without creating a dependency loop with desispec.

Since a lot of code already gets it from desispec, we could ease the transition by temporarily leaving `desispec.log.get_logger()` behind.  It would print a DeprecationWarning and then proceed to return the desiutil logger.  Go ahead and change any usage of `desispec.log` that we know about, but don't immediately break it.

Other refactoring requests:
* Add hooks to turn timestamps on/off (see #348 and #346)
* Could we make this work like the astropy logger where you just do `from astropy import log` and then you can immediately use it with `log.warn(...)` etc.?  i.e. get rid of the call to `get_logger` entirely?

Issue #348, opened 2017-02-21
"Phantom" calls to desispec.log.get_logger()?


weaverba137:
Theoretically, when I initially start a Python program, `desispec.log.desi_logger` should be `None`.  It becomes defined the first time `desispec.log.get_logger()` is called.  However, what I'm finding is that `desispec.log.desi_logger` is already defined when a program starts up.  In particular:

```python
assert desispec.log.desi_logger is None
```

throws an AssertionError before the first time I call `get_logger()`.  Until this is solved, I can't implement an optional timestamp in the log infrastructure because `desi_logger` is already formatted.

weaverba137:
I temporarily forced `desispec.log.get_logger()` to raise an exception if it was called with settings I didn't like.  Turns out it is being called by `desispec.frame` at import time.

This is a great reason why you should import inside functions, not inside modules.

```
Traceback (most recent call last):
  File "/home/weaver/Documents/Code/git/desihub/desispec/bin/desi_load_quicksurvey", line 3, in <module>
    from desispec.io.quicksurvey import main
  File "/home/weaver/Documents/Code/git/desihub/desispec/py/desispec/io/__init__.py", line 20, in <module>
    from .frame import read_frame, write_frame
  File "/home/weaver/Documents/Code/git/desihub/desispec/py/desispec/io/frame.py", line 16, in <module>
    from ..frame import Frame
  File "/home/weaver/Documents/Code/git/desihub/desispec/py/desispec/frame.py", line 15, in <module>
    log = get_logger()
  File "/home/weaver/Documents/Code/git/desihub/desispec/py/desispec/log.py", line 42, in get_logger
    raise ValueError("I got called prematurely!")
ValueError: I got called prematurely!
```

weaverba137:
O, and it gets even better!  `desispec.frame` was importing and calling `desispec.log.get_logger()` at the module level, **AND THEN NOT EVEN USING IT!**

Sloppy code, people, sloppy code.

weaverba137:
`get_logger()` was also being called at the module level in:

* `desispec.io.frame` (and only one of the functions in that module uses the log)
* `desispec.io.qa`
* `desispec.qa.qa_exposure` (and wasn't using it at all)
* `desispec.qa.qa_frame` (only one small method in the module uses it)
* `desispec.qa.qa_prod`
* `desispec.preproc`
* `desispec.io.raw`

[...]

PR #356, opened 2017-02-28
Quicksurvey database

### LINKED ISSUE #320: Create schema for pipeline database
Based on @sbailey's slides from the OSU meeting this database should have this approximate structure:

* a brick table, with brickid or brickname as the primary key.
* a target table with targetid as the primary key.
* a truth table with arbitrary primary key, and a foreign key to the target table.
* a fiberassign/frame table with foreign keys to target and tile.
* a obslist table with a foreign key to the tile table.
* a tile table with tileid as the primary key.
* a zcat table with a foreign key to the target table.

This brings up a few questions that I can't find in my notes:

1. What is the relationship between target and truth? 1:1? 1:many? many:many?
2. In the fiberassign table, is every row a fiber?
3. The obslist table seems very similar to the frame table I developed for the raw data.  We need to capture the fact that a tile may be observed multiple times, possibly with different fiber assignments; in the frame table, the primary key was spectrograph arm plus exposure number.

### LINKED ISSUE #321: Create Implementation-independent SQL schemas
We should develop a framework that can create and connect to SQLite and Postgresql databases.  This involves writing SQL schemas (`CREATE TABLE`, etc.) in a generic form of SQL.  The API for the Python `sqlite` and `psycopg2` modules is nearly the same, so hopefully all one would need to do is have an easy means of choosing the database type.

### LINKED ISSUE #328: Load December 2016 Data Challenge into pipeline database
Once several individual issues are completed (or *as* they are completed), we should demonstrate the pipeline database functionality by loading the December 2016 data challenge.  For reference, the other issues are: #323, #321, #320, #310, #93.

weaverba137:
I'm creating this PR to be able to easily see the changes, but this is not ready to merge.

* Closes #320.
* Closes #321.
* Closes #328.

I will leave it up to reviewers as to whether this PR should also address #349, #348, #310 (probably not that one, but just thought I'd throw that out there).


sbailey:
General comments:
* This PR includes a lot of PEP8 stylistic cleanup and import methodology refactoring that isn't directly related to the core features being implemented.  Let's try to keep those changes separated in the future so that a PR about DB features doesn't touch hundreds of lines across 27 files, thus obscuring the most important changes to be reviewed.
* This PR introduces a dependency upon sqlalchemy, but that should be isolated as an optional dependency such that you don't need it unless you are actually loading a DB with results.  I added a check for sqlalchemy in `test_io` so that it will skip the database tests if sqlalchemy isn't installed.  Note that Travis *does* install sqlalchemy and run those tests -- good.
* `desispec.io.quicksurvey` is unfortunately special purpose for a specific data challenge and is applicable only to those quick simulations, not to normal data running.  I won't be dogmatic about it this time, but in the future we should make simulation runs produce something that looks more like the real directory structure such that loading it can be done without requiring quicksurvey specific code.  At that point we should refactor this to load a DB from "real" files/directories and deprecate/remove this quicksurvey specific code.
* This PR refactors how `desispec.log.get_logger` is imported and called, but let's do the swap to `desiutil.log.get_logger` in a separate PR; there are enough changes here already.  (related to #348 and #349)
* Similarly, let's not move `desispec.brick` (issue #310) in this PR.

@weaverba137 ok to merge; can you take care of the merge conflict with `doc/changes.rst` and do the merge by hand at the command line?

weaverba137:
@sbailey, thank you for the comments.  A couple of items:

> @weaverba137 ok to merge; can you take care of the merge conflict with doc/changes.rst and do the merge by hand at the command line?

I don't understand. I can fix the merge conflict on the command line, but why can't I then merge the PR in GitHub?

> This PR introduces a dependency upon sqlalchemy, but that should be isolated as an optional dependency such that you don't need it unless you are actually loading a DB with results.

On the one hand, I can understand not wanting more and more dependencies, but sqlalchemy is just a conda package.  It's *already* installed at NERSC, and is trivial to install in any anaconda distribution.  Are we attempting to bend over backwards to support people who for some reason don't want to install Anaconda?

sbailey:
Maybe I'm missing something simple about how to work around `doc/changes.rst` merge conflicts, but when github finds a merge conflict it points to instructions about manually doing the merge from the command line and pushing to master and bypassing the GUI.  Did you just edit `doc/changes.rst` by hand to match what was in master so that the merge conflict would go away, thus re-enabling the ability of the GUI to do the merge?

Regarding dependencies: I'm trying to avoid a situation where someone has working code, pulls from master, and then suddenly it no longer works because of a missing dependency that they don't actually need.  Even if it is trivial to install, that just trips up our development workflow.  It is useful to keep the distinction between our core external dependencies (e.g. numpy, scipy, astropy) and optional dependencies (e.g. sqlalchemy) that are needed only for special-case operations not used by most people.

PR #352, opened 2017-02-27
Changes that allow the pipeline to run within a shifter container.

tskisner:
desi_pipe now generates slurm scripts which use a specified docker image containing the desi software and dependencies.

PR #359, opened 2017-03-03
work around libc stdout/stderr unit test prob on Mac

sbailey:
PR #352 broke unit tests on a mac due to this code in `desispec.parallel`:
```python
libc = ctypes.CDLL(None)
c_stdout = ctypes.c_void_p.in_dll(libc, 'stdout')
c_stderr = ctypes.c_void_p.in_dll(libc, 'stderr')
```
Both @tskisner and Travis use Linux where this works, but on a Mac this causes:
```
...
  File "/Users/sbailey/desi/git/desispec/py/desispec/test/test_util.py", line 13, in <module>
    import desispec.parallel as dpl
  File "/Users/sbailey/desi/git/desispec/py/desispec/parallel.py", line 27, in <module>
    c_stdout = ctypes.c_void_p.in_dll(libc, 'stdout')
ValueError: dlsym(RTLD_DEFAULT, stdout): symbol not found
```
This PR provides the minimal workaround of only running that code on Linux.  `libc`, `c_stdout`, and `c_stderr` are only used in one place later to flush C stdout/stderr, and I think that code is only really used for full pipeline productions at NERSC.  But there might be a cleaner "real" way to do this for mac that would really do the flushing there too.

Assigning to @tskisner for review.

### DIFF
--- a/py/desispec/parallel.py
+++ b/py/desispec/parallel.py
@@ -13,8 +13,7 @@ import time
 import io
 from contextlib import contextmanager
 import logging
-
-import ctypes
+import platform
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


tskisner:
@sbailey can you try this latest commit on your mac?

sbailey:
Yep, works on a mac.  Thanks.  Merging now.

PR #361, opened 2017-03-13
Quicklook

rstaten:
This PR:

- Adds an integration test for Quicklook, creating the necessary input files and running the pipeline (only for r and z cameras, b to be added as well as other additions such as bright time etc.). This should catch issues such as the previous preprocessor change, which caused Quicklook to fail.
- Adds an XWSigma unit test.


py/desispec/test/test_ql_qa.py
```diff
@@ -286,19 +284,197 @@ def testGetRMS(self):
...
+        #- make the test pixfile, fibermap file
+        img_pix = rawimg
+        img_ivar = np.ones_like(img_pix) / 3.0**2
+        img_mask = np.zeros(img_pix.shape, dtype=np.uint32)
+        img_mask[200] = 1
+
+        #- manually insert sky peaks for xwsigma
+        zpeaks = np.array([8401.5,8432.4,8467.5,9479.4,9505.6,9521.8])
+        fibers = np.arange(5)
+        peak1 = PSF.xy(psf,fibers,zpeaks[0])
```
sbailey: This code should be
```python
peak1 = psf.xy(fibers, zpeaks[0])
```
`PSF` subclasses aren't supposed to override `.xy`, but if they ever did then calling `PSF.xy(psf, fibers, zpeaks[0])` would be calling the wrong `.xy` calculation.  Same comment applies to the lines below.

i.e. you shouldn't need to `import PSF` at the top.

OTOH, if `psf.xy(fibers, zpeaks[0])` doesn't work, then we have a bug in that subclass and should fix it there.

py/desispec/test/integration_test_quicklook.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # verify that quicklook pipeline runs
+        com = "desi_quicklook -n {} -c {} -e {} -f {} --psfboot {} --fiberflat {} --rawdata_dir {} --specprod_dir {}".format(night,camera,expid,'dark',psffile,fiberflatfile,datadir,'$DESI_SPECTRO_SIM')
+        if runcmd(com, clobber=clobber) != 0:
+            raise RuntimeError('quicklook pipeline failed for camera {}'.format(camera))
+
+        os.remove('lastframe-{}-{:08d}.fits'.format(camera,expid))
+
+    # remove all output
+    output_dir = os.path.join(os.environ['DESI_SPECTRO_SIM'],'exposures',night)
```
sbailey: Please add an option for whether the output files should be cleaned up or not.  It can be useful to run an integration test and have the output say behind to inspect.

Normally raw data simulations should be in `$DESI_SPECTRO_SIM/$PIXPROD`, with any processing of those simulations in a separate directory such as `$DESI_QUICKLOOK_DIR` or `$DESI_SPECTRO_REDUX/$QL_DIR`.  Quicklook should not write output into the raw data directory, and it shouldn't know that your simulated raw data aren't actually real raw data.

This cleanup as-is could blow away files that existed prior to the test being run.   Please isolate the test output (simulation and processing) into separate directories and raise an error if they exist at the start of the test (or provide a clobber option that must be set to overwrite a pre-existing directory).  Then you can safely cleanup those directories.

Note: the normal DESI spectro pipeline integration test also doesn't isolate its output into a test directory (bad) but at least it doesn't auto-cleanup everything and risk blowing something else away (good).

Cleanup should also be structured such that it will run (if requested) even if earlier code raises an exception.  e.g.
```python
try:
    # run the tests...
except Exception as e:
    print('Gack! Something failed while running the tests')
    raise e
finally:
    #- always excecuted, even upon exception
    if cleanup:
        # cleanup files
```

gdhungana:
The most recent commit does not address the changes requested. @rstaten will confirm them. The recent commit adds a function adopting from offline code, to use fiber/wavelength resolution on compute sky. Direct call to offline compute sky is not used so as to provide future developments of this inside QL.

rstaten:
@sbailey The requested changes have been made:
- There is no issue with the psf subclass for the XWSigma fix.
- The integration test can now overwrite and/or delete output written by this test. The simulated files are written as usual and I created a 'ql_output' directory under $DESI_SPECTRO_REDUX for the Quicklook output. I know there has been some discussion recently about naming/location of Quicklook output, so I just created a generic directory that will be updated once these decisions have been made.

rstaten:
Also, with this update each Quicklook QA now has a unit test.

sbailey:
In `integration_test_quicklook` the lines 120-122, you should read the desimodel PSFs with
```python
psf_b = desimodel.io.load_psf('b')
psf_r = desimodel.io.load_psf('r')
psf_z = desimodel.io.load_psf('z')
```
instead of explicitly constructing the path relative to `$DESI_PRODUCT_ROOT`. This also means that in your environment variable checks around line 37 you need `$DESI_ROOT` set but not `$DESI_PRODUCT_ROOT`.

Unrelated to that, when I run `integration_test_quicklook`, it calls
```
desi_bootcalib --fiberflat /data/desi/spectro/sim/debug/20160728/pix-r0-00000001.fits \
  --arcfile /data/desi/spectro/sim/debug/20160728/pix-r0-00000000.fits \
  --outfile /data/desi/spectro/sim/debug/20160728/psfboot-r0.fits
```
but then runs quicklook with
```
desi_quicklook -n 20160728 -c r0 -e 2 -f dark \
  --psfboot /data/desi/spectro/sim/exposures/20160728/psfboot-r0.fits \
  --fiberflat /data/desi/spectro/sim/exposures/20160728/fiberflat-r0-00000001.fits \
  --rawdata_dir /data/desi/spectro/sim/exposures \
  --specprod_dir /data/desi/spectro/redux/ql_output
```
Note that the `desi_bootcalib` psf output is sent to a different location than `desi_quicklook` is told to look for it, which causes `desi_quicklook` to crash.

In my case, `$DESI_SPECTRO_SIM/$SPECPROD` = `/data/desi/spectro/sim/debug`, i.e. the test is not isolated to a quicklook-only set of directories and it is potentially overwriting or conflicting with pre-existing files.

Normally pixel simulations would be written to `$DESI_SPECTRO_SIM/$PIXPROD` (not `$SPECPROD`), but the test should ensure that it goes somewhere else that can't conflict with other files.

rkehoe:
@sbailey  Relevant to your prior comment, we have started to implement a directory structure for quicklook which mirrors the offline structure, but is parallel to it.  So outputs will go to $QL_SPECPROD instead of $SPECPROD.  (We will use the same $DESI_SPECTRO_DATA as the offline, since we use the same raw inputs and pipeline algorithms as the offline.)  Within $QL_SPECPROD, there will be exposures/ and calib2d/ trees for the outputs from processing science and calibration exposures, respectively, similarly to offline.  The QA outputs (eg. .yaml or .pdf files) will be sent to the same place as the pipeline outputs (eg. frame files), again as in offline.  

We will need the $QL_SPECPROD environment variable defined so travis tests know where to find files.  Who do we ask to add that?

Ryan will elaborate on the code update shortly.

sbailey:
+1 for defining a separate but parallel quicklook output location.  For context, the offline spectro pipeline writes to `$DESI_SPECTRO_REDUX/$SPECPROD` by default; this can be overridden when setting up a production run.  The idea is that `$DESI_SPECTRO_REDUX` defines a typically unchanging base directory location, while `$SPECPROD` gets updated for different runs, e.g. with different code versions or different simulation inputs.  Directories like "exposures" and "calib2d" appear underneath that (i.e. not directly under `$DESI_SPECTRO_REDUX` itself.

For consideration: output quicklook to `$DESI_SPECTRO_REDUX/QL/$SPECPROD` by default (note the extra "QL" directory), thus not *requiring* any new environment variables.  If `$QL_SPECPROD` is set, use that instead as an override directory location, but leave it up to the user to set that if they want to override the default location.  Tests could set `$QL_SPECPROD` to a test directory to not clobber any other previous QL runs.  At KPNO we would probably always set `$QL_SPECPROD` as part of the quicklook machine environment.

Details: under the hood at NERSC, `source /project/projectdirs/desi/software/desi_environment.sh` calls `module load desimodules`, which loads `desitree`, which is actually the module that sets the basic default environment variables like `$DESI_ROOT` and `$DESI_SPECTRO_REDUX`.

rstaten:
@sbailey I am using this method of implementing the psf files because desi_extract_spectra takes files as input rather than psf objects and I only need these for extraction to make fiberflats. If I should still use desimodel.io.load_psf I can amend this.

I have changed the output of this test to go to $DESI_SPECTRO_REDUX/QL/$SPECPROD by default, with an optional command line argument that sends it to $DESI_SPECTRO_REDUX/$QL_SPECPROD if this environment variable is set.

The simspec files are written to $DESI_SPECTRO_SIM/$PIXPROD and raw files, fibermap files, and pix files are sent to $DESI_SPECTRO_DATA/$PIXPROD.

An error is raised if there are already files located in any of these locations, with the options to overwrite or change the night simulated.

rkehoe:
@sbailey The use of a QL subdirectory under $DESI_SPECTRO_REDUX was avoided after communication with @sbailey a couple weeks ago indicating a fully parallel arrangement was preferred.  I prefer this for reasons you state in your first paragraph in your last comment -- we use $QL_SPECPROD by default, and can override for specific production runs.

Until we resolve this conclusively, I would prefer not to merge this PR.

sbailey:
The intension of my suggestion was that `$QL_SPECPROD` should override the complete combination of `$DESI_SPECTRO_REDUX/QL/$SPECPROD` not just the `QL/$SPECPROD` part.  That allows `$QL_SPECPROD` to safely point to a completely isolated test directory.  You could also implement something like `$QL_SPECDATA` to override `$DESI_SPECTRO_DATA` so that the test can also generate its inputs in isolation.

rkehoe:
@sbailey After taking a further look at this, it does seem sensible to put in a default using established environment variables ($DESI_SPECTRO_REDUX, $SPECPROD, etc.), and then we can override that with the new variables.  Thanks for clarifying that $QL_SPECPROD would cover the complete path.  In order to reflect the 'redux' and 'data' naming convention, we're implementing two environment variables, $QL_SPECREDUX and $QL_SPECDATA.  In terms of default directories, quicklook will point to $DESI_SPECTRO_REDUX/QL/$SPECPROD as you suggest, but also $DESI_SPECTRO_DATA/QL when $QL_SPECDATA is not set.

rstaten:
@sbailey After sorting out the environment variable structure and fixing the docstring issue that was occurring, if the changes made based on your requests are sufficient this should be ready to be merged.

rkehoe:
@sbailey  I think all comments were accounted for, and I think this should be merged.

sbailey:
This would still benefit from more environment variable and input/output directory cleanup such that a single canonical location would determine the paths, analogous to `desispec.io.findfile`, `desispec.io.radata_root`, and `desispec.io.specprod_root`.  As it is, there are several places that construct the same paths based on various combinations of environment variables, making it fragile to maintain (especially if we decide to update the QL directory paths in the future).

However, this meets the minimum requirement of tests not blowing away pre-existing files or auto-deleting them so I'll go ahead and merge now.

PR #366, opened 2017-03-28
Use desiutil.log

### LINKED ISSUE #349: move desispec.log into desiutil.log

weaverba137:
This PR:

* Closes #349 
* Removes all references to `desispec.log`, and indeed the module itself.


weaverba137:
The only problem with `DeprecationWarning` is that it is ignored by default.  Do you actually want a `warnings.warn('message', DeprecationWarning)`, or just a `log.warn()` message?

sbailey:
I thought that `warnings.warn('message', DeprecationWarning)` would print a message but not raise an exception that had to be caught.  Whether it is that or `log.warn()` or just a big scary print statement, the desired effect is to print something to the log/screen that a human might notice (telling them to switch to `desiutil.log`) while not completely breaking the code that had been calling `desispec.log.get_logger()`.

weaverba137:
> I thought that warnings.warn('message', DeprecationWarning) would print a message but not raise an exception that had to be caught.

Yes, in Python **2**.  In Python 3 `, `DeprecationWarning` is ignored by default.  See https://docs.python.org/3/library/warnings.html#the-warnings-filter

In any case, I've already got a solution that uses `log.warn()` instead.

sbailey:
I agree that that py3 documentation says that DeprecationWarnings are ignored.  My experience says otherwise:
```python
In [1]: import sys, warnings

In [2]: sys.version
Out[2]: '3.5.2 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:52:12) \n[GCC 4.2.1 Compatible Apple LLVM 4.2 (clang-425.0.28)]'

In [3]: warnings.warn('blatfoo', DeprecationWarning)
/Users/sbailey/anaconda/envs/desi/bin/ipython:1: DeprecationWarning: blatfoo
  #!/bin/bash /Users/sbailey/anaconda/envs/desi/bin/python.app
```
Either way, using `log.warn` is a safer way to ensure that it does get printed.  Go ahead and merge when ready.

weaverba137:
iPython almost certainly alters the default set of warnings that are ignored.  Try running this script:
```python
#!/usr/bin/env python3
import warnings
warnings.warn("deprecated!", DeprecationWarning)
```

sbailey:
For the record:
```
(desi) [master*] etc $ python -V
Python 3.5.2 :: Continuum Analytics, Inc.
(desi) [master*] etc $ ipython -V
5.1.0
(desi) [master*] etc $ python -c 'import warnings; warnings.warn("deprecated!", DeprecationWarning)'
(desi) [master*] etc $ ipython -c 'import warnings; warnings.warn("deprecated!", DeprecationWarning)'
/Users/sbailey/anaconda/envs/desi/bin/ipython:1: DeprecationWarning: deprecated!
  #!/bin/bash /Users/sbailey/anaconda/envs/desi/bin/python.app
```

PR #367, opened 2017-03-31
Test DTS delivery

### LINKED ISSUE #322: Create DTS delivery scripts
Create DTS delivery script(s) that can receive command-line arguments from DTS (easily configured in DTS configuration files) and submit pipeline processing jobs.

The script(s) should be able to handle start and end-of-night as well as routine data processing jobs.

For initial tests, use the 'desi' pseudo-user at NERSC, for a clean environment.

weaverba137:
**Don't merge until we've thoroughly reviewed this!**

This PR:

* Closes #322 
* Fleshes out the DTS delivery script.
* Adds a DTS simulator script that calls the delivery script.

This is nearly done, but could probably use some tweaks, so I want to get the review process started. I *think* that most of the remaining work is configuration details; the bulk of the necessary *mechanisms* should be in place.

Here's how this currently works.

1. Log in to one of the data transfer nodes and become the 'desi' pseudo-user.  Set up environment to point at this branch of desispec.  In other words, everything below runs on the 'desi' account.
2. `etc/desi_dts_delivery_test.sh` This script copies simulated raw data files to a spool directory (a.k.a. "inbox") to simulate the arrival of files via DTS. It also calls the delivery script after every file is copied (just like DTS would).
3. `bin/desi_dts_delivery` is the delivery script.  It calls `desi_start_night` or `desi_end_night` unconditionally when it receives the corresponding notification from DTS itself (or the DTS simulator described above).  If a file arrives, and it is neither the start nor the end of the night, the delivery script only calls `desi_update_night` if all files associated with a particular exposure number have arrived.
4. The delivery script starts the night scripts on (currently) edisongrid (so I can easily log in to that node and see the scripts running) via ssh.  Some care is required to redirect stdin, stdout & stderr so that the night script "detaches" cleanly from the delivery script, which can then exit effectively immediately.
5. The "night" scripts modify the standard `desiutil.log` system to write to a file instead of standard output.  The exact filename and directory remains to be defined, but the important thing is that the mechanism for modifying the logging system to do this is working.

### DIFF
--- a/bin/desi_end_night
+++ b/bin/desi_end_night
@@ -1,4 +1,4 @@
 #!/usr/bin/env python
 from sys import exit
-from desispec.scripts.night import end_main
-exit(end_main())
+from desispec.scripts.night import main
+exit(main())
--- a/bin/desi_start_night
+++ b/bin/desi_start_night
@@ -1,4 +1,4 @@
 #!/usr/bin/env python
 from sys import exit
-from desispec.scripts.night import start_main
-exit(start_main())
+from desispec.scripts.night import main
+exit(main())
--- a/bin/desi_update_night
+++ b/bin/desi_update_night
@@ -1,4 +1,4 @@
 #!/usr/bin/env python
 from sys import exit
-from desispec.scripts.night import update_main
-exit(update_main())
+from desispec.scripts.night import main
+exit(main())
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -8,6 +8,9 @@ desispec Change Log
 * Replace all instances of :mod:`desispec.log` with ``desiutil.log``;
   :func:`~desispec.log.get_logger` now prints a warning that users need
   to switch.
+* Working DTS delivery script and DTS simulator (PR `#367`_).
+
+.. _`#365`: https://github.com/desihub/desispec/pull/367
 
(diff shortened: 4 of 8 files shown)


sbailey:
Ben and I discussed this over a hangout with screen share.  Green light to merge when ready.  Topics that came up for followup:

* We may want to call `desi_start_night` from a cronjob at NERSC rather than relying upon ICS or DTS to notice when it is transferring the first file for a new night.  We know we want to run it at a fixed time (in UTC, not CA localtime with daylight savings mess) every day.
* We need an additional hook for "all calibration data have been taken and we are proceeding with science data" so that the pipeline knows it can proceed with coadding the PSFs.  This might fit more naturally with ICS operations.  Need to discuss with Ted and Klaus.
* `desi_end_night` really means "all science exposures have been received and you can proceed with redshift fitting".  It is possible that additional morning calibration data will arrive after that.  The daily processing will not retroactively use them, but later full reprocessings might.

weaverba137:
A couple of TO DO items for me before I merge:

* Fix a math bug in the DTS simulator.
* Copy delivered files to `$DESI_SPECTRO_DATA/NIGHT`.
* Write night log files to someplace besides `$HOME`.

PR #370, opened 2017-04-11
Teststand preprocessing update

julienguy:
This PR to merge recent changes to preprocessing of teststand data.
Only minor changes to preprocessing that won't affect simulations

 - Use EXPREQ instead of EXPTIME if key exists in raw data for darks
 - Updated ccd_calibration.yaml with gains and new darks
 - Option to preproc to add header keywords from other HDUs in raw data
 - Optional cross-talk correction


sbailey:
The change of letting EXPREQ trump EXPTIME smells like something that could bite us a couple of years from now if we forget about it.  Is there a cleaner way for us to handle EXPREQ vs. EXPTIME for the teststand data?  In the long term I think the data will have both keywords and at some point EXPTIME will be fixed to be reliable and should be the default again.  Can we make a change now that will automatically work then, i.e. allowing the teststand to be the special case instead of having to refix this later?

julienguy:
OK. I'll add the exptime keyword to the yaml file (default being EXPTIME).
So it will use EXPREQ only if specified there.


2017-04-11 19:20 GMT+02:00 sbailey <notifications@github.com>:

sbailey:
I merged this PR but left the branch behind because I wasn't sure if you wanted to continue development on this branch or start over from a new branch off of master.  Please delete this teststand branch from github if you are done with it for now.

PR #379, opened 2017-04-20
new fibermap columns LOCATION,DEVICE_LOC,PETAL_LOC

sbailey:
This PR adds 3 new columns to the fibermap based on [DESI-0530](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=530) and discussions with ICS (Klaus) and the focal plane (Joe):
  * PETAL_LOC [0-9]: location of the petal (wedge) on the focal plane
  * DEVICE_LOC [0-542]: location of the device on the petal
  * LOCATION: for convenience, 1000*PETAL_LOC + DEVICE_LOC

It also fills in more default values in `desispec.io.fibermap.empty_fibermap()`, in particular the mapping of FIBER to LOCATION, using `desimodel.io.load_fiberpos()` which returns the canonical mapping from desimodel.

Notes:
  * POSITIONER [0-5000] is deprecated; use LOCATION or (PETAL_LOC, DEVICE_LOC) instead
  * (PETAL_LOC, DEVICE_LOC) will become the canonical way that we reference an object on the focal plane, e.g. for fiber assignments communicated to ICS
  * There are DEVICE_LOC values > 500 since the device locations include non-positioner locations such as fiducials.  The fibermap only tracks devices attached to spectrograph fibers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -152,6 +152,9 @@ install:
     # packages are installed, thus desiutil is not installed in that script.
     - for p in $DESIHUB_PIP_DEPENDENCIES; do r=$(echo $p | cut -d= -f1); v=$(echo $p | cut -d= -f2); pip install git+https://github.com/desihub/${r}.git@${v}#egg=${r}; done
     # - source etc/travis_install_specex.sh
+    - export DESIMODEL=${HOME}/desimodel/${DESIMODEL_VERSION}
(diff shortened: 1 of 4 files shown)


sbailey:
Self merging, but feel free to comment post-facto if you think something needs to be cleaned up or corrected here.

PR #380, opened 2017-04-24
fix fibermap dtype for astropy 1.3

sbailey:
This PR works around a change in behavior from astropy 1.2 to 1.3, likely related to the units issue we had to work around desihub/specsim#62:

Under astropy 1.2, assigning a value to a table column would retain the original dtype of that column.  Under astropy 1.3, the column gets the new dtype of the right-hand-side value.  This breaks some consistency checks for fibermap columns though I don't think it broke any actual functionality (for desispec).

This PR also fixes two smaller issues that I found while debugging this:
  * previously `desispec.io.Brick` couldn't be created with a filename that didn't include a path (it would parse the path as '' and then try to create '' and complain)
  * typo in #379 (new fibermap columns) resulted in both `LAMBDAREF` and `LAMBDA_REF`

I have tested with astropy 1.3 on edison and astropy 1.2 on my laptop.  I'm Leaving `.travis.yml` at astropy 1.2.1 for now; the nightly cronjob unit tests on edison test master with astropy 1.3.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -7,9 +7,10 @@ desispec Change Log
 
 * Set default brick size to 0.25 sq. deg. in desispec.brick (PR `#378`_).
 * Added function to calculate BRICKID at a given location (PR `#378`_).
-* Additional LOCATION, DEVICE_LOC, and PETAL_LOC columns for fibermap.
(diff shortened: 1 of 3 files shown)


sbailey:
I'm going to go ahead and merge this now to get tests passing again under astropy 1.3 at NERSC, but feel free to make retroactive comments if you find something wrong.

PR #381, opened 2017-04-25
Pipeline Improvements

### LINKED ISSUE #277: Slicing spectrographs from dependency graph must also track bricks
For some productions, we wish to prune the inputs to just a subset of the spectrographs.  This also needs to account for the fact that some bricks may only be observed by fibers from some spectrographs.

### LINKED ISSUE #375: Pipeline changes for use with quickgen (needed for 2% data challenge)
When creating the nightly dependency graph, the pipeline uses desispec.io.get_raw_files to get all the images for a given exposure.  In the case of quickgen use, no pix files exist.  We should have an option to the production creation that simply assumes that all frames for a given exposure exists (rather than querying the filesystem).

### LINKED ISSUE #376: Adjust pipeline step concurrency more intelligently (2% data challenge)
Several small things are needed for smoother operation of the pipeline, and for supporting redrock:

- When generating slurm scripts using multiprocessing, ensure that OMP_NUM_THREADS=1 and that the special KMP affinity keyword is set.

- When choosing number of workers, use estimate of time for one task and set max nodes.  If it exceeds a limit for the given machine, increase run time and reduce nodes to make as fully packed as possible.

- Output a new helper shell script which loops over days and submits each day's slurm script.

tskisner:
This pull request:

- Passes the list of selected spectrographs down into the nightly planning function.  Only bricks from fibers with a matching spectrograph ID are accumulated.  This fixes a longstanding bug with creating productions that use only a subset of the spectrographs.

- Adds a --fakepix option to desi_pipe and lower level functions, which skips checking of the pix data files.  This allows creating a production compatible with using quickgen in place of the usual psf / extract steps.

- Add high level helper scripts to submit chains of nightly processing.  This allows running all nights of processing as independent chains of jobs prior to running redshift fitting.

- Changes the way that the size of pipeline jobs are computed.  Given a maximum number of nodes, processes per task, and number of tasks, the number of workers is adjusted so that each worker has (as close as possible) a uniform number of tasks. 

This pull request closes #277, closes #375, and closes #376 .

### DIFF
--- a/py/desispec/scripts/pipe_prod.py
+++ b/py/desispec/scripts/pipe_prod.py
@@ -20,6 +20,43 @@ from ..parallel import dist_discrete
 from .. import pipeline as pipe
 
 
+def job_size(ntask, taskprocs, tasktime, nodeproc, maxnodes):
+    """
+    Internal helper function used only in the desi_pipe
+    script to compute the size of batch jobs.
+    """
+    maxprocs = nodeproc * maxnodes
+
+    fullprocs = ntask * taskprocs
+
+    nodes = None
+    time = None
+    procs = None
+    
+    if fullprocs <= maxprocs:
+        # we can run fully unpacked, with every process group
+        # assigned one task.
+        time = tasktime
+        procs = fullprocs
+    else:
+        # we want to repack as efficiently as possible
+        nworker = maxprocs // taskprocs
+
+        taskperworker = ntask // nworker
+        if ntask % nworker != 0:
+            taskperworker += 1
+
+        nworker = ntask // taskperworker
+        if ntask % nworker != 0:
+            nworker += 1
+
+        time = tasktime * taskperworker
+        procs = nworker * taskprocs
+    
+    nodes = procs // nodeproc + 1
+    return (nodes, procs, time)
+
+
 def step_props(first, last, specs, night):
     """
     Internal helper function used only in the desi_pipe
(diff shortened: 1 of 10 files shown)


py/desispec/test/test_pipeline_graph.py
```diff
@@ -25,6 +25,7 @@ def setUp(self):
         self.raw = ph.fake_raw()
         self.redux = ph.fake_redux(self.prod)
         ph.fake_env(self.raw, self.redux, self.prod, self.redux)
+        self.specs = [ x for x in range(10) ]
```
sbailey: FYI, `list(range(10))` also works to convert the py3 iterator into a plain list.

py/desispec/scripts/pipe_prod.py
```diff
@@ -20,6 +20,43 @@
...
+            taskperworker += 1
+
+        nworker = ntask // taskperworker
+        if ntask % nworker != 0:
+            nworker += 1
+
+        time = tasktime * taskperworker
+        procs = nworker * taskprocs
+    
+    nodes = procs // nodeproc + 1
```
sbailey: If `procs % nodeproc == 0` then you will end up requesting one more node than you actually need, correct?
Should this be `nodes = max(1, (procs-1) // nodeproc)` instead?

Having writing similar code in the past and been humbled by how many different ways I screwed up the edge cases, I would recommend writing some unit tests to check cases, especially cases like number of processes exactly fitting on an integer number of nodes, just barely exceeding or just barely under subscribing, and just barely exceeding maxnodes.  This looks close enough to right that we could proceed with a merge, but eventually we'll want to check this more closely.

PR #387, opened 2017-05-22
compute_flux_calibration using frame mask

julienguy:
PR with minor changes to the routine compute_flux_calibration 
 - use frame mask in ivar
 - propagate mask to calibration files

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -850,23 +850,32 @@ def compute_flux_calibration(frame, input_model_wave,input_model_flux,input_mode
         convolved_model_flux[fiber]=stdstars.R[fiber].dot(model_flux[fiber])
 
     # iterative fitting and clipping to get precise mean spectrum
-    current_ivar=stdstars.ivar.copy()
-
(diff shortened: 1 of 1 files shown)


py/desispec/fluxcalibration.py
```diff
@@ -850,23 +850,32 @@ def compute_flux_calibration(frame, input_model_wave,input_model_flux,input_mode
...
-    # chi2 = sum w ( data_flux - R*(calib*model_flux))**2
-    # chi2 = sum (sqrtw*data_flux -diag(sqrtw)*R*diag(model_flux)*calib)
+    chi2=np.zeros((stdstars.flux.shape))  
 
+    for fiber in range(nstds) :
+        M = median_calib*stdstars.R[fiber].dot(model_flux[fiber])
+        
+        try:
+            pol=np.poly1d(np.polyfit(dwave,stdstars.flux[fiber]/(M+(M==0)),deg=deg,w=current_ivar[fiber]*M**2))
+        except:
```
sbailey: What's the specific exception that you might encounter here?   perhaps np.linalg.LinAlgError?  Please catch that and only that so that any other unexpected exceptions still crash.

PR #391, opened 2017-06-02
ivar coadds

ngbusca:
This is an initial PR with code that handles a simple, fast ivar coadding. I've done many tweaks not specifically related to the goal of the branch but required to get desi_make_bricks and desi_update_coadds to work (they seemed to be broken due to changes in brick.py and other places). 

The algorithm is (I think) the same as redrock: simply ivar average fluxes and resolution matrices. The resolution matrix is still meaningful: it correctly maps the model to the data.

The algorithm is implemented in coaddition.py, with the choice of doing full inversion or ivar coaddition governed by the "fast" option, now passed to the finalize(fast=False) function. Note that some potentially large matrices, only needed for spectro-perf, are still computed but not used in the final step if fast=True (is this ok?). 

There are still things to fix before actually merging, but I need some input. 

Here's a summary of the changes:

* added the option --fast to desi_update_coadds to do do ivar coaddition instead of spectro perf. The name of the output files is still "coadd-...". Do we want to make it explicit that these were "ivar-coadds" or add a keyword or something else?

* scripts/makebricks.py: added an option the override the rawdata directory and add some flexibility wrt the location of input cframes

* the INDEX field was missing from the output bricks created by io/brick.py (causing update_coadds on those bricks to crash). It is now filled to arange(num_spectra). Is this what's supposed to be? This might have consequences elsewhere.

* The TARGETID keyword is used to match spectra. At least in BOSS (I'm using boss2desi data to test this branch), sky fibers have TARGETID=-1 so their spectra would get coadded as coming from the same object. This doesn't make much sense... How is this handled in DESI? Do sky fibers have a TARGETID?

* The output files have the incorrect number of objects, set to max(INDEX). Since index counts the rows, the number of ojbects corresponds to the number of spectra while we want the number of unique ids. In practice we end up with many many empty spectra.

* There's a lot of overhead not related to the coaddition step. Could it be related to pyfits?

* various minor fixes here and there in there

* I'll post some comparison plots

### DIFF
--- a/py/desispec/coaddition.py
+++ b/py/desispec/coaddition.py
@@ -44,6 +44,8 @@ class Spectrum(object):
         self.wave = wave
         self.flux = flux
         self.ivar = ivar
+        if mask is not None:
+            self.ivar *= (mask==0)
         self.mask = mask
         # if mask is None:
         #     self.mask = np.zeros(self.flux.shape, dtype=np.uint32)
@@ -58,13 +60,19 @@ class Spectrum(object):
             n = len(wave)
             self.Cinv = np.zeros((n,n))
             self.Cinv_f = np.zeros((n,))
+            self.sum_ivar = np.zeros(n)
+            self.ivar_f = np.zeros(n)
+            self.ivar_R = np.zeros((n,n))
         else:
             assert flux is not None and resolution is not None,'Missing flux and/or resolution.'
             diag_ivar = scipy.sparse.dia_matrix((ivar[np.newaxis,:],[0]),resolution.shape)
             self.Cinv = self.resolution.T.dot(diag_ivar.dot(self.resolution))
             self.Cinv_f = self.resolution.T.dot(self.ivar*self.flux)
+            self.ivar_f = self.ivar*self.flux
+            self.ivar_R = self.resolution.toarray()*self.ivar[:,None]
+            self.sum_ivar = self.ivar*1.
 
-    def finalize(self):
+    def finalize(self,fast=False):
         """Calculates the flux, inverse variance and resolution for this spectrum.
 
         Uses the accumulated data from all += operations so far but does not prevent
@@ -88,12 +96,21 @@ class Spectrum(object):
         self.ivar = np.zeros_like(self.Cinv_f)
         R = np.zeros_like(self.Cinv)
         # Calculate the deconvolved flux,ivar and resolution for ivar > 0 pixels.
-        self.ivar[mask],R[keep_t,keep] = decorrelate(self.Cinv[keep_t,keep])
-        try:
-            R_it = scipy.linalg.inv(R[keep_t,keep].T)
-            self.flux[mask] = R_it.dot(self.Cinv_f[mask])/self.ivar[mask]
-        except np.linalg.linalg.LinAlgError:
-            self.log.warning('resolution matrix is singular so no coadded fluxes available.')
+        if not fast:
+            self.ivar[mask],R[keep_t,keep] = decorrelate(self.Cinv[keep_t,keep])
+            try:
+                R_it = scipy.linalg.inv(R[keep_t,keep].T)
+                self.flux[mask] = R_it.dot(self.Cinv_f[mask])/self.ivar[mask]
+            except np.linalg.linalg.LinAlgError:
+                self.log.warning('resolution matrix is singular so no coadded fluxes available.')
+        else:
+            self.ivar = self.sum_ivar
+            self.flux = self.ivar_f
+            R = self.ivar_R
+            w = self.ivar > 0
+            self.flux[w]/=self.ivar[w]
+            R[w,:]/=self.ivar[w,None]
+
         # Convert R from a dense matrix to a sparse one.
         self.resolution = desispec.resolution.Resolution(R)
 
@@ -114,12 +131,18 @@ class Spectrum(object):
         if np.array_equal(self.wave,other.wave):
             self.Cinv = self.Cinv + other.Cinv
             self.Cinv_f += other.Cinv_f
+            self.ivar_f += other.ivar_f
+            self.sum_ivar += other.ivar
+            self.ivar_R += other.ivar_R
             if (self.mask is not None) and (other.mask is not None):
                 self.mask |= other.mask
         else:
             resampler = get_resampling_matrix(self.wave,other.wave)
             self.Cinv = self.Cinv + resampler.T.dot(other.Cinv.dot(resampler))
             self.Cinv_f += resampler.T.dot(other.Cinv_f)
+            self.ivar_f += resampler.T.dot(other.ivar_f)
+            self.sum_ivar += resampler.T.dot(other.ivar)
+            self.ivar_R += resampler.T.dot(other.ivar_R).dot(resampler)
             if (self.mask is not None) and (other.mask is not None):
                 mask_resampler = (resampler != 0).T
                 self.mask |= mask_resampler.T.dot(other.mask)
@@ -139,7 +162,7 @@ so we add an extra bin to the end of the global wavelength grid to fully contain
 Note that we use a linear grid (rather than a log-lambda grid, for example) so that
 co-added spectra have a roughly constant FWHM/BINSIZE.
 """
-global_wavelength_grid = np.arange(3579.0,9826.0,1.0)
+global_wavelength_grid = np.arange(3579.0,10001.0,1.0)
 
 def get_resampling_matrix(global_grid,local_grid):
     """Build the rectangular matrix that linearly resamples from the global grid to a local grid.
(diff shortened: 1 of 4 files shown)


ngbusca:
Here's an example comparing the full deconvolution with the ivar weight:

coadd example

sbailey:
Thanks for posting this.  Commenting on some of your bullets:

> added the option --fast to desi_update_coadds to do do ivar coaddition instead of spectro perf. The name of the output files is still "coadd-...". Do we want to make it explicit that these were "ivar-coadds" or add a keyword or something else?

Good idea to add a keyword.  FITS 8-character keyword limit is always a fun challenge.  Perhaps `COADDTYP` or just `COADD`, with options "ivar" or "deconvolve"?

> the INDEX field was missing from the output bricks created by io/brick.py (causing update_coadds on those bricks to crash). It is now filled to arange(num_spectra). Is this what's supposed to be? This might have consequences elsewhere.

INDEX was removed because it is redundant with now number (i.e. arange(num_spectra)).  It would be better to fix update_coadds to not need INDEX instead of reintroducing the INDEX column.

> The TARGETID keyword is used to match spectra. At least in BOSS (I'm using boss2desi data to test this branch), sky fibers have TARGETID=-1 so their spectra would get coadded as coming from the same object. This doesn't make much sense... How is this handled in DESI? Do sky fibers have a TARGETID?

Yes, in DESI sky targets will have unique TARGETIDs too.  For boss2desi you could assign them to have a random 64-bit integer with an extremely low probability of having an actual collision.  To be really safe, you could use `np.in1d` to check if any of your sky TARGETIDs are already in the non-sky target ID list.

> The output files have the incorrect number of objects, set to max(INDEX). Since index counts the rows, the number of ojbects corresponds to the number of spectra while we want the number of unique ids. In practice we end up with many many empty spectra.

Agreed that we want the output to to have the number of unique TARGETIDs, not max(INDEX) with a bunch of blank spectra.

> There's a lot of overhead not related to the coaddition step. Could it be related to pyfits?

Perhaps related: we (Ted) are refactoring the brick making, and in the process deprecating the Brick class which entangles the data in memory with its representation on disk.  Especially for this fast coadded I'd prefer to factor this out as
* read data into memory
* coadd data from unique TARGETIDs
* write it back out

i.e. bypassing the CoAddedBrick class.

A brickmaking test is failing:
```
FAIL: test_brick (desispec.test.test_io.TestIO)
Test desispec.io.brick.Brick objects.
  File "/home/travis/build/desihub/desispec/py/desispec/test/test_io.py", line 350, in test_brick
    self.assertEqual(len(fibermap.colnames)+2, len(brick_fibermap.colnames))
AssertionError: 28 != 29
```

ngbusca:
Thanks @sbailey for the responses. The "INDEX" issue is now fixed (this was causing the travis test to fail due to the extra column). The output coadds now have the right number of objects.

I also found the reason for the overhear (twofold):
* when requesting a single target the code was still looping over all targets (correctly ignoring those unrequested but still making it unnecessarily slow)
* zero len(wavelength)^2 arrays (instead of sparse arrays) where instantiated for every spectra filling up memory pretty quickly and slowing things down due to swapping

The fast option is now pass to the Spectrum object (instead of its finalize function) to avoid instantiating the other fast=True-useless arrays.

I'd still like to make a few checks to make sure the output metadata is correctly aligned with the flux, ivar, etc arrays.

ngbusca:
Incidentally, I also changed the coaddition.global_wavelength_grid from linear in lambda to linear in log-lambda (travis doesn't seem to like this)

ngbusca:
Here're some performance tests over boss2desi plate 4199 (five expossures) on a 12 cores:

time desi_update_coadds --brick 4199 --fast:
real    13m31.643s
user    9m9.693s
sys     4m14.256s

time desi_update_coadds --brick 4199
real    1186m45.195s
user    12019m9.634s
sys     150m34.725s

Here's a plot comparing the mean sky residuals on all fibers on all exposures in this plate with ivar caodds and full spectro. perf. coadds. The spectro perf options reduces the noise by about 25%.

sky_ivar_vs_spect_perf

ngbusca:
Some caveats: 

* I've modified makebricks to zero the ivar of masked pixels, since bricks don't have a mask hdu. Is this ok? this might have an impact elsewhere...
* the code assumes that the same objects are present in bricks from all bands. 
* with the --fast option, the final coaddition over the global grid is done via interpolation, which introduces some correlation between neighboring pixels.

dkirkby:
@ngbusca Why did you change the wavelength grid from the default linear spacing?  Its was originally log-spaced, like BOSS, but I changed it to linear to better match the DESI spectrograph resolution.  We discussed this a while ago in [desi-data 923](https://desi.lbl.gov/trac/tracmailman/browser/private/desi-data/2015-March/000922.html).

dkirkby:
> with the --fast option, the final coaddition over the global grid is done via interpolation, which 
> introduces some correlation between neighboring pixels.

It would be helpful to quantify these correlations, e.g., relative to the resolution correlations that have nominally been removed.

ngbusca:
@dkirkby I'm using boss2desi data to test this branch and in the red end the 1A binsize is about half the binsize of the loglam grid. I can revert back to linear.

ngbusca:
>It would be helpful to quantify these correlations, e.g., relative to the resolution correlations that have nominally been removed.

The correlation is only between first neighbors. The worst case is when an interpolation grid pixel is exactly at the midpoint of two original grid pixels. This case gives ~50% correlation in the interpolated fluxes.

ngbusca:
I really don't like the final interpolation step. Best would be to have overlapping blue and red wavelength grids, which would allow us to do a global ivar coadd. Would that make sense?

By the way, I don't understand the travis failure...

sbailey:
We should derive the coadd wavelength grid from the input spectra and not resample nor enforce a specific pre-defined global wavelength grid.  Options:
* Coadd per-band but not across b,r,z bands (perhaps doesn't gain the convenience people are seeking in a coadd if they still have to deal with 3 separate bands)
* Require our extractions to use an aligned wavelength grid for b,r,z.  Since different bands have different typical resolutions, this is not ideal (but is what we are currently doing while testing).  See plot below.
* Define a transition wavelength between bands, and just ignore data above/below that from a given band during coaddition.  e.g. a coadd could use just wavelength<5800 for b and >5800 for r.  Also not optimal, but we aren't claiming that coadds are optimal.
* Other?

[Sorting out these types of details is perhaps why we don't have a "trivial" coadder yet...]

Spectrograph sigma vs. wavelength for fiber 0 (black solid) and fiber 250 (black dotted), and CCD pixel size in Angstroms (blue):
brz-sigma

Travis error: `doc/conf.py` defines a set of "autodoc_mock_imports" that it pretends exists without actually requiring that they be installed for the sphinx documentation to be built.  Your Travis sphinx failure is that numpy is on the mock list (i.e. isn't installed by Travis when just building docs) but you are using it to make a module-level variable:
```
Warning, treated as error:
/home/travis/build/desihub/desispec/doc/api.rst:19: WARNING: autodoc: failed to import module 'desispec.coaddition'; the following exception was raised:
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/Sphinx-1.5.6-py3.5.egg/sphinx/ext/autodoc.py", line 560, in import_object
    __import__(self.modname)
  File "/home/travis/build/desihub/desispec/py/desispec/coaddition.py", line 178, in <module>
    global_wavelength_grid = 3579.*10**(np.arange(4500)*1e-4)
TypeError: unsupported operand type(s) for *: '_MockModule' and 'float'
```
You could wrap this in a try/except block so that sphinx won't die when it can't use `np.arange` for real.  There must be some way to tell Travis to install numpy prior to building the docs, but that exceeds my current knowledge of astropy ci-helpers and Travis, and you can google as well as I can...

weaverba137:
Why do you need a global variable?  Why not a function that returns that value?

ngbusca:
@sbailey The code generates coadds per band that use the wavelength grid from the bricks. There is no interpolation problem for those files. I also generates all-band coadds that use the global grid.

I vote for aligned extraction grids ;^). But I like also the idea of transition wavelengths, it reduces the problem to the overlap regions where we could use a coarser grid and resample flux.

Let's agree on a solution to this and then I'll fix the use of a global variable.

ngbusca:
There's another possibility for the final coadd: interlace the original grids. This is actually a way to "trick" the user into using the bands separately but in a transparent way. It actually works also with the resolution matrix by extending them with zeros appropriately

sbailey:
I'm not exactly sure what you mean by how to handle the resolution matrix in that case.  If it is easy to implement, please put together a demo and distribute a file that has the interleaved coadd including the resolution matrix.

I'm also fine with requiring that this version of the coadds will only work if the grids are aligned, and just print an error message and stop if they aren't.

sbailey:
Let's wrap this up and provide a coadder that requires the different channels to be aligned to the same wavelength spacing and phase, and we can reopen the issue of how to do coadds if we later use different wavelength grids for different channels.

sbailey:
PR #805 revived coaddition.  Closing this PR without merging.

PR #393, opened 2017-06-05
temporary xwsigma fix

rstaten:
This PR:
- Provides a temporary fix for the xwsigma failure. Further debugging will continue to fully solve this issue.


profxj:
I can't say that I understand this test well enough to comment
on it, but am fine merging anyhow.  So I am.

PR #394, opened 2017-06-08
Spectral regrouping by HEALPix pixel

tskisner:
This work is focused on grouping spectra by healpix pixel and performing that re-grouping on either a full production or on an incremental update from a few frames:

 - A new Spectra class can hold spectral data (wavelength, flux, ivar,
   mask, resolution, etc) for one or more arbitrary bands.  There are
   methods for returning a selection of spectra and for updating or
   appending new spectral data.  Each band may have different number
   of diagonals in the resolution matrix.

 - The pipeline is now designed to work with these spectra objects rather
   than bricks.

 - A new MPI-aware entry point (desispec.scripts.group_spectra) is added
   which distributes spectral group files among processes to perform the
   regrouping.  Each process reads (and optionally caches) the frames
   needed for its spectral groups.

 - Pipeline production creation now builds nightly chains of jobs which
   can run in parallel.

 - Redshift fitting will be broken until Redrock can be fully integrated
   and using the new spectral group files.

### DIFF
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -1,0 +1,285 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+"""
+desispec.io.spectra
+=====================
+
+I/O routines for working with spectral grouping files.
+
+"""
+from __future__ import absolute_import, division, print_function
+
+import os
+import re
+import warnings
+
+import numpy as np
+import astropy.io.fits as fits
+from astropy.table import Table
+
+from desiutil.depend import add_dependencies
+from desiutil.io import encode_table
+
+from .util import fitsheader, native_endian
+
+from .frame import read_frame
+
+from ..spectra import (Spectra, spectra_columns, 
+    spectra_comments, spectra_dtype)
+
+
+def write_spectra(outfile, spec, units=None):
+    """
+    Write Spectra object to FITS file.
+
+    This places the metadata into the header of the (empty) primary HDU.
+    The first extension contains the fibermap, and then HDUs are created for
+    the different data arrays for each band.
+
+    Floating point data is converted to 32 bits before writing.
+
+    Args:
+        outfile (str): path to write
+        spec (Spectra): the object containing the data
+        units (str): optional string to use for the BUNIT key of the flux
+            HDUs for each band.
+
+    Returns:
+        The absolute path to the file that was written.
+
+    """
+
+    outfile = os.path.abspath(outfile)
+
+    # Create the parent directory, if necessary.
+    dir, base = os.path.split(outfile)
+    if not os.path.exists(dir):
+        os.makedirs(dir)
+        
+    # Create HDUs from the data
+    all_hdus = fits.HDUList()
+
+    # metadata goes in empty primary HDU
+    hdr = fitsheader(spec.meta)
+    add_dependencies(hdr)
+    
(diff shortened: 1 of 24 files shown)


sbailey:
Regarding failing Travis tests:

@weaver do you know why travis tests aren't finding healpy even though it is included in the PIP_DEPENDENCIES?

@tskisner have you tried putting healpy into CONDA_DEPENDENCIES or CONDA_ALL_DEPENCENCIES instead?

moustakas:
I'm not at my computer but check where I added the dependency in the
sky-and-QA branch of desitarget, which worked for me.

py/desispec/io/spectra.py
```diff
@@ -0,0 +1,285 @@
...
+    hdus = fits.open(infile, mode="readonly")
+    nhdu = len(hdus)
+
+    # load the metadata.  FITS will capitalize all our metadata
+    # which is annoying.  Here we lower it all.
+
+    meta = {}
+    for k, v in hdus[0].header.items():
+        if k != "EXTNAME":
+            meta[k.lower()] = v
```
sbailey: Elsewhere we leave these as upper case.  Let's just cast this to a dictionary via `meta = dict(hdus[0].header)` and leave them as is.  Or you could even leave it as a `astropy.io.fits.header.Header` object in which case either upper or lowercase keys will work.  Header object are much less efficient to access, but that only matters if we are looking up keywords many many times.

weaverba137:
@sbailey, it might help if you point out a specific test.

weaverba137:
PS, I was going to say, first of all, that healpy should be a conda package.

weaverba137:
* For the sphinx problem, you need to add healpy to `autodoc_mock_imports` in doc/conf.py (see line 136).
* For the test failures, there are additional failures that appear to be unrelated to healpy, so you'll need to fix those as well.

sbailey:
sphinx fixed and tests can see healpy now, but now we have our favorite astropy making random changes on us type of error:
```
ERROR: test_empty (desispec.test.test_spectra.TestSpectra)
  File "/home/travis/build/desihub/desispec/py/desispec/test/test_spectra.py", line 137, in test_empty
    path = write_spectra(self.filebuild, spec)
  File "/home/travis/build/desihub/desispec/py/desispec/io/spectra.py", line 120, in write_spectra
    all_hdus.writeto("{}.tmp".format(outfile), overwrite=True, checksum=True)
TypeError: writeto() got an unexpected keyword argument 'overwrite'
```
overwrite is astropy 1.3, clobber is astropy 1.2, we should support both until 1.4 is out. :(  Other codes have handled this with try/except blocks on the writing.

tskisner:
Hi folks, I don't believe healpy is available as a conda package, which is why I included it in pip dependencies.  I'll go add boilerplate for astropy.  Super annoying...

weaverba137:
@tskisner, just so you know, the tests that required healpy did pass when it was installed with conda, so I guess that means there's a package.

tskisner:
Your travis builds must have access to other conda channels then.  It is not in the default one:
```
$ which conda
/home/kisner/software/intel/idp/bin/conda
```
```
$ conda --version
conda 4.3.21
```
```
$ conda config --get channels
--add channels 'defaults'   # lowest priority
--add channels 'file:///home/kisner/software/intel/idp/conda-bld'   # highest priority
```
```
$ conda search healpy
Fetching package metadata ...........
$
```

A web search seems to show one version available on conda-forge though.

tskisner:
I'll merge this unless there are further changes you would like...

PR #395, opened 2017-06-08
Quicklook config

gdhungana:
Replaces #389 @rkehoe Please confirm if this addresses enough on the topics we discussed.


rkehoe:
@gdhungana  As we discussed, this is fine.   Will need to follow up with @afausti before recommending merge.

py/desispec/qa/qa_quicklook.py
```diff
@@ -120,7 +121,7 @@ def run_qa(self,image,paname=None,amps=False,qafile=None, qafig=None,qlf=False):
 
         if qafile is not None:
             f=open(qafile,"w")
-            yaml.dump(retval,f)
+            f.write(yaml.dump(yamlify(retval)))
```
profxj: I should think it would be better to have QA I/O handled
by desispec.io.qa

Consider it.

py/desispec/quicklook/qlconfig.py
```diff
@@ -123,17 +127,27 @@ def paargs(self,psfspfile=None):
...
+           
         if filetype in ["fframe","sframe"]: #- fiberflat fielded or sky subtracted intermediate files       
             pafile=os.path.join(self.specprod_dir,'exposures',self.night,"{:08d}".format(self.expid),"{}-{}-{:08d}.fits".format(filetype,self.camera,self.expid))
         else:
             pafile=findfile(filetype,night=self.night,expid=self.expid, camera=self.camera, rawdata_dir=self.rawdata_dir,specprod_dir=self.specprod_dir,outdir=self.outdir)
-            
+
         return pafile
 
     def dump_qa(self):
```
profxj: Same comment as above.  I should prefer I/O be handled
in desispec.io.qa

py/desispec/quicklook/qlconfig.py
```diff
@@ -172,32 +189,39 @@ def qaargs(self):
...
-                             PER_RESID=95.,   # Percentile for residual distribution
-                             BIN_SZ=0.1,) # Bin size for residual/sigma histogram
+        if self.algorithms is not None:
+            for PA in self.palist:
+                if qa in self.qalist[PA]:
+                    params[qa]=self.algorithms[PA]['QA'][qa]['PARAMS']
+
         else:
-            params[qa]= None
+            if qa == 'Count_Pixels':
```
profxj: Are these different from the offline?

Either way, I feel these should be written a file (e.g. desispec_params.yaml)
and read from disk.   Such parameter files can be archived and far more
easily modified than searching for and changing these parameters in this code.

py/desispec/quicklook/qlconfig.py
```diff
@@ -244,95 +268,143 @@ def io_qa(self,qaname):
...
+        outconfig['Camera'] = self.camera
+        outconfig['Expid'] = self.expid
+        outconfig['DumpIntermediates'] = self.dumpintermediates
+        outconfig['FiberMap']=self.fibermap
+        outconfig['FiberFlatFile'] = self.fiberflat
+        outconfig['PSFFile'] = self.psf
+        outconfig['Period'] = self.period
+
+        pipeline = []
+        for ii,PA in enumerate(self.palist):
```
profxj: I have to admit that I don't know what "PA" is.  But 
I probably should..

py/desispec/quicklook/qlconfig.py
```diff
@@ -244,95 +268,143 @@ def io_qa(self,qaname):
...
-            sys.exit(0)
+        
+        if self.thislist is not None:
+            pa_list=self.thislist
+        else: #- construct palist
+            if self.flavor == "arcs":
+                pa_list=['Initialize','Preproc','BoxcarExtract'] #- class names for respective PAs (see desispec.quicklook.procalgs)
+            elif self.flavor == "flat":
+                pa_list=['Initialize','Preproc','BoxcarExtract', 'ComputeFiberFlat_QL']
+            elif self.flavor in ['dark','elg','lrg','qso','bright','grey','gray','bgs','mws']:
```
profxj: Many of these flavors are now defunct.  
See PR #386.

py/desispec/qa/qa_quicklook.py
rkehoe: @profxj  Agreed.  We will use desi spec.io.qa.

py/desispec/quicklook/qlconfig.py
rkehoe: Same here also.  We had not wanted to add further instances flagrantly, but these are conventions based on those already existing and we've decided to stay with that.  We'll add relevant .fits and qa cases.

py/desispec/quicklook/qlconfig.py
rkehoe: The params for quicklook are controlled by an input configuration object, which in normal (production) usage is set by a configuration file.  As you say, such files are important for easy modification and archiving purposes, as well as just humans reading them.  We have had the params stuff partially in place since fall, but now have implemented more completely.  The configuration file example currently in this branch is called qlconfig_dark.yaml for dark time exposures.  We will be adding  configuration examples appropriate to bright time and calibration exposure pipelines soon.  While this particular branch is not complete yet, this configuration file example will be more informative than the version in prior versions.

py/desispec/quicklook/qlconfig.py
rkehoe: 'PA' is what we have always used to refer to pipeline algorithms.  In QuickLook, the QA algorithms are independent of pipeline algorithm step (code wise) and can be moved around.  It has been useful for us, and the pipeline is built Around a configurable list of PAs each with configurable list of QAs attached.  A modern example is in qlconfig_dark.yaml, but it has been there for a long time.  So we keep separate labels for PA and QA algorithms and steps

py/desispec/quicklook/qlconfig.py
rkehoe: @profxj  Yes, we realized this change was coming.  We need to keep track of the difference between dark, bright and grey exposures (and also arc, continuum lamp, etc) because each pipeline will be configured differently.  The science pipelines will all likely run the same algorithms, but it is likely the params will be different (eg. targets to focus on will certainly change), so we need a term to distinguish dark, bright and grey.  We will stop using 'flavor' for this and instead use 'obstype' for observation type unless the recent changes in #386 also provide a new term for this (this was not obvious to me).

afausti:
@rkehoe I have tested the **quicklook_config** branch with qlf and it is working fine.

rkehoe:
@afausti Thanks.  Looking it over before merging.

rkehoe:
This request includes substantial updates from @gdhungana and @rstaten.  It accomplishes several tasks associated with 1) migrating from the prototype QuickLook to a production version, 2) requirements to improve performance when integrated with QuickLook Framework, 3) further implementation of QA information particularly associated with discussions in Utah, and 4) other improvements including addressing comments from @profxj.  This update includes:

- Factoring QL configuration from dependencies on exposure ID, night and camera.  This includes removing all file names so no camera-specific information is given in the QL configuration.  These changes allow one QL configuration file to be used for all cameras and all exposures.  The intention is also to facilitate archiving of the file which should be tracked in association with the resulting QA outputs. 
- QuickLook configuration dramatically shortened by omission of all redundant information that can be determined at run time. These include paths and file names that now strictly adhere to the DESI data model.  They include the command line argument list given to each pipeline algorithm (PA) and QA, which are no longer specified in the standard configuration file.  QuickLook will still allow such defaults to be overridden in some cases to facililate needs during testing, but for now few overrides are implemented. 
- Remove other pipeline execution details such as names of classes and omission of the strict syntax used to set up the pipeline. This and above changes are meant to 1) facilitate archiving of the non-redundant configuration information needed for QL execution, and 2) make the configuration more substantially human readable (and therefore editable), which will be necessary during commissioning.
- use of standard DESI environment variables so no paths are in the QL configuration.
- Implementation of PARAMs for QAs into configuration file.   These are passed thru to QA outputs according to the offline/QL data model discussions in Fall.  Their inclusion in the configuration also satisfies the archiving role of the configuration file as it encapsulates all nonredundant information going into QA (and PA) execution.   Not all configured PARAMs are acted upon yet, but these will be implemented in near-term updates.
- Finish implementing merged QA output for a particular QL pipeline.  The data model of this merged output still needs to be documented in desidatamodel, but it conforms to the model offline/QL model discussed with @xprofxj in Fall.  There may be reason to revisit this data model based on recent comments from @afausti, but the current data model should be used for merged outputs until a new one is developed.
- rationalize intermediate frame file names (fframe, sframe) and implement these in desispec.io.qa.  Remove 'lastframe' and designate final QL output after sky subtraction as 'sframe'.
- Implement several switches to control the output of intermediate files: pix files, intermediate frame files except for 'sframe', sky model files and PA-level QA outputs can all be turned off independently now.  Implemented switch to disable static output plots which are not needed when run via QLF.
- Further implement run-time parameters and QA specifications into PARAMs blocks from QAs.  Implement further scalar metrics and final warning flags for when these metrics are outside of tolerances given in PARAMs.  These include results from discussions w/@kdawson in May.  
- Provide switch to turn off use of spectral resolution.  This permits a 6-8x reduction in the output frame file thru reduction in the dimensionality of the included resolution matrix.  Further updates will require a more intelligent storage of the resolution information to allow its use under QLF usage and testing.
- Implement PROGRAM and 'new' FLAVOR keywords in FITS headers
- Fix problems in numpy data format for QA outputs to 1) make them human readable, and 2) remove Inf entries.  The latter is accomplished for now by placing imaging magnitudes at 26.0 when they are absent in the fibermap file.  A more robust convention may be needed before commissioning.  There do not currently appear to be any other sources of Inf's beyond this omission in the fibermap.
- Initial fix to address problem in XWSigma QA where it was crashing when fitting lines in the z-arm.  These lines are omitted for now.  'This must be studied and XWSigma made more robust in fitting in future updates.

Further work is needed to consolidate these updates, but this version is expected to significantly improve ease of use of quicklook both in standalone and QLF mode.  Following @afausti  confirming QLF and QL work together in this version, I will merge now.

gdhungana:
@sbailey  I am just looking at this now. When this PR was merged, python 3.5 tests were terminated because they were not finished inside the specified time limit. The most recent PRs seem okay. Is this anything to worry about now?

PR #398, opened 2017-06-13
8x faster in 2.5x less memory desi_zcatalog

sbailey:
This PR changes desi_zcatalog to use fitsio instead of astropy.io.table.Table to be 8x faster with 2.5x less memory when run on the 2% survey.

Example output on edison@NERSC:
`/scratch2/scratchdirs/sjbailey/desi/dc17a/spectro/redux/dc17a2/zcat-dc17a2.fits`
producted with
`/usr/bin/time desi_zcatalog --indir spectra-64/ --outfile blat.fits --verbose`


rainwoodman:
congratulations!

PR #399, opened 2017-06-13
Update changes.rst and disable single unit test

### LINKED ISSUE #397: desispec.parallel tests broken on edison login nodes
desispec.parallel unit tests are failing on edison login nodes:
```
[edison06 master] python py/desispec/test/test_parallel.py 
.[Mon Jun 12 12:12:03 2017] [unknown] Fatal error in PMPI_Init_thread: Other MPI error, error stack:
MPIR_Init_thread(525): 
MPID_Init(228).......: channel initialization failed
MPID_Init(601).......:  PMI2 init failed: 1 
Aborted
```
They pass on my laptop which does not have mpi4py installed.  I suspect this came in via Ted's pipeline refactor in PR #394 and is caused by some check that tries to use mpi4py if it is installed or a fake communicator otherwise.  NERSC login nodes are in an evil middle ground: mpi4py is installed but will crash if you try to load it since it tries to initialize MPI and that is forbidden on the login nodes.

@tskisner please investigate and fix.  Getting our tests working again for the collaboration meeting is high priority.

tskisner:
Small changes that close #397 .  In the future, we may have more unit tests that genuinely cannot be run without MPI.  These should eventually be run with a separate optional test command.  Until then, comment out a unit test for a function that is not actually used anywhere in the code.

### DIFF
--- a/bin/desi_group_spectra
+++ b/bin/desi_group_spectra
@@ -8,12 +8,14 @@
 Re-group spectral data from cframe files into healpix indexed groups.
 """
 
+from desispec.parallel import use_mpi
+
 comm = None
-try:
-    import mpi4py.MPI as MPI
+if use_mpi:
+    from mpi4py import MPI
     comm = MPI.COMM_WORLD
-except ImportError:
-    pass
+else:
+    print("mpi4py not found, using only one process")
 
 import desispec.scripts.group_spectra as group_spectra
 
--- a/bin/desi_mpi_zfind
+++ b/bin/desi_mpi_zfind
@@ -1,19 +1,17 @@
 #!/usr/bin/env python
 # See top-level LICENSE.rst file for Copyright information
 
-import desispec.scripts.zfind as zfind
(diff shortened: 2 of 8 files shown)


sbailey:
For consideration:
```
is_nersc_login = ('NERSC_HOST' in os.environ) and ('SLURM_JOB_NAME' not in os.environ)
```
That is tied to NERSC using SLURM, but should be valid for several years at least (I hope...)

tskisner:
I've implemented a global solution like this, am running integration tests on cori login node.  Will push for further review after they complete.

tskisner:
Ready for a fresh review from @sbailey.  Unit tests and integration test run on the login nodes.

sbailey:
Code looks good.  Thanks for the effort to fix this in general and not just to make this test pass.  Will merge once tests finish on Travis.  Travis tests have recently started taking bizarrely long (and 2x variable in runtime) but that isn't unique to this PR.

PR #400, opened 2017-06-14
Database refactor

weaverba137:
This PR:

- Moves database code out of `desispec.io` into its own subpackage `desispec.database`.
- Supports loading both quicksurvey and pipeline results into both SQLite & PostgreSQL databases.

**This PR still needs an additional load test before merging.**  Specifically loading pipeline outputs into PostgreSQL still needs to be tested.


sbailey:
Code looks fine, though I am confused about the naming and roles of `desispec.database.spectra` vs. `desispec.database.datachallenge`.  Which one is for loading quicksurvey results (and thus will be deprecated when quicksurvey outputs converge to the "real" pipeline output formats)?

weaverba137:
It actually says in the docstring for those modules, but the short version is
- spectra.py is intended to be the code that loads "SpectraDB", but right now it is optimized for quicksurvey.
- datachallenge.py is intended to be deprecated, but right now it is optimized to load the pipeline outputs, *i.e.* dc17a.

sbailey:
> * spectra.py is intended to be the code that loads "SpectraDB", but right now it is optimized for quicksurvey.
> * datachallenge.py is intended to be deprecated, but right now it is optimized to load the pipeline outputs, i.e. dc17a.

Thus my confusion, because that seems backwards.  The datachallenge outputs like dc17a2 use the real pipeline to generate files that mimic what the real data outputs will look like, and thus that DB loader should not be the one that is deprecated, and it shouldn't have "datachallenge" in the name since it will evolve into the real DB loader.  The quicksurvey results are the special case that bypasses the pipeline and generates a similar but somewhat different set of outputs; we want to adapt quicksurvey to match the real pipeline (and thus the data challenges).

The caveat is that right now the datachallenge / pipeline outputs don't match what we eventually want (e.g. missing NUMOBS -> NUMTILES, and S/N per band), but it is the WIP draft toward what we do want.

weaverba137:
If you *really* don't like the name, I can rename spectra.py *back* to qucksurvey.py, then rename the two merged files *back* to spectra.py.  Can I just get away with the documentation in the docstrings for now?

weaverba137:
PS, also note that the command line scripts are called `desi_load_quicksurvey` and `desi_load_datachallenge`, so there is less ambiguity there.

weaverba137:
@sbailey, I completely agree that there will need to be renaming.  The goal was to get something that worked, & that could potentially be shown at the collaboration meeting, not to, ah, optimize the namespace. The load test I mentioned is going on right now.

weaverba137:
I'm reasonably sure the load test will complete OK, so I'm merging this now to get it into the next tag.

PR #402, opened 2017-06-16
Load dc17a2

weaverba137:
This PR deals with some minor changes needed to load the healpix based zbest files in `dc17a-twopct`.  In addition to filename changes, this also deals more intelligently with the duplicate `TARGETID` values that crept into these files.  Instead of rejecting the whole file if duplicates are found, it just rejects the rows in the file that are duplicates.


weaverba137:
Successfully loaded the healpix-style zbest files, so merging now.

PR #405, opened 2017-06-17
Enable desi_group_spectra to run without pipeline infrastructure.

tskisner:
This commit: 

- Allows a production to be created without any
   calibration exposures.

 - Fixes some bugs with cori slurm script generation
   since now the "--constraint" keyword is required.

 - Adds a command line switch to desi_group_spectra
   which allows it to get the spectra <--> frame
   dependency information from either the pipeline
   or the filesystem.


sbailey:
Looks good.  Thanks.  I made one change to avoid using lonlat keyword in healpy.ang2pix(), since that wasn't available in healpy 1.9.1 which is the latest version available via the openastronomy conda channel documented on the [healpy installation page](http://healpy.readthedocs.io/en/latest/install.html).  The consequence is needing desimodel 0.7.0 for the wrapper function I used.

sbailey:
Forgot to comment: I've verified that this works with the notebook in desihub/desisim#244, and will post an update to that soon.

thanks for the late night coding fixes.

PR #408, opened 2017-06-18
Notebook tutorial: introduction to reading and manipulating DESI spectra

geordie666:


sbailey:
I made one small change for the fmap -> fibermap rename in PR #407; otherwise looks good.  Thanks.

PR #410, opened 2017-06-19
Found a couple of differences between data on thumb drive and tutoria…

geordie666:
…l [ci skip]


geordie666:
Just an update to the notebook for the collaboration meeting. It's needed this afternoon, and shouldn't break anything, so I'm self-merging.

Issue #77, opened 2015-10-01
resample_flux edge effects


sbailey:
test_resample.py now includes a unitttest flagged as `expectedFailure` because it identifies an edge effect bug in `resample_flux`.  Note how the last resampled bin is way lower that it apparently should be.

```
import numpy as np
from desispec.interpolation import resample_flux

x = np.arange(0.0, 30)
y = np.sin(x/4)+1
xx = np.linspace(1, 29, 13)
yy = resample_flux(xx, x, y)

clf()
plot(x, y, 'b.-')
plot(xx, yy, 'r.-')
```

![screen shot 2015-10-01 at 3 38 26 pm](https://cloud.githubusercontent.com/assets/218471/10235584/7b63a6e2-6852-11e5-83ce-e62818c26ce4.png)

julienguy:
This is not a bug but a feature of the routine resample_flux where, as written in the docstring,

The input is interpreted as the nodes positions and node values of
    a piece-wise linear function y(x) = sum_i y_i * f_i(x)
and
the output value is the average flux density in a bin
        flux_out(j) = int_{x>(x_{j-1}+x_j)/2}^{x<(x_j+x_{j+1})/2} y(x) dx /  0.5*(x_{j+1}+x_{j-1})

This is however an issue for some use of this routine.
For this purpose, I have added an option "extrapolate" which considers that the input flux density is the same as the first (last) input values for output x smaller (larger) than the input x range.
This is in branch "interpolate". When applied to the above case, this gives 
![figure_1](https://user-images.githubusercontent.com/5192160/27867366-bb6ad8cc-6199-11e7-9c6c-31d2fb44c7b0.png)

moustakas:
Fixed in #415.

Issue #414, opened 2017-07-04
resample_flux is doing odd things when extrapolating


moustakas:
In the pre-2% data challenge spectra redrock was failing dramatically on QSOs, which was eventually tracked down to an issue with the templates.  Cell [9] of this notebook demonstrates the issue:
https://github.com/desihub/desisim/blob/qso-extrap/doc/nb/QSO_templates.ipynb
Notice especially the correct extrapolation of the template based on the last "measurement" but then the funny step function that drops the spectrum to zero at large wavelengths (and then back up, sometimes).

I'm using `interpolation.resample_flux` to do the interpolation, so there must be a bug there.  @julienguy did you write this code?  If so, can you please take a look when you can?

julienguy:
This is fixed in branch "interpolate" if you use resample_flux with the option extrapolate=True.
I tested this on your QSO simulations.

see also related issue #77 also fixed in branch "interpolate"

moustakas:
Extrapolating in this way will create problems depending on the derivative of the spectrum on the edges.  In the QSO spectrum examples your method will probably work fine because the spectra are pretty flat at long wavelengths.  

Would it be possible to simply replicate the last value (perhaps as another extrapolation option) to fill out the array?  

Note that I'm going to be experimenting with Ian's QSO templates soon at which point all the templates will have UV-mid-IR wavelength coverage, so this will no longer be an issue.

julienguy:
John, what I did is exactly equivalent to what you are saying if I am not mistaken. You should try the code.

However, if you have some specific extrapolation needs for QSOs, it would be better that you write a custom routine better suited for your purpose. 

There are zillions of different ways to interpolate data. The specific role of resample_flux (at least when I wrote it) is to project a high resolution model spectrum on a data wavelength grid where the input spectrum is considered a piece-wise continuous flux density and the output the average flux density in bins centered on the output wavelength array (because this is what we need to interpret the data, once convolved the resolution). I added the extrapolation for convenience, but now I realize that we'd better write several 1D interpolators in interpolation.py

moustakas:
#415 fixes this issue beautifully, thank you, @julienguy!

PR #415, opened 2017-07-05
Interpolation

julienguy:
Fixes issues #77 and #414 in interpolation.resample_flux by adding an optional argument "extrapolate".


julienguy:
What is the easiest way to test the docstring formatting before a pull request? It would be useful to put this information on the wiki.

sbailey:
Thanks for this fix.  Code looks fine; we'll still need to check for uses of `resample_flux` and consider updating them to `extrapolate=True` on a case-by-case basis.  Ideally we'd fix travis+numpy prior to merging, but ok to merge now if this is a blocking factor for something else.

I thought `python setup.py build_sphinx` should build the docs and report formatting errors, but now I'm getting errors importing `desiutil.log` even though I have desiutil in my path.
```
$ python setup.py build_sphinx
...
  File "/Users/sbailey/desi/git/desispec/py/desispec/util.py", line 16, in <module>
    from desiutil.log import get_logger, INFO
ImportError: No module named 'desiutil.log'

$ pydoc desiutil.log | head -10
Help on module desiutil.log in desiutil:

NAME
    desiutil.log

DESCRIPTION
    ============
    desiutil.log
    ============
 ```

**Comment by sbailey** [comment]
@weaverba137 could you take a look at these sphinx failures related to desispec/database/datachallenge.py?  I don't understand why they didn't crop up when you last updated the database code (PR #402 I think).  I already added a bunch of sqlalchemy related autodoc_mock_imports that I thought should have been previously necessary, and I'm still getting some pretty basic sphinx failures on my laptop, though different from the failure in Travis.

Travis sphinx failure:
```
Warning, treated as error:
/home/travis/build/desihub/desispec/doc/api.rst:25: WARNING: autodoc: failed to import module 'desispec.database.datachallenge'; the following exception was raised:
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/Sphinx-1.5.6-py3.5.egg/sphinx/ext/autodoc.py", line 560, in import_object
    __import__(self.modname)
  File "/home/travis/build/desihub/desispec/py/desispec/database/datachallenge.py", line 42, in <module>
    class Truth(SchemaMixin, Base):
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

example laptop sphinx failure from `python setup.py build_sphinx`, where it appears to be actually running code and not just parsing it:
```
WARNING: /Users/sbailey/desi/git/desispec/doc/api.rst:10: (WARNING/2) autodoc: failed to import module 'desispec.bootcalib'; the following exception was raised:
  File "/Users/sbailey/anaconda/envs/desi/lib/python3.5/site-packages/Sphinx-1.6.2-py3.5.egg/sphinx/ext/autodoc.py", line 657, in import_object
    __import__(self.modname)
  File "/Users/sbailey/desi/git/desispec/py/desispec/bootcalib.py", line 28, in <module>
    from astropy.modeling import models, fitting
  File "/Users/sbailey/anaconda/envs/desi/lib/python3.5/site-packages/astropy/modeling/__init__.py", line 11, in <module>
    from . import models
  File "/Users/sbailey/anaconda/envs/desi/lib/python3.5/site-packages/astropy/modeling/models.py", line 10, in <module>
    from .core import custom_model  # pylint: disable=W0611
  File "/Users/sbailey/anaconda/envs/desi/lib/python3.5/site-packages/astropy/modeling/core.py", line 36, in <module>
    from ..table import Table
  File "/Users/sbailey/anaconda/envs/desi/lib/python3.5/site-packages/astropy/table/__init__.py", line 28, in <module>
    from .column import Column, MaskedColumn
  File "/Users/sbailey/anaconda/envs/desi/lib/python3.5/site-packages/astropy/table/column.py", line 13, in <module>
    from ..units import Unit, Quantity
  File "/Users/sbailey/anaconda/envs/desi/lib/python3.5/site-packages/astropy/units/__init__.py", line 14, in <module>
    from .core import *
  File "/Users/sbailey/anaconda/envs/desi/lib/python3.5/site-packages/astropy/units/core.py", line 2319, in <module>
    dimensionless_unscaled = CompositeUnit(1, [], [], _error_check=False)
  File "/Users/sbailey/anaconda/envs/desi/lib/python3.5/site-packages/astropy/units/core.py", line 1995, in __init__
    self._expand_and_gather(decompose=decompose, bases=decompose_bases)
  File "/Users/sbailey/anaconda/envs/desi/lib/python3.5/site-packages/astropy/units/core.py", line 2066, in _expand_and_gather
    for b, p in zip(self.bases, self.powers):
TypeError: iter() returned non-iterator of type 'NoneType'
```

**Comment by weaverba137** [comment]
@sbailey, looking at this now.  What version of astropy are you using on your laptop?

**Comment by sbailey** [comment]
```
(desi) ~ $ python -c "import astropy; print(astropy.__version__)"
1.2.1
```

**Comment by weaverba137** [comment]
I'm making some progress, at least on my own systems:

- Sphinx v1.6.2
- Anaconda environment only contains packages needed for `python setup.py build_sphinx`.  That is, I'm relying on the dummy modules defined in doc/conf.py.
- With that setup, I only get two warnings, related to SQLAlchemy, & those are going to be tricky, but at least it's narrowed things down.

Not directly related, but it seems a bit odd that we're defining desiutil modules in doc/conf.py.  It's not possible to do `python setup.py` at all without desiutil installed, so the desiutil imports should *always* work.

**Comment by sbailey** [comment]
I was also surprised by the apparent necessity of desiutil mock modules.  I added that when I was in a state where desiutil was in my path and could be loaded, but "python setup.py build_sphinx" would raise an ImportError on desiutil.log.  I have not been able to reproduce that on master (many many other errors though...)

**Comment by weaverba137** [comment]
So there seems to be two different sets of warnings/errors:

1. If you `build_sphinx` in an empty environment, and rely on mocked-up modules, you get the error related to metaclass conflicts as noted by @sbailey.  This is annoying, but not really surprising, because we can't expect mock modules to do what SQLAlchemy is doing under the hood to support its ORM.
2. If you `build_sphinx` in an environment that has SQLAlchemy, astropy, etc. installed, then you get these errors related to an iterator returning `NoneType` (as also noted above).  However, you do *not* get the metaclass conflicts, presumably because SQLAlchemy takes care of that for you.  

So the second set of errors is actually more serious, because that behavior *is* surprising.

**Comment by weaverba137** [comment]
As a sanity check, I went back to the master branch and did a clean build with no errors with Sphinx = **1.5.6**, and a full environment (astropy, numpy, scipy, etc. all installed).

**Comment by weaverba137** [comment]
Another anomaly: I just noticed that some of the earlier builds were showing a warning about specter being unavailable, even though it is explicitly on the list of things to install.

**Comment by weaverba137** [comment]
Interesting.  It's starting to look like we might be using [autodoc_mock_imports](http://www.sphinx-doc.org/en/stable/ext/autodoc.html?highlight=sphinx.ext.autodoc#confval-autodoc_mock_imports) *too much*.

The fundamental question is this:  When Sphinx interprets the `autodoc_mock_imports` option, does it create a mock module *even if the real module is installed in the environment*, or does it always prefer the real module to a mock module?  If it is doing the former, that would explain a lot of what I'm seeing.

It think the correct way to proceed, for the time being is:

* Install all the packages that would normally be installed for `python setup.py test`.
* Do not use Sphinx 1.6.
* If you don't want to install sqlalchemy on a "home" system, ignore any warnings about metaclasses, etc. when doing `python setup.py build_sphinx` "at home".

**Comment by weaverba137** [comment]
I believe this will be ready to merge after the latest round of tests complete.  There are still some sqlalchemy warnings when doing `build_sphinx`, but these are both harmless and the `build_sphinx` command does not interpret these as errors.

**Comment by sbailey** [comment]
Thanks for tracking down these sphinx problems.  The remaining odd messages to ignore (and which don't trigger test failure) are:
```
/home/travis/miniconda/envs/test/lib/python3.5/site-packages/sqlalchemy/ext/declarative/api.py:182: SAWarning: Unmanaged access of declarative attribute __tablename__ from non-mapped class SchemaMixin
  (desc.fget.__name__, cls.__name__))
/home/travis/miniconda/envs/test/lib/python3.5/site-packages/sqlalchemy/ext/declarative/api.py:182: SAWarning: Unmanaged access of declarative attribute __table_args__ from non-mapped class SchemaMixin
  (desc.fget.__name__, cls.__name__))
```
On my laptop, I was able to revert to sphinx 1.5.6 via
```
conda install sphinx=1.5
```
and then `python setup.py build_sphinx` worked (SAWarnings but not pages of errors).

PR #422, opened 2017-08-08
QA Prod updates

profxj:
Some refactoring of QA Prod to run on 2% test

Includes a new read_meta_frame() method with test.

Stripped out a significant number of SKYSUB QA that
was inserted by QL.  Changes should not affect QL QA.

Has passed tests on Travis (healpy is not running well on my Python 3.6)


gdhungana:
Should not affect QL. Few of these changes are recent ones, so pinging @rstaten as well. As it seems significant updates are striped, upcoming updates may/ shouldn't reside there. A future refactor may handle that.

py/desispec/io/frame.py
```diff
@@ -91,6 +91,19 @@ def write_frame(outfile, frame, header=None, fibermap=None, units=None):
 
     return outfile
 
+def read_meta_frame(filename):
+    """ 
+    Args:
+        filename: path to a file, or (night, expid, camera) tuple where
```
sbailey: Is this function really needed as a future abstraction layer, or could you just call `astropy.io.fits.getheader(filename)` for now?

If you do keep it, at minimum you should also include an optional `extname` and correct the docstring (the code doesn't accept a (night,expid,camera) tuple but the docstring says it does).

profxj: I prefer how we have the I/O module handle most/all
of the I/O.  Could be we'll change the format of saving
Frame files (e.g hdf5), although I suppose that isn't too likely.

Anyhow, I've kept this and made the requested changes.

profxj:
@rstaten -- Please let me know if you are ok with this PR.
If so, I'll merge.  Thanks.

rstaten:
@profxj I checked out the branch and QL runs fine and all tests pass locally. I have no problem with merging this PR.

profxj:
Merging

PR #423, opened 2017-08-10
Minor modifications to teststand

julienguy:
This PR to synchronize the teststand branch with master.
It's only minor modifications that only affect the preprocessing of test stand data.
The goal is to merge this before including new developments on the pixel level flat field.


sbailey:
Looks good.  Go ahead and merge as soon as the travis tests pass.

PR #424, opened 2017-08-11
QA frame script

profxj:
Introduces a desi_qa_frame script to generate QA for an input frame file.

Adds new get_nights() and search_for_framefile() methods with tests.
Adds documentation on QA scripts (qa.rst)

Tested successfully on 2% challenge files.

### DIFF
--- a/py/desispec/io/frame.py
+++ b/py/desispec/io/frame.py
@@ -171,8 +171,16 @@ def read_frame(filename, nspec=None):
 
 
 def search_for_framefile(frame_file):
+    """ Search for an input frame_file in the desispec redux hierarchy
+    Args:
+        frame_file:  str
+
+    Returns:
+        mfile: str,  full path to frame_file if found else raise error
+
+    """
     log=get_logger()
-    # Parse frame file
+    # Parse frame file name
     ifile = frame_file.split('/')[-1]
     splits = ifile.split('-')
     root = splits[0]
--- a/py/desispec/io/frame.py
+++ b/py/desispec/io/frame.py
@@ -180,8 +180,8 @@ def search_for_framefile(frame_file):
 
     """
     log=get_logger()
-    # Parse frame file name
-    ifile = frame_file.split('/')[-1]
+    # Parse frame file
+    path, ifile = os.path.split(frame_file)
     splits = ifile.split('-')
     root = splits[0]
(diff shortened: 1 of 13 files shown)


py/desispec/io/frame.py
```diff
@@ -169,3 +168,32 @@ def read_frame(filename, nspec=None):
...
+    ifile = frame_file.split('/')[-1]
+    splits = ifile.split('-')
+    root = splits[0]
+    camera = splits[1]
+    fexposure = int(splits[2].split('.')[0])
+
+    # Loop on nights
+    nights = get_nights()
+    for night in nights:
+        for exposure in get_exposures(night):
```
sbailey: it appears that you could just directly jump to
```
mfile = findfile(root, camera=camera, night=night, expid=fexposure)
if os.path.exists(mfile):
    ...
```
i.e. no need to loop over `get_exposures(night)` if you already know the exposure number you are looking for.  Unfortunately you do still need to loop over nights since we don't have a mapping from exposure -> night.

py/desispec/io/frame.py
```diff
@@ -169,3 +168,32 @@ def read_frame(filename, nspec=None):
...
+        mfile: str,  full path to frame_file if found else raise error
+
+    """
+    log=get_logger()
+    # Parse frame file name
+    ifile = frame_file.split('/')[-1]
+    splits = ifile.split('-')
+    root = splits[0]
+    camera = splits[1]
+    fexposure = int(splits[2].split('.')[0])
```
sbailey: minor: `os.path.split()` and `os.path.splitext()` are more standard ways of doing some of this logic.

profxj: ok, implemented os.path.split()

py/desispec/io/frame.py
profxj: Well, I still need to catch cases when the exposure folder exists
but not the frame file.  Have done so.

profxj: Actually, I prefer my way (headed back to it).
This way the error throws that the exposure folder was found
but no frame file in it.

profxj:
@julienguy -- let me know if you wish to review this before I merge.  You may be
one of the few who would actually use the script..

profxj:
Merging

PR #425, opened 2017-08-14
Offline QA for SkySub Residuals

profxj:
A series of changes to implement new QA (figures)
for skysub residuals.

* Add frame_meta to parameters (for slurping the Frame headers)
* Add get_reduced_frames() method to io.meta
* Modifies QA_Prod meta file output to be JSON
* Add load_meta() method to QA_Exposure
* Add time_series ploting to desi_qa_prod
* Add several new plots for skysub residuals
* Adds method to generate QA Table for Prod
* Refactor of skysubresid script

Here are a couple of example PNGs:

1. Exposure (not the best)
qa_skyresid_wave_expid_95b

2.  Production level 

skyresid_prod_dual_r

3.  Gaussianity

skyresid_prod_gauss_z


sbailey:
Looks good.  Highlighting one change where we should coordinate with online QA:  this switches from YAML to JSON, presumably for performance when slurping in a full production worth of QA files.  We can be temporarily out of sync while trying different format ideas, but eventually we should converge on the same basic format and structure for both online and offline QA.

PR #428, opened 2017-08-17
fix integration test for QA updates

sbailey:
This PR fixes some problems with the integration test and its compatibility with the latest QA code.  In particular, the QA code is correctly becoming more picky about allowable exposure flavors and the integration test was sloppy about program 'dark' vs.  flavor 'science'.

This remains an area where we need to coordinate with ICS to formalize the allowable values, but at least this gets us back into a self-consistent old integration test state.

It also fixes a simspec data model incompatibility in the full pipeline integration test, though that doesn't yet have redrock integration so it still doesn't complete without errors.  That will be a separate PR when redrock is integrated with pipeline.

### DIFF
--- a/py/desispec/qa/qa_exposure.py
+++ b/py/desispec/qa/qa_exposure.py
@@ -36,7 +36,7 @@ class QA_Exposure(object):
             All input args become object attributes.
         """
         desi_params = read_params()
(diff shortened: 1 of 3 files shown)


sbailey:
self merging to get this in before the nightly update+tests at NERSC.

PR #429, opened 2017-08-18
Offline QA Refactor

profxj:
This PR moves offline QA files out of the main calib2d and exposures
folders of the redux tree and puts them in a parallel tree named QA.

Other changes include:

- Code to generate static HTML with links to QA figures
- Exposure level QA on fiberflat  (see example figure below)
- Normalizing the fiber flat by fiber area
- Fixed ZP calculation to include exposure time
- Documentation and tests were updated

Travis is failing as fiber_area_arcsec2() is now in desimodel
and this change hasn't yet propagated.

QA generated with:

desi_qa_exposure --expid=96 --qatype=fibermap --channels=b

qa_00000096_b_fibermap


py/desispec/scripts/qa_prod.py
```diff
@@ -6,7 +6,7 @@
 
 
 def parse(options=None):
-    parser = argparse.ArgumentParser(description="Generate/Analyze Production Level QA [v1.3]")
+    parser = argparse.ArgumentParser(description="Generate/Analyze Production Level QA [v1.4.1]")
```
sbailey: Is QA versioning separate from the desispec versioning needed?  Perhaps, but let's be explicit about that choice and reasons.

Maintaining this version string is going to be a pain, e.g. elsewhere in this PR you refer to version 1.4 instead of 1.4.1...  Let's add a QA version variable, and populate the usage/help messages from that so that updating the version number only requires updating one place.

doc/qa.rst
```diff
@@ -124,8 +156,9 @@ Here is the usage::
     usage: desi_qa_prod [-h] [--reduxdir REDUXDIR] [--make_frameqa MAKE_FRAMEQA]
                         [--slurp] [--remove] [--clobber]
                         [--channel_hist CHANNEL_HIST] [--time_series TIME_SERIES]
+                        [--html HTML]
 
-    Generate/Analyze Production Level QA [v1.3]
+    Generate/Analyze Production Level QA [v1.4]
```
sbailey: Here's the other place with 1.4 instead of 1.4.1

profxj:
Added a QA offline versioning approach.

PR #434, opened 2017-09-01
Offline QA: Bright/dark and MED_SKY metric

profxj:
Adds a new metric to QA SkySubtraction (MED_SKY).

Also allows for bright/dark split in production level,
time series QA.

updated docs


gdhungana:
@profxj It's just merged, but useful metric for monitoring sky level. We should use this for QL as well. Thanks.

PR #436, opened 2017-09-05
QA flat fix

profxj:
Modify SkySub QA (post-pipeline) to flat-field prior
to subtraction.

This was the cause of the large residuals seen in the QA plots.

### DIFF
--- a/py/desispec/qa/qa_frame.py
+++ b/py/desispec/qa/qa_frame.py
@@ -254,8 +254,8 @@ def qaframe_from_frame(frame_file, specprod_dir=None, make_plots=False):
         # Flat field first
         dummy_fiberflat_fil = meta.findfile('fiberflat', night=night, camera=camera, expid=expid,
                                       specprod_dir=specprod_dir) # This is dummy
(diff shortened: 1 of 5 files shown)


julienguy:
- I get an error because the code cannot file the fiberflat file (this is my fault because I had to duplicate the directory structure for writing the QA files). But the code error message is not very helpful 
```
File "/global/homes/j/jguy/software/desispec/bin/desi_qa_frame", line 16, in <module>
    qa_frame.main(args)
  File "/global/homes/j/jguy/software/desispec/py/desispec/scripts/qa_frame.py", line 40, in main
    _ = qaframe_from_frame(args.frame_file, specprod_dir=specprod_dir, make_plots=args.make_plots)
  File "/global/homes/j/jguy/software/desispec/py/desispec/qa/qa_frame.py", line 261, in qaframe_from_frame
    fiberflat = read_fiberflat(fiberflat_files[0])

```
 - It would be useful to add to desi_qa_frame an argument for the output directory (to avoid this step of duplicating the directory structure and making sym links).

julienguy:
After fixing my sym links, I was able to run the code successfully.

PR #438, opened 2017-09-19
Sky Continuum in offline QA

profxj:
This PR has two objectives:

- Implement the Quicklook QA algorithm for sky continuum into offline
- Begin moving QA parameters into desispec_param.yml

The first item required refactoring the main algorithm into qalib.py
The second cleans up the organization of QA.

Travis is giving occasional failures related to downloading speclite (not on 
specific tests related to this repo).

The test_binscripts test tickles all of the new code.


gdhungana:
@profxj Thanks, internal refactor looks good. For future, we should also move other QL params to the desispec_param.yml as well taking this example. @rstaten Please confirm on the results with an example QL test. Otherwise looks fine.

profxj:
Yes, assuredly; I plan to move all the other QA params there but I wanted
your buy in before doing so.

rstaten:
Everything looks fine by me, QL runs on all channels without a problem.

profxj:
Ready to merge @rkehoe ?

sbailey:
Test failure is due to sphinx version problem that was fixed with PR #439 . This should be OK to merge once you guys are in agreement that the code is ok.

profxj:
Am waiting for @rkehoe to give his input.

rkehoe:
@gdhungana, @profxj The QL params are all included in the QuickLook qlconfig_dark.yaml already, which is needed by QuickLook at startup time.  Down the road we will want to consider whether online and offline can use a common configuration system.  But for now, they accomplish somewhat different tasks and are used very differently in their respective pipelines.

I'm really glad to see this effort to integrate more QA code across the online/offline divide.  I'm taking a look at the modifications and will provide some feedback shortly.

profxj:
Merging.

PR #439, opened 2017-09-21
Redrock pipeline integration

tskisner:
Integrate the redrock redshift fitting tool into the pipeline:

  - Update the redrock worker class to match upstream interface

  - Unify the slurm script generation between all NERSC machines.

  - Compute the redrock batch times based on empirically
    determined run times.

Note that these changes are designed to work with the current MPI branch of redrock (which will eventually migrate into master).  We can pre-emptively make these changes to the pipeline since the current redshift fitting is essentially unusable, so there is nothing to lose.


tskisner:
One note:  these changes still are not very "smart" about changing runtimes for the different machines for each pipeline step.  That is something that will be added soon.

tskisner:
Hmm, not sure why the sphinx build is failing now, will look into it.

PR #440, opened 2017-09-22
Pipeline for new specex

tskisner:
This is Julien's pipeline changes to be compatible with recent specex work.  Branch is rebased against current master.  Opening this PR for final testing.


tskisner:
I have tested this on a small production with the corresponding version of specex (just merged in that repo).  This all works great and is much faster than the previous code.

PR #441, opened 2017-09-22
Teststand

julienguy:
Improvements/cleanup in codes to prepare the new season of tests at WINLIGHT

Add
 - desi_compute_gain (photon transfer curve analysis based on flatfield slit data)

Improvements on 
 - desi_compute_crosstalk (use several image and average)
 - desi_compute_pixel_flatfield (can first fit for mean spectrum and transmission)
 - preprocessing (add options to desi_preproc , update yaml file with improved cross-talk)

Fix minor bug in desispec.io.raw.py


sbailey:
OK to merge for now, but longer term we should factor out the useful pieces of the new bin scripts into functions that could be called from other programs too.

Another note for the record: we're currently keeping the CCD calibration data in a yaml file in desispec.  This is convenient to not have to sync desispec with a different repo, but as we enter operations we'll want to separate that out so that we can update gains, etc. without having to cut a new version of the code.

PR #448, opened 2017-10-17
enable optional MPI parallelism for desi_extract_spectra script

sbailey:
This PR adds a `--mpi` option to the `desi_extract_spectra` script to enable MPI parallelism at the command line level.  Previously this was only available when running extractions within the context of the pipeline, which directly calls `desispec.scripts.extract.main_mpi()`.

e.g.
```
srun -n 32 -c 2 --cpu_bind=cores desi_extract_spectra --mpi ...
```

It also fixes an inconsistency between the default wavelength step for the MPI vs. non-MPI versions.

### DIFF
--- a/py/desispec/scripts/extract.py
+++ b/py/desispec/scripts/extract.py
@@ -50,6 +50,7 @@ def parse(options=None):
     parser.add_argument("--nwavestep", type=int, required=False, default=50,
                         help="number of wavelength steps per divide-and-conquer extraction step")
     parser.add_argument("-v", "--verbose", action="store_true", help="print more stuff")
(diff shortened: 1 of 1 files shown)


tskisner:
This is a small change and obviously fine, but should we also go back and unify the mpi and non-mpi code into a single main()?  This particular code was the very first one we did this to (adding mpi communicator to the entrypoint after moving the entrypoint into the package).  Many of the other codes in desispec and desisim simply have one main with an optional communicator.  However, we can also leave that for another day- no need to couple it to this small change.

sbailey:
Yes, I would like to merge those to different main functions, but purposefully didn't chase that in this PR.  "another day..."

PR #449, opened 2017-10-25
QL Merge QA schema update

rkehoe:
This PR updates the hierarchical QuickLook merged QA file to incorporate feedback from @afausti.  Specifically, it specifies keyword:value pairs for night, expid, and camera, which are implemented as lists of nights, exposures and cameras.  It also fixes a problem in outputting these files in Py 3.5.

This merged QA output is implemented for QL pipelines for dark time science exposures, and for arc lamp exposures already configured in master.  As before, it can be accessed via the --mergedQA command line option.

@profxj Please review this PR and provide feedback.  We would like to merge to master as soon as possible.

(This version is not currently intended for near-term QLF development, as it is not based on the desispec 0.16.0 reference release.  It should not be confused with merge_qa_0.16.0 designed for that purpose specifically based on requirements in last week's QLF Display Workshop.)


py/desispec/quicklook/quicklook.py
```diff
@@ -260,12 +260,16 @@ def runpipeline(pl,convdict,conf,mergeQA=False):
     qlog=qllogger.QLLogger("QuickLook",0)
     log=qlog.getlog()
     passqadict=None #- pass this dict to QAs downstream
-
+    StepsArr=[]
+    mergedQADict={'NIGHTS':[{'NIGHT':conf['Night'],'EXPOSURES':[{'EXPID':conf['Expid'],'FLAVOR':conf['Flavor'],'CAMERAS':[{'CAMERA':conf['Camera'],'PIPELINE_STEPS':StepsArr}]}]}]}
```
profxj: Not the easiest code to read, but I suspect it is fine.

py/desispec/quicklook/util.py
```diff
@@ -5,51 +5,60 @@
 put pipeline related utility functions here.
 """
 
+import os
 import numpy as np
 import yaml
+from desispec.quicklook import qllogger
 
-def merge_QAs(qaresult):
+def merge_QAs(qaresult,config):
```
profxj: Are we going to merge the same way between QL and offline?

If so, this method should be in a different place.

py/desispec/quicklook/quicklook.py
rkehoe: This line is providing the schema for the top level merged dictionary.  It should be directly related to the structure used for offline.

rkehoe: @profxj Attached please find merged QA files for dark science, and arc lamp exposure processing.  They show the structure produced by this branch.  Please note that the order of the keys is scrambled in the yaml when it is read, but the format closely follows the discussed data model refinements.  

[mergedQA.tar.gz](https://github.com/desihub/desispec/files/1415439/mergedQA.tar.gz)

py/desispec/quicklook/util.py
rkehoe: We also concluded merge_QAs was not the place to do this, but perhaps for different reasons.  

In QL the merge is very brief, being specified just by the schema lines above, and gathering the individual QA dictionaries as the pipeline is executed.  So the util.py code is no longer executed, and it would be removed either when we merge the PR or in a later branch.

Perhaps the most important thing for the merged QA implementation is to have a common place for the definition of the schema, to avoid diverging implementations.  Let me know if you agree.

rkehoe:
@profxj There may have been a misunderstanding.  We were going to implement the new model for QL.  This does not entail modification of any code shared with offline, but provides an example.  I believe a wider group of those in the Data group would be asked to comment on this.  As I indicated, if it will make sense we can consider whether to factor the scheme so both can use it.

py/desispec/quicklook/util.py
samikama: This method is obsolete and should be removed.

rkehoe:
@profxj After discussing with @samikama he has produced an implementation that provides reusable code you could tap.  A file permission issue is preventing finishing testing this, but he will commit it to the branch either this evening or tomorrow.

samikama:
With new commit I added a new class so that Offline can use it to create the same schema. Quicklook is already making use of the new class. Use is rather straight forward. Top level object is instantiated with parameters such as night, exposure id, flavor and camera and then at each pipeline step a new entry to schema is added by addPipelineStep() call which returns another object. QA params and metrics dictionaries for given step added to this object by addParams() and addMetrics() calls respectively. @profxj can use these classes to create identical structure in offline as well.

rkehoe:
Thanks!  

@profxj This adds substantial lines of code to the QL implementation, but the hope is this will assist offline efforts, and future compatibility insurance for online and offline merged QAs for QLF use.  Please indicate if it accomplishes the former.

This PR will need a merge from master, plus a unit test ensuring the mergedQA file is output when requested on the command line.  @rstaten I will wait for that, and further comment from @profxj, before merging.

py/desispec/quicklook/merger.py
```diff
@@ -0,0 +1,45 @@
+#
+# A class to merge quicklook qa outputs.
+#
```
profxj: If this is intended to be used for offline (soon)
as I believe it should, then I strongly request it
be written as a stand-alone bit in qa.utils
and then you can wrap it within a QL object in
quicklook to your heart's desire.

Thanks.

profxj:
ps.  Extra credit to Team SMU if someone actually tries
this on the offline QA.  Part of the idea was for them to
do so in order to become more familiar with the offline QA.

As I have been doing while refactoring bits and pieces
of the QL QA.   Double Q scores 20!

py/desispec/quicklook/merger.py
samikama: @profxj  Could you please clarify this comment? This is a standalone class, it doesn't have any dependencies other than yaml, desiutil.io.yamlify and if you want to get json output as well, json module. It is imported into a project by 
from desispec.quicklook.merger import QL_QAMerger

What do you mean standalone? Do you mean you don't want to use line above?

Regards,
Sami

profxj: I mean it has QL written all over it, @samikama 

That at least implies it is to be used for QL,
not offline QA.

rkehoe:
@profxj Sorry for my delay, lots of mtgs/teaching yesterday/this morning.  I would be glad to help out with offline more in the near term if I could, as I have done in the past.  Unfortunately, I don't have any spare cycles right now to actually integrate this schema into offline.  This is why I could not commit to this in Sep. also.  The best we could do given your Monday request is to abstract the implementation into a class as @samikama has done.   My earlier understanding was that you wanted to see our implementation before working on offline.  I hope this will be of use to you and I would encourage you to adapt it to help ensure QLF sees the same schema.

The purpose of this PR is to a) update QL for the small changes in the hierarchical mergedQA data model proposed earlier, and b) to fix a problem in Py 3.5 running with it's output.  The PR accomplishes these items in addition to providing the class for offline work following your initial request.  Since it is important for us to complete this PR for other pending work which QLF will require, and there don't appear to be any technical issues, I would like to go forward and merge this today.  It will be more appropriate to address the offline update in a separate dedicated branch.

rkehoe:
Oh yeah, as a follow up to you @profxj:  I totally agree with the value of us working with the offline QA and getting more team members experience with that!  Thank you for your efforts, also.

rkehoe:
Thanks @rstaten this looks ready to merge.

PR #450, opened 2017-10-25
Ql calib config

rkehoe:
This PR combines several modest changes relevant to running on different exposure types, including

- initial configurations for bias and dark exposures, in addition to the existing arc pipeline configuration
- configurations for bright and grey time, in addition to the existing dark time pipeline
- adoption of naming 'dark/grey/brightsurvey' for science exposures, and 'dark current' for darks
- clarify use of useResolution switch in arc processing

In meta.py, findfile will eventually need to be worked on to simplify the long list of output types, but this will come in a later branch.  @gdhungana Please take a look at these edits.  I would like to merge to master today.


py/desispec/quicklook/qlconfig.py
```diff
@@ -252,14 +256,22 @@ def io_qa_pa(self,paname):
...
-                 'Preproc': 'ql_preproc',
-                 'BootCalibration': 'ql_bootcalib',
-                 'BoxcarExtract': 'ql_boxextract',
-                 'ResolutionFit': 'ql_resfit',
-                 'ApplyFiberFlat_QL': 'ql_fiberflat',
-                 'SkySub_QL': 'ql_skysub'
-                 }
+        if self.conf["Flavor"] == 'arcs':
+            filemap={'Initialize': 'ql_initial_arc',
+                     'Preproc': 'ql_preproc_arc',
```
gdhungana: How is "ql_intitial_arc" different from ql_initial?

py/desispec/quicklook/qlconfig.py
```diff
@@ -278,17 +290,25 @@ def io_qa(self,qaname):
...
-                 'CountSpectralBins': 'ql_countbins',
-                 'Sky_Continuum': 'ql_skycont',
-                 'Sky_Peaks': 'ql_skypeak',
-                 'Sky_Residual': 'ql_skyresid',
-                 'Integrate_Spec': 'ql_integ',
-                 'Calculate_SNR': 'ql_snr'
-                 }
+        if self.conf["Flavor"] == 'arcs':
+            filemap={'Bias_From_Overscan': 'ql_getbias_arc',
+                     'Get_RMS' : 'ql_getrms_arc',
```
gdhungana: Same here. Do we explicitly need "arc" string in the file names? Isn't unique expid sufficient. It seems there are many of these.

gdhungana:
@rkehoe It seems most of the updates are placeholders for early calibration. I am not sure the simulations are available and are tested with QL pipeline. Otherwise, I only comment on the files written out are getting quite many. So that may need to be figured out internally.

rstaten:
@gdhungana These extra files are included because they have a different path in findfile. While the science output is sent to exposures, calibration output is sent to calib2d.

rkehoe:
@gdhungana Thanks for your feedback.  You pointed out a couple other areas where the file multiplicity is becoming an issue.  We do need to simplify this, but I'll move that to a future branch.
The bias and dark configurations are sort of placeholders until their inputs arrive, but they can be run mechanically on regular existing synthetic inputs.  This will allow QLF to toggle them and work on their display.  Bright survey and greysurvey of course exist, so these can be run on those files.  The run-time params are not tuned yet.  

I'll go ahead and merge next.

PR #451, opened 2017-10-25
use specter ex2d nsubbundles option

sbailey:
This PR uses the specter.extract.ex2d nsubbundles option for faster extractions.  It uses a default of 6 subbundles per bundle, based upon timing tests on a Cori Haswell node.
runtime

This requires >=specter/0.8.0 to run real extractions, and >=specter/0.8.1 to pass tests (there was a bug with really small test extractions that were requesting more subbundles than fibers...).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ env:
         - SPHINX_VERSION=1.5
         - DESIUTIL_VERSION=1.9.5
         - SPECLITE_VERSION=0.5
-        - SPECTER_VERSION=0.8.0
+        - SPECTER_VERSION=0.8.1
         # - DESIMODEL_VERSION=0.7.0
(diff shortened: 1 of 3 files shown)


julienguy:
Tested in prod at NERSC with desi_pipe_run_mpi on b camera. Same gain as reported by Stephen on cori,
a smaller gain on edison. This difference is not understood yet but we will merge anyway.

 - edison (using 24 cores) nsubbundles=1 -> 10min14sec per core per camera bundle of 25 fibers
 - edison (using 24 cores) nsubbundles=6 ->  8min13sec
 - cori (using 32 cores) nsubbundles=1 -> 8min0sec
 - cori (using 32 cores) nsubbundles=6 -> 5min20sec

weaverba137:
Is there any reason not to delete this branch?

PR #458, opened 2017-11-07
Sky line QA and more

profxj:
This PR does the following:

- Refactors the sky line residual QA to work for both QL and offline
- Turns off amps in QL QA
- Adds a new skyline QA (not fully implemented yet) as per @julienguy 's idea

On the latter, one can run on a night of data with:

```
desi_qa_skyresid --skyline --nights 20190902 --channels b
```

to produce a plot like:

skyline_b

This one isn't so interesting.  But maybe the sky subtraction is quite
good here.


julienguy:
- code tested on latest pixsim prod
 -  there is an error when writing the image if directory QA doesn't exist
 -  improve axis range to better see deviations from 1
 - we would need to see the whole wavelength range.  Systematic sky residuals are seen in current sims on edges of camera wavelength range because of the wavelength solution extrapolation uncertainties (beyond the last arc lamp lines).

sbailey:
@julienguy could you point us to the outputs of your processing of /global/projecta/projectdirs/desi/spectro/sim/month/ prior to the specex wavelength calibration fix (if those still exist)?  We know those had big sky subtraction problems in the z-band, and we should make sure that sky line QA would actually catch that.

julienguy:
- large sky residuals in ```/global/cscratch1/sd/jguy/sim2017/redux/month```
 - smaller sky residuals in ```/global/cscratch1/sd/jguy/sim2017/redux/month3```

The sky residuals are due to statistical fluctuations of the wavelength solution. Work in progress.

sbailey:
It is possible for me to run this on Julien's directory without it trying to clobber something that might already be there?  I don't see a way to set/override the output directory.

If @profxj or @julienguy have already run this on the known bad sky subtraction in z-band in /global/cscratch1/sd/jguy/sim2017/redux/month, please post results.  This update looks fine, but it would be nice to confirm that it actually catches the types of sky subtraction problems that we know we have (or had, they have been improved...)

sbailey:
Please fix these before merging:
* The `--night` option doesn't work; it still tries to load all nights.  But the `--nights` option does work.  Is the `--night` option needed?
* The `--reduxdir` option doesn't work to override `$DESI_SPECTRO_REDUX/$SPECPROD`.
* Running it on my own prod confirmed that it tries to write output to `$(pwd)/QA` and that that directory must exist.  It would be helpful to provide an option to override that location and create the directory if needed.

Knowing that it was safe to run on Julien's prod with known problems, I confirmed that they do stand out in the plot (if you know that the plot should be 1 everywhere):

/global/cscratch1/sd/jguy/sim2017/redux/month (pixsim-based production with known wavelength solution problems):
skyline_z

/global/cscratch1/sd/jguy/sim2017/redux/month3 (pixsim-based production with improved wavelength solution but still bad residuals):
skyline_z

Specsim production with perfect wavelength solution: 
skyline_z

profxj:
Have addressed @sbailey 's issues and merged with 
parallel QL development.

Passing Travis tests, but I ask @sbailey to quickly try 
the code again.

sbailey:
`--out_dir` option was still not working (would crash), and even then it was even only trying to work for the skyline output and not for the other outputs.  Rather than dragging out the PR I went ahead and fixed this myself.  I also fixed the `__offline_qa_version__` usage throughout the code instead of just in the one place that this PR fixed it.  I think this is ready to merge after tests pass.

PR #459, opened 2017-11-08
Running a month of data2

julienguy:
Two improvements to the pipeline 
 - desispec/pipeline/task.py run desi_psf_fit with option --trace-prior-deg 4. This forces the Legendre coefficients of degree 4 and higher to be the same for all fibers of the same bundle to reduce statistical fluctuations in the wavelength solution. This requires an update of specex (changes already in master). This modification reduced sky residuals and improved significantly the redshift efficiency of redrock for ELGS.
 - computes a contribution to the variance of the sky model from flatfielding uncertainties (0.5%) and wavelength solution uncertainties (>=0.005 A). The sky model variance  is only modified on sky emission lines. The wavelength solution uncertainty is increased if needed to get a reduced chi2 on a given emission line <=1. This modification improved a bit the ELG redshift efficiency.


julienguy:
Another modification I forgot to mention is the use of more arc lamp lines (Kr) in the call to desi_fit_psf  after the change to the arc lamp template in desisim which includes Kr lines (the template is based on observations at WINLIGHT).

PR #463, opened 2017-11-18
Ql res

rkehoe:
This PR addresses QL's handling of resolution information in the frame object and file such that it solves the data size problem that hinders acceptable performance with 30 cameras via QLF.  With current frame file sizes ~150 Mb, this resulted in ~5Gb outputs for final frame files from QL pipeline execution.   The update solves this by use of a fast module, QuickResolution, to go from Wsigma's to the resolution matrix.  The calculation from wavelength fit coefficients to Wsigma's is performed in the frame object.  A speed improvement for these calculations was needed to provide acceptable performance in the context of QuickLook.  Specifically, this PR includes:

- implementation of wcoeff from psf file into frame object to support calculation of the resolution matrix each exposure
- fast algorithm to handle computation of the  resolution matrix from 80 s with existing calculations to 1 s
- removal of resolution data from output final frame file to reduce file size from ~150 Mb to ~17 Mb.  This corresponds to a reduction in the size of the resolution information from 105 Mb to 50 Kb.  The final size of 17 Mb corresponds to the size with resolution turned OFF in qlconfig, which has provided acceptable performance and has been the default since performance slowdowns were first observed with QLF on a full 30 camera array.
- ensure backward compatibility with existing mode of retaining resolution_data in the frame file.  Use of resolution_data has priority unless specified otherwise, and the code is compatible with offline running.

This modification follows discussion with @sbailey in which the need to find a smaller frame size was necessary for QL performance.  @sbailey, particularly given the work on the frame code, I think it would be best if you could have a look and indicate your comments.  I imagine some tweaks, at least, to our approach might come up, for instance the inclusion of icoeffs with wcoeffs in the frame extension.  Once these are finalized, we will document this QL frame data model.


gdhungana:
@samikama Please check and confirm. I did not make the config file default to use the functionality of all this. A couple of switches need to be adjusted. Potentially that is to be made on several config types and may be the scope of future updates.

sbailey:
I like the spirit of this, but I'm not thrilled about having the `Frame` constructor get 7 new arguments (albeit optional) to support this.  Please try to think of a simpler interface, and/or give me a bit of time to think on that (and ping me again if it falls off my radar, review prep is pretty distracting).

Please compare your `QuickResolution` implementation with `desispec.resolution.Resolution` when it is given a 1D array of sigmas as input (read the code; that option seems to have been missed in the docstring).  Could that have been used instead?  Also see `desisim.pixelsplines.gauss_blur_matrix` for another version of the same concept, though we don't want to have desispec depend upon desisim.

samikama:
Hi @sbailey ,
I agree that it is a long list of arguments. Only other solution I can think of right now would be putting all these into a data structure and passing this structure as an optional argument. Unfortunately, this is the only way to save about 2000x(105MB->50KB) on file representation. Alternatively we can save sigma array for every fiber in which case saving would be ~20x(105MB->5MB) but it would require only one extra parameter. 

About the comparison with desispec.resolution.Resolution object. In my tests, construction of diagonals from the 1D sigma array with Resoultion object took about 80s for an image. Since I didn't want to change other peoples code, I implemented another version in QuickResolution. Effective calculation is identical but QuickResolution is optimized to use c/c++ side of the numpy thus can calculate same diagonals for whole image in around 1s. 

Let us know what you prefer.

sbailey:
Let's go with saving a 1D sigma array per fiber, which is still a significant savings over the full resolution matrix data while being much simpler in the interface and also much more straightforward for end users than interpreting a bunch of coefficients.  For comparison, SDSS/BOSS kept the uncalibrated frames with some parameters as Legendre coefficients like this, but only a few experts became adept at interpreting those; the calibrated frames kept the full 1D arrays and were much more user friendly.

Interesting about the 80s vs. 1s for Resolution vs. QuickResolution.  Do they actually produce the same answer in the end?  If so, we should port this implementation to Resolution too.  In the short term OK to merge into QuickResolution so that we can compare them in more detail later, but please leave behind a comment in QuickResolution about the conceptual similarity to Resolution and why this is different.

rkehoe:
@sbailey  Thanks for the quick look at this.  I would really strongly prefer not to loose information by going to a coarse array per fiber.  To accommodate your concerns, could we instead encapsulate the added optional keywords in an optional input list or dict?  Or perhaps we could save the 10 arrays in addition to the coefficients -- the coefficients only take 50 Kb.  If nobody else uses them, at least QL can.  Other people using the frame outputs will then have the arrays.  The implementation we have yields the same results as the slower resolution calculation, which means we loose no information.  @samikama and I had discussed the issue of relevance to the offline calculations, and we can put the comments into the code where appropriate.   

You indicated it was okay to merge, but I think you at least want us to include the 10 arrays for non-QL users, correct?

sbailey:
I didn't mean to imply a coarse array, but rather a finely sampled array of the same length as the wavelength array, giving the sigma at that wavelength.  This makes the sigma array the same size as the flux array, and has a direct and easy correspondence to relating sigmas to wavlengths and fluxes.

I'm not understanding what the 10 is in "we could save the 10 arrays in addition to the coefficients".

rkehoe:
Wow, my eyes are literally going.  1D looked like 10 (as in '10 bin array' if you don't read carefully enough!) and then I misinterpreted from there, so ignore that particular comment.

Now I understand what you are suggesting.  This is actually not far from my original idea, to reduce the size of the resolution data, though your suggestion does not have fewer bin contents but the WSigmas themselves and so does avoid my concern.  But I still do prefer @samikama 's approach as it saves significant space (150 Mb per 30 camera exposure, and ~3 Gb per night), taking each frame from 17 Mb to ~22Mb.  Also, I find looking at the coefficients easier to interpret than the resolution data, but that may be my idiosyncracy.

This isn't to argue interminably.  I understand your preference and we are starting a derivative branch to implement the wsigma arrays approach in case the above comments don't seem sufficient to you.   I am not sure we will have time with thanksgiving and @samikama leaving to complete it promptly, but we will try.

rkehoe:
This is an obsolete PR and we have moved way past the issues raised here, in part by adopting a different data model now with the Qframe file.  We kept this PR open in case we wanted the alternative solution to the resolution information model contained here, but this is not necessary now.  I will close this PR without merging.

sbailey:
@rkehoe Sounds good.  Please also delete the QLRes branch from github after closing this.

rkehoe:
Closing this PR without merge.

PR #464, opened 2017-11-20
Fussing a bit with QA

profxj:
A little polishing of the docs.
Also had to allow for frames to be missing what I believe are
required keywords (e.g AIRMASS).

And I had to turn off the generation of the frame_skychi figure.
The method has simply vanished.  I worry it was part of branch
and got eaten on a merge.  Will buy anyone a beer that can
find it on GitHub...

Haven't tested on NERSC yet.  I imagine someone can do so
faster than I but I'll give it a shot tomorrow if not.

### DIFF
--- a/doc/qa.rst
+++ b/doc/qa.rst
@@ -56,7 +56,8 @@ Generate the QA YAML file and figures::
 desi_qa_exposure
 ++++++++++++++++
 
-Generates Exposure level QA.
+Generates Exposure level QA.   The current
+implementation is only for the flat flux.
 
 usage
 -----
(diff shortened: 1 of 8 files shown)


sbailey:
Merging now to get this into the nightly NERSC master update.  We definitely need to get the QA code into standard usage so that we don't require a weekend firedrill for non-experts to learn how to use it and experts to fix minor issues that we should have caught when they first arose.

AIRMASS: calibration arc and flat exposures don't have AIRMASS, but science exposures have that keyword in HDU0.  If you found a science exposure that is missing it, or feel that it should be in a different HDU (FIBERMAP HDU, perhaps?), holler.

frame_skychi: The call to it was added in PR #438 (Sky continuum in offline QA), but the matching code that should have come from `desispec.qa.qa_plots` was not included in that commit.  It's not a good sign for our procedures that we didn't notice until now.  `git grep "def frame_skychi" $(git rev-list --all)` can't find it in any revision (recipe from [stackoverflow](https://stackoverflow.com/questions/2928584/how-to-grep-search-committed-code-in-the-git-history); I'm not that good of a git ninja...), but I'm not sure if that can track code from branches that never made it into master.

Please open new tickets if we try this at NERSC and hit trouble.

PR #466, opened 2017-11-20
Running a month of data3

julienguy:
- option to reject cosmics in 1D spectra (if gradient sharper than PSF from resolution matrix) in fiberflat, sky model, and frame processing.
 - py/desispec/scripts/extract.py , change default option: do not decorrelate fibers (because this is biasing)

those changes allow a <5% catastrophic failure rate for all targets. tested on >1million targets.

### DIFF
--- a/py/desispec/cosmics.py
+++ b/py/desispec/cosmics.py
@@ -10,7 +10,89 @@ from desiutil.log import get_logger
 import numpy as np
 import math
 import copy
+import scipy.ndimage
 from desispec.maskbits import ccdmask
+from desispec.maskbits import specmask
+
+def reject_cosmic_rays_1d(frame,nsig=3,psferr=0.05) :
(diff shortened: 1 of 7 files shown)


julienguy:
- Tests where failing for a very good reason: the multiprocessing and mpi extractions were inconsistent
   because of a hard coded decorr=False in multiprocessing and =True in MPI
 - This is my fault, but it is a code design issue to have two independent main functions. I fixed this by adding the option at the argument parsing level.
 - This is really annoying because we have known for 2 years that decorrelating fibers was not a good idea as it introduces back the fiber cross-correlation bias. I thought I had done the change in the code, but unfortunately it was in the wrong main! (mpi version is used in the pipeline).
 - **I am rerunning extractions to see if this improves the chi2 on sky lines.**

julienguy:
Checked does not improve sky residuals. Merging.

PR #476, opened 2017-12-12
add option to join otehr tables such as truth.fits or target.fits

julienguy:
Option of redshift catalog to add data from the target table or truth file.
Example

```
desi_zcatalog -i spectro/redux/mini/spectra-64 -o spectro/redux/mini/zcat.fits --match targets/targets.fits  targets/truth.fits 
```

### DIFF
--- a/bin/desi_zcatalog
+++ b/bin/desi_zcatalog
@@ -18,37 +18,78 @@ from __future__ import absolute_import, division, print_function
 import sys, os
 import numpy as np
 from astropy.table import Table, vstack
-
+from desiutil.log import get_logger,DEBUG
 from desispec import io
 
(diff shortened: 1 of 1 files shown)


sbailey:
+1 for the idea of this.  Have you tested it on anything larger than the minitest?  I'm concerned that the explicit python loops won't scale very well for larger datasets.  astropy Tables have join operations built in, though come with their own overhead and issues.

sbailey:
We'll likely need to revisit the efficiency of the matching algorithm for larger datasets, but this doesn't impact the prior functionality and does work for smaller datasets, so I'll go ahead and merge now so that it can be included in upcoming tags.

PR #480, opened 2017-12-19
Ql update qas

rkehoe:
This PR implements a requested modification by @rstaten to the SNR QA which calculates the residual of each target with respect to the fit of SNR vs. imaging mag.  These residuals are provided now in the output list of QA metrics.  A plot of residual in RA, Dec is also provided to indicate any gradient with position in the focal plane.  

Results corresponding to the current implementation can be found at:

      https://desi.lbl.gov/trac/wiki/Pipeline/QuickLook/Results

This update is intended to be the last before we request a desispec tag to freeze versions for the next round of QLF development.  Once outstanding questions are addressed, I will merge this PR.


sbailey:
I'm sure we'll want to continue to refine these plots, but this is a good first step so I'll merge now so that it can be included in the impending set of tags.

PR #489, opened 2018-01-17
QA Edits for mini Test

profxj:
As titled.

Added a test for 'bad' files in the redux directory.


profxj:
Failing Travis but passes all tests when running 
python setup.py test
at NERSC

sbailey:
I think the travis test problems are the same as what Adam reported yesterday; the fix requires an update to desiutil 1.9.9 (which I have done in .travis.yaml; tests are rerunning now).  OK to merge after tests pass.

profxj:
Need to fuss a bit further (failing mini Notebook test).
Will reopen.

PR #505, opened 2018-02-12
new desi_group_spectra

sbailey:
Opening this PR for review while I do the final cleanup (e.g. adding unit tests).

This PR is a complete rewrite of `desi_group_spectra` for several reasons:
  * propagates SCORES tables from cframe files into spectra files
  * 40x faster by using fitsio, optimized data appending, and smarter "do I really need to do this frame?" logic
  * isolates "which spectra files need to be updated" logic into a single function, which could be replaced in a future update with a spectro-processing DB query

Known issues I want to fix before merging:
  * no unit tests (I believe that was true of the previous version too)
  * `--nights` option became `--night` (easy to fix)

Known issues that may wait for a further update.
  * appending a new night that is completely missing one spectrograph channel for
  every frame, when the prior data did have that spectrograph channel.

In the name of optimization, this bypasses the standard `Frame` and `Spectra` objects and `deispec.io.read/write` routines.  The `FrameLite` and `SpectraLite` objects here might be useful for other purposes (like QA) but are currently isolated into `pixgroup.py` and are not intended to be generally useful objects with the full features of `Frame` and `Spectra`.  They are optimized for fast I/O and regrouping, but need unit tests to help ensure that they remain consistent with the standard more full-featured objects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ env:
         # These packages will always be installed.
         - CONDA_DEPENDENCIES="qt=4"
         # These packages will only be installed if we really need them.
-        - CONDA_ALL_DEPENDENCIES="scipy matplotlib==2.0 sqlalchemy coverage==3.7.1 pyyaml healpy numba numpy"
+        - CONDA_ALL_DEPENDENCIES="scipy matplotlib==2.0 sqlalchemy coverage==3.7.1 pyyaml healpy numba numpy fitsio"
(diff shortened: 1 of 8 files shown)


sbailey:
I changed the option back to `--nights` and added unit tests (and satisfyingly fixed a few bugs that were found...).  Thanks for the check @julienguy.  I'm ready to merge when travis tests pass.

Issue #368, opened 2017-04-03
Implement radial variation of sky brightness


julienguy:
Fiber flat-fielding will have a residual radial pattern (see https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=2761 ) that has to be accounted for during the sky model fit.

julienguy:
There are actually three different topic here
 1) calibration of the dome screen non-uniformity w.r.t. sky. this will be done during commissioning using dome flats and  blank sky exposures, and saved as a new fiberflat calibration file (TBD). We expect this to be stable, because it's a geometric effect (but we'll check during commissioning).
 2) sky non-uniformities which have no reason to follow a fov radial pattern. this has to be implemented as a polynomial of fov coordinates when computing the sky model.
 3) throughput variations due to FRD variation with fiber flexure that cannot be calibrated out, and that we have to absorb in sky subtraction. we should prepare the code for this eventuality as an option in sky subtraction (coefficient correction per fiber).

julienguy:
The point 3. above has been implemented in branch skysub.
 - When subtracting the sky spectrum, an achromatic throughput correction can be fit.
 - This is based on 2 parameter fit around each sky line; 1 normalization factor for the sky line + 1 constant to account for the target spectrum.
 - The width of the wavelength region around each sky line has been tuned on simulations and found to be +- 4A.
 - The normalization coefficients of all sky lines are optimally combined, there is an outlier rejection loop per fiber to discard lines potentially affected by cosmics or target emission line.

A test run on one exposure/spectrograph (dark program, with LRG,ELG,STD stars...) gives encouraging results a with precision better than 0.1% rms on the (sky) throughput correction. Worse case is a 2% error. There is no significant bias with target brightness. The figure below shows the coefficient for each fiber (b,r and z from top to bottom).

![corr](https://user-images.githubusercontent.com/5192160/35589892-11cd8fba-05ba-11e8-9495-a0a415b099a9.png)

sbailey:
Please spot check the ~2% outliers to see why the outlier rejection of the normalization coefficients didn't work as well as we might expect.  Would a simple median of the normalization coefficients do better?  What causes enough of the normalization coefficients to be wrong to result in a 2% overall incorrect answer?

julienguy:
[...]

PR #506, opened 2018-02-14
Non uniform sky

julienguy:
Implemented several routines to fit a non-uniform sky.
See issue #368 for details.

### DIFF
--- a/py/desispec/scripts/sky.py
+++ b/py/desispec/scripts/sky.py
@@ -32,7 +32,9 @@ def parse(options=None):
                         help = 'n sigma rejection for cosmics in 1D (default, no rejection)')
     parser.add_argument('--no-extra-variance', action='store_true',
                         help = 'do not increase sky model variance based on chi2 on sky lines')
-     
+    parser.add_argument('--fp-corr-deg', type = int, default = 0, required = False,
+                        help = 'Focal plane variation degree')
+    
     args = None
     if options is None:
         args = parser.parse_args()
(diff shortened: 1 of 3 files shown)


sbailey:
Tested using the 17.12 reference run data that does not have sky variation over the focal plane.  Comparing to the "truth" with no angular / chromatic variations that takes 5 seconds to run:
* `sky.compute_sky(frame, angular_variation_deg=2)` takes 1 minute is consistent to 0.1%
* `sky.compute_sky(frame, angular_variation_deg=2, chromatic_variation_deg=2)` also takes 1 minutes and is consistent to 0.2%
* `sky.compute_sky(frame, angular_variation_deg=2, chromatic_variation_deg=-1)` (i.e. each wavelength is separately solved) takes 15 minutes and is only consistent to 10%.  Let's not use this one... :)

We could go ahead and merge this as work in the right direction; thanks for keeping the various options separated into different functions for comparison.  For real use though we may need to simultaneously solve all 10 petals with consistency constraints along the boundaries.  Of course the different spectrographs themselves will have differences in the uncalibrated sky due to throughput variations, but I'm wondering if the higher order terms should be constrained to be continuous.  e.g. here are the 10 different independent sky solutions with angular_variation_deg=2, shown relative to the solution without any extra degrees of freedom:

skypoly

sbailey:
Merging now in the spirit of directionally correct progress that doesn't break backwards compatibility / performance.  We can continue to update this with new branches and PRs even if this non-uniform sky isn't ready to be default yet.

PR #507, opened 2018-02-15
fix --reduxdir; fix fitsio compressed output

sbailey:
This PR fixes two bugs with spectral regrouping that were introduced in PR #505:
* `--reduxdir` option was broken
* output spectra files didn't work with redrock because uncompressed uint32 HDUs can't be read with memmapping due to the BZERO / BSCALE transformations.

The updated implementation here also works around a fitsio bug esheldon/fitsio#150.

My penance for introducing the bugs in PR #505 was to write additional unit tests in this PR that would have caught them.


sbailey:
Merging this now to get the fix in place for the nightly integration test (which originally caught the problem).  Retroactive review is still welcome.

PR #508, opened 2018-02-15
Fix logger redirection

tskisner:
When redirecting stdout to a file, loop over all registered loggers and redirect their outputs.  When exiting the context manager, verify that no new loggers have been added during the redirection.

This fixes https://github.com/desihub/desiutil/issues/98 as long as user code does not create new loggers while being redirected.  The redirection code will raise an exception if this happens.  A second-order fix to this would be to pre-create all possible loggers.

### DIFF
--- a/py/desispec/parallel.py
+++ b/py/desispec/parallel.py
@@ -18,6 +18,7 @@ import warnings
 
 import numpy as np
 
+import desiutil.log
 from desiutil.log import get_logger
 
 
@@ -271,13 +272,22 @@ def stdouterr_redirected(to=None, comm=None):
         to (str): The output file name.
         comm (mpi4py.MPI.Comm): The optional MPI communicator.
     """
+    nproc = 1
+    rank = 0
+    if comm is not None:
+        nproc = comm.size
+        rank = comm.rank
 
     # The currently active POSIX file descriptors
     fd_out = sys.stdout.fileno()
     fd_err = sys.stderr.fileno()
 
-    # The DESI logger
-    log = get_logger()
+    # The DESI loggers.  This is the list of registered loggers.  We are going
+    # to redirect all of them and then restore them.  We will also check that
+    # the list of loggers has not changed after we return from running the
+    # decorated code.
+
+    desi_loggers = list(sorted(desiutil.log._desiutil_log_root.keys()))
 
     def _redirect(out_to, err_to):
 
@@ -310,21 +320,27 @@ def stdouterr_redirected(to=None, comm=None):
             sys.stderr = io.TextIOWrapper(os.fdopen(fd_err, 'wb'))
 
         # update DESI logging to use new stdout
-        while len(log.handlers) > 0:
-            h = log.handlers[0]
-            log.removeHandler(h)
-        # Add the current stdout.
-        ch = logging.StreamHandler(sys.stdout)
-        formatter = logging.Formatter("%(levelname)s:%(filename)s:%(lineno)s:%(funcName)s: %(message)s")
-        ch.setFormatter(formatter)
-        log.addHandler(ch)
+        for loggername in desi_loggers:
+            logger = desiutil.log._desiutil_log_root[loggername]
+            hformat = None
+            while len(logger.handlers) > 0:
+                h = logger.handlers[0]
+                if hformat is None:
+                    hformat = h.formatter._fmt
+                logger.removeHandler(h)
+            # Add the current stdout.
+            ch = logging.StreamHandler(sys.stdout)
+            formatter = logging.Formatter(hformat, datefmt='%Y-%m-%dT%H:%M:%S')
+            ch.setFormatter(formatter)
+            logger.addHandler(ch)
 
     # redirect both stdout and stderr to the same file
 
     if to is None:
         to = "/dev/null"
 
-    if (comm is None) or (comm.rank == 0):
+    if rank == 0:
+        log = get_logger()
         log.debug("Begin log redirection to {} at {}".format(to, time.asctime()))
 
     # Save the original file descriptors so we can restore them later
@@ -333,9 +349,8 @@ def stdouterr_redirected(to=None, comm=None):
 
     try:
         pto = to
-        if comm is not None:
-            if to != "/dev/null":
-                pto = "{}_{}".format(to, comm.rank)
+        if to != "/dev/null":
+            pto = "{}_{}".format(to, rank)
 
         # open python file, which creates low-level POSIX file
         # descriptor.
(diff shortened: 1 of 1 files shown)


julienguy:
Checked that it's working in prod.

sbailey:
This seems fragile because
1. it depends upon a private variable in a different repo (`desiutil.log._desiutil_log_root`)
2. the requirement that no new loggers can be added during redirection, without any knowledge or guarantee that that would be the case.  And if it isn't, it raises an exception rather than gracefully un-redirecting the loggers that it knew about at the start while leaving the others alone.

Previously I thought that `_desiutil_log_root` was keeping the various singleton loggers for each level, but looking more closely it is only keeping track of different combinations and timestamps and delimeters, not different loglevels.  So I'm not sure if pre-caching the different loglevels would work here.

weaverba137:
In fact desiutil.log keeps track of all possible combinations of log level, delimiter & timestamp. Please re-read the code if you don’t believe me, & point out specific lines that contradict me.

sbailey:
`_desiutil_log_root` is updated in `_configure_root_logger(timestamp=False, delimiter=':')` which doesn't know about the requested loglevel.  `get_logger` only does stuff with the level after calling `_configure_root_logger`.  Empirically requesting a different log level doesn't increase the number of keys but requesting a different delimiter does:
```
In [1]: import desiutil.log

In [2]: desiutil.log._desiutil_log_root.keys()
Out[2]: dict_keys(['desiutil.log.dlm58'])

In [3]: log_debug = desiutil.log.get_logger('DEBUG')

In [4]: log_info = desiutil.log.get_logger('INFO')

In [5]: log_error = desiutil.log.get_logger('ERROR')

In [6]: desiutil.log._desiutil_log_root.keys()
Out[6]: dict_keys(['desiutil.log.dlm58'])

In [7]: desiutil.__version__
Out[7]: '1.9.9.dev596'

In [8]: log_blat = desiutil.log.get_logger('INFO', delimiter='/')

In [9]: desiutil.log._desiutil_log_root.keys()
Out[9]: dict_keys(['desiutil.log.dlm47', 'desiutil.log.dlm58'])
```

weaverba137:
That cannot be correct. If desiutil.log did not keep track of all possible combinations, then it wouldn’t even pass the unit tests, let alone do what it is supposed to.

Please please believe me here. Please take the time to actually read the code. Please take into account the work that I have done and the care that I have taken.

sbailey:
Please take a look yourself.  I read the code and that changed my understanding of what was kept in `_desiutil_log_root`, and then I empirically verified that with test code that I put in the previous comment.

Note: the logger code does work as advertised because each different loglevel comes in as `log = logging.getLogger(root_name + '.' + _level_children[gl])`, so you do get a different singleton per loglevel.  It just that those singletons aren't kept in `_desi_log_root` because that only tracks the `root_name` part, not the `+ '.' + _level_children[gl]` part.

tskisner:
Ok, well I agree that touching a package-wide variable that starts with an underscore (indicating that it is an internal variable) is generally a bad idea (especially doing that from an external package!)  

However, it is true that regardless of how that dictionary is populated, it contains all the logger instances that are currently in use.  So iterating over them and redirecting them and restoring them does currently work.

Line 154 reads:
```
root_name = "desiutil.log.dlm" + ''.join(map(str, map(ord, delimiter)))
```
So the root_name (i.e. the key to the dictionary) is made of a constant string plus a concatenation of the string representations of the ordinal values of each character in the delimiter.

For example, if delimiter = "abc", then the root_name would be:
```
root_name =  "desiutil.log.dlm" + "97" + "98" + "99" ->  "desiutil.log.dlm979899"
```
Plus an optional timestamp (effectively guaranteeing that a new logger would be added to the dictionary).

I don't have any opinion about these names, I just want to be able to iterate over them and redirect them.  Maybe if we made the dictionary a publicly documented package member?

sbailey:
`_desiutil_log_root` caches the *root* loggers (thus the name...) keyed by the values of `root_name`.  It does *not* cache the actual singleton loggers themselves, which are obtained via
```
log = logging.getLogger(root_name + '.' + _level_children[gl])
```
It appears that redirecting stdout/stderr of the root logger will also redirect stdout/stderr of anything with the name `rootname+'.'+something`, whether those are created before or after the redirection started.  That may have been obvious to @tskisner and @weaverba137, but it wasn't obvious to me, thus perhaps leading to the confusion.

I tested it with a simple script (below) and indeed the child loggers are redirected whether or not they were first created before or after the root logger was redirected.

However, even in the case of a new logger with a different delimiter being created, the redirection remained in place and the messages went where intended (and then the redirector raises an exception because of the new root logger).  It's unclear to me whether raising an exception if a new logger shows up is actually needed.
```python
#- Testing MPI log redirection
from desiutil.log import get_logger
from desispec.parallel import stdouterr_redirected
from mpi4py import MPI
import time
import os

def dashdash_function(a):
    log = get_logger('DEBUG', delimiter='--')
    log.debug(a)
    log.info(a)

def debug_function(a):
    log = get_logger('DEBUG')
    log.debug(a)
    log.info(a)
    dashdash_function(a)

def info_function(a):
    log = get_logger('INFO')
    log.debug(a)
    log.info(a)
    debug_function(a)

comm = MPI.COMM_WORLD
rank = comm.rank
size = comm.size

outfile = 'blat.log'
with stdouterr_redirected(to=outfile, comm=comm):
    log = get_logger('INFO')
    message = 'Rank {}'.format(rank)
    info_function(message)

print('Rank {} all done'.format(rank))
```
Has individual log files like
```
INFO:mpilog.py:22:info_function: Rank 0
DEBUG:mpilog.py:15:debug_function: Rank 0
INFO:mpilog.py:16:debug_function: Rank 0
DEBUG--mpilog.py--10--dashdash_function-- Rank 0
INFO--mpilog.py--11--dashdash_function-- Rank 0
  File "mpilog.py", line 33, in <module>
    info_function(message)
  File "/global/common/cori/contrib/desi/desiconda/20170613-1.1.4-spectro/code/desiconda/20170613-1.1.4-spectro_conda/lib/python3.5/contextlib.py", line 66, in __exit__
    next(self.gen)
  File "/global/cscratch1/sd/sjbailey/desi/code/desispec/py/desispec/parallel.py", line 375, in stdouterr_redirected
    raise RuntimeError("Some desi loggers were added during "
RuntimeError: Some desi loggers were added during redirection
```
Removing the check on the exit_loggers seemed to work just fine.

I'll touch base with Ted in person about this in the morning and we can merge this one way or another.  As he pointed out this afternoon, even if this is a fragile solution, at least it is taking us from a completely non-working state (current master) into something that does work for the current cases where we use it.

tskisner:
I'm fine with that proposal (remove the check on exit).  I'll update this PR.

weaverba137:
I was not able to actually look at the code yesterday and it was difficult to communicate.  But yes, the *root* loggers are cached, and that has the *effect* of covering all possible combinations, even if it is difficult to see that explicitly.

sbailey:
I confirmed that the lastest update works as desired: even if new loggers are created during the redirection, they also go to the redirected location (presumably because C stdout was pointing there and they grabbed that) and they get reset back to the "normal" terminal stdout upon completion.  Thanks for fleshing this out.  Merging now.

PR #517, opened 2018-03-05
Ql drilldown metrics update

gdhungana:
This PR 
- updates the keys for the QL QA metrics as defined in the new model
- removes the amp level granularity of the 1D level QAs
Note: Few of the new keys are placeholders that preserves the data structure but need to be calculated correctly. That will be addressed in subsequent update. At the moment this will facilitate for the QLF implementation. @rkehoe @Srheft Please review and comment if you see any inconsistencies or other issues.

### DIFF
--- a/py/desispec/qa/qa_plots_ql.py
+++ b/py/desispec/qa/qa_plots_ql.py
@@ -52,12 +52,9 @@ def plot_countspectralbins(qa_dict,outfile):
     plt.suptitle("Count spectral bins after {}, Camera: {}, ExpID: {}".format(paname,camera,expid),fontsize=10,y=0.99)
 
     gs=GridSpec(7,6)
-    ax1=fig.add_subplot(gs[1:4,:2])
-    ax2=fig.add_subplot(gs[1:4,2:4])
-    ax3=fig.add_subplot(gs[1:4,4:])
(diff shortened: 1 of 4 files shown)


Srheft:
@gdhungana 

Change the FIBER_AMG back to INTEG so that the plot doesn't show empty.

gdhungana:
@Srheft Comments taken into account. Let me know if there are further issues.

Srheft:
@gdhungana  
Just to confirm:

This [first attachement] is an integ plot that was being created before [i.e., using master] :
image

and this [second attachement] is the integ plot that this branch [QL_drills] creates:

image

rkehoe:
Thanks @gdhungana.  This has some useful cleanup, and really help regularize things and ease usage downstream.

Thanks also @srheft for catching the missing fibermag information.  This isn't calculated correctly, but I'd rather preserve the start of the final calculation than disable or delete it.  So I'm glad we have something there for the moment.  Once we have a fast flux calibration in place, we can complete the fiber magnitude/imaging magnitude comparison.

I think there are some second order naming effects we should consider, but that can come as part of another PR.  This one is ready to merge, which I will do now.

weaverba137:
Is there any reason not to delete this branch?

Srheft:
It is merged to master so I don't see any reason not to.

PR #519, opened 2018-03-07
Astropy 2 compatibility

### LINKED ISSUE #486: update desispec to astropy 2.x
Update desispec to be astropy 2.x compatible.  Maintain astropy 1.3.x compatibility if possible.

weaverba137:
This PR fixes #486.  It is not yet ready to merge, but I want to track changes and discussion in a PR.

- [x] `desispec.test.test_io.TestIO.test_fibermap_rw` is the one and only one test I have found that is broken with Astropy 2.  It is broken due to a known issue with round-tripping `astropy.table.Table` to FITS, astropy/astropy#5267.
- [x] Fix broken download test.  The download test is trying to download a file that exists, but is currently inaccessible.  Travis tests skip this test, so we haven't noticed it.
- [x] Documentation builds on Travis are failing, but in theory, they should succeed on ReadTheDocs.  I have been able to build the documentation in a minimalist environment on my laptop.


weaverba137:
Thank you to @julienguy for making necessary permission changes to permit the download test to succeed.

weaverba137:
`desispec.test.test_io` is written in such a way that it relies on certain tests to create temporary directories instead of creating those directories in the `setUp` or `setUpClass` methods.  This means the success or failure of certain tests could depend on the order in which tests are run, or on whether certain tests are skipped.  It's going to take a while to disentangle this.

It also looks like only *some* of the various functions in `desispec.io` call `desispec.io.util.makepath`.

sbailey:
I'm confused about `desispec.test.test_io.test_fibermap_rw`.  The problem in astropy/astropy#5267 isn't new to astropy 2.  Why wasn't this test failing on py3 + astropy 1.3.x?  Ditto for documentation building.  That was working on Travis previously.

weaverba137:
And the documentation builds were taking place in an environment significantly different from ReadTheDocs, so in some sense, that test was too easy.

sbailey:
It could be nice for the bootcalib test files to be placed somewhere nicer than the top level directory, but please *don't* delete them when done — it is convenient to have them around from a previous run without having to re-download every time the tests are run (especially if one is on an airplane or at a hotel with a slow wifi connection).

weaverba137:
OK, I'll revert that and make a note of why they aren't cleaned up.

weaverba137:
I think this is ready for final review now.  For the record:

* I restored the full build environment for documentation builds.  I don't want to go into gory technical details, but astropy imports do a lot of introspection under the hood, including searching for the presence of other packages (*e.g.* scipy).  In documentation builds, where some of those packages may be mocked-up, that can cause exceptions.
* I don't know the astropy Table round-trip problem in fibermap got fixed, but I noticed that `desispec.test.test_io` in generally started behaving better once I fixed an unclosed file error in `desispec.io.frame`.

Issue #530, opened 2018-03-15
pipeline DB: improve desi_pipe sync


julienguy:
desi_pipe sync should run postprocessing() on tasks that are done.

julienguy:
solved in PR #543

PR #543, opened 2018-03-19
DB sync

julienguy:
Improved sync. DB based a prod on disk.
This is useful for simulation runs where frames or cframes are simulated directly.
Tested on an instance of the mini test.

### DIFF
--- a/py/desispec/pipeline/db.py
+++ b/py/desispec/pipeline/db.py
@@ -73,29 +73,39 @@ def all_tasks(night, nside):
 
     for ex in sorted(expid):
 
-        log.debug("read fibermap for exposure {}".format(ex))
-
+        
         # get the fibermap for this exposure
         fibermap = io.get_raw_files("fibermap", night, ex)
+        
+        log.info("read {}".format(fibermap))
 
         #fmdata = io.read_fibermap(fibermap)
         #flavor = fmdata.meta["FLAVOR"]
 
         fmdata,header = fitsio.read(fibermap,header=True)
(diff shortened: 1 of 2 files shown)


julienguy:
And this solves issue #530

tskisner:
Hell @julienguy , I thought by definition the "sync" command only updated the DB task states based on the filesystem (as opposed to the "update" command, which actually adds tasks to the DB).  What is the intention of the "force" option?

tskisner:
In other words, if the production does not exist, then running "sync" is an error.

tskisner:
Alternatively- running "sync" on a production with zero tasks is basically a non-operation (does nothing), since there are no tasks to check the state of.

julienguy:
The --force is to force the db creation on an existing redux prod. This is needed for simulations where frames or cframes are generated by the sims. If you think of a better approach to this don't hesistate to modify this branch.
As an example I have done,
```
desi_pipe create --data /global/cscratch1/sd/jguy/sim2017/mini7/spectro/sim/mini --redux /global/cscratch1/sd/jguy/sim2017/mini7/spectro/redux --prod mini --db-sqlite

source /global/cscratch1/sd/jguy/sim2017/mini7/spectro/redux/mini/setup.sh

desi_pipe sync

```

julienguy:
I forgot the --force in the desi_pipe create command.

tskisner:
I talked to @julienguy offline, and I had misunderstood what was going on.  This PR looks fine to me- sorry for the noise.

Issue #525, opened 2018-03-15
pipeline DB: desi_pipe chain


julienguy:
Chain tasks and launch jobs with dependencies

tskisner:
Implemented as of 8f89f161d03f36de1d459001fcbe51cbbd39b88d

PR #547, opened 2018-03-20
Initial implementation of nightly script

### LINKED ISSUE #304: High level scripts for nightly processing
This issue cannot be addressed until the merging of #300.  This issue tracks the "pipeline side" of the requirements for WBS 1863_80, "Integrate data transfer --> pipeline".  Basically we have almost all the pieces in place.  I propose 3 new entry points in desispec.scripts:

desi_night_start (run after calibration data exists):
 - Inputs are the night, and and properly set raw / production environment variables.
 - Calls create_prod to update the production location and add a plan for data up through fiberflats.
 - Submit the generated slurm scripts to process calibration data.

One natural output of these jobs (which use the desi_pipe_run executable) is a state file stored in yaml.  Eventually that will be a database, but that can be tracked in a separate issue.

desi_night_update (run whenever there are new data):
 - Inputs are the night, and properly set raw / production environment variables.  
 - NEW CODE: add option to scripts.pipe_prod.py that tells it to generate per-cframe job scripts.  
 - Call this new code to update plans for the specified night and generate per-cframe slurm scripts.
 - NEW CODE: add option to pipeline.run_steps that specifies a single target object. This code 
      will then prune the graph before running the steps leading up to that object.
 - NEW CODE: add option to pipeline.graph_slice that recursively extracts all dependencies, not only
      direct ancestors.
 - Much like pipe_status.py, load the state of the graph from the newest state file.
 - Submit all per-cframe slurm scripts whose output does not exist AND which are not already
    in-progress from another job.

desi_night_end (run once we have all the data):
 - Inputs are the night, and properly set raw / production environment variables.
 - Load the state and verify that all cframes have been generated.
 - Run just the makebricks and zfind slurm scripts.

tskisner:
Do not merge yet, this needs more testing.  Just opening a PR to track this.  Once merged, this will close #304 and #525 .

### DIFF
--- a/py/desispec/pipeline/db.py
+++ b/py/desispec/pipeline/db.py
@@ -387,7 +387,7 @@ class DataBase:
             cur.execute( 'select name, state from {}'.format(tasktype))
             for name, intstate in cur.fetchall():
                 state_count[task_int_to_state[intstate]] += 1
-        
+
         return state_count
 
     def get_states(self, tasks):
@@ -539,8 +539,6 @@ class DataBase:
             night (str): The night to scan for updates.
 
         """
-
-
         tasks_in_db = None
         # Grab existing nightly tasks
         with self.cursor() as cur:
@@ -568,8 +566,8 @@ class DataBase:
     def cleanup(self, cleanfailed=False):
         """Reset states of tasks.
 
-        Any tasks that are marked as "running" will have their state
-        reset to "ready".  This can be called if a job dies before
+        Any tasks that are marked as "running" or "queued" will have their
+        state reset to "ready".  This can be called if a job dies before
         completing all tasks.
 
         """
@@ -580,13 +578,15 @@ class DataBase:
             for tt in task_types():
                 if cleanfailed:
                     cur.execute(\
-                        "select name from {} where state = {} or state = {}"\
+                        "select name from {} where state = {} or state = {} or state = {}"\
                         .format(tt, task_state_to_int["running"],
-                        task_state_to_int["fail"]))
+                        task_state_to_int["queued"],
+                        task_state_to_int["failed"]))
                 else:
                     cur.execute(\
-                        "select name from {} where state = {}"\
-                        .format(tt, task_state_to_int["running"]))
+                        "select name from {} where state = {} or state = {}"\
+                        .format(tt, task_state_to_int["running"],
+                        task_state_to_int["queued"]))
                 tasks_running[tt] = [ x for (x, ) in cur.fetchall() ]
 
         # For each task type, check status WITHOUT the DB, then set state.
--- a/py/desispec/pipeline/db.py
+++ b/py/desispec/pipeline/db.py
@@ -482,6 +482,104 @@ class DataBase:
         return
 
 
+    def get_submitted(self, tasks):
+        """Return the submitted flag for the list of tasks.
+
+        Args:
+            tasks (list): list of task names.
+
+        Returns:
+            (dict): the boolean submitted state of each task (True means that
+                the task has been submitted).
+
+        """
+        from .tasks.base import task_type
+        # First find the type of each task.
+        ttypes = dict()
+        for tsk in tasks:
+            tt = task_type(tsk)
+            if (tt == "spectra") or (tt == "redshift"):
+                raise RuntimeError("spectra and redshift tasks do not have submitted flag.")
+            ttypes[tsk] = tt
+
+        # Sort tasks into types
+        taskbytype = dict()
+        for t in task_types():
+            taskbytype[t] = list()
+        for tsk in tasks:
+            taskbytype[ttypes[tsk]].append(tsk)
+
+        # Process each type
+
+        submitted = dict()
+        for t, tlist in taskbytype.items():
+            if len(tlist) > 0:
+                namelist = ",".join([ "'{}'".format(x) for x in tlist ])
+                with self.cursor() as cur:
+                    cur.execute(\
+                        'select name, submitted from {} where name in ({})'.format(t, namelist))
+                    sb = cur.fetchall()
+                    submitted.update({ x[0] : x[1] for x in sb })
+        return submitted
+
+
+    def set_submitted_type(self, tasktype, tasks, unset=False):
+        """Flag a list of tasks of a single type as submitted.
+
+        Args:
+            tasktype (str): the type of these tasks.
+            tasks (list): list of task names.
+            unset (bool): if True, invert the behavior and unset the submitted
+                flag for these tasks.
+
+        Returns:
+            Nothing.
+
+        """
+        val = 1
+        if unset:
+            val = 0
+        with self.cursor() as cur:
+            for tsk in tasks:
+                cur.execute("update {} set submitted = {} where name = '{}'".format(tasktype, val, tsk))
+        return
+
+
+    def set_submitted(self, tasks, unset=False):
+        """Flag a list of tasks as submitted.
+
+        Args:
+            tasks (list): list of task names.
+            unset (bool): if True, invert the behavior and unset the submitted
+                flag for these tasks.
+
+        Returns:
+            Nothing.
+
+        """
+        from .tasks.base import task_type
+        # First find the type of each task.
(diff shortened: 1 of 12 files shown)


tskisner:
Ah yes, multiple problems in this work.  Will address this later today.

tskisner:
This is now ready for review.  Closes #525 and (finally) closes #304.  There will be further refinements to the nightly scripts- like having desi_update_night select only a single exposure for update.  However, I believe those refinements should be opened as new tickets.

tskisner:
@julienguy, the default value of --nersc_queue is "regular" which can submit 13 jobs fine.  For debug use, there is no choice but to run subsets of the pipeline steps (which is often fine for debugging).  Do want to just print an error message if a user tries to do this?

julienguy:
yes, ok. testing several nights in regular queue as well. 
It would also be convenient to add a job dependency option ```desi_pipe chain --after JOBID``` that would be used when submitting the first job of the chain.

tskisner:
Some notes from offline conversations with @julienguy and @sbailey:

- desi_pipe chain should have an option for starting dependency
- nightly script should call update at beginning
- nightly script should get NERSC_HOST from environment
- print warning if user tries to chain too many jobs
- fail --> failed
- add "queued" state
- desi_pipe change_state takes a list of tasks and sets state.  Used for setting state to "queued" in nightly scripts

tskisner:
I have implemented and am testing all changes above, except for using NERSC_HOST anywhere.  I also added a ```--once``` option to desi_pipe top for @sbailey :-)  Will update this branch soon.

tskisner:
Ok, those changes implemented.  Since the list of possible task states has changed, **ALL PRODUCTION DATABASES MUST BE RECREATED**.  You should be able to do this with "desi_pipe create --force".

sbailey:
> Currently the nightly scripts run a set of bash scripts unless the --nersc option is specified. In that case it uses the specified NERSC machine (edison | cori | coriknl). This is the most straightforward solution. If we want to use $NERSC_HOST to guess this information, that means that the default on NERSC machines would be to use slurm scripts, and we would need a commandline option to override that to use bash scripts. Also, $NERSC_HOST is the same on cori and coriknl, so we would still need some extra option to disambiguate those two.
> 
> @sbailey, how strongly do you want this feature?

Given that the dominant (only?) purpose of the nightly scripts is for batch processing data at NERSC, I'd really like the default simple options to automatically support that, and have "run everything as bash scripts spawned from this shell" be a special case option.  I propose:
```
if args.nersc is not None:
    # use args.nersc to pick machine and submit batch jobs like current usage
else:
    if os.getenv('NERSC_HOST') == 'edison':
        # submit edison batch jobs
    elif os.getenv('NERSC_HOST') == 'cori':
        # submit cori haswell (not knl) batch jobs
    else:
        # error until NERSC 9 exists
```
i.e. `--nersc coriknl` can be used to force submission to KNL, but otherwise it will just default to submitting batch jobs for the NERSC host that you are on.

sbailey:
Independently of the `--nersc` option, please don't merge this until one of us has cleanly verified that the following works as expected:
1. `desi_pipe create ...` for a blank production
2. Add some arc and flat exposures to `$DESI_SPECTRO_DATA`
3. `desi_start_night ...`
4. Add some additional science exposures to `$DESI_SPECTRO_DATA`
5. `desi_update_night ...`
6. `desi_end_night ...`

My first attempt at doing that yesterday worked through `desi_start_night`, but appeared to not run some steps for `desi_update_night`.  It might have been user error or something like a missing `desi_pipe getready`, but let's try this again with the latest version to verify the whole chain.

tskisner:
Ok, working on implementing the above, and I also realized that for the debug queue, we now have support for running one chain, getting the final job ID, and then later running a new chain with that starting job ID dependency.

However, the nightly scripts are designed to run a fixed chain of jobs that is longer than the debug limit.  So they will not run in the debug queue without:

1.  an option to override the list of tasks with something smaller

OR

2.  support in the pipeline for packing multiple desi_pipe_exec commands into a single job.

While not super complex, those changes would require more testing that we can't finish for the immediate 18.3 release.  I suggest we open a ticket for that packing feature and punt it to 18.4.

sbailey:
OK to punt those to 18.4; yesterday running in the edison regular queue worked ok for testing.

tskisner:
ok @sbailey, nightly script uses NERSC_HOST by default, but can be overridden with --nersc.  Also, a special override of ```--nersc shell``` will generate normal bash scripts.  I think that feature might still be useful if (for example) we are doing nightly processing on another MPI cluster during a NERSC outage.

Also desi_pipe cleanup now resets both queued and running tasks back to "waiting" and then calls getready() to check which tasks should have state moved to "ready".

julienguy:
bug in option --depjobs  (and I don't see what's wrong in the code) 
ex: ```desi_pipe chain --night 20191001 --nersc edison --nersc_queue debug  --depjobs 8758210```
```subprocess.CalledProcessError: Command 'sbatch -d afterany:8:7:5:8:2:1:0 /global/cscratch1/sd/jguy/sim2017/redux/month9/run/scripts/fiberflat_20180323-10:55:48/edison.slurm' returned non-zero exit status 1.```

**Comment by julienguy** [comment]
I pushed a fix of the --depjobs options, add default task list and some messages.

**Comment by tskisner** [comment]
@julienguy, I already saw that problem and fixed it- now I'm concerned that one of us is out of date...  Will go double check my local checkout before pulling your changes.

**Comment by tskisner** [comment]
@julienguy, working with my branch **before** your commits, I successfully did:
```
$ desi_pipe chain --tasktypes preproc --night 20191001 --nersc edison --nersc_queue debug
8759239
$ desi_pipe chain --tasktypes psf --night 20191001 --nersc edison --nersc_queue debug --depjobs 8759239
8759245
```
with no problems.  I'll go look through your commits and see what was changed.

tskisner:
OK, I looked through the commits, and it is just extra logging and the one critical fix (splitting the deps string before passing to _run_scripts() ).  So this is fine- just be aware that the default tasktypes (i.e. all of them), will fail if using --nersc_queue debug.

tskisner:
Ok, had an offline discussion with @julienguy and now I understand the problem better.  Will move the queued to a new DB column (with boolean value) and rename to "submitted".  This column will indicate whether the task has ever been submitted in the past (and hence should be ignored by further calls to the nightly script).  This submitted flag will be reset with a new option to desi_pipe cleanup.

sbailey:
I verified that the sequence of `desi_start_night`, `desi_update_night`, and `desi_end_night` works when trickling raw data in as long as I waited for all jobs from one step to finish before starting the next step.  I'm OK with either merging as-is and then fixing the `queued` logic in a separate branch+PR, or if Ted is already mid-way through a fix in this current branch we could wait for that if that will be ready by Monday  noon.

Wait for a future branch+PR:
* combining multiple steps in a single job
* tuning of job sizes and runtimes.

tskisner:
Ok, let's see if I can get that done by Monday morning.  Would be really great to have, but at the very least we should revert the addition of the "queued" state before merging.

tskisner:
Question for @julienguy , I believe the new "submitted" DB column should only apply to per-night tasks (not pixel based tasks), correct?  The purpose of this flag is very specific:  so that per-exposure runs of desi_update_night do not try to re-run tasks that are already queued.  Since spectral regrouping and redshift fitting must re-run the same tasks when there is new data, I think we basically always queue those tasks.

julienguy:
yes. it does not apply to spectra and redshift tasks.

tskisner:
Ok @julienguy, I made these changes and tested that desi_pipe chain can optionally ignore tasks with the submitted flag set.  Also tested that ```desi_pipe cleanup --submitted``` clears the flags.  I also removed a couple of debugging print statements that were confusing the job IDs returned by desi_pipe chain to the nightly scripts.

julienguy:
Verified jobs chaining, nightly script (nice to see new jobs being scheduled at each step desi_start_night , desi_update_night , desi_end_night), with new 'submitted' mode.
Ok for me to merge after travis tests ended.

PR #552, opened 2018-03-22
Fix integration test

sbailey:
In prep for the 18.3 software release, this PR fixes the desispec integration test and makes improvements to the user-friendliness of its outputs.
* support python 3.5 (previous sqlite3 DB only worked with python 3.6)
* Adds `desispec.pipeline.db.DataBase.count_task_states(tasktype)` utility function to count how many tasks are in each state for a given tasktype.
* Print number of tasks in each state before running each step* `desi_pipe` pre-flight check failures will exit with non-zero error codes
* Adds missing trace shift step (and standardizes on which steps are run in which order via `desispec.pipeline.tasks.base.default_task_chain`
* Don't clobber the production unless `clobber==True`
* If final zbest output files are missing, don't try to print redshift success table (which previously would crash with messy traceback).

### DIFF
--- a/py/desispec/pipeline/db.py
+++ b/py/desispec/pipeline/db.py
@@ -14,6 +14,7 @@ from __future__ import absolute_import, division, print_function
 import os
 
 import re
+from collections import OrderedDict
 
 from contextlib import contextmanager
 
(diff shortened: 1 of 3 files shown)


tskisner:
@sbailey , all changes look good to me.

PR #556, opened 2018-03-26
Ql snr

rstaten:
This PR includes QuickLook updates, most significantly pertaining to the signal vs. noise calculation:

SNR updates:

- Use the correct astronomical equation to calculate SNR.
- Remove infs, NaNs, etc. from magnitude and SNR arrays for fitting/plotting.
- Update SNR for bright and grey time.
- Consolidate per target output metrics into a single metric containing N target arrays.
- Change residual SNR values to fractional differences between calculated and fit values.
- Add configurable N sigma cut to residual values.
- Apply basic flux calibration to QA results, so that they can be applied to the SNR equation.

Other updates:

- Add QL ability to run on a single QA rather than full pipeline.
- Store failing XW sigma values and continue fitting.
- Obtain sky continuum ranges from configuration file.
- Amend percentile calculations in bias from overscan region QA.
- Update QL plots for newer versions of matplotlib.

@Srheft, @rkehoe, please review these updates and provide any necessary comments.


rstaten:
With recent commits, QL is now compatible with recent file/path changes. Calibration files are now appropriately generated and found. @Srheft, could you also verify that this is the case?

Srheft:
@rstaten have you uploaded the new config for arcs? if yes, where?

Srheft:
Thanks @rstaten; here are my comments:

Let's have a complete set of configuration files by adding preproc config (i.e.,qlconfig_flat_preproc.yaml ) in https://github.com/desihub/desispec/tree/ql_snr/py/desispec/data/quicklook


After updating the configuration files that has the new key *PSFExpid* used as expid for arc exposures must have been specified due to the recent path/name changes.

Starting from ground zero (running QL on a flat, arc and science frame):

1- pre-processing:
The fiber flat file was successfully created in the NEW address and now is called **preproc**:
$QL_SPEC_REDUX/**preproc**/20191016/00003874/preproc-b1-00003874.fits
----------------------------------------------------------

2- Arc-processing/spectral extraction  [resolution ON]:
This runs fine in that the ResolutionFit performs successfully to get the resolution from wsigmas from arc lines PSF
 QuickLook WARNING : Output filename is None and has a object of <class 'desispec.frame.Frame'>. SKIPPING FINAL OUTPUT

This was purposefully done to avoid the non-existing calibnight folder and the name that requires exposure id [correct?]. 

The merged JSON is being created here:
~/desi/spectro/redux/QL/calib2d/20191016/ql-mergedQAarc-b1-00003871.json

NOTE:
Ryan, as you have indicated in our discussion, quicklook needs to enforce  PSFType: psfboot for this step to work as we are used to run it (for instance, desi_quicklook -i qlconfig_arcs.yaml -n 20191016 -c b1 -e 3871).
 
----------------------------------------------------------
3- fiber flattening :
The fibermap was created in this address [with the **EXPID** in the name due to recent developments]:
~/$QL_SPEC_REDUX/exposures/20191016/00003874/**fiberflat-$camera-00003874.fits**
merged JSON file is created in the same address as well (~$QL_SPEC_REDUX/exposures/20191016/00003874/ql-mergedQA-b1-00003874.json)

NOTE:
The PSFtype key in the qlconfig_flat.yaml  was enforced to be PSFType: psfboot which negates the purpose of having UseResolution = True. If we set PSFtype=psfnight, then it complains that **~/desi/spectro/redux/QL/calibnight/20191016/psfnight-$camera-20191016.fits** doesn't exist.

--------------------------------------------------------
4- Sky Subtraction
Using PSFType=psfboot it ran fine and had the sky subtracted spectra and the merhed and 10 individual JSONs as well as 10 png plots written here:
~/desi/spectro/redux/QL/exposures/20191016/00003880/sframe-b1-00003880.fits
~/desi/spectro/redux/QL/exposures/20191016/00003880/ql-mergedQA-b1-00003880.json
All 10 plots were created.
JSONs have TGT instead of array for indivirual targets except for the fiber IDs that will be merged into FIBERID_TGT in the next release.

runtime: 2-3 min [on my 4 cpu machine]
---------------------------------------------------

I confirm that with the try & except addition to the XWsigma, the frequent failure of this QA has been resolved. It could get computationally expensive but how about [in the next round of the implementation] for those fibers that fail to have gauss-fit in both directions, we only try the x-direction [treating it as if we don't have the w info] and see if they at least get the curvefit result in one direction [i.e. assuming either wcoef*cos(arctan(x/w)) or xcoef*cos(arctan(w/x)) to be negligible for now and see how many failed fibers get recovered].

@rkehoe we need this to be merged today. Review/comment/merge ASAP. The need for the QL to be able to use the psfnight [currently we only use psfboot] is absolutely essential and hopefully will be implemented in the clean-up branch.  


Regards,
srh

Srheft:
I should mention that the bold faces and crossed letters in my comments are unintentional.

rstaten:
@Srheft the configuration file for flat preprocessing is now in desispec/data/quicklook. The warning you mentioned means the frame file was not written, which can be addressed in the clean up branch. Psfboot and psf files are generated by this configuration.

Srheft:
@rstaten I think I should clarify that PSFtype = psf  works as well as PSFtype = psfboot. It sounded like I'm not emphasizing enough on the fact that we Do use the resolution information but we DO.

sbailey:
Merging now so that we can include this in the desispec tag for the 18.3 software release.  If @rkehoe or others have post-facto comments or fixes, please open new issues / PRs and we can make another desispec tag later if needed.

PR #559, opened 2018-03-28
spec pipeline file/dir naming cleanup

sbailey:
This PR provides some file and directory naming cleanup:
* remove colons from spectro pipeline output directory and file names since they are a special character for Unix.
  * old: `run/scripts/redshift_20180328-00:28:04/run_20180328-00:28:04.log`
  * new: `run/scripts/redshift_20180328-002804/run_20180328-002804.log`
  * the new scheme is admitted less obviously readable as a time, but easier to work with when cutting and pasting and browsing directories with tab completion.
* use `desispec.io.findfile()` to get correct redrock hdf5 output file name

Additional features that came along for the ride while debugging the above:
* Fixed "fail" vs. "failed" typos in pipeline state tracking
* Added a `--skip-psf` option to `desispec.test.integration_test` for testing without spexec installed

### DIFF
--- a/py/desispec/pipeline/defs.py
+++ b/py/desispec/pipeline/defs.py
@@ -34,7 +34,7 @@ task_int_to_state = {
     1 : "ready",
     2 : "running",
     3 : "done",
-    4 : "fail"
+    4 : "failed"
 }
 
 
@@ -43,7 +43,7 @@ state_colors = {
     "ready" : "#0000ff",
     "running": "#ffff00",
     "done": "#00ff00",
-    "fail": "#ff0000",
+    "failed": "#ff0000",
 }
 """State colors used for visualization."""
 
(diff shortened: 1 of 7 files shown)


julienguy:
Stephen, why did you remove the backward compatibility of PSF I/O in traceshifts.py .
I would rather keep it for now because we have EM spectrograph PSFs in the old format on disk.

sbailey:
It did not have backwards compatibility (it assumed only X/YTRACE and didn't look for X/YCOEFF).  I tried adding it, which sort of worked, but resulted in extractions that has resolution matrices normalized to ~0.5 instead of ~1.0, thus eventually causing redshift fitting to fail due to masking.  Rather than go further down that rabbit hole, I backed out my changes so I wouldn't leave messy if/then code that was only a partial solution anyway.  i.e. there is a commit message about undoing changes for old PSFs, but in the end this PR doesn't impact any previous PSF format functionality (working or otherwise).

julienguy:
ok. thanks for the clarification. merging.

PR #564, opened 2018-03-28
module file set DESI_CCD_CALIBRATION_DATA

sbailey:
The new spectro pipeline needs $DESI_CCD_CALIBRATION_DATA to be set for PSF fitting (and maybe trace shifting too?).  This PR updates the module file to set that to $DESI_ROOT/spectro/ccd_calibration_data/trunk .

In the future for real productions we should tie this to tags of that product, but this gets a placeholder in there to enable desiInstall'ed modules to have a working environment.


weaverba137:
Looks good to me.

PR #577, opened 2018-04-06
Ql Clean up 1

gdhungana:
This PR addresses following updates as part of QL clean up
- Fix Continuum and Peaks QA params for all program config files
- Manage uniformity for status keys ALARM/ WARN -> NORMAL/WARN 
- Fix the remaining MAGDIFF -> DELTAMAG keys
- Remove redundant arguments for sky continuum QA

@rkehoe @Srheft Please review and suggest if this addresses the items discussed for this PR

### DIFF
--- a/py/desispec/data/quicklook/qlconfig_brightsurvey.yaml
+++ b/py/desispec/data/quicklook/qlconfig_brightsurvey.yaml
@@ -35,7 +35,7 @@ Algorithms:
             Get_RMS:
                 PARAMS: {NOISE_NORMAL_RANGE: [-1.0, 1.0], NOISE_WARN_RANGE: [-2.0, 2.0]}
             Count_Pixels:
-                PARAMS: {CUTHI: 10, CUTLO: 3, NPIX_WARN_RANGE: [200.0, 500.0], NPIX_ALARM_RANGE: [50.0, 650.0]}
+                PARAMS: {CUTHI: 10, CUTLO: 3, NPIX_NORMAL_RANGE: [200.0, 500.0], NPIX_WARN_RANGE: [50.0, 650.0]}
             Calc_XWSigma:
                 PARAMS: {B_PEAKS: [3914.4, 5199.3, 5578.9],
                          R_PEAKS: [6301.9, 6365.4, 7318.2, 7342.8, 7371.3],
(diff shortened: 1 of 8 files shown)


gdhungana:
The recent commits update the QA STATUS checks with the template file using the QA base class. This update provides examples of obtaining status for 3 QAs. This also adds a by-hand list of imaging mag in the merged dictionary to facilitate QLF implementations. @rkehoe @Srheft Please review.

rkehoe:
These updates are primarily items leftover from desispec-18.3.  The STATUS fix has been needed.  This appears ready to merge.

gdhungana:
@Srheft : I got things like 

WARNING: AstropyDeprecationWarning: "clobber" was deprecated in version 2.0 and will be removed in a future version. Use argument "overwrite" instead. [desispec.io.frame] 

running with python 3.6.4 with astropy 3.0.1.

I'd expect you'd get the same with master. But I did not get any failure as you are suggesting. I tested QL pipeline, individual QA, and also the QL tests in desispec. Travis also didn't fail. Do you have some specific setup locally? Can you send me more details on what test you got the QAs to fail? I am surprised as why they would pass with master and not with this branch. I do not think I have modified any of the I/O.

Srheft:
Deprecation warnings weren't worrisome and I've seen them in master too. But I don't recall them result  in failure of any QA last time I checked on the same machine I'm using now. Python 2.7 doesn't seem to have any issue with it but python 3 did give me a failure on bias_from_overscan:

 QuickLook WARNING : Failed to run QA Bias_From_Overscan. Got Exception Invalid slice for Image objects
  File "~/desi/software/desispec/py/desispec/quicklook/quicklook.py", line 305, in runpipeline
    res=qa(inp,**qargs)
  File "~/desi/software/desispec/py/desispec/quicklook/qas.py", line 25, in __call__
    res=self.run(*args,**kwargs)
  File "~/desi/software/desispec/py/desispec/qa/qa_quicklook.py", line 161, in run
    return self.run_qa(raw,camera,paname=paname,amps=amps, qafile=qafile,qafig=qafig, param=param, qlf=qlf, refmetrics=refmetrics)
  File "~/desi/software/desispec/py/desispec/qa/qa_quicklook.py", line 165, in run_qa
    rawimage=raw[camera.upper()].data
  File "~/desi/software/desispec/py/desispec/image.py", line 60, in __getitem__
    raise ValueError('Invalid slice for Image objects')
ValueError: Invalid slice for Image objects

Srheft:
As a side note: We must make sure that all the config files are updated to the latest pipeline developments.

gdhungana:
Hmm. That looks odd. Did you check the same run with master? that shouldn't be checking desispec.image, as rawimage in line 165 as quoted above should be only numpy.ndarray coming out of astropy.io.fits.hdu.hdulist.HDUList. Perhaps you should check your simulated file. Or you are running QL parsing different arguments, and probably one or more non-default mode is checking this and we do not have a test for that.

sbailey:
Regarding the questions about testing on python 2.7 vs. 3: we only support python 3 now, so please don't use python 2.7 as the default environment for development and testing.  Develop and test only using python 3.6.x .

Srheft:
@sbailey all the test are done with python 3. It was me just checking whether a problem I faced was also seen in old versions. The developments and all the tests are done with the latest version of python packages 3+

rkehoe:
@srheft  Thanks for committing the updated configuration.  In the future, we should try to ensure that each PR includes complete set of functioning configs (arc, flat, dark time, dark survey, greysurvey, bright survey, preproc).    I'll merge now...

PR #585, opened 2018-04-09
allow redrock to use a full node per healpix

### LINKED ISSUE #573: desi_pipe redshift step takes longer than wrap-redrock equivalent
For the 18.3 reference run of the minitest notebook, the redshift fitting step took ~35 minutes.  Processing the same spectra files with `wrap-redrock` takes ~20 minutes.  Figure out why and fix so that the pipeline is as fast as `wrap-redrock`.  Possibilities:
* random disk delays; they actually are the same speed
* non-optimal packing of healpix into nodes (`desispec.parallel.dist_discreet` vs. `weighted_partition` in `redrock/bin/wrap-redrock`)
* non-optimal allocation of ranks to individual healpix.

sbailey:
This PR fixes #573 by allowing the pipeline wrapper of redrock to use the full number of cores per node instead of just 20 (which I'm guessing was a leftover from copying the extraction task as a template for the redshift task).

This was especially problematic for the minitest where the largest healpix dominates the overall runtime of the redshift fitting step.  For a large production runs with approximately equally sized pixels it wouldn't have mattered as much, though splitting pixels across nodes would probably still carry some performance penalty.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,9 @@ desispec Change Log
 0.20.1 (unreleased)
 -------------------
 
-* No changes yet.
+* Pipeline fix to allow redrock to use a full node per healpix (PR `#585`_).
+
+.. _`#585`: https://github.com/desihub/desispec/pull/585
 
 0.20.0 (2018-03-29)
 -------------------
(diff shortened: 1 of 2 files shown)


tskisner:
That is definitely a typo.  Merge away!  :-)

Issue #531, opened 2018-03-15
pipeline DB: documentation


julienguy:
Document pipeline usage

tskisner:
Just to expand on this topic before merging #590.  The desispec documentation has two relevant sections, "pipeline use" and "pipeline development".  The "pipeline use" section should document every command and option provided by the desi_pipe tool, as well as examples of common use cases, including:

- Creating / updating / syncing a production
- Looking at tasks and their states
- Running specific tasks (the "script" and "run" commands) and examining logs
- High-level jobs with desi_pipe chain
- Emergency recovery techniques (database problems)

The "pipeline development" section should go into details about:

- Pipeline tasks and their class methods
- DataBase class:  methods, how task state is tracked, **how the healpix_frame table is used** to manage the special dependency case of spectra and redshift tasks.

moustakas:
I realize that the pipeline is still being developed and that documentation (cf this issue) is understandably lagging, but I wonder if there's a README, simple tutorial, or other general guidelines for running the latest version of the pipeline anywhere.

tskisner:
Hello @moustakas, point taken...  The calling options to the main "command and control" program (desi_pipe) are fairly stable at this point.  I will see if I can get some minimal docs on that done this weekend, and leave the "developer docs" out for now (and leave this ticket open).

akremin:
`desi_pipe` is now deprecated, so I will close this ticket.

PR #590, opened 2018-04-17
Better sync of pixel tasks

### LINKED ISSUE #558: desi_pipe sync waiting/ready inconsistencies
**Use case**: I wanted to re-run some pipeline steps, so I deleted their output files and re-ran `desi_pipe sync` intending it to update the database tasks from "done" back to "waiting" and "ready".

**Problem**: Some tasktypes went back to "ready" even though they should have gone back to "waiting".

**Workaround**: I suspect that ready vs. waiting doesn't matter for this case, but I was able to get back to the desired state by deleting the sqlite file and beginning again with `desi_pipe create --db-sqlite --force` and `desi_pipe sync`.

**Example**
```
[cori03 ~] desi_pipe top --once
----------------+---------+---------+---------+---------+---------+---------+
   Task Type    | waiting | ready   | running | done    | failed  | submit  |
----------------+---------+---------+---------+---------+---------+---------+
preproc         |     1260|        0|        0|        0|        0|        0|
psf             |      360|        0|        0|        0|        0|        0|
psfnight        |      120|        0|        0|        0|        0|        0|
traceshift      |      900|        0|        0|        0|        0|        0|
extract         |        0|        0|        0|      900|        0|        0|
fiberflat       |        0|        0|        0|      360|        0|      360|
fiberflatnight  |        0|        0|        0|      120|        0|      120|
sky             |        0|        0|        0|      540|        0|      540|
starfit         |        0|        0|        0|      180|        0|      180|
fluxcalib       |        0|        0|        0|      540|        0|      540|
cframe          |        0|        0|        0|      540|        0|      540|
spectra         |        0|       46|        0|        0|        0|       NA|
redshift        |       46|        0|        0|        0|        0|       NA|
----------------+---------+---------+---------+---------+---------+---------+
```

remove a bunch of files in a different window, then...

```
[cori03 ~] desi_pipe sync
[cori03 ~] desi_pipe top --once
----------------+---------+---------+---------+---------+---------+---------+
   Task Type    | waiting | ready   | running | done    | failed  | submit  |
----------------+---------+---------+---------+---------+---------+---------+
preproc         |     1260|        0|        0|        0|        0|        0|
psf             |      360|        0|        0|        0|        0|        0|
psfnight        |      120|        0|        0|        0|        0|        0|
traceshift      |      900|        0|        0|        0|        0|        0|
extract         |        0|        0|        0|      900|        0|        0|
fiberflat       |        0|      360|        0|        0|        0|      360|
fiberflatnight  |      120|        0|        0|        0|        0|      120|
sky             |        0|      540|        0|        0|        0|      540|
starfit         |      180|        0|        0|        0|        0|      180|
fluxcalib       |      540|        0|        0|        0|        0|      540|
cframe          |      540|        0|        0|        0|        0|      540|
spectra         |        0|       46|        0|        0|        0|       NA|
redshift        |       46|        0|        0|        0|        0|       NA|
----------------+---------+---------+---------+---------+---------+---------+
```
Note that the sky state is in "ready", which can't really be true if fiberflat and fiberflat night still need to be run.

### LINKED ISSUE #563: desi_pipe sync doesn't sync redrock output
My redrock job timed out leaving me partially done on the redshift tasks.  `desi_pipe sync` doesn't recognize that and still leaves all of my tasks in the "ready" state instead of moving those with file output into the "done" state.

Optimistically assigning this to 18.3 'cause it would be really really nice to have this working for this major release.

tskisner:
The DataBase.sync() method scans the filesystem and rebuilds the database state of all tasks.  This works fine for all tasks through cframes.  For spectra and redshift tasks, the dependencies on cframes are tracked in the healpix_frame table.  This PR attempts to reconstruct the state of these tasks (i.e. waiting, ready, done) based on the filesystem **and** on the current status of the healpix_frame table.  The logic here is the desire to do:
```
$>  find spectra-64 -name 'zbest*' -exec rm '{}' \;
$>  desi_pipe sync
```
And have the state of the spectra and redshift tasks be correct.  The sync command now also calls the getready() method.

Additionally there was a bug in the order of tasktypes returned by db.task_types() (used in several places).  The tasktype order was not the same as the execution order.  This should close #558 and close #563 .

### DIFF
--- a/py/desispec/pipeline/db.py
+++ b/py/desispec/pipeline/db.py
@@ -698,8 +698,9 @@ class DataBase:
 
         def set_hpx_frame_1(row, spec, red, cur):
             self.update_healpix_frame_state(row, 1, cur)
-            task_classes["spectra"].state_set(
-                self, spec, "ready", cur)
+            # getready() will do this for us:
+            #task_classes["spectra"].state_set(
+            #    self, spec, "ready", cur)
             task_classes["redshift"].state_set(
                 self, red, "waiting", cur)
             return
@@ -708,8 +709,9 @@ class DataBase:
             self.update_healpix_frame_state(row, 2, cur)
             task_classes["spectra"].state_set(
                 self, spec, "done", cur)
-            task_classes["redshift"].state_set(
-                self, red, "ready", cur)
+            # getready() will do this:
+            #task_classes["redshift"].state_set(
+            #    self, red, "ready", cur)
             return
 
         def set_hpx_frame_3(row, spec, red, cur):
--- a/py/desispec/pipeline/db.py
+++ b/py/desispec/pipeline/db.py
@@ -36,8 +36,11 @@ def task_types():
         list: The list of supported task types.
 
     """
-    from .tasks.base import task_classes, task_type
-    return list(sorted(task_classes.keys()))
+    from . import tasks
+    from .tasks.base import default_task_chain
+    ttypes = ["fibermap", "rawdata"]
+    ttypes.extend(tasks.base.default_task_chain)
+    return ttypes
 
 
 def all_tasks(night, nside):
@@ -74,7 +77,6 @@ def all_tasks(night, nside):
 
     for ex in sorted(expid):
 
-
         # get the fibermap for this exposure
         fibermap = io.get_raw_files("fibermap", night, ex)
 
@@ -89,12 +91,8 @@ def all_tasks(night, nside):
             log.error("Do not know what do to with fibermap flavor '{}' for file '{}".format(flavor,fibermap))
             raise ValueError("Do not know what do to with fibermap flavor '{}' for file '{}".format(flavor,fibermap))
 
-
         fmpix = dict()
         if (flavor != "arc") and (flavor != "flat"):
-
-
-
             # This will be used to track which healpix pixels are
             # touched by fibers from each spectrograph.
             ra = np.array(fmdata["RA_TARGET"], dtype=np.float64)
@@ -136,8 +134,6 @@ def all_tasks(night, nside):
                         break
                 if not exists : full["redshift"].append(props)
 
-
-
         fmprops = dict()
         fmprops["night"]  = int(night)
         fmprops["expid"]  = int(ex)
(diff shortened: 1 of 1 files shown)


tskisner:
Offline feedback from @sbailey:  for this short term fix, if the spectra and redshift files exist at all, then we should assume they are done.  

This will give us the wrong answer if (for example) we have completed cframes that have not yet been added to spectra and redshifts, and then we run ```desi_pipe sync```.  However, it would handle the case of files being created outside the pipeline that we then want to include in the DB.  

In the case where files were removed, we would still move the state back to a previous one cleanly.  

The long term solution is obviously to open the fibermaps in each file and see if the cframe contents are complete.

tskisner:
@sbailey, this should now implement your desired behavior- can you test again on your mini sim?  I verified that using one day of a month production that I can add / remove spectra and zbest files and sync sets the expected state.  One note- if spectra files are missing but zbest files exist, then it resets the state back to "spectra = ready, redshift = waiting".  Presumably if you are creating zbest files outside of the pipeline then you also have spectra files, so this should not be a problem.

sbailey:
I confirm that it now works for my minitest case.  Thanks.

Please add some comments in the code about what the 1/2/3/4 magic number states mean.  Some of those are mentioned `# We are all done (state 3)` but others are more implicit "`# We are only at state 2`".  I think the mapping is:

* 0 = some cframe inputs are missing (but not necessarily all)
* 1 = all required cframes have been generated
* 2 = spectra file has been updated
* 3 = zbest file has been updated

When new data arrive, do all covered healpix reset back to state 0 because some cframe files are missing, even if others exist and have been processed all the way through redshifts?  Answer that with code comments, not just by answering me here.  If those states are already clearly documented in comments elsewhere, it's ok if this updating code just references "see x.y.z for explanation of the states".

tskisner:
The issue of comments and documentation is certainly valid, and documenting the states in the healpix_frame table and when those are changed should be done.  It is a much bigger issue than just the DataBase.sync() method changed here (see #531).

For the record here:  completion of new cframes changes pixel state 0 --> 1.  Completion of spectra changes 1 --> 2.  Completion of redshifts changes 2 --> 3.

So no spectra files will be moved from "done" back to "ready" until after a new cframe file that touches that pixel is completed.

Issue #504, opened 2018-02-08
make QA faster


sbailey:
Make `desi_qa_prod` and `desi_qa_skyresid` faster.  They currently take ~30 minutes each on the minitest notebook, which only has 18 exposures of 10 tiles (i.e. less than half a night of DESI observing).  We need something much much faster to be viable for full production running.

The I/O could be made much faster (e.g. with fitsio and not reading the resolution matrix that isn't used for QA), but even that doesn't appear to be the dominant factor.  A simple loop over all 540 frame files calling `desispec.io.read_frame` was able to read all of them in 2.8 minutes.

profxj:
Sufficiently addressed by PR #593 (for now)

PR #593, opened 2018-04-18
Speedup offline QA without Truth

profxj:
Changes to offline QA code, primarily to speed up performance
but also to enable testing within the mini Notebook's.  Here are
the specific changes:

* Add option to load Frame without reading Resolution matrix
* Refactor offline QA to use qaprod_dir more sensibly
* Include hooks in QA to previous fiberflat file location (calib2d)
* Inhibit scatter plot in skyredidual QA

Offline QA code figure generation now takes 7.5min on reference run 18.2 on cori

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,6 +5,10 @@ desispec Change Log
 0.20.1 (unreleased)
 -------------------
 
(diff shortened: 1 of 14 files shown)


profxj:
Addresses Issue #504

PR #611, opened 2018-05-14
Update QL SNR fit

rstaten:
This PR:
- Updates the SNR fit to include exposure time.
- Updates the conversion of imaging magnitude to flux.
- Includes minor updates for SNR plots.
- Adds imaging magnitudes to SNR output.
- Returns the median of sky fibers over sky peaks for QA Sky_Peaks.

### DIFF
--- a/py/desispec/qa/qa_plots_ql.py
+++ b/py/desispec/qa/qa_plots_ql.py
@@ -590,7 +590,7 @@ def plot_SNR(qa_dict,outfile,objlist,badfibs,fitsnr,rescut,sigmacut):
 
         if i == 0:
             ax.set_ylabel('Median S/N**2',fontsize=8)
-        ax.set_xlabel('{} Mag ({})'.format(objtype,thisfilter),fontsize=6)
+        ax.set_xlabel('{} Mag ({})\na={:4f}, B={:4f}'.format(objtype,thisfilter,fitval[0],fitval[1]),fontsize=6)
         ax.set_xlim(16,24)
         ax.tick_params(axis='x',labelsize=6)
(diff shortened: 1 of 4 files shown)


rkehoe:
@rstaten, @Srheft This branch looks good to me.  I can't run a test right now, but aside from doing that it looks like it is ready to merge.  Please go ahead with that as appropriate.

Srheft:
@rstaten: thank you for this. SNR evaluation is starting to come together nicely.  Could you:

in line 586 of qalib.py, use the json file as opposed to yaml. We no longer produce yamls in the next release but now that we have both in current master, let's see the migration in full effect.

Srheft:
@rstaten:  please also make this change->  in line 587 of qalib.py  NOISE_OVER -> NOISE

rstaten:
@Srheft I fixed/added points 2 and 3. As 1 will involve more changes than the line you mention, I will do this in a subsequent branch. Regarding 4, currently SNR fits each object type, so choosing a single magnitude range won't work for all objects (specifically STD). This can be reconciled if we decide to only fit certain object types, but for now it seems to make sense to fit the whole magnitude range.

rkehoe:
@rstaten  About the item 4, I would suggest to consider a third option, for a future branch though since we want this PR merged soon.  This third option might be to have the algorithm auto determine the magnitude range per object, based on some general principles (eg. # of objects in 0.25 mag bins, no less than 2 mag range).  I don't think we want to add configuration parameters for each target type, but it seems likely we will want to know the results for all target types during data taking.   Perhaps you can include this with the branch where you take care of the JSON migration point from @Srheft ?

Srheft:
Merged.

PR #615, opened 2018-05-20
Qlresample

rkehoe:
This PR addresses shortcomings in QL when resampling to a fixed grid.  Results on arc and continuum lamp sims now agree at all wavelengths with input native grid, as in attached plots.

[QLgrid.pdf](https://github.com/desihub/desispec/files/2020818/QLgrid.pdf)

@Srheft,  @rstaten, please test this rigorously and on science data.

### DIFF
--- a/py/desispec/qa/qa_quicklook.py
+++ b/py/desispec/qa/qa_quicklook.py
@@ -252,7 +252,7 @@ class Bias_From_Overscan(MonitoringAlg):
         if not self.is_compatible(type(args[0])):
             raise qlexceptions.ParameterException("Incompatible input. Was expecting {} got {}".format(type(self.__inpType__),type(args[0])))
 
-        if kwargs["singleqa"] == 'Bias_From_Ovescan':
+        if kwargs["singleqa"] == 'Bias_From_Overscan':
             night = kwargs['night']
             expid = '{:08d}'.format(kwargs['expid'])
             camera = kwargs['camera']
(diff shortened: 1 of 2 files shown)


Srheft:
Thank you @rkehoe. Few comments:

- I ran this branch with all the configuration files, processing few flats, arcs, Science frames [dark,grey,bright programs]. All successful runs. Needed to make a small change in qa_quicklook for a minor failure but the git setting did not let me to push the change to qlresample; saying "Please make sure you have the correct access rights and the repository exists". I will make that change in another PR as it is not a critical failure.

- I did create fiberflat for few different cameras: z2, z7, r1, b9 and confirmed that the infamous ellipse features are gone [could attach a pic here if you like to document it here].

- In line 67-69 of palib.py: Please add a comment on the "newvar" and its relation to "maintaining Total S/N".  It was not clear how the newvar is maintaining the total SNR. Does it mean:
 sum(newivar)  =? sum(median(flux*sqrt(ivar)))
Note: (median(flux*sqrt(ivar))  is how quicklook calculates SNR through the function  SN_ratio(flux,ivar) in qalib.py.

rkehoe:
@srheft  The error was coming ultimately from cases where the native grid did not contribute to the first entries of the resampled grid.  I've put a catch for that to avoid the divide by zero, but this needs a better approach.  I'll merge this branch, but will improve this further in a subsequent branch.

Not sure what happened with the GitHub access.  You should have been able to push.  Let's take a look at this once the dust settles on existing PRs.

PR #617, opened 2018-05-22
Add Flexure test to QL

rstaten:
This PR:
 - Adds a QL pipeline algorithm to generate a modified PSF file (using desi_compute_trace_shifts).
 - Includes a placeholder QA for trace shifts to be updated when proper metrics are determined.
 - Updates the FIBER_MAG conversion from integrated counts to magnitudes.
 - Removes list from merged json file.

@Srheft, @rkehoe Please review these updates and add any necessary changes or comments.


Srheft:
Thanks @rstaten for the good fixes added in this PR. I tested this branch on flat/arc science exposures and they were all successful. as expected, adding flexure to the pipeline steps added ~1.5 min to the science processing time [that is to be reported to the authors of that code for possible runtime reduction].
With this PR, I am happy to announce that all the 7 tickets coming from the "Utah Agreement" are in!

rkehoe:
@rstaten  This is a very useful PR!  Thanks.  I am still going thru it, but a note at the moment.  Like with the qlconfig_darksurvey, where the Flexure PA is given separately and the Trace_Shift QA is hanging off of it, the bright survey and greysurvey configs need to be implemented this way also.  In addition, any of the pipelines that get to extraction (flat and arc) should also have this PA/QA implemented between preproc and extraction.  So Trace_Shift should not hang off of Preproc PA, but off of Flexure.   I document this here so others may familiarize with it, but I suspect it was just a typo when updating to the PA.  Could you address that in the GitHub configs? I will keep going over this...

rkehoe:
@rstaten  I finished looking.  Looks good, and assuming my prior note on configs is implemented, this is ready to merge.  

I have one point, though, that should be dealt with in a subsequent update.  The QA scalar metrics in the code appear to be defined as TRACE_SHIFTS, while we have separately referred to them as XWSHIFT and DELTA_X/Y (see @srheft comment here).  I don't have a strong opinion over what they should be called (TRACEXY?, XYSHIFT, DELTAXY), but it would be important not to have two separate scalar metrics here.  As in XWSIGMA, please implement the scalar metric as a two-component metric, to take the same specifications for each coordinate.  Also, the name of the scalar metric should be the same as the specification, minus the WARN/NORMAL_RANGE.  So TRACE_SHIFT might make for a very long spec name.

rstaten:
@rkehoe There are actually no metrics defined as we haven't yet determined exactly what they will be, but they can certainly be whatever is decided. Right now metrics for this QA just returns an empty dictionary. The ranges are defined as TRACE_x_RANGE, but this is just a placeholder until the metrics are determined.

Srheft:
Merged.

PR #623, opened 2018-05-25
Data discussion updates

rstaten:
This PR:
 - Updates xwsigma scalar metric.
 - Adds night for reference template.
 - Uses merged json file instead of yaml.

@Srheft, @rkehoe Please review and make sure this satisfies requirements from the data meeting.


rkehoe:
Looks good.  This makes some useful updates following yesterday's discussion.  @srheft, from my end this is ready to merge after your tests.

Srheft:
Tests are returning no failure. Template night looks for a merged JSON to set references but we don't make the QL crash for not finding the ref JSON. Certain measures and decisions to refuse running a class if the params are not set, will be placed for the future release.

PR #632, opened 2018-06-06
Faster trace shifts

julienguy:
Improvements to trace shifts
 - faster: from 3min30sec to 20sec on a dark time RED CCD image on single edison core.
 - header keywords in hdu PSF of output shifted psf fits file giving the mean,min and max shifts.
More details in issue #622

PR #633, opened 2018-06-06
QA Data Task

profxj:
Primarily a new QADataTask in the tasks framework
of desi_pipe.

Tested on NERSC by JXP and Julien.

Gets the ball rolling for re-integrating QA within the main pipeline.


tskisner:
Hi @profxj , I see now that the comments I left in #632 probably should have been here- I think maybe @julienguy merged these into his traceshift working branch.  Basically one change:

In db.all_task_types(), I think we want the order to be:

fibermap,
rawdata,
<default_task_chain up to cframe>
qadata
<the rest of default_task_chain>

profxj:
ok, I believe I've accomplished this.  
To be honest, I don't think we need qdata
right after cframe (after redshifts and spectra would be fine too).

tskisner:
Looks good, thanks!

Issue #622, opened 2018-05-24
Flexure modification for QL purpose


Srheft:
@julienguy:
 Currently, QL only uses flexure to report the XYshifts of the traces.  A wider discussion would be whether QL should start using those PSFs [another ticket?] but for now, is it possible to turn off the PSF creation and instead spit out an array of (X,Y)shifts per fiber?

sbailey:
We should use the shifted PSFs for quicklook, unless they turn out to be unstable with occasional catastrophic failures (which we would need to fix anyway, but they wouldn't necessarily have to cause boxcar extractions to fail).

The output PSFs should include the pre-calculated metrics for QL to use, rather than having a special mode for outputting the shifts without outputting the updated PSF itself.

Assigning to @julienguy, and I will talk to him in person since he may not have noticed this.

julienguy:
Work in progress to run the trace shifts faster in branch faster-trace-shifts 
 - faster trace-shift along cross-dispersion (essentially thanks to a rebinning of the input image along the other axis, the dispersion axis)
- much faster boxcar extraction ( **1.5sec per frame** !!!) with numba , see
   https://github.com/desihub/desispec/blob/37449a27dcd0b09067418e3fdc56858791dee8a5/py/desispec/trace_shifts.py#L235
- the shifts of traces along x and y now takes 45s instead of 5min. 

This is still on-going work. I expect more computation time gain with a more systematic use of numba in this code

julienguy:
Finished implementation of faster trace shifts in branch faster-trace-shifts.
For the following example run:
```time desi_compute_trace_shifts --image ~/sim2017/redux/month7/exposures/20191001/00003577/pix-r1-00003577.fits --psf ~/sim2017/redux/month7/calib2d/psf/20191001/psfnight-r1.fits --outpsf psf-shifted-dx-r1-00003577.fits```
 It now takes 20 sec compared to 3 min 30 sec with the master version on one edison core.

**Comment by julienguy** [comment]
Added header keywords in HDU PSF of the shifted PSF fits file.
Mean , min and max shifts of the traces across all fibers and the full wavelength.
For the above example (CCD r1 in a dark exposure)

```
MEANDX  = -0.00127486349305683
MINDX   = -0.04011489853837702
MAXDX   =   0.0233941410119769
MEANDY  = -0.00312482008538481
MINDY   = -0.02353019516999666
MAXDY   =   0.0116099172896611
```
(units are CCD pixels)

[...]

PR #634, opened 2018-06-06
faster trace shifts

julienguy:
New PR for the trace shift acceleration (from 3min 30s to 20s, using numba).
Added trace shifts values in header of shifted PSF for the quicklook pipeline.
More details in issue #622.


tskisner:
This all looks good to me.  Looking forward to faster traceshifts.

tskisner:
One comment- can you adjust the traceshift task time to be reasonable for this new code?  If you are not sure of the number then just leave it and we can tweak it after testing on edison.

PR #638, opened 2018-06-08
Data base loading in integration test

### LINKED ISSUE #637: redshift DB loading errors
The desispec integration test (python -m desispec.test.old_integration_test) is generating non-fatal error messages related to redshift DB loading:

```
ERROR:redshift.py:529:load_zcat: (sqlite3.IntegrityError) NOT NULL constraint failed: zcat.coeff_0 [SQL: 'INSERT INTO zcat (targetid, chi2, z, zerr, zwarn, npixels, spectype, subtype, ncoeff, deltachi2, brickname, numexp, numtile) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)'] [parameters: ((185670190927646600, 7720.9840676784515, 0.9347282742773337, 2.4721931985478964e-05, 0, 7650, 'GALAXY', '', 10, 406.79132664203644, '3350p205', 1, 1), (445603930984159622, 2523.589103899896, 0.7152646488307819, 0.0004004767718361138, 4, 7646, 'GALAXY', '', 10, 0.5871839812025428, '3350p207', 1, 1), (1292730898444268586, 2471.687471103649, 0.01429586432758536, 0.00013395120602628793, 4, 7674, 'GALAXY', '', 10, 0.6003796133172727, '3350p205', 1, 1), (2250805110042921454, 7715.686710610986, 0.9995244913002634, 9.898407386153824e-05, 4, 7685, 'GALAXY', '', 10, 0.4147386010736227, '3352p207', 1, 1), (2667388932819333481, 332.10182214248744, 0.0003372087377787315, 4.016325143348069e-06, 0, 9118, 'STAR
 ', 'F', 5, 1078.5971626589799, '3353p205', 1, 1))] (Background on this error at: http://sqlalche.me/e/gkpj)
```

weaverba137:
This PR fixes #637, and cleans up the old database loading scripts.


weaverba137:
PS, I ran the integration test myself on cori, and it worked, but it wouldn't hurt to have someone else run it as well.

py/desispec/database/redshift.py
```diff
@@ -474,7 +474,7 @@ def load_file(filepath, tcls, hdu=1, expand=None, convert=None, index=None,
     return
 
 
-def load_zcat(datapath=None, q3c=False):
+def load_zbest(datapath=None, q3c=False):
```
sbailey: I'm confused about the DB loading roadmap here.  This function docstring says that it is deprecated since there should now be a single redshift catalog file, but I don't see where the code is for loading that file.  Previously we had specifically moved away from loading N>>1 individual zbest files (e.g. to avoid the hardcoded "spectra-64/*/*/zbest-64*.fits assumptions).  Where is the target catalog loading now done (for when the target catalog exists)?

To be clear: DB loading should match the data available from "real" processing like the minitest notebook.  If it is also able to support the subset of files available from the desispec integration test, great.  If not, we shouldn't go to great lengths to alter the DB loading to match a non-standard case.

weaverba137: Assuming the zcatalog file *exists*, it is loaded by `desispec.database.redshift.load_file()`, just like the target, truth and exposure files.  If it does not exist, then a subset of the data can be loaded from the zbest files, and that's what I've updated.  I can remove that deprecation warning if necessary.

However, it sounds like you want the integration test to actually create the zcatalog file.  Is it even possible to do that without targeting data?  If so, we can try that.

sbailey:
Thanks for the explanation.  This is good enough for now for the integration test.

PR #642, opened 2018-06-20
Logging QA metric status

felipelm:
Suggestion to log QA status so we can color petals on QLF.


Srheft:
@felipelm: 

Due to the test failure, I added your suggested screen log  to the ql_TagUpdate_0.21.0  together with a change for tests to accommodate this. If QLF got what they need from the most recent commit, please close this PR. 
Thanks.

felipelm:
@Srheft thanks I'll close this now

Issue #612, opened 2018-05-16
reduce preproc cosmics memory usage


sbailey:
`desi_preproc` currently uses 4.6 GB of memory to pre-process a single camera.  This is dominated by the cosmics finding algorithm; using `--nocosmic` reduces the memory footprint to 1 GB.

This prevents us from preprocessing all 30 cameras in parallel on a single cori or edison node.  That's not terribly problematic for the spectro pipeline, but it could be problematic for quicklook which needs to process all 30 cameras on a single node at KPNO ASAP.  Although quicklook tests (not at NERSC) have survived full parallelism, they are likely making use of virtual memory that could be bogging down the speed.

julienguy:
Fixed in branch light-cosmics using numba to compute gradients per pixel instead of gradient images.
 - gain of a factor 4 in memory usage (from 4GB to 1GB for the most memory consuming images which are the high S/N continuum images).
- gain in CPU time varies depending on image type. max gain is a factor 2 (from 18s to 9s to mask the cosmic rays, on a simulated dark time exposure).

PR #644, opened 2018-06-26
cosmic ray finder is faster uses less memory with numba

julienguy:
Solves issue #612 . Reduced memory usage by a factor 4  using numba in the cosmic ray finder.

### DIFF
--- a/py/desispec/cosmics.py
+++ b/py/desispec/cosmics.py
@@ -11,8 +11,14 @@ import numpy as np
 import math
 import copy
 import scipy.ndimage
+import time
+
+import numba
+
 from desispec.maskbits import ccdmask
 from desispec.maskbits import specmask
+
+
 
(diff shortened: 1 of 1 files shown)


sbailey:
Tested on an r5 image on cori, I find 20.5 sec -> 12.5 sec, and 2.2G -> 1.2G.  The pixels, ivar, and readnoise are identical, while a tiny fraction of the masked pixels changed their mask value.  Most but not all of those were edge pixels; there were 9 interior pixels whose mask value changed.  That's good enough for me, but holler if you want to check those for completeness.

sbailey:
For the record: Julien pointed out that the flat lamp exposures have higher S/N and the original algorithm used more memory in that case, so his 4x savings in memory isn't an exaggeration: it is indeed a 4x win for flats, which is huge because that will enable us to run all 30 cameras simultaneously.

Issue #514, opened 2018-02-27
Cryptic message when using desispec.interpolation.resample_flux for size one array


londumas:
When trying to resample an array with size one, the code gives a cryptic error.
We could add a `try/except` to clarify the error.
```
Traceback (most recent call last):
  File "test_nb_resample.py", line 16, in <module>
    flux, ivar = resample_flux(wave, twave[:1], tflux[:1], tivar[:1])
  File "/global/homes/h/hdumasde/Programs/desi/code/desispec/py/desispec/interpolation.py", line 117, in resample_flux
    a = _unweighted_resample(xout, x, flux*ivar, extrapolate=False)
  File "/global/homes/h/hdumasde/Programs/desi/code/desispec/py/desispec/interpolation.py", line 193, in _unweighted_resample
    ixmin=1.5*ix[0]-0.5*ix[1]  # = ix[0]-(ix[1]-ix[0])/2
IndexError: index 1 is out of bounds for axis 0 with size 1
```

```
import scipy as sp
from desispec.interpolation import resample_flux

### Get data
spec = sp.loadtxt('issue_spectrum.txt')
twave = spec[:,0]
tflux = spec[:,1]
tivar = spec[:,2]

### Resample
CRVAL1 = 2.6534
CDELT1 = 0.0001
NBLL   = 13637
wave   = 10**(CRVAL1+sp.arange(NBLL)*CDELT1)

flux, ivar = resample_flux(wave, twave[:1], tflux[:1], tivar[:1])
print(flux, ivar)
```

[issue_spectrum.txt](https://github.com/desihub/desispec/files/1764081/issue_spectrum.txt)

londumas:
The issue comes from the following line:
https://github.com/desihub/desispec/blob/f32955b95534cb1864a425dcb0ec8fb9c5ba9c74/py/desispec/interpolation.py#L195 that gives a negative ivar for this spectrum

julienguy:
IndexError: index 1 is out of bounds for axis 0 with size 1 
is pretty clear for me.

PR #645, opened 2018-06-27
handle exception if input array has a size of 1

londumas:
Fix issue https://github.com/desihub/desispec/issues/514 by adding an error message if the input array is of size 1.


julienguy:
Please re-raise the same exception. In fact I am not sure this commit is really useful.

```
try :
    ...
except IndexError as e :
    print("oups")
    raise e 
```

londumas:
ok, let's forget it then.

PR #648, opened 2018-06-28
Move mock observing updates into master

rstaten:
This PR:

- Adds functionality to map from new to old fibermap format (via @sbailey)
- Adds expid to raw/fibermap path
- Pythonically opens/closes configuration file
- Consolidates QL findfile arguments
- Adds output plot of sframe spectra (one spectrum per target type)
- Updates deltamag calculation to include all fibers
- Uses linalg if cholesky fails for fiber flats
- Prints status result per QA


sbailey:
As a heads up, the disruptive change will be that `desispec.io.findfile` will look for fibermaps and desi* raw data in per-exposure directories instead of per-night directories.  To read previously simulated data after this PR we will need to add simlinks (thanks to @julienguy for the suggestion).  But the per-exposure subdirs are what ICS will provide us so we need to move in that direction.

Related: desihub/desispec#384 to write simulated data in the new subdirs.

@rstaten has a few more updates that we might get in tonight so not merging just yet...

sbailey:
For the record, other changes coming in with this PR:
* separate QL config files into a `$QL_CONFIG_DIR` so that they can be updated independently of the code installation
* temporarily get PSF and fiberflat reference files from `$QL_CALIB_DIR` until we converge with @julienguy about formats for those in `$DESI_CCD_CALIBRATION_DATA` svn repo
* Add file locking when reading the QL config file.  For reasons we don't understand, this appears to be necessary when running with docker+NFS at KPNO.  Without the locks, sometimes the config file comes back as an empty string (!).  We haven't been able to reproduce the problem in separate tests.  This PR uses the [filelock](https://filelock.readthedocs.io/en/latest/) library, but it is implemented as an optional lock so that you don't have to have it installed.

Tests pass on my laptop; we'll see if Travis agrees...

PR #650, opened 2018-06-29
edison realtime queue 25 nodes

sbailey:
This PR updates the edison realtime queue config to be 25 nodes instead of the previous 10.  This is needed for testing of the data transfer -> spectro pipeline.

@tskisner and/or @weaverba137 may have already done this in a separate branch, in which case they can close this PR.

Note: the testing will need this PR or equivalent, but it doesn't require the latest desiconda which isn't installed on edison yet.

### DIFF
--- a/py/desispec/pipeline/scriptgen.py
+++ b/py/desispec/pipeline/scriptgen.py
@@ -57,7 +57,7 @@ def nersc_machine(name, queue):
             props["submitlimit"] = 5000
             props["sbatch"].append("#SBATCH --partition=regular")
         elif queue == "realtime":
-            props["maxnodes"] = 10
+            props["maxnodes"] = 25
             props["maxtime"] = 120
             props["submitlimit"] = 5000
             props["sbatch"].append("#SBATCH --exclusive")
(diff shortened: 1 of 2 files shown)


tskisner:
Hi @sbailey, this change looks fine to me, thanks for updating the max nodes for the realtime queue.

weaverba137:
I’m seriously confused by this. Does the command-line option to `desi_night` not work? I had already set that explicitly for the test.

weaverba137:
To clarify, you're saying that `desi_night update ... --nersc edison --nersc_queue realtime --nersc_maxnodes 25` is invalid because the number of nodes was hardcoded?

tskisner:
The command line option for --nersc_maxnodes is a way of restricting the number to be something smaller than the "hard" maximum, which is set by NERSC for each queue.  The realtime queue is sort of a special case, since NERSC can change that for us.  This PR changes that "hard" maximum to be what NERSC has recently configured for us

weaverba137:
Well, then it sounds like the realtime parameters shouldn't be hardcoded at all!

tskisner:
In practice, once we are set up for actual operations, that number will be fixed.  It is only in these early days that it is changing.  This hardcoded maximum is the fallback value if --nersc_maxnodes is not specified.  This number is set for each queue, in order to relieve the user of knowing the details about the particular NERSC queue limits.

weaverba137:
OK, let's get this in, because I'll have to merge it with my dts-scripts branch.

PR #653, opened 2018-06-30
DTS scripts and pipeline handoff

### LINKED ISSUE #327: Verify DTS delivery scripts
Upon completion of #322, we also need to test that the delivery scripts work when called by DTS.

It should be sufficient to test transfer a few nights worth of data from NOAO to NERSC.

weaverba137:
This PR closes #327.  We have a working DTS transfer script that copies (for now, simulated) data from KPNO to NERSC and hands off the data to the pipeline.  I can archive the log file of the test that ran today (29 June 2018) if necessary.  We should double-check that all the pipeline stages actually ran, though from the point of view of the data transfer script, all of the `desi_night` jobs appear to have run successfully.

In addition, this PR fixes some problems with the pipeline not being *quite* ready for some data model changes implemented for consistency with the quicklook & ICS teams.

Future work will refine this considerably, but for now I think it would be good to get this into master.


sbailey:
I have some changes to request to the new functionality (in particular how the environment is configured when data transfer calls desi_night on edison), but we can merge this now for the updated `get_exposures` and I'll document my requests in a new issue later (today, I hope).

PR #663, opened 2018-07-18
Add row filter capability to database loader

weaverba137:
This PR fixes desihub/desitest#13 by allowing rows to be filtered using an arbitrary function.  Specifically this is to exclude rows in the ZCATALOG file that have TARGETID = -1.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,10 @@ desispec Change Log
 0.22.1 (unreleased)
 -------------------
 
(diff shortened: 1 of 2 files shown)


weaverba137:
I'll need to test this in the notebook before merging, so let me do the merge.

weaverba137:
I've tested this on the minitest data, and the filter works fine.  I'll merge after tests pass.

PR #666, opened 2018-07-20
Small pipeline usability improvements

tskisner:
This work: 

 - Add new "desi_pipe go" command, which attempts to submit
   a full set of jobs to process an entire production.  This
   runs 3 jobs per night and then a final redshift job. The
   dependencies are tracked within a night, and the redshift
   job depends on the calibration jobs from all nights.

 - If --nersc option is not specified, check NERSC_HOST
   environment variable.

 - Improve help message for desi_pipe, to better describe
   commands as "high-level" vs "low-level".

You can now do:
```
$>  desi_pipe create <options>
$>  desi_pipe go
```
and all jobs needed to process the full production will be submitted in the regular queue on the current host.  @sbailey and @julienguy , comments welcome.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,9 @@ desispec Change Log
 0.22.2 (unreleased)
 -------------------
 
-* No changes yet.
+* Adds `desi_pipe go` for production running (PR `#666`_).
+
+.. _`#666`: https://github.com/desihub/desispec/pull/666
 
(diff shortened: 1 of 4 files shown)


sbailey:
Yay, much simpler to run a prod.  Thanks.  Works great for submitting jobs and now I'm in the queue.  Will merge now and open new ticket if something else comes up.

Issue #672, opened 2018-07-31
desispec.io.spectra.write_spectra is not consistent with desispec.pixgroup.SpecLite.write.


weaverba137:
It appears that there are at least two different methods for writing what we would generally consider 1-d spectra:

1. `desispec.io.spectra.write_spectra()`
2. `desispec.pixgroup.SpecLite.write()`

In particular, `write_spectra()` does not apply units (`BUNIT` header) to the inverse variance, while `SpecLite.write()` does. 

Also, neither applies units to the `[BRZ]_RESOLUTION` HDU.  Is the resolution truly dimensionless?  If it's resolution in the delta lambda/lambda sense, then sure.  But I sometimes get that confused with the dispersion in the wavelength solution.

sbailey:
Both `desispec.io.spectra.write_spectra()` and `desispec.pixgroup.SpecLite.write()` should write units.

Context: In practice `desispec.io.spectra.write_spectra()` is not used by the main pipeline but it is kept as an end-user convenience, e.g. if someone wants to read in spectra, filter them, and then write the subset back out.  `desispec.pixgroup.SpecLite.write()` is the memory- and I/O- optimized code for the specific purpose of reading in Frames (not Spectra), regrouping them by healpix, and writing them back out.

Resolution data are unitless in the sense that when you multiply a spectrum by a resolution matrix, the result has the same units as the original spectrum.  In detail they are the the wavelength-dependent convolution kernels used to convolve a high resolution spectrum into a lower-resolution spectrum, stored as the diagonals of the sparse convolution matrices.  They are *not* Gaussian sigmas (e.g. with units of Angstroms or bins) and they are *not* delta lambda / lambda either.

weaverba137:
Closed by #673.

PR #673, opened 2018-07-31
Make sure units are FITS-compliant

### LINKED ISSUE #665: Use FITS-standard flux units
Just looking at the data model, there are a variety of ways that the standard flux units, `10^-17 erg/(s cm2 Angstrom)` are written in FITS files.  We should stick with one standard way to write the flux units **and** make sure that the standard way also complies with the [FITS standard](https://fits.gsfc.nasa.gov/standard30/fits_standard30aa.pdf), which has the bonus of also making the units Astropy-compatible.

These are FITS-standard compatible (based on running, *e.g.*, `f = astropy.units.Unit('10^-17 erg/(s*cm2*Angstrom)', format='fits)`):

* `10^-17 erg/(s*cm^2*Angstrom)`
* `10**-17 erg/(s*cm**2*Angstrom)`
* `10-17 erg/(s cm2 Angstrom)`

These are *not* compatible:

* `1e-17 erg/(s cm2 Angstrom)`
* `10^-17 ergs/(s cm2 Angstrom)`

This is technically compatible, but raises a warning:

* `10-17 erg/s/cm2/Angstrom`

This is compatible but **dangerous** because `A` gets parsed as "ampere" not "Angstrom".

* `10-17 erg/(s cm2 A)`

weaverba137:
This PR closes #665.  I'd like to get an answer to #672 before merging though.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,6 +6,9 @@ desispec Change Log
 -------------------
 
 * Integrates QL S/N QA into offline QA Frame object
+* Ensure that FITS files specify FITS-standard-compliant units (PR `#666`_).
+
+.. _`#666`: https://github.com/desihub/desispec/pull/666
 
 0.23.1 (2018-08-09)
 -------------------
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,17 +5,18 @@ desispec Change Log
 0.23.2 (unreleased)
 -------------------
 
-* Integrates QL S/N QA into offline QA Frame object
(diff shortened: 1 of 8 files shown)


weaverba137:
OK, I still need to update `desispec.io.spectra.write_spectra()` to write units, but that should be quick.

weaverba137:
@sbailey, I've got the final changes in, but let's make sure tests pass.  You might also want to recheck `desispec.io.spectra.write_spectra()`.

PR #675, opened 2018-08-12
Offline S/N QA

profxj:
Starts integrating S/N metrics in offline QA.

Integrates the methods first introduced for QL QA
into the QA Frame object.
Also generates a first plot, but only in Notebook form.

The fitted S/N model on the data looks spurious.  
I ask the QL team to think about this.  Could be my
fault, but I'm not sure how at this stage.

Failing Travis because of a doc string error which
I have no hope of resolving.


profxj:
Here is the Notebook:

https://github.com/desihub/desispec/blob/offline_s2n/doc/nb/QA_S2N_Develop.ipynb

profxj:
Found my error in the fit.  Was running on the frame (i.e. not
sky-subtracted) instead of the cframe.

Notebook has been updated to show a good example with 18.7

rstaten:
@profxj This looks fine to me. One thing to note, in your 'some progress' commit (line 1898 of qa_quicklook.py), we will be removing this if statement involving hard coded params for all QAs. Would you mind providing a list of all QAs where you will call desi_params for offline QAs so that we don't remove something you need?

PR #676, opened 2018-08-14
add ability to choose between using legval cache (default) or no_cache

lastephey:
This pull request corresponds to the changes in specter pull request 70 (https://github.com/desihub/specter/pull/70) for the re-structuring of _xypix. Some minor changes were required here to allow the user to specify if he or she would like the default (to precompute legval values for each patch) or to use the original method (by specifying no_cache).


sbailey:
If caching produces identical results while being 50% faster, I don't think we need to expose that option here.  Just make caching default on the specter side and then desi_extract_spectra becomes 50% faster, yay!  We can leave this open until desihub/specter#70 is resolved, but I'm inclined to think that we don't need this here.

sbailey:
After in person discussion, we're going to handle this on the specter side and not make it an option on the desispec side.  Closing.

Issue #657, opened 2018-07-11
Adding a new QA for check resolution 


Srheft:
Trivia: Quick Look writes the first three coefficients of the Legendre polynomial as the best fit parameters of the ResolutionFit, into the output psf file.

The goal of this proposed QA is to make an assessment of the first attempt to extract resolution (WSIGMA) information from the 1D arc spectra.  The assessment should speak to the stability and consistency of the results coming from the lines, as well as their consistency with established references for these quantities. 

Possible scalar metrics might be:
a) The RMS of each of the three fitting parameters taken over all fibers, OR
b) Number of outliers 2xRMS away from the median value of each parameter 

drill down could be the list of the fiber IDs for those with dramatically different best fit parameters from the norm.

Any suggestions for a more informative quantity for fitting quality assessment than the ones mentioned above, are most welcome.

PR #677, opened 2018-08-16
New QL calibration QAs

rstaten:
This PR includes two new QuickLook QA algorithms developed by @eguzman that examine the final outputs from processing calibration exposures. Specifically, they:
- Add a QA for arc processing that checks 'w' coefficients after resolution fit. These metrics are meant as a 'quick' assessment of whether calibration exposures provide final calibration results - in this case the wavelength resolutions.
- Add a dummy QA for flat processing to validate continuum lamp exposures (to be updated with metrics in a future branch).
- Take monthly psf files as input for resolution fit instead of deprecated bootcalib files.
- Note: Further optimization of the included metrics will benefit from simulations with real world effects and commissioning data.

@Srheft, @rkehoe Please review these edits.


rkehoe:
This is valuable to complete the last QAs for calibration processing.  The update looks good, but I don't favor the use of an extra 'Data' dictionary out of the run method into the plotting.  The coefficients should just be taken by the plotting from the psf object input. 

However, that is under-the-hood and not important at the moment.  I'd rather merge this now and then can clean that up later, maybe when fleshing out the fiber flat metric. 

@Srheft Thanks ahead for following up with your test to a merge.

Srheft:
@rstaten, @eguzman  thanks for your work on this. As discussed internally:

* The currently executed idea for this QA has been suggested in ticket #657  to which no comments were received thus we moved on with place holders.

* This PR adds two QAs called **Check_FiberFlat** with scalar metric called FFLMEAN and **Check_Resolution** with three metrics: Medians, RMS, NBADCOEFFS (see qa_quicklook.py )

* The feasibility of the suggested quantities to be reported in these QAs is not obvious to me/some. The metrics in these QAs will be completed and/or altered upon access to faulty exposures where the traces of the *bad* fibers exhibit measurable characteristic behavior; one of which could be an out-of-norm set of best-fit Legendre polynomial or a flat-out failure to find a solution for the fit, etc. The metric to check the quality of fiber flat also needs a full revision.

Issue #681, opened 2018-08-18
New PSF format fails QL


Srheft:
Using the MockObs format of the desi calibration data (located in $DESI_CCD_CALIBRATION_DATA)
QL fails to process arc exposures. See the running example from 20191001 below:

**time desi_quicklook -i ./config_files/qlconfig_arc.yaml -n 20191001 -c b3 -e 3573 --plots**
QuickLook CRITICAL : Failed to run PA ResolutionFit error was no ysig coefficents were read in the PSF
Traceback (most recent call last):
  ...
  File "/desi/software/desispec/py/desispec/quicklook/qlpsf.py", line 130, in ysigma
    return self.traceset.ysig_vs_wave(ispec,wave)
  File "/desi/software/desispec/py/desispec/xytraceset.py", line 61, in ysig_vs_wave
    raise RuntimeError("no ysig coefficents were read in the PSF")
RuntimeError: no ysig coefficents were read in the PSF
Failed to run PA ResolutionFit

julienguy:
Can you please provide more details for me to reproduce the bug at NERSC.
 - path to the data
 - env. variables
 - code version (master?)

sbailey:
NERSC is down today and tomorrow, but I think the relevant files are also at KPNO.  @Srheft can you provide cut-and-paste instructions for how to reproduce the problem starting with a fresh login of quicklook@desi-1.kpno.noao.edu?  Thanks.

rstaten:
@julienguy QL uses XSIGMA and WSIGMA rather than the XSIG and YSIG HDUs searched for in the read_xytraceset function of io/xytraceset.py. However, these HDUs (XSIGMA and WIGMA) do not contain wavemin and wavemax information, so a failure still occurs in _traceset_from_image even if the correct HDUs are found. If this information is found, QL finishes processing for arcs (and science).

Srheft:
Running quicklook in master:

**Exact command**: 
time desi_quicklook -i qlconfig_arc.yaml -n 20191001 -c b3 -e 3573 --plots

**Exact error**: 

2018-08-20 21:43:46,660 QuickLook INFO : Running ResolutionFit
desispec.psf is DEPRECATED, PLEASE USE desispec.xytraceset
INFO:xytraceset.py:107:read_xytraceset: npix_y=4096
2018-08-20 21:43:47,279 desiutil.log.dlm58.info INFO : npix_y=4096
INFO:xytraceset.py:115:read_xytraceset: psf is a 'psf'
[...]

PR #682, opened 2018-08-22
xytraceset in quicklook

julienguy:
Replace all instances of quicklook.qlpsf.PSF by in quicklook , modify read_xytraceset and add write_xytraceset.
This PR fixes the issue #681 .

### DIFF
--- a/py/desispec/io/__init__.py
+++ b/py/desispec/io/__init__.py
@@ -18,6 +18,7 @@ from .fluxcalibration import (read_stdstar_templates, write_stdstar_models,
                               write_flux_calibration)
 from .spectra import read_spectra, write_spectra, read_frame_as_spectra
 from .frame import read_meta_frame, read_frame, write_frame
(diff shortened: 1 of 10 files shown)


julienguy:
I realize my description of the pull request was not intelligible, sorry for that. Let me try again:
- I replaced everywhere in quicklook the use of the class qlpsf.PSF by the class desispec.xytraceset which is used in the offline pipeline. xytraceset is a container for the trace coordinates and the PSF sigma values.
- I added write_xytraceset in desispec.io.xytraceset
- This is to make sure we use only one set of IO routines for spectral traces in desispec, in order to avoid inconsistencies as reported in issue #681.
- This turned out to touch a lot of files in quicklook.

rstaten:
I have tested this branch on arcs and science and no longer receive the errors we were previously seeing. Thanks for this fix.

sbailey:
Thanks for testing @rstaten.  Merging now.

PR #695, opened 2018-09-13
fix [OII] flux parsing for integration test

sbailey:
Recent desisim PRs updated the simspec format to keep object metadata in different HDUs for different template types, since they have different columns.  This PR fixes the desispec integration tests to look for the OIIFLUX in the new TRUTH_ELG HDU instead of the original TRUTH HDU.

Note: LRGs also have OIIFLUX in their truth metadata, but that isn't used by the integration test so this just checks in the ELG case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,15 +68,15 @@ env:
         - REDROCK_VERSION=0.9.0
         - MAIN_CMD='python setup.py'
         # Additional conda channels to use.
-        - CONDA_CHANNELS="openastronomy"
+        - CONDA_CHANNELS="astropy"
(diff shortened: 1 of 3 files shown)


moustakas:
> Note: LRGs also have OIIFLUX in their truth metadata, but that isn't used by the integration test so this just checks in the ELG case.

FYI -- BGS, LRG, and ELG targets all have `OIIFLUX` (and other emission-line quantities) in their metadata table because those templates are all generated by the `desisim.templates.GALAXY` class, which is subclassed by the individual `ELG`, `LRG`, etc. classes.

PR #696, opened 2018-09-25
Do not compress image masks

### LINKED ISSUE #546: Review how mask HDUs are compressed
In reviewing the data model for release 18.3, I've found several files that have FITS-compressed HDUs.  These HDUs are all masks:

* fiberflat files (EXTNAME = MASK)
* frame files (EXTNAME = MASK)
* spectra-NSIDE files (EXTNAME = `[BRZ]_MASK`)

The compression does not appear to be applied consistently on these files.  In particular, the spectra-NSIDE files may be adding extraneous header keywords that may cause `astropy.io.fits` to interpret the compressed HDUs as compressed *primary* HDUs, instead of compressed *extension* HDUs (see astropy/astropy#7312).

In addition, fiberflat and frame files use `RICE_1`, while spectra-NSIDE files use `GZIP_1`.  Is there a reason for this choice?  All of the masks are 2D images, so the dimensionality is not the reason.

Note, I don't think this issue needs to be resolved for release 18.3, but it is  something that could bite us in the future.

### LINKED ISSUE #603: No compression for mask
having compressed image mask for frame and spectra is an unnecessary complexity for data users that do not use python with astropy.io.fits.

weaverba137:
This PR replaces all instances of `CompImageHDU` with `ImageHDU` (but should not change the compression of raw data).

* Fixes #546 
* Fixes #603


sbailey:
minor: I updated changes.rst to use 0.24.0 as the next tag (instead of 0.23.2) since this is a datamodel change.  ci-skipped.

weaverba137:
@sbailey, will do.

PR #715, opened 2018-11-03
preproc

julienguy:
- add header keywords and comments related to gain values
 - improve bias and dark measurements

### DIFF
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -411,6 +411,8 @@ def preproc(rawimage, header, primary_header, bias=True, dark=True, pixflat=True
     '''
     log=get_logger()
 
+    header = header.copy()
+
     calibration_data = None
     
     if ccd_calibration_filename is not False :
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -565,9 +565,16 @@ def preproc(rawimage, header, primary_header, bias=True, dark=True, pixflat=True
         for j in range(nrows) :
             readnoise[kk][j] = rdnoise[j]
 
-        header['OVERSCN'+amp] = median_overscan
-        header['OBSRDN'+amp] = median_rdnoise
-
+        header['OVERSCN'+amp] = (median_overscan,'ADUs (gain not applied)')
+        if gain != 1 :
+            rdnoise_message = 'electrons (gain is applied)'
+            gain_message    = 'e/ADU (gain applied to image)'
+        else :
+            rdnoise_message = 'ADUs (gain not applied)'
+            gain_message    = 'gain not applied to image'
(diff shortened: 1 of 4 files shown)


julienguy:
Can someone have a look at this matplotlib issue with bootcalib (which has nothing to do with this PR)?

py/desispec/preproc.py
```diff
@@ -565,9 +565,16 @@ def preproc(rawimage, header, primary_header, bias=True, dark=True, pixflat=True
...
-
+        header['OVERSCN'+amp] = (median_overscan,'ADUs (gain not applied)')
+        if gain != 1 :
+            rdnoise_message = 'electrons (gain is applied)'
+            gain_message    = 'e/ADU (gain applied to image)'
+        else :
+            rdnoise_message = 'ADUs (gain not applied)'
+            gain_message    = 'gain not applied to image'
+        header['OBSRDN'+amp] = (median_rdnoise,rdnoise_message)
+        header['GAIN'+amp] = (gain,gain_message)
```
sbailey: This trick didn't work.  You are ending up with files with gains like "(1.0, 'gain not applied to image')" rather than a gain of 1.0 with that comment.  That is causing the Travis test failures like "TypeError: Cannot cast ufunc multiply output from dtype('<U32') to dtype('float64') with casting rule 'same_kind'" which should be reproducible by running `python py/desispec/test/test_preproc.py` on your laptop.

sbailey:
One of your Travis test failures is
```
ImportError: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /home/travis/miniconda/envs/test/lib/python3.5/site-packages/matplotlib/_path.cpython-35m-x86_64-linux-gnu.so)
```
which is suspiciously similar to what @moustakas hit in desihub/desitarget#416.
```
ImportError: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /home/travis/miniconda/envs/test/lib/python3.5/site-packages/scipy/sparse/_sparsetools.cpython-35m-x86_64-linux-gnu.so)
```
Let's keep the conversation over there, and then apply whatever fix we find here.  It certainly looks like a new bug from a change to the Travis environment and/or conda and/or a pip package.

py/desispec/preproc.py
julienguy: It works fine when preprocessing images. The header are correct, with values and comments.

The issue was in the test itself that runs several times preproc(..., self.header, ....)
Gains tuples (value,message) are set at the first run, then used in the second and that's where the code fails because it expects a scalar and not a tuple.

I fixed that by forcing a copy of the input header the preproc routine.

Issue #679, opened 2018-08-17
Update fibermap to match targeting inputs


sbailey:
Related to desihub/desidatamodel#41 (datamodel consistency), but opening a self-assigned desispec-specific issue to track this: update fibermap to use a structure that matches input targeting, and fix code like flux calibration that will break as a result (e.g. MAG[i] -> FLUX_G/R/Z)

sbailey:
Changes implemented in PR #717 for consistency with target catalogs and fiber assignment:

## Unchanged columns
```
TARGETID, DESI_TARGET, BGS_TARGET, MWS_TARGET
FIBER, PETAL_LOC, DEVICE_LOC, LOCATION
LAMBDAREF
SPECTROID
BRICKNAME
```

## Columns that changed name or meaning
```
Old                     ->  New
OBJTYPE (ELG, LRG, ...)     OBJTYPE  (TGT, SKY, BAD)
RA_TARGET, DEC_TARGET       TARGET_RA, TARGET_DEC
MAG, FILTER                 FLUX_G, FLUX_R, FLUX_Z, FLUX_W1, FLUX_W2
RA_OBS, DEC_OBS             FIBER_RA, FIBER_DEC
X_TARGET, Y_TARGET          DESIGN_X, DESIGN_Y
X_FVCOBS, Y_FVCOBS          DELTA_X, DELTA_Y (wrt intended position)
                            --> should have been DELTA_XFPA
X_FVCERR, Y_FVCERR          DELTA_X_IVAR, DELTA_Y_IVAR
POSITIONER                  DEVICE_LOC
TARGETCAT                   (dropped)
```

## New Columns propagated from tractor catalogs
```
BRICKID (int)
TARGET_RA_IVAR, TARGET_DEC_IVAR  (originally RA_IVAR, DEC_IVAR)
BRICK_OBJID (originally OBJID from Tractor)
MORPHTYPE (originally TYPE from Tractor)
REF_ID, PMRA, PMDEC, PMRA_IVAR, PMDEC_IVAR (proper motion info)
FLUX_IVAR_[G,R,Z,W1,W2]  (flux inverse variances)
FIBERFLUX_[G,R,Z,W1,W2]  (model flux in fiber)
FIBERTOTFLUX_[G,R,Z,W1,W2]  (model+overlapping models flux in fiber)
MWTRANSMISSION_[G,R,Z] (Milky Way dust transmission, 0-1)
EBV (galactic dust reddening E(B-V))
[...]

PR #717, opened 2018-11-06
Update to new fibermap format

sbailey:
This PR is the long awaited update to the new fibermap format.  Most changes are harmless renames (`RA_TARGET` -> `TARGET_RA`), but there are two substantive changes that required algorithm updates:
* switching from `MAG[5]` to `FLUX_G`, `FLUX_R`, `FLUX_Z`, `FLUX_W1`, `FLUX_W2`
* `OBJTYPE` values are now just `TGT`, `SKY`, and `BAD`, instead of `ELG`, `LRG`, `QSO`, etc. to avoid ambiguity when a given target passes more than one cut.  Use `DESI_TARGET` and `desitarget.targetmask.desi_mask` instead.

A reference run is available at `/global/cscratch1/sd/sjbailey/desi/dev/end2end/`

Companion PRs for desisim and desitest will be submitted shortly.  I will updated issue #679 will all the fibermap column changes.

@julienguy please evaluate
* the flux calibration updates
* redshift success rates / meeting requirements
* whether you agree with the functionality tradeoffs in the caveats below

Now for the caveats:
* With sincere apologies, this breaks offline QA and quicklook due to the `OBJTYPE` change.  I'm aware that I'm breaking my own rules with this.  I tried to fix this in this branch, but other QL and desisim changes were coming in so fast that I couldn't keep up, and we ended up in a state where desitarget+fiberassign master are out of sync with desisim+desispec master for the core end-to-end integration.  Further postponing this fibermap update was becoming more and more problematic, so I suggest proceeding with this now and then fixing the broken pieces in separate smaller pull requests.
* redshift database loading is broken.  I have some updates from Ben that were working on an earlier version of this, but this branch is still broken.  Again, let's fix that as a separate smaller PR instead of postponing the core of this one.
* I'm sure that more detailed review will reveal other updates that need to be made.


sbailey:
The minitest run for this branch + desisim/newfibermap-redo + master of everything else is available at: https://github.com/desihub/desitest/blob/newfibermap-redo/mini/minitest.ipynb .  A few things to note there:
* ELG global redshift efficiency has dropped, but I think that is due to the newer mocks having fainter ELGs.  The offline with-truth QA with [OII] flux cuts still reports excellent redshift success.  Other classes still have good redshift success.
* database loading and offline no-truth QA are broken as previously noted.
* there is an odd gap in the QSO tracer redshift distribution; I think that is a separate issue with mocks / target selection and unreleased to these PRs.

sbailey:
Oops: output files have both `LAMBDAREF` and `LAMBDA_REF`

julienguy:
Notes on flux calibration
 - I checked the calibration vectors are similar to those of the redwood prod (so no big issue).
 - I note you do not redden the stellar models prior to the fit, I am not sure why. It seems to me to be less accurate. Given the fact the median filtering continuum subtraction is not perfect, I perform twice the calibration to avoid the median filtering potential bias. Not applying the dust extinction to the model prior to the fit removes some of the benefit of this.
 - We may merge this PR as is for the sake of efficiency, given it addresses so many other issues, but we should then open a ticket to deal with this later.

julienguy:
Will need to update desi_zcatalog --match option to collect the information from the truth present in the various HDUs ...

julienguy:
Commenting on my previous comment. Not applying extinction to the stellar models prior to the fit is probably not an issue at all, because the second fit, which includes a first estimation of the calibration, will absorb this extinction.

julienguy:
The ELG redshift efficiency is similar to redwood.
elg-eff

sbailey:
@weaverba137 can you tell what is wrong with sphinx for this PR?  Thanks.

Issue #718, opened 2018-11-07
Reference flux calibration vectors


julienguy:
- Define data model for reference calibration vector (including airmass term). Data will be part of the calibration SVN directory (currently pointed to by DESI_CCD_CALIBRATION_DATA).
- Write code to average flux calibration from previous observations and fit for the airmass term. 
- Later: develop a model that separate atmospheric variations from stable instrument throughput.
- This is needed for Quicklook.

julienguy:
Implemented in branch average_flux_calib

The script ```desi_average_flux_calibration``` reads a series of ```calib-*.fits``` files from the standard pipeline processing and computes:
 - an average high resolution calibration vector
 - a low resolution (i.e. smooth) seeing dependent term
 - a low resolution (i.e. smooth) atmospheric extinction, except in the wavelength range where telluric features are expected,  where the model is high resolution.

The seeing dependent term and atmospheric extinction are in the form of magnitude:
```cal(lambda) = average_cal(lambda) * 10^( -0.4*(   (seeing-pivot_seeing)*seeing_term(lambda) + (airmass-pivot_airmass)*atm_ext(lambda) ))``` 

Example for camera z3, based on one month of data from the DC17b data challenge
```
desi_average_flux_calibration -i /project/projectdirs/desi/datachallenge/dc17b/spectro/redux/dc17b/exposures/*/*/calib-z3*fits -o calib-z3-dc17b.fits --plot
```
Mean calibrations derived from the DC17b data challenge:

b3 camera
![mean-calib-b3-dc17b](https://user-images.githubusercontent.com/5192160/49107693-46d29c00-f23b-11e8-9c19-277b0486d66f.png)

r3 camera
![mean-calib-r3-dc17b](https://user-images.githubusercontent.com/5192160/49107704-4b975000-f23b-11e8-9e9d-b8c6c64ecd40.png)

z3 camera
![mean-calib-z3-dc17b](https://user-images.githubusercontent.com/5192160/49107710-4f2ad700-f23b-11e8-9f80-09c1a863b36e.png)

julienguy:
Draft data model in https://github.com/desihub/desidatamodel/blob/average_flux_calib/doc/DESI_SPECTRO_CALIB/fluxcalib-CAMERA.rst

julienguy:
Calibration is saved in `$DESI_CCD_CALIBRATION` data and can be used with `desi_qproc` (for the moment in branch average_flux_calib).
Example
[...]

PR #713, opened 2018-10-31
Ql qproc - wrap up of integration of QP reduction algorithms and Qframe into QuickLook

rkehoe:
This PR contains the last items to finish the integration of the QProc data reduction algorithms into QuickLook.  Previously, these algorithms were integrated, but were not run by default.  To complete
this work, this PR includes the following:

- conversion of qlconfig files to utilize the QProc extraction, flat, sky subtraction reductions
- reversion of prior QL reductions (extraction, flat, sky subtraction) to run as non-defaults.  This will allow continued testing and comparison of the two sets of reductions (see note below on cosmics removal).
- reliance on qframe object for default QL running.

There were a couple fixes, including a modification to the Extract_QP and ApplyFiberFlat_QP code writing out the qframe file (i.e. io.write_qframe(dumpfile...) --> write_qframe(dumpfile....); as at current lines 944 and 1023  of quicklook/procalgs.py).

@felipelm This PR is now ready for testing with QLF, at least for science and flat exposures.  We have a modification yet coming for arcs, but please don't feel the need to wait for that.  All of the changes are under-the-hood, so you should see no difference in the outputs from QL, although numerical values will change.  This might change a bit the red petals, so we'll need to track that down if observed.  You will notice the pipeline executes considerably faster, however.  Can you do your normal QL/F testing to ensure no breakage has occurred, in preparation for merging this PR?

@julienguy Following some tests of QL and QP reductions, we have noticed at least one difference worth exploring.  In both versions of the pipeline, the masking of cosmic rays is performed in the extraction pipeline algorithm step.  Currently, applying the mask is coded differently in the original QL code, than for the QP code.  An example line location, after subtraction, is shown here.

qlnqpcosmics

Not all line locations look like this, but there seems to a systematic oversubtraction in both cases, with significantly more over subtraction from the QP removal version.  Is this behavior expected?  If further tests continue to show this, we could do an initial update to improve on it.

PR #720, opened 2018-11-09
QL Qproc related cleanup

rkehoe:
This PR involves items related to the migration of some QL data reduction algorithms to use the
qproc reductions.  Specifically,

- changing the name of the HDU provided in the qframe output data model formatted for use in QL.  This HDU was called 'SIGMA' when it specifically refers to the line spread function in the wavelength direction.  Following an exchange with @jguy, this will be renamed to 'YSIGMA'  in the code, and also in the qframe data model documentation.  Also, this HDU is being dropped in the final cframe file, and will be retained by this PR.
- We observed that the PSF HDU is being inadvertently dropped at the extraction pipeline step, and will include it in the final qframe output for completeness.
- As was observed in PR #713, there appears to be a difference in the cosmic removal effectiveness when comparing the qproc based algorithm with the original QL based one.  While we are now using qproc boxcar extraction, we want to retain the best cosmic rejection possible, so we will study if the apparent difference is systematic, and make an adjustment to cosmic removal accordingly.

@felipelm, @sbailey and other users of QL, this PR should not produce any changes that current users will notice in the QL outputs.  Down the road, those accessing the cframe fits file may want to look at the resolution information being stored, but that is not being looked at by users now.  None of the changes in this PR corresponds to a change of datamodel.  This PR should also not be impacted by the pending update to the new fiber map file, or the standard flux calibration vector input related to Issue #718.

### DIFF
--- a/py/desispec/qproc/io.py
+++ b/py/desispec/qproc/io.py
@@ -67,7 +67,7 @@ def write_qframe(outfile, qframe, header=None, fibermap=None, units=None):
 
     if qframe.sigma is None :
         qframe.sigma=np.zeros(qframe.flux.shape,dtype=np.float)
-    hdus.append( fits.ImageHDU(qframe.sigma.astype('f4'), name='SIGMA') )
+    hdus.append( fits.ImageHDU(qframe.sigma.astype('f4'), name='YSIGMA') )
 
     hdus.append( fits.ImageHDU(qframe.wave.astype('f8'), name='WAVELENGTH') )
     hdus[-1].header['BUNIT'] = 'Angstrom'


rkehoe:
@sbailey. I had not realized you might merge.  This PR was still in implementation — the comments were up front as asked for.  I will be more clear on this going forward.  Bob

sbailey:
@rkehoe I misinterpreted; I thought this was a clean set of under-the-hood changes that were ready to merge.  Please start another branch to continue the work.

When you open a new pull request, please use the WIP label (work in progress) to flag something that isn't ready to merge and then remove that label and say something like "ready for final review and merge" when ready.  Thanks.

rkehoe:
No problem @sbailey.  I should have tagged with WIP, as I did with our other branch.  Won’t make that mistake again.  It’s okay, the branch was working and so master is not screwed up.  Two notes for the moment, though.  

-This branch included integration of the new fibermap, which was not as intended in my original points.  This does change the outputs for QLF a bit, so we will document those in the follow up cleanup branch to this one.
- I did not want to keep QLF developers waiting as we piled up changes, so @rstaten has gone ahead and modified the flux calibration module to give a warning when the input flux calibration file is missing, and use a placeholder in it’s place for now.  This will allow them, once we merge the cleanup PR, to test QL and not be stopped by the flux calibration error, so they can accomodate changed metric names, etc from the fibermap change.

I will itemize all of this when the cleanup PR is done.

felipelm:
> I did not want to keep QLF developers waiting as we piled up changes, so @rstaten has gone ahead and modified the flux calibration module to give a warning when the input flux calibration file is missing, and use a placeholder in it’s place for now. This will allow them, once we merge the cleanup PR, to test QL and not be stopped by the flux calibration error, so they can accomodate changed metric names, etc from the fibermap change.

Great! Let me know when we can test and integrate.

PR #723, opened 2018-11-13
fixes desi_zcatalog --fibermap for new format fibermap

sbailey:
This PR fixes desihub/desisim#448, where `desi_zcatalog --fibermap` didn't work for the new format fibermap.  This implementation works on zbest files with both old and new format fibermap tables, and adds columns `RA`, `DEC`, `FLUX_G`, `FLUX_R`, `FLUX_Z` in both cases, even though the old version used to write `MAG` instead.

Unlike the discussion in desihub/desisim#448 this morning, I actually have tested it this time, using the 18.7 (old) and 18.11 (new) reference runs in /project/projectdirs//desi/datachallenge/reference_runs/ .

@tetourneau please take a look if you'd like.  Otherwise I'll merge tomorrow morning.

### DIFF
--- a/bin/desi_zcatalog
+++ b/bin/desi_zcatalog
@@ -88,7 +88,13 @@ for zbestfile in zbestfiles:
     if args.fibermap :
         fibermap = fitsio.read(zbestfile, 'FIBERMAP')
         # new zbest structured array with two more columns ...
-        ndtype=np.dtype( zbest.dtype.descr + [("RA",str(fibermap["RA_TARGET"].dtype)),("DEC",str(fibermap["DEC_TARGET"].dtype)),("MAG",str(fibermap["MAG"].dtype),fibermap["MAG"].shape[-1])])
+        ndtype=np.dtype( zbest.dtype.descr + [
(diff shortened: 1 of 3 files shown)


TEtourneau:
I tested desi_zcatalog --fibermap, and it works.
Thanks for fixing the bug!

Just a practical question, what's the best way to use different branchs in the desi environment ?
To use desisim and desispec, I use this command:
source /project/projectdirs/desi/software/desi_environment.sh master
But I didn't know how to test desi_zcatalog from the branch target_ra, so I just git clone in my home directory desispec and switched to the branch target_ra.

sbailey:
Thanks for testing @TEtourneau .  Here's what I do for testing branches:
```
#- Use pre-installed master versions of all packages (updated nightly)
source /project/projectdirs/desi/software/desi_environment.sh master

#- Remove desispec/master from my path
module unload desispec

#- Replace it with my copy
cd $SCRATCH/desi/code/desispec
git checkout target_ra
export PATH=$(pwd)/bin:$PATH
```

I've actually wrapped this in my own module file so that I can `module swap desispec/sjb` to replace the standard version with my own, but it is equivalent to the above.

PR #732, opened 2018-11-16
Additional transfer scripts

weaverba137:
This PR:

* Adds a script, `desi_kpno_transfer.sh` for copying arbitrary data from KPNO to NERSC.  In particular, data that does not follow the NIGHT/EXPID model used by the raw data.
* Adds a script `wrap_desi_night.sh` that sets up the DESI environment properly when the raw data transfer script activates the pipeline via ssh.
* Substantial changes to `desi_dts.sh`:
   - Added a kill switch.  The existence of `${HOME}/stop_dts` will shut down the data transfer daemon and prevent it from restarting.
   - Allow for the possibility of transferring additional data sets that do follow the NIGHT/EXPID model.

Look around, kick the tires, but don't merge yet.  Some things that need to be done before merging:

- [ ] Update change log.
- [ ] Move completed exposures into `DESI_SPECTRO_DATA`, instead of symlinking them.
- [ ] Clean up configuration left over from tests this week.


weaverba137:
OK, I think this is ready now, but please review closely.

bin/desi_init_dts.sh
```diff
@@ -3,6 +3,10 @@
 # Sean McManus 2015/11/17
 # Adapted for DESI: Benjamin Weaver 2018-0627
 #
+# DESI Environment
+source /global/common/software/desi/desi_environment.sh test-release
+module use -a ${HOME}/software/${NERSC_HOST}/modulefiles
+module load desispec/my-master
```
sbailey: Do we need "test-release", "${HOME}/software/${NERSC_HOST}/modulefiles", and "my-master", or can that be moved to a more standard release / tag?

weaverba137: Is "master" an acceptable release?

weaverba137: Actually, how about if I leave the release blank, so it will pick up whatever the default release on datatran is?

sbailey: master is fine, or using datatran default.

PR #725, opened 2018-11-13
QL under-the-hood updates/cleanup

rkehoe:
This PR follows on PR #720 to implement qproc-related updates.  These will include:

- use qframe in flux_calibration step — i.e. Sample calibration vector to individual fiber native grids
- complete process of propagating new fibermap input names to QAs and, where relevant, to QA outputs.

There are several code cleanups also.  When this PR is ready, I will itemize the differences that will be evident to the user (eg. QLF).

Issue #698, opened 2018-10-11
Redwood fibermap vs Mockobs Fibermap


Srheft:
Quicklook finds all the keys it needs in both fibermap formats but switching from one fibermap format to the other would result in "UNKNOWN" status for SNR. Fixing it for one format would result in its happening for the other format. 

The reason lies in OBJLIST=['SCIENCE','STD'] versus OBJLIST={'QSO','LRG','ELG','STD'] difference between theses two fibermaps and the fact that QL takes reference values for fitting results from the configuration file. If there are two target type, QL requires two references and if there are four target types QL requires four ref values. Which OBJLIST is the new format to which QL should conform?

Srheft:
Also, what prevents image.meta to include 'PROGRAM' as a key fot the science exposure?

 QL doesn't read in the raw exposure in QAs but it needs to know what program is being processed to use the program-specifc reference values from the configuration file.

rkehoe:
@srheft, @sbailey, @julienguy  This issue was accommodated in PR 720, and final propagation of relevant parameters into QL algorithms to their outputs will be completed in PR #725.  I believe we can close this Issue when that PR is successfully merged, hopefully early next week.

Srheft:
@rkehoe  ok. let's keep this open until PR725 is merged into master. I think it would be nice for documentation purposes to include the resulting SNR plot here given the recent changes and discuss any unanswered questions.

rstaten:
@Srheft Here is a current SNR plot based on PR 725.
![screen shot 2018-11-21 at 12 32 12 pm](https://user-images.githubusercontent.com/15023569/48861719-f3b79f80-ed89-11e8-82df-9d99eeae4d99.png)

Srheft:
@rstaten Thank you. After the use of flux calibrated frames is implemented in SNRcalc,  given that we observed a bit of an "off" fit when combining all the targets as 'SCIENCE' , it is constructive to see how much of a better fit we get for all 'SCINECE' targets [to know the target type, one has to utilize targetting bits and not sure QL wants/has to/can do that]. Testing whether avoiding the scipy fits brings about any improvement to the fit is also beneficial.

rkehoe:
[...]

Issue #685, opened 2018-08-22
Use the QFrame object in quick look pipeline


julienguy:
- The quick look pipeline performs a boxcar extraction giving one flux value per CCD row per fiber.
- The natural output from this algorithm is a QFrame (see https://desidatamodel.readthedocs.io/en/latest/DESI_SPECTRO_REDUX/SPECPROD/exposures/NIGHT/EXPID/qframe-CAMERA-EXPID.html).
- However the quick look pipeline currently uses internally a Frame object (see https://desidatamodel.readthedocs.io/en/latest/DESI_SPECTRO_REDUX/SPECPROD/exposures/NIGHT/EXPID/frame-CAMERA-EXPID.html). The drawbacks of this approach are: 
  - a ill-defined, made-up resolution matrix
  - correlated fluxes due to the resampling of fiber spectra to the same wavelength grid
  - CPU/mem. inefficiencies because of all of this
  - an unnecessary and misleading modification to the Frame data model with an optional wsigma HDU, and a dependency in the Frame class to the QuickResolution function, see https://github.com/desihub/desispec/blob/42982e43bab9e5f085f40cd786a2428fed5655ee/py/desispec/frame.py#L132
 
It would be better to use the QFrame object in quick look.

julienguy:
Yes. sigma is the LSF sigma in pixel units. This is described in the docstring of the init function and in the data model https://desidatamodel.readthedocs.io/en/latest/DESI_SPECTRO_REDUX/SPECPROD/exposures/NIGHT/EXPID/qframe-CAMERA-EXPID.html . But, I admit a more explicit name for this parameter might be better. Feel free to change this.

julienguy:
yes, and please the datamodel accordingly

rkehoe:
@julienguy, @sbailey. Now that PR #734 is merged, I would like to close this Issue later today.  Let me know if you prefer to keep this open and what question you feel has yet to be resolved.

PR #734, opened 2018-11-20
QL Cosmics

rkehoe:
This PR will attempt to improve on cosmic ray removal when running qproc reductions.  It will also upgrade the XWSigma QA to more carefully handle poorly fit lines, including those impacted by cosmics.  A more specific feature list will be itemized as it is pushed here.


rkehoe:
I’ve changed this PR’s course to focus on addressing the identified issue of QuickLook use of speclite, which occurs in order to calculate a spectral magnitude using broadband filter response.  This PR addresses a problem by which prior attempts to access the input response file would crash, as identified in PR #725.  This has now been fixed using the first approach suggested by @sbailey.  Specifically, this PR contains the following updates:

- use of resource_file to access the speclite response file.  An error is reported if the file is not found.  
- an update to the XWSigma QA to cap the width of line fits.

This PR was tested with QuickLook standalone running, and running via QLF.  In both cases, running with master produced the error, and running with this PR avoided it.  Thanks @rstaten for testing this!

@felipelm. This PR should produce no difference in output relative to what was observed in PR #725.  Please have a look at it also.  If you confirm all is in order, I will merge this PR.  (We will punt the cosmics update to a later update.)

rkehoe:
@felipelm.  I forgot to mention the tag versions that this PR QuickLook was tested with.  Since different code versions did show up recently, maybe this will be useful:

desiutil 1.9.14
desitarget 025.0
desisim 0.31.0
desimodel 0.9.9
specter 0.9.1
speclite v0.8
qlf master

@rstaten sent a version of the mergedQA on the ql-qlf-int mailing list in case you want to make a comparison.

felipelm:
Thanks for the dependencies list @rkehoe was really helpful.

Science, arc and flats ran fine but I miss the OBJTYPE in the mergedQA. Was it moved somewhere else?

`TASKS -> CHECK_SPECTRA -> METRICS -> OBJLIST`

This breaks the science SNR plot.

rkehoe:
@felipelm Sure, no problem. 

I think your comment on the 'OBJTYPE' is that the OBJLIST is missing from recent QL's, correct?  The change happened already in master, and shouldn't have.  We're improving our tests with QLF to check more thoroughly so we catch this right away.  We should have some more flexible tools/scripts by which we can easily compare a current JSON with a reference one for the sake of debugging.  I think this will help with anybody's use/testing of this system.  @rstaten and I will take a start, but I'll draft a ticket on this once we get past the current items to field comments.

In the meantime, @rstaten will update the PR for the missing list soon.

rkehoe:
Thanks @felipelm.  I will merge this.

@sbailey, @srheft, @julienguy This PR is the last loose end from the integration of qframe and the new fibermap file.  As such, Issue #685 and Issue #698 can now be closed.

PR #736, opened 2018-12-02
use new desiutil.dust module during flux-calibration

moustakas:
Use the new dust module implemented in https://github.com/desihub/desiutil/pull/128 during flux-calibration.


moustakas:
Pinging @julienguy on this very simple PR, or maybe @sbailey if Julien is underwater with other work.

sbailey:
I updated it to be more selective about error catching and print an ERROR instead of a WARNING, but otherwise looks good.  OK to merge once tests pass again.

PR #737, opened 2018-12-07
Refactor QA Offline for targeting bits

profxj:
Fussed with offline S/N QA to handle new data model
for QA metrics (which already accommodates new 
`DESI_TARGET` bits in `SNRFit`.

Now back where we were.  Example QA using 18.11 is attached.

exposure_s2n

qa-s2n-b1-00000023

Other QA appears to have been unaffected.


sbailey:
This looks great.  Let's merge after Travis tests are done.  Thanks.

profxj:
Passed.  Merging.

Issue #739, opened 2018-12-07
QL refactor for configurable plotting


rkehoe:
Currently, QuickLook for DESI employs several QA algorithms to measure the data performance and assess data quality.  Each of these algorithms has several metrics as outputs, and these metrics change as the needs of the software or relevant studies have changed.  This need to add to/remove and modify these metrics will expand dramatically during Commissioning and subsequent phases of early data taking, and must be an efficiently achievable process.

In QL, these metrics are plotted by dedicated functions associated with each QA.  These are static/noninteractive plots that are used for testing or any application of QL that does not require use of the Online system’s QL Framework.  This has worked well for such exercises, however there is a significant development bottleneck that comes from the fact that such plots are effectively hard coded in Python.  Therefore, if one wants to add a new metric to QL, which is easily achieved by adding the calculation to the most relevant QA (or writing a new one and adding it to the qlconfiguration), one still also has to modify the plotting functions to provide the plotting capability.  This is at least annoying and wastes development time uselessly.  More importantly, it presents more potential points of failure by not reusing code.  In most respects, the number of different _kinds_ of plots, for instance, is very few, while different metrics will be involved in different instances of each of these kinds. An analogous problem also happens with QLF, which must currently hardcode which metrics to plot and how to make plots, and this results in significant brittleness when updating QL and QLF together with, say, a new or changed metric.

In order to address this in the QL domain, I.e. the efficient production of static plots, we require a ‘plotting configuration’ for QL.  Such a capability would abstract the plotting code to a few generic kinds, which are then called for specific instances and metrics arguments based on concise description of the plot to be made.  For instance, there are several plots which present a 2x2 grid of squares containing in the z-axis the values of metrics assessed per CCD amp.  These include noise level, or bias level.  Give this kind of plot a name, say PATCH, then one can describe generically what to plot and how as:

	noise_per_amp: {TYPE: PATCH, VALS: NOISE_AMP, TITLE: Noise in each CCD amp, 
		GRID: [2,2]}

Where TYPE indicates what kind of plot, VALS indicates what metric or quantity to plot, and GRID indicates the size of the grid of patches (here 2x2).

A more typical plot would be a 2D plot, such as of sky continuum vs. fiber:

    skycont: {TYPE: 2DPLOT, XVALS: SKYFIBERID, YVALS: SKYCONT_FIBER, 
	TITLE: Sky Continua vs. Fiber, XTITLE: Fiber ID, YTITLE: Sky continuum (counts), 	
[...]

PR #740, opened 2018-12-07
QL Plot config

rkehoe:
This PR is focused on refactoring QL so that the making of static plots is more developmentally efficient, as described in Issue #739.  As such, it can also facilitate the ability to promptly observe plots in QLF as soon as QL is updated.

- a new plotting configuration file that describes those that are meant for the QL static plots.  This file includes rudimentary layout information, which we ignore here to focus on the plots first.  Currently has descriptions of several example plots, previously hard coded in QL.

- implementation of one or more examples for three plot types: PATCH, 2DPLOT and 3DPLOT.  (See below)

- includes one example where a metric has been changed and it appears with only changes to configuration, not plotting code.

In addition to these, some refactoring has been done to move the information in the GENERAL_INFO portion of the normal mergedQA structure so that it is gathered at the beginning of the pipeline.  This is one necessary step to generalizing QL so all plotting is configurable.

@dschlegel, @julienguy, @sbailey, @felipelm. Please find here examples of different plots that are now created by the configuration.  This is in qlconfig_plots.yaml currently:

https://github.com/desihub/desispec/blob/ql_calibs/py/desispec/data/quicklook/qlconfig_plots.yaml

This configuration produces the following plots:

PATCH plot:

1da89327-c0c6-4c0e-a66c-d6a14628599a

2DPLOT:

6d0a396d-d3fe-4feb-8b99-7f03da0fb260

3DPLOT:

e162fccf-2cd2-41eb-8667-fb547d6c10ae

It is worth noting that this 3DPLOT was never implemented in the old list of plots and is a new plot enabled by configuration.

An important part of the function of this PR is to be able to make a modification to QL QAs and be able to see the plot promptly without making separate modification to the plotting code.  Using master, the output of the NOISE metrics comes out as:

5075f4b3-9981-4783-9c4a-23f69f693924

In this PR, a new dummy metric was created called SIGMA_AMP which equals the square root of 
The noise per amp.  This metric did not exist before, is included in the METRICS block of the QA.
The plotting function finds this metric, and the instructions to plot if from the configuration, and adds it to the plots for this QA:

635d8e8d-025e-4b7c-bd45-24b2f54d53fa

Currently, the implementation is not generalized all QAs, or all metrics in a QA.  Completing this transition is not difficult, but it is best if there is consensus on the manner of this configuration or syntax before proceeding.  So please provide your questions or comments here.  In particular @felipelm, it will be best for the same plotting configuration to be used for both QL in standalone mode, and by QLF.  If that is acceptable, then this can be implemented in QLF as well.

### DIFF
--- a/py/desispec/data/quicklook/qlconfig_arc.yaml
+++ b/py/desispec/data/quicklook/qlconfig_arc.yaml
@@ -16,6 +16,7 @@ Timeout: 120.0
 Pipeline: [Initialize, Preproc, Flexure, BoxcarExtract, ResolutionFit]
 Algorithms:
     Initialize:
+        PEAKS: {B_PEAKS: [4047.7,4078.8,4359.6,4801.2,5087.2,5426.2], R_PEAKS: [6144.7,6404.0,6508.3,6931.4,7034.3,7247.1], Z_PEAKS: [8302.6,8379.9,8497.7,8856.3,9151.1,9668.0]}
         QA: 
             Check_HDUs:
                 PARAMS: {}
--- a/py/desispec/data/quicklook/qlconfig_plots.yaml
+++ b/py/desispec/data/quicklook/qlconfig_plots.yaml
@@ -2,7 +2,8 @@
 PAGE1:
     Title: Get RMS metrics after Preprocessing
     noise_per_amp: {TYPE: PATCH, XVALS: NOISE_AMP, TITLE: Noise per amp, GRID: [2,2]}
-    overscan_noise_per_amp: {TYPE: PATCH, XVALS: NOISE_AMP, TITLE: Overscan Noise per amp, GRID: [2,2]}
+    overscan_noise_per_amp: {TYPE: PATCH, XVALS: NOISE_OVERSCAN_AMP, TITLE: Overscan Noise per amp, GRID: [2,2]}
+    sigma_amp: {TYPE: PATCH, XVALS: SIGMA_AMP, TITLE: Sigma amp, GRID: [2,2]}
 
 PAGE2:
     Title: Bias from Overscan metrics after Preprocessing
--- a/py/desispec/data/quicklook/qlconfig_plots.yaml
+++ b/py/desispec/data/quicklook/qlconfig_plots.yaml
@@ -1,15 +1,13 @@
 # Example Configuration to plot desired metrics
-#- Metrics to be plotted: check qa_quicklook.py, qalib.py, or output jsons for available metrics
-METRICS: NOISE_AMP
-#- Type of plot: AMP (per amp plots), FIBER (per fiber plots), PETAL (focal plane plot per camera)
-TYPE: AMP
-#- Title of plot/x & y variables
-TITLE: RMS Noise per amp
-XTITLE: None
-YTITLE: None
-#- Does plot require heatmap?
-HEATMAP: True
-#- Limit x & y ranges if applicable
-XLIMIT: None
-YLIMIT: None
+PAGE1:
+    noise_per_amp: {TYPE: PATCH, XVALS: NOISE_AMP, TITLE: Noise per amp, GRID: [2,2]}
+
+PAGE2:
+    bias_per_amp: {TYPE: PATCH, XVALS: BIAS_AMP, TITLE: Bias per amp, GRID: [2,2]}
+
+PAGE3:
+    skycont: {TYPE: 2DPLOT, XVALS: SKYFIBERID, YVALS: SKYCONT_FIBER, TITLE: Sky Continua, XTITLE: Fiber ID, YTITLE: Sky continuum (counts), YRANGE: [140,180]}
+
+PAGE4:
+    specmag_petal: {TYPE: 3DPLOT, XVALS: RA, YVALS: DEC, ZVALS: SPEC_MAGS, TITLE: Spectral Magnitudes on the Focal Plane, XTITLE: RA, YTITLE: DEC, HEAT: Blues}
 
--- a/py/desispec/data/quicklook/qlconfig_plots.yaml
+++ b/py/desispec/data/quicklook/qlconfig_plots.yaml
@@ -1,13 +1,18 @@
 # Example Configuration to plot desired metrics
 PAGE1:
+    Title: Get RMS metrics after Preprocessing
     noise_per_amp: {TYPE: PATCH, XVALS: NOISE_AMP, TITLE: Noise per amp, GRID: [2,2]}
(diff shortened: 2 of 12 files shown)


felipelm:
Looks good to me I would add the full path to the metric in the mergedQA e.g. `TASKS->CHECK_SPECTRA->METRICS->PEAKCOUNT`

sbailey:
It looks like `plotconf` has become a required parameter for (almost?) all plot functions in qa_plots_ql.py, even for those that don't use it at all.  For those functions that do use it, I see code wrapped in `if plotconf:` implying that it wasn't intended to be required.  Could you make it optional via
```python
def plot_bias_overscan(qa_dict,outfile,plotconf=None):
...
if plotconf:
    ...
```
Please also add docstrings explaining what `plotconf` is and where to find an example of valid keys/values.

rkehoe:
@felipelm. Glad to hear it.  Next steps will be to generalize to all QL QA plotting.

About using a pathname, this could be necessary if metric names are duplicated across QAs.  However, we have all agreed along time ago to ensure this never happens.  For instance, only one QA calculates sky continuum, and all it’s variants.  In QL right now, it is sufficient just to find the metric in the QA dictionary overall.  Do you think this will create a noticeable performance issue?

The other issue here will be when we discuss layout.  To make the plotconfig suitably general, one will need to be able to specify the ‘steps’ or ‘tasks’ by which to group metrics for the display.  In QL, this is still currently by QA/PA step, although we can update this to mimic QLF’s grouping according to tasks (‘CHECK_SPECTRA’, ‘CHECK_FIBERS’, etc).  But one can imagine during the height of commissioning different, or additional, breakdowns.  Eg.  One might want to add a TEST task, or maybe regroup by observable like CHECK_SKY, CHECK_GALS, CHECK_STARS.  One would need to support such structures regardless of exactly how the information is stored in the mergedQA.  Alot will change between now and the end of the year.  So it would seem that requiring the path may inadvertently perpetuate a portion of the issue we want to solve.  It would be good to have your comment on this and whether you feel this is a critical modification needed now.

rkehoe:
@sbailey  Thanks for the suggestion.  

The implementation at this moment was not complete, just enough to demonstrate the ideas mentioned.  (Thanks for nice work, @rstaten!)  It seems that there is agreement to the plotting syntax proposed, so we will complete the implementation now.  All QAs will want the option of plotting.  When we run QL as part of QL/F, the it does not make plots since the ‘—p’ command line option is by default off there (QLF makes the plots in that case.)   So each QA will have an ‘if’.  Once we have completed the implementation for QAs, this ‘if’ will be at the level of whether to call the plotting functions in QL.  So in that ultimate state, plot_bias_overscan would only be invoked when plotting is asked for, and it will need the plotconf argument going forward.  I think there should be no use case in which QL makes plots unless it is told to by command line option.  Seem reasonable?

Docstrings and other documentation will be added as you suggest.

felipelm:
@rkehoe ok it's a pretty big refactor for us and I'll discuss with the team tomorrow.

rkehoe:
@felipelm. One thing to keep in mind, it is probably not necessary to refactor everything all at once.  If you can update a few representative cases, this will also provide lessons for further implementations. You know that your existing plots do work, and the need to have this ‘on the fly’ capability is not needed immediately.  What’s more, it will be important that everything work for the January QL/F review, so if the refactor jeopardizes that you may want to consider a partial update for now.  

Those are my 2 cents.  I hope they are useful.  In our case, updating all the QAs is probably not that time-consuming, so we may just do it completely now.

rkehoe:
Thanks @rstaten!  This now provides a complete implementation of the code that makes plotting purely configurable in QL.  Hopefully, these can be useful examples for QLF as well.  One can run QuickLook to make plots by configuration in the following way:

desi_quicklook -i $DESI_PRODUCT_ROOT/desispec/py/desispec/data/quicklook/qlconfig_science.yaml 
            -n 20191001 -c r0 -e 3577 --plotconfig    
            $DESI_PRODUCT_ROOT/desispec/py/desispec/data/quicklook/qlplot_config.yaml        

@stephenbailey last week voiced support to retain the use of the hardcoded plots, since they are already working, in addition to the configurable option, perhaps as a backup.  These plots, in some cases, have some features that are not accommodated by the configurable options.  For instance, the overlaid fit of the SNR vs. magnitude plots are not in place.  Some features, like this one, may be added as configurable options because they have very clear need in QL/F.  But we do not want to implement an independent plotting library by this configuration, to support every possible plotting option.  This last point is a particularly important one, and I will try to follow up more on this in Issue #739.   

Another reason to  retain the hardcoded plots, at least for now, is that running QL might occur where a user does not have the plotconf file, but still wants to see the results.  We want to keep this option, at least for now.

The approach taken in this PR is that when calling quicklook with the -p option (without specifying a plotting configuration file) one obtains the existing hardcoded plots for all QAs.  When the plotconf is given with the --plotconf option, then for those QAs where metrics are given all of their plots are controlled by the plotconf.  For QAs that are without configured plots, they are still plotted with hardcoded commands for now.  A later PR will populate a fuller plotconf.

Following some tests with QLF, this should be mergeable today.

rkehoe:
Should have been @sbailey tagged in the comment above.

rstaten:
I have tested this branch using QLF and it runs fine. Note that QLF generates its own plots and runs QL without this feature as default, but the new plotting configuration can be implemented if desired. @felipeaoli If you have any questions please let us know. Also feel free to test this, but we will be merging soon.

rkehoe:
Looks good.  Thanks @rstaten. 

While this PR has been focused on the plot configuration capability, it includes an under-the-hood change that is worth nothing also:

- ability to add or change plots in a separate plotting configuration file (qlplot_config.yaml) which effectively eliminates the need to code the vast majority of QL plots.  Functionality supports the three primary forms of plots that have been utilized for QL (and in QLF): patch plots, 2D and 3D plots.

- if plotting is requested by command-line option, but a configuration file not given, the default hardcoded plots will be provided.

- the internal gathering of information for the GENERAL_INFO block of the mergedQA output JSON file is now done in the Initialize step rather than the merger.py.  This helped to permit the generalized plotting needed for the plot configuration.  

- both XWSigma and Sky_Peaks QAs now use exactly the same array of sky line locations in science exposures.  Previously they accessed separate lists that were copies of each other.

Since this PR has had no travis errors and passed QLF testing, and the code diffs look reasonable, I will wait for the latest travis test and merge after lunch.  

@julienguy, @felipelm, @felipeaoli, @djschlegel, @sbailey I will leave the branch undeleted during the break in case you want to go thru as documentation after the break.

PR #743, opened 2018-12-11
allow null RA,DEC etc. columns in DB for calib exp

sbailey:
This PR updates the DB loading to support the latest surveysim exposures format, which has NaN values for RA, DEC, EBV, SEEING, and MOON* parameters for arc and flat calibration exposures.  I think it is reasonable to allow these to be NaN for calibration exposures where they are meaningless, instead of forcing them to have dummy values.

Tested with the minitest integration test, which was previously failing with the latest surveysim due to NOT NULL constraints on those columns.

### DIFF
--- a/py/desispec/database/redshift.py
+++ b/py/desispec/database/redshift.py
@@ -204,18 +204,18 @@ class ObsList(SchemaMixin, Base):
     expid = Column(Integer, primary_key=True, autoincrement=False)
     tileid = Column(Integer, nullable=False)
     passnum = Column(Integer, nullable=False)
-    ra = Column(Float, nullable=False)
-    dec = Column(Float, nullable=False)
-    ebmv = Column(Float, nullable=False)
+    ra = Column(Float, nullable=True)   #- Calib exposures don't have RA, dec
+    dec = Column(Float, nullable=True)
+    ebmv = Column(Float, nullable=True)
     night = Column(String, nullable=False)
     mjd = Column(Float, nullable=False)
     exptime = Column(Float, nullable=False)
-    seeing = Column(Float, nullable=False)
-    transparency = Column(Float, nullable=False)
-    airmass = Column(Float, nullable=False)
-    moonfrac = Column(Float, nullable=False)
-    moonalt = Column(Float, nullable=False)
-    moonsep = Column(Float, nullable=False)
+    seeing = Column(Float, nullable=True)
+    transparency = Column(Float, nullable=True)
+    airmass = Column(Float, nullable=True)
+    moonfrac = Column(Float, nullable=True)
+    moonalt = Column(Float, nullable=True)
+    moonsep = Column(Float, nullable=True)
     program = Column(String, nullable=False)
     flavor = Column(String, nullable=False)
     # dateobs = Column(DateTime(timezone=True), nullable=False)
@@ -362,7 +362,7 @@ def load_file(filepath, tcls, hdu=1, expand=None, convert=None, index=None,
         return
     if maxrows == 0:
         maxrows = len(data)
-    log.info("Read data from %s.", filepath)
+    log.info("Read data from {} HDU {}.".format(filepath, hdu))
     try:
         colnames = data.names
     except AttributeError:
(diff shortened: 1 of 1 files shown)


py/desispec/database/redshift.py
```diff
@@ -362,7 +362,7 @@ def load_file(filepath, tcls, hdu=1, expand=None, convert=None, index=None,
         return
     if maxrows == 0:
         maxrows = len(data)
-    log.info("Read data from %s.", filepath)
+    log.info("Read data from {} HDU {}.".format(filepath, hdu))
```
weaverba137: Please don't change the formatting of the log messages.  There is a very specific reason for using `%`-style formatting.  See https://docs.python.org/3/howto/logging.html#logging-variable-data.  In theory other formatting methods are supported, but would require modifications to the `desiutil.log` module.  In addition, the logging infrastructure is smart enough to *not* format strings that it doesn't need to, as long as you are following the formatting style it is configured for.

sbailey: I don't understand this objection.  The new-style formatting is generating a string that gets passed to the logger, without the logger knowing or caring how that string was generated.  i.e. this doesn't require any modifications to `desiutil.log`; it is equivalent to:
```python
message = "Read data from {} HDU {}.".format(filepath, hdu)
log.info(message)
```

sbailey:
What type of additional tests are you thinking of?  The particular test that I was using was that in current master the minitest DB loading fails, and with this PR it now works again.  Some end-user queries may become more complex, but hopefully no one should be relying upon a filter on `RA` or `MOONFRAC` for arc/flat exposures anyway.

weaverba137:
I missed the part where you already tested the DB loading.  Please fix the log message formatting, then OK to merge.

py/desispec/database/redshift.py
weaverba137: The logger does know and does care.  Please read the link I sent in detail.

sbailey: I have empirically verified by running the code and looking at the log output that the formatting in this PR works just fine.  My understanding of that link you sent (and its links) is that the logger knows and cares only if you want the logger do the string formatting for you by passing in additional positional arguments.  If you format the string ahead of time before passing it to the logger (like I do here; note that there are no positional arguments or string parsing for the logger to do), then the logger doesn't know or care how you generated the string that you provide it.

For the record, I switched to the new-style formatting because I wanted to log the `hdu` but it can either be a string or an integer and I didn't realize until just now that old-style string formatting knows how to cast an integer to a string (it won't do the reverse).  But it just doesn't seem worth going back and changing working code to different working code.

weaverba137: Contact me on Slack.  It is important that you understand this.

tskisner:
Adding to the noise, I often do:
```
msg = "my log message number {}".format(5)
log.debug(msg)
```
To work around the silly logging thing.  Kind of annoying, but not DESI's fault.

weaverba137:
Ted, you are also welcome to contact me on Slack.  This is not "silly".

tskisner:
Sorry- "silly" in the sense that it is a shame that python3 logging attempts to maintain backwards compatibility and force the use of percent-style string formatting in the middle of modern code that uses format().   I understand their arguments, and obviously this is not something that DESI controls.  This is a feature of python logging.  My previous comment was just stating what I do (not just in DESI code) to keep the string formatting consistent across the code while still using the python logger.  Nothing else to say- I'll move along now ;-)

weaverba137:
The articles I cited do make this clear, but ultimately, this has nothing to do with the specific formatting used. This has to do with how the `logging` infrastructure decides what to log.

Here's a toy example.  This:
```python
log.debug('my message number %d', 2)
```
is *not* the same as
```python
message = 'my message number %d' % 2
log.debug(message)
```
Internally, `log.debug()` does something like this:
```python
def debug(message, *args):
    if logLevel <= DEBUG:
        emit_log_message(message % args)
    return
```
Note that the string is only formatted if it needs to be.  If it doesn't need to be then `log.debug()` is a no-op.

This means that if you format a string before passing it to a log function, you are wasting the format operation.  Let the logging infrastructure decide whether a string needs to be formatted.

sbailey:
Summarizing: your suggested practice is a runtime efficiency issue, not that the pre-formatting method I suggested doesn't work.  In this particular case, the runtime efficiency of the logging doesn't matter at all, and we have been remarkably inefficient in our human developer time discussing this, but I acknowledge that being in the habit of writing efficient patterns is useful.  Since I did ask you for a review and as a goodwill gesture, I will revert to your suggested formatting so that we can move on with 18.12.

Issue #744, opened 2018-12-13
desi_qa_prod crashing


sbailey:
The following crashes:
```
source /project/projectdirs/desi/software/desi_environment.sh master
export DESI_SPECTRO_DATA=/global/cscratch1/sd/sjbailey/desi/dev/qacrash-end2end/spectro/sim/mini
export DESI_SPECTRO_REDUX=/global/cscratch1/sd/sjbailey/desi/dev/qacrash-end2end/spectro/redux
export SPECPROD=mini
desi_qa_prod --make_frameqa=3 --qaprod_dir=$SCRATCH/temp/QA
```

The crash happens in `desispec.qa.qa_plots_ql.plot_SNR` while making a scatter plot, but the core problem is that  the input `qa_dict["METRICS"]["SNR_RESID"]` has 457 entries while there are 458 fibers in the `qa_dict['METRICS']['ELG/LRG/QSO/STAR_FIBERID']` lists used to create the `ra`, `dec` lists.  I checked that the fibers are unique -- this isn't a case of a target being both a QSO and an ELG.  Punting to @profxj to debug from here.

Note: The command specifies `--qaprod_dir=$SCRATCH/temp/QA` but the code still writes to `$DESI_SPECTRO_REDUX/$SPECPROD/QA/` i.e. you may need to make your own copy of the input data and reset the environment variables to actually test this so that it isn't trying to write into my production area.

I think running `desi_qa_prod --make_frameqa=2 --qaprod_dir=$SCRATCH/temp/QA` will get to the crash faster, but that may skip over the step that created a QA file with mis-matched array lengths.

profxj:
working on it..

sbailey:
Fixed by PR #746.

PR #746, opened 2018-12-15
S/N fix (offline)

profxj:
Addresses Issue #744 and includes a refactor of QA S/N handling
for exposures.  Appears to have run successfully on `qacrash-end2end`

Mis-match in array sizes was due to the 'cutting' of fibers that generate
NAN in `SNRfit`.  Have kept these fiber outputs for offline QA.  I suggest
the QL team consider doing the same as they will also have this array mis-match.


rstaten:
@profxj Thanks for fixing this, I will look into implementing this into QL.

PR #751, opened 2019-01-07
QL targeting update

rkehoe:
This PR is meant to update QuickLook for targeting updates in desispec.  As such it will include:

- propagation of use of target bits into all QAs for which needed
- reintroduction of bright/grey targets for SNR QA


rkehoe:
Okay, thanks @rstaten.  This looks good!

This PR implements the following updates around the topic of targeting and survey time:

- finish re-enabling bright/grey time target types following earlier merge down to one science config.  
- further use of targeting bits where needed, i.e. algorithm calculating spectral magnitudes.
- complete task of using three reference levels (dark/grey/bright) for assessment of scalar metrics.  This was already coded, but the configurable references and specifications needed to be updated.  The implications of this update is that the QL assessments are now able to be more sensitive, more along lines they were meant for.  See below for further information.
- finish calibration propagation to spectral magnitude following prior updates including average flux calibration vectors and broadband filter transmission.  DELTAMAG differences are now about 0.5-1 magnitude.  Further work will be needed to translate to comment on fiber throughput, etc.

As a note, one thing that is not yet re-enabled is different references for each camera.  This used to work by an individual camera comparing to an individual reference file input for each camera.  It is straightforward to implement this in the current scheme, but I would prefer a way that does not blow up the qlconfig with 30 references for each metric.  Have to come back to this, probably after next week.

Concerning the issue of the QL assessments, the specifications around the references for warning and alarm status have been tightened to reflect a by-eye measure of 2-sigma and 3-sigma variation of each scalar metric.  This yields mostly NORMAL status.  Currently, in 'r' arm for r0, all are NORMAL except bias, which is in ALARM due (I think) to a faulty sim, and DELTAMAG, which is 'WARN'.  Depending on where DESI ultimately wants to set these thresholds, having an occasional 'WARN' in one exposure is not necessarily a problem and is expected given the # of channels.  But this level's ultimate choice is not in the purvue of QL development itself and has to be determined by 'higher ups'.  It may be that [2,3] sigma is too tight given the number of cameras, and [3,4] is more appropriate.  @rstaten let's iterate this once more in the next PR.  I think it will be best if we can give users the first 'first stab' at what these should be.

@sbailey This PR is the first of two needed I mentioned earlier.  I will check a couple items and merge this one in a few minutes.  The second one will start today and we aim to finish early next week.

PR #754, opened 2019-01-18
QL stability

rkehoe:
This PR is meant to address two items that are observed in testing this week with current master to Need improvement.  One is a small edit on the SNR fitting area, and the other is better specifications for QAs to set warnings and alarms.

This PR will be merged today so that final update of QL/F can occur on Monday.

@felipelm this is needed for the installation for the review.  Can we make a final update Monday?


rkehoe:
Thanks for the update @rstaten.  These are looking more sensible given the underlying distributions.  

It is problematic to get references and specs from a sim sample that has bogus bias information, but the other observables seem stable.  The issue of bias can be followed up with better ‘good’ samples, I hope this week.

This brings up another point that qlconfigs (and now the plotconfig) have not been moved to the new repo created by @sbailey for that use.  This would allow these specs to be updated the way they were meant to — independent of the underlying code.  Given the pending review, I hesitate to make this change at the last minute with the risk of breaking something.  But it would be better to have this properly done.  

A decision can be made once the loop is closed with QLF.  This PR is working and passes tests, so I will merge now.

rkehoe:
@felipelm When I merge this PR, please use the resulting master as the basis for final integration of QL/F.  It has necessary updates.  Thank you.

felipelm:
@rkehoe there is a problem on SNR:

```
2019-01-21 15:24:44,332 Calculate_SNR INFO : FIDSNR_TGT_STATUS: UNKNOWN
2019-01-21 15:24:44,332 Calculate_SNR CRITICAL : QL Calculate_SNR : REFERENCE(4) and RESULT(3) are of different length!
  resid = (med_snr-fit_snr)/fit_snr
/app/desispec/py/desispec/qa/qalib.py:763: RuntimeWarning: invalid value encountered in true_divide
  resid = (med_snr-fit_snr)/fit_snr
/app/desispec/py/desispec/qa/qalib.py:763: RuntimeWarning: divide by zero encountered in true_divide
2019-01-21 15:24:42,728 QuickLook INFO : Running Calculate_SNR
  "astro":lambda x,a,b:(exptime*a*x)/np.sqrt(exptime*(a*x+b))
/app/desispec/py/desispec/qa/qalib.py:561: RuntimeWarning: invalid value encountered in true_divide
2019-01-21 15:24:37,653 QuickLook INFO : Running Calculate_SNR
```

rkehoe:
@felipelm That INFO message for UNKNOWN is being set when one of the target types does not complete its fit, but the other three do.  This currently therefore given as ‘unknown’ status.  However, in this situation if LRGs don’t fit, but ELGs, QSOs and STDs do fit and yield NORMAL status (which is what is happening) then it is clear the status is NORMAL.  There will be an update today for this.

PR #756, opened 2019-01-23
minimize warnings for resolution fit

rstaten:
This PR minimizes the warnings generated by resolution fit. The algorithm runs fine, but if a line is not able to be fit, we do not need to log extra information.

@rkehoe Please check this and merge when possible.
@felipelm At your convenience, please include this update in your latest QLF version.


rkehoe:
@rstaten. Thanks for getting this out there.  This solution, of dropping the fiber when one of the lines does not fit, is expedient now.  But it would be better to preserve the information per fiber of how many lines were fit, and keep fibers unless the percentage failing is less than some fraction.  One does need quite a few lines to get a decent Legendre polynomial fit, but missing one line out of, say, 16, is not going to fundamentally cripple that usually.

Needs a study!  So I’ll merge this and hopefully somebody gets to tune this up!

PR #766, opened 2019-03-11
KPNO

julienguy:
Merge several minor modifications to the handling of calibration data.
The most important one is the use of the keywords CCDCFG and CCDTMING to select calibration configurations IF they are present in the yaml files (so that things still work with pre-existing data and simulated data without those keywords) instead of DOSVER and FEEVER.
CCDCFG and CCDTMING give the configuration file names used by the FEE boxes (CCD readout).


julienguy:
Merging to keep consistency with calib svn

PR #767, opened 2019-03-14
Ql SNR Fit

rkehoe:
This PR improves the fit of SNR vs. fiber magnitude by increasing the granularity of the ‘a’ parameter in the fit from 0.01 to 0.0001.  This improves the fit itself, and also facilitates convergence on extreme cases where fits might fail.

Thanks @rrstaten for this update!

### DIFF
--- a/py/desispec/qa/qa_quicklook.py
+++ b/py/desispec/qa/qa_quicklook.py
@@ -84,11 +84,11 @@ def get_image(filetype,night,expid,camera,specdir):
 
     #- Create necessary input for desispec.image
     image = fits.open(imagefile)
-    pix = image[0].data
-    ivar = image[1].data
-    mask = image[2].data
-    readnoise = image[3].data
-    meta = image[0].header
+    pix = image['IMAGE'].data
+    ivar = image['IVAR'].data
+    mask = image['MASK'].data
+    readnoise = image['READNOISE'].data
+    meta = image['IMAGE'].header
 
     #- Create image object
     imageobj = im(pix,ivar,mask=mask,readnoise=readnoise,camera=camera,meta=meta)
@@ -103,12 +103,12 @@ def get_frame(filetype,night,expid,camera,specdir):
 
     #- Create necessary input for desispec.frame
     frame = fits.open(framefile)
-    wave = frame[4].data
-    flux = frame[0].data
-    ivar = frame[1].data
-    fibermap = frame[5].data
+    wave = frame['WAVE'].data
+    flux = frame['FLUX'].data
+    ivar = frame['IVAR'].data
+    fibermap = frame['FIBERMAP'].data
     fibers = fibermap['FIBER']
-    meta = frame[0].header
+    meta = frame['FLUX'].header
 
     #- Create frame object
     frameobj = QFrame(wave,flux,ivar,fibers=fibers,fibermap=fibermap,meta=meta)
--- a/py/desispec/qa/qa_quicklook.py
+++ b/py/desispec/qa/qa_quicklook.py
@@ -103,7 +103,7 @@ def get_frame(filetype,night,expid,camera,specdir):
 
     #- Create necessary input for desispec.frame
     frame = fits.open(framefile)
-    wave = frame[3].data
+    wave = frame[4].data
     flux = frame[0].data
     ivar = frame[1].data
     fibermap = frame[5].data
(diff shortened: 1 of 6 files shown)


rstaten:
Here is an example comparing master vs. this branch. This is a very extreme case, but it shows the improvement:
ql-snr-z0-00003590
ql-snr-z0-00003590_qlsnr

@profxj This should not affect offline, but please let us know if that is not the case or if you have any comments.

py/desispec/qa/qa_quicklook.py
```diff
@@ -103,15 +103,15 @@ def get_frame(filetype,night,expid,camera,specdir):
 
     #- Create necessary input for desispec.frame
     frame = fits.open(framefile)
-    wave = frame[3].data
+    wave = frame[4].data
```
sbailey: This PR just highlighted that you are reading HDUs by number rather than name, and this change from 3->4 implies that QL will work for some version of the data but not others.  It is slightly mixing topics in this PR, but please update this code to read by EXTNAME instead of numbers.  e.g.
```
wave = frame['WAVE'].data
flux = frame['FLUX'].data
...
```
See the EXTNAME keyword of each HDU for what the names should be.  This allows us to add or reorder HDUs and still have code like this work.

py/desispec/quicklook/__init__.py
```diff
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, division, print_function
 
 #SE: needed to maintain quicklook-specific version 
-__qlversion__ = "10.18.0"
+__qlversion__ = "02.19.0"
```
sbailey: Suggestion: reorder your QL version numbering schedule to be YY.MM.minor so that they will be sorted in time.  02.19.0 implies that this version is earlier than 10.18.0 until a human interprets it as a date.  i.e. let's switch this version to 19.2.0

py/desispec/quicklook/quicklook.py
```diff
@@ -159,7 +159,7 @@ def runpipeline(pl,convdict,conf):
     else:
         import numpy as np
         qa=None
-        qas=['Check_HDUs',['Bias_From_Overscan','Get_RMS','Count_Pixels','Calc_XWSigma'],'Trace_Shifts','CountSpectralBins',['Sky_Continuum','Sky_Peaks'],[],['Sky_Rband','Integrate_Spec','Calculate_SNR']]
+        qas=[[],['Bias_From_Overscan','Get_RMS','Count_Pixels','Calc_XWSigma'],'Trace_Shifts','CountSpectralBins',['Sky_Continuum','Sky_Peaks'],['Calculate_SNR'],['Sky_Rband','Integrate_Spec']]
```
sbailey: What is the impact (if any) of having a blank list in this list?  Will the removal of "Check_HDUs" and the regrouping of "Calculate_SNR" have any impact on the structure of the output files (and thus on QLF)?

py/desispec/qa/qalib.py
```diff
@@ -738,8 +738,36 @@ def SNRFit(frame,night,camera,expid,params,fidboundary=None,
                             minchi2=chi2
                             fita=guess[0]
                             fitb=guess[1]
+                #- Increase granualarity of 'a' by a factor of 10
+                for c in range(100):
+                    for d in range(100):
```
sbailey: OK for now, but for the record, this brute force search looks slow, and without outlier rejection I would expect it to have convergence problems on real data.  I would expect that one of the routines in scipy.optimize could be much faster, enabling a few iterations of outlier rejection. Consider that for a future PR.

rstaten:
@sbailey The most recent push should address the requests. Regarding the reading of HDUs and blank list, these are both only used when trying to run a single QA at a time (it helps when trying to develop QAs like SNR that are at the end of the pipeline). These areas don't affect the running of the actual pipeline at all.

For SNR, we used to use scipy.optimize.curve_fit, but it was requested in the fall that we didn't use that, which is when this brute force method was implemented. This can certainly be changed though and I agree that this isn't the best way, just wanted to make an improvement here.

rkehoe:
@sbailey   Thanks for merging.  This PR hadn’t been tested against QLF yet — that was my next step.  In this case, however, the change does not touch anything on the interface for QL/F, so I’m sure it’s now problem.  Just wanted to remind you of this step just in case.

sbailey:
@rkehoe if the intension was to open a PR for review but not yet for merging, please flag it with the WIP tag (work in progress) and explicitly mention "before merging, we should ...".  Given that it didn't appear to change the output details, I thought it was safe to merge, but I would have been fine with waiting if I had known other tests were coming.

rkehoe:
@sbailey. Okay.  Next time one of us will put a note that ‘pending QL/F tests’ or some such.  I didn’t flag with WIP because this was a small edit and was to be closed off quickly.  My sense is we generally want to use WIP for long term items that will be ongoing for a while.  I’ll try to be more explicit about tests in comments.

PR #771, opened 2019-03-23
import speed of light from scipy

londumas:
move to a common definition of speedoflight. Step toward fixing issue https://github.com/desihub/desisim/issues/477.
The remaining one I can find, but I don't yet understand is the following.
@sbailey, what is that number `2.9979e18`? Is that `c in A/s`
```
<me>/Programs/desi/code/desispec> grep '2.99' -ri * --include \*.py
py/desispec/fluxcalibration.py:    ZP_fnu = ZP_flambda * wave**2 / (2.9979e18)  # c in A/s
```

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -19,6 +19,12 @@ from astropy import units
 import multiprocessing
 from pkg_resources import resource_exists, resource_filename
 
+try:
+    from scipy import constants
+    C_LIGHT = constants.c/1000.0
+except TypeError: # This can happen during documentation builds.
+    C_LIGHT = 299792458.0/1000.0
+
 def isStdStar(desi_target, bright=None):
     """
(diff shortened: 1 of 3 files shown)


londumas:
@julienguy, can you do a review?

julienguy:
- I agree we need a single place where physical constants are kept. 
 - However, I don't see why it is more reliable to use an external package for this.
 - I would personnally rather have a python file constants.py in desiutil for this purpose, used by both desisim and desispec

londumas:
@julienguy, I have no strong feeling about importing versus defining. However, I really would like the speed of light to be exactly `299792458.0` everywhere.

weaverba137:
I should add that @londumas is not doing anything new here.  Other DESI packages have already transitioned to importing the speed of light from scipy.

julienguy:
Then that's fine. I am merging.

PR #773, opened 2019-04-03
Update API documentation

weaverba137:
While working on something else, I noticed that the doc/api.rst file had not been updated in quite a long time, and as a result, many docstrings had never been tested.  There are no changes to code, but some docstrings had to be substantially modified.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -319,7 +319,10 @@ Full desispec API Reference
 .. automodule:: desispec.scripts.bootcalib
     :members:
 
-.. automodule:: desispec.scripts.dts
+.. .. automodule:: desispec.scripts.dts
+..     :members:
+
+.. automodule:: desispec.scripts.delivery
     :members:
 
 .. automodule:: desispec.scripts.dts_status
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -7,10 +7,13 @@ Full desispec API Reference
 .. automodule:: desispec
     :members:
 
+.. automodule:: desispec.averagefluxcalibration
+    :members:
+
 .. automodule:: desispec.bootcalib
     :members:
 
-.. automodule:: desispec.quicklook.qlboxcar
+.. automodule:: desispec.calibfinder
     :members:
 
 .. automodule:: desispec.coaddition
@@ -29,6 +32,9 @@ Full desispec API Reference
     :members:
 
 .. automodule:: desispec.database.util
+    :members:
+
+.. automodule:: desispec.dustextinction
     :members:
 
 .. automodule:: desispec.fiberflat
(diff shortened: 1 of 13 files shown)


sbailey:
Thanks for updating this.  A few style questions / comments:

In several places you replaced [Google-style docstrings](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) with the more verbose [numpy-style docstrings](https://numpydoc.readthedocs.io/en/latest/format.html).  Was this a sphinx requirement, or do Google-style docstrings also work, or only a subset of Google-style or was this a personal preference or something else?

Are there any tools that would auto-walk a module tree and generate the appropriate API listings, instead requiring us to manually keep `doc/api.rst` in sync with any code updates?  (which we obviously have not been diligent about)

In many places you used the `:class:` formatting prefix for built-in basic python types like `dict`, `list`, and `str`.  Although that enables a link to the python documentation in the generated readthedocs HTML, it adds visual clutter when reading the code and when using pydoc or ipython/jupyter %help magic to read the docstrings, which I think are the more common ways for most developers to view the API docstrings.  i.e. let's not decorate basic python types with `:class:`, and reserve that for less familiar types or ones that we define ourselves (and even then I don't think we should require it).

weaverba137:
* Google -> numpy: I've noticed that docstrings -- independent of style -- are very fiddly about line spacing.  In those particular cases, it was faster for me to rewrite the the docstring in a style I'm intimately familiar with (and thus know the line spacing gotchas) than debug the original style.
* I'm not aware that this capability exists.  If it did, the Astropy project probably would have adopted it a while ago.  They use an enhanced version of `automodule` called `automodapi`, but that still works module by module.  In theory though, api.rst is a very simple document, and could be generated by a script that scrapes the package tree.  That would have to be re-run as new modules are added.  I think that's another PR though...
* Class cross-references: Are you speaking based on a survey of the entire collaboration or just anecdotally?  I personally don't use pydoc (at all) or %help (extremely rarely), and prefer to view API documentation on a webpage (*i.e.* ReadTheDocs) with lots of cross-references.  Are you asking me to *remove* those notations from this PR?

sbailey:
Google vs. numpy:  OK.  Let's try to continue supporting both, but ok to convert from one to the other when that is the most efficient way to debug picky line spacing.

Auto-generating api.rst: I'm surprised this isn't a common stock tool, but ok.  Agreed that any updates in that direction are a future PR, not this one.

`:class:` usage and pydoc / %help / readthedocs: I'm speaking anecdotally, but it is based on a sample larger than just myself, and the circumstantial evidence that our readthedocs API docs wouldn't have gotten so out of date if that was the primary method that most active developers were using.  My *impression* is that most developers just read the docstrings directly in the code, and some use ipython/jupyter %help (when working in ipython/jupyter, but they don't go there just to read documentation), and even fewer use pydoc (though I use that a lot).  I'm not asking you to do the busy work of removing those `:class:` from this PR, but I am asking that you *not* add them in future PRs for really basic python types like list, dict, str, int, float.  And I might remove them in some future update.  OK to use that for links to classes within our own packages.

Issue #762, opened 2019-02-28
Unflagged cosmics in desi_preproc


londumas:
The following data are using SP2, but the results are similar for SP0.
I preproc using the following for the series of 10 zeros (`199,200,201,202,203,204,205,207,208,209`), i.e. not using the `206`:
```
desi_preproc -i /exposures/desi/sps/20190227/00000208/sp2-00000*.fits.fz
--outdir /n/home/datasystems/users/hdumasde/20190227/00000208/
--cameras b2,r2,z2
```
Here is the histograms of data. You can clearly see that there are a lot of unflagged cosmics, spetially in r2 and z2.

![histo_b2_zero](https://user-images.githubusercontent.com/3166436/53539441-d7d14180-3ace-11e9-8ca9-3f569d23bd39.png)

![histo_r2_zero](https://user-images.githubusercontent.com/3166436/53539440-d7d14180-3ace-11e9-9a27-ddf17970469d.png)

![histo_z2_zero](https://user-images.githubusercontent.com/3166436/53539439-d7d14180-3ace-11e9-8075-e64d1fa95f15.png)

Looking at the images, most of them seem to be associated with flagged cosmics. Some are isolated though.
In white good data with `value<40`, in violet masked data, in color the log of the value of the unflagged cosmics.

![r2_202_cosmic_3](https://user-images.githubusercontent.com/3166436/53539643-a86f0480-3acf-11e9-973e-4e742f4c1277.png)

![r2_202_cosmic_2](https://user-images.githubusercontent.com/3166436/53539644-a86f0480-3acf-11e9-8db3-5ab6d9ba02de.png)

![r2_202_cosmic](https://user-images.githubusercontent.com/3166436/53539645-a86f0480-3acf-11e9-87c1-23e96d7719ed.png)

![r2_199_cosmic_isolated](https://user-images.githubusercontent.com/3166436/53539753-13b8d680-3ad0-11e9-8a34-41803d0452b0.png)

londumas:
Updated results using 50 zeros: `[1159,1208]`:
 - b0 and b2 seems to have all of their cosmic-rays flagged.
The outlier pixels are different observation of pixel `x = 54, y = 778`; it hat has a strange behavior (large variance?).
 - r0 and r2 have a large number of unflagged cosmic.
 - z0 and z2 have some unflagged cosmics but way less than r*.

[...]

PR #775, opened 2019-04-20
Improve cosmics

julienguy:
- new gradient parameters from average of SP2 PSF
- new default value =0.5 for the parameter "c2fudge" which encodes the gradient threshold in units of the PSF gradient.
See issue #762 for more information.


londumas:
Works well, does nothing for `b`, since all cosmics were flagged, improves a lot `r` , and not much `z`.
This is compatible with the fact that finding a balance between purity and completeness of CCD pixels.
In the following plots blue is master and orange is this branch for a zero frame precessed via desi_preproc.
r2
r0

z2
z0

Issue #774, opened 2019-04-15
Unit tests don't clean up QA fits files automatically


lastephey:
I discovered that if you are running the full unit test suite in desispec, i.e.

python setup.py test

and you quit the tests partway through, several QA tests fail when you run the unit tests again. This is because there are residual files left behind that are not automatically deleted at the beginning of the tests. These files are:

test-pix-abcde.fits
test-raw-abcde.fits
test_arc.fits.gz
test_flat.fits.gz

which are written in the top directory of desispec. 

If these leftover files exist at the beginning of the unit tests, errors like these occur:

ERROR: testCountPixels (desispec.test.test_ql_qa.TestQL_QA)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/global/cscratch1/sd/stephey/git_repo/desispec/py/desispec/test/test_ql_qa.py", line 148, in setUp
    desispec.io.write_raw(self.rawfile,rawimg,hdr,camera=self.camera)
  File "/global/cscratch1/sd/stephey/git_repo/desispec/py/desispec/io/raw.py", line 198, in write_raw
    raise ValueError('Camera {} already in {}'.format(camera, filename))
ValueError: Camera z1 already in test-raw-abcde.fits

weaverba137:
I asked a very similar question a while ago, and at that time `test_(arc|flat).fits.gz` were meant to be cached between tests.  However, `test-(pix|raw)-abcde.fits` files are new to me.  Have you seen failures related to the presence of the `test_(arc|flat).fits.gz` files or only `test-(pix|raw)-abcde.fits` files?

sbailey:
Fixed in PR #776.

@weaverba137 FYI the test_(arc|flat).fits.gz files are big and auto-downloaded, and tests still work if they are left behind.  The test*-abcde.fits files were created on the fly and would cause problems if they were leftover from a previously crashed test (because the auto-creation would step on itself and fail).

PR #776, opened 2019-04-26
remove leftover files when rerunning unit tests

rstaten:
This PR addresses #774, removing temporary files generated by QuickLook unit tests that are not removed if test_ql_qa.py is killed while running.

- Note: only the files test-pix-abcde.fits and test-raw-abcde.fits mentioned in #774 are removed as the other files are not generated by test_ql_ql.py.

@rkehoe, @sbailey  Please review and/or provide any necessary additional comments.

### DIFF
--- a/py/desispec/test/test_ql_qa.py
+++ b/py/desispec/test/test_ql_qa.py
@@ -3,6 +3,8 @@ tests for Quicklook QA class and functions. It also indludes tests on low level 
 """
 
 import unittest
+import shutil
+import tempfile
 import numpy as np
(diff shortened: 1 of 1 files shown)


rkehoe:
Thanks @rstaten. Looks fine as a first pass.  As we discussed, a more robust solution may be to  develop a signal handler for this, but that should be contingent on QL development questions still TBD elsewhere. This at least removes the item that was thwarting others.  So if @sbailey is okay with it as a first pass, I’m okay with it.

PR #780, opened 2019-05-21
support findfile camera+spectrograph as b,r,z + 0-0 instead of b0,r1..z9

### LINKED ISSUE #779: Spectrograph number missing from cframe filepath returned by desispec.io.meta.findfile?
```findfile('cframe', night=20200317, expid=18, nside=64, camera='r', spectrograph=9, rawdata_dir=None, specprod_dir=None, download=False, outdir=None, qaprod_dir=None)```

returns:

```/global/cscratch1/sd/mjwilson/minitest-19.2/spectro/redux/mini/exposures/20200317/00000018/cframe-r-00000018.fits```

even though spectrograph is provided.  Each camera band, in this case r, should be accompanied by spectrograph number.  In this case, real path ends with ```cframe-r1-00000018.fits``` to ```cframe-r9-00000018.fits```.

It might make sense to add a file exists check before returning the filepath?

sbailey:
This PR fixes #779, to support `io.findfile(..., camera='r9')` and `io.findfile(..., camera='r', spectrograph=9)`.  Previously only the former was supported.

For context, the `spectrograph` argument was previously only used for stdstars, which span b/r/z and thus don't request a camera.

### DIFF
--- a/py/desispec/io/meta.py
+++ b/py/desispec/io/meta.py
@@ -118,6 +118,11 @@ def findfile(filetype, night=None, expid=None, camera=None, groupname=None,
     else:
         specprod = None
 
+    #- Normally camera is b0, r1, .. z9, but also support b,r,z + spectrograph
+    #- Combine camera + spectrograph if needed
+    if camera is not None and spectrograph is not None and len(camera) == 1:
+        camera = camera + str(spectrograph)
+
     actual_inputs = {
         'specprod_dir':specprod_dir, 'specprod':specprod, 'qaprod_dir':qaprod_dir,
         'night':night, 'expid':expid, 'camera':camera, 'groupname':groupname,
--- a/py/desispec/io/meta.py
+++ b/py/desispec/io/meta.py
@@ -118,10 +118,10 @@ def findfile(filetype, night=None, expid=None, camera=None, groupname=None,
     else:
         specprod = None
 
-    #- Normally camera is b0, r1, .. z9, but also support b,r,z + spectrograph
-    #- Combine camera + spectrograph if needed
-    if camera is not None and spectrograph is not None and len(camera) == 1:
-        camera = camera + str(spectrograph)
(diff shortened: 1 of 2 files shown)


sbailey:
Having submitted this, I am a bit ambivalent about it, and welcome comments.  Previously "camera" meant b0, r1 .. z9, and we have used "arm" or "channel" or "band" to refer to just the b,r,z part.  This PR supports calling "camera" just b,r,z if also specifying "spectrograph", which could be a convenience, or it could just encourage non-standard nomenclature and cause confusion if we have to start thinking about "in this code does camera mean b,r,z or b0, r1, .. z9"?

Opinions?

michaelJwilson:
On checking the existence of the path ...  you could at least check the existence of the parent dirs?  I'd suggest the current format works well if you know what you're doing, but otherwise...  For something labelled 'findfile' it seems reasonable to default to something that already exists, as opposed to writing.  If it was me, I'd add a write=False flag to check on only parent dirs and otherwise check on the file path itself.  

The other point isn't a big one.  But asking for spectrograph does seem confusing in this context.  I can close the other ticket.

weaverba137:
I do not like the using 'camera' to mean 'b,r,z'.  We already have plenty of good names for 'b,r,z'.

sbailey:
After discussion, reverting this PR to *not* support camera='b', 'r', 'z' without the spectrograph number.  But I did add a small check to print a slightly more informative error message if the user tries camera='b', spectrograph=9.  It purposefully does not check for the existence of the file or automatically create the directory.  If you want that, use `os.path.isfile(filename)` and `os.makedirs(os.path.dirname(filename), exist_ok=True)`.

sbailey:
The smallest things always unravel...

The tests correctly caught that we sometimes take advantage of wildcards in camera names to create a string to be used by globs for finding multiple files at once.  Updated checks to not reject that...

Issue #703, opened 2018-10-23
Heliocentric redshift


julienguy:
David S. reminded me we have to implement a heliocentric redshift correction in the pipeline.
This would be trivial if applied after the redshift fit but we cannot do this. We have to scale the wavelength, because observations of the same target can be several month apart. It is also important for Lyman-alpha. 

- the wavelength correction has to happen **after** flux calibration because we define the instrument calibration in the observer frame. this means the std stars peculiar velocities have to be corrected after the fit.
- this is an important change because it means we need to either resample the spectra to the same wavelength grid (ideally with spectroperf 1D which is CPU intensive) or have different wavelength per spectrum in the spectra files (I don't think we want that).

sbailey:
Documenting a lunch conversation: another option is to extract each frame on a custom grid such that after the heliocentric correction is applied it ends up back on the common grid.  The uncalibrated frame files would all be on different grids, but the calibrated cframe files would be heliocentric corrected and on a common grid.  There is a second order effect of variation of the correction across the field, but I think that is tiny (should confirm).

sbailey:
Implemented in PR #790.  Closing this issue.

PR #790, opened 2019-06-10
Heliocentric correction 

julienguy:
- I resorted to use the routines of astropy.coordinates (and astropy.time). It is quite slow, slower than what a custom implementation would be, but it has the advantage of being debugged (hopefully).
- In the routine heliocentric.py , there is a main() function to compare with the IDL idlutils routine used for the BOSS/eBOSS projects. After correcting for a difference of sign due to different definition, the max difference I obtained for a range of coordinates is <0.2 km/s.
- The routine is called in scripts/extract.py 
  - the heliocentric correction is computed based on the RA,DEC,MJD of the observations found in the preprocessed image header (coordinates of the center of the field of view)
  - the wavelength of the extraction grid are the nominal wavelength array *divided* by the correction factor
  - the final wavelength is the wavelength in the extraction grid *multiplied* by the correction factor so that the final wavelength array is the same for all spectra, while we do not introduce any extra resampling (flux values remain uncorrelated)

### DIFF
--- a/py/desispec/heliocentric.py
+++ b/py/desispec/heliocentric.py
@@ -10,12 +10,13 @@ import numpy as np
 import astropy.units as u
 from astropy.time import Time
 from astropy.coordinates import SkyCoord, EarthLocation
+import astropy.constants
 
 kpno = EarthLocation.from_geodetic(lat=31.96403 * u.deg,\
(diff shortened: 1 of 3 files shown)


julienguy:
This addresses issue #703.

Issue #792, opened 2019-06-24
Study overscan subtraction problems


sbailey:
Study the overscan subtraction of CCD raw data as part of the preprocessing.  e.g. 20190621/00016474 z0 has a clear step between amplifiers that might be an algorithmic robustness problem:
![image](https://user-images.githubusercontent.com/218471/60057834-d4c01980-969a-11e9-90a3-e7e8cef058d7.png)
and amps C and D of 20190306/00001076 z2 have row-to-row variations that look like we need row-by-row subtraction (or maybe better bias image):
![image](https://user-images.githubusercontent.com/218471/60057893-0fc24d00-969b-11e9-946e-9b8ac18ba0c6.png)

Track these down and see if any algorithmic updates are needed, or otherwise feedback to the hardware team if it appears to be a raw data problem.

profxj:
I am digging in on this.

profxj:
I believe the first issue as regards overscan subtraction has been mititaged by PR #793.
But, we need to re-characterize the gain of `z0` to confirm (I believe the values will
change).  Camera `z2` is looking good.

See discussion in PR #793 on the second issue raised here.

sbailey:
What's the status of the z0 gain updates?  desispec was updated at KPNO and monitoring restarted after PR #793 was merged, but we're still getting amplifier jump problems on z0, e.g. 20190625/00016618:

![image](https://user-images.githubusercontent.com/218471/60206724-25558500-9809-11e9-85d6-fe658b08cd59.png)

sbailey:
Regarding this comment from PR #793:

> The second issue raised in #792 is partly mitigated by doing a by-row
removal of the overscan_col region, but I agree with @julienguy that the
extra noise incurred is not worth the slight improvement in appearance.

Could you quantify the size of the residual wiggles vs. the amount of noise introduced by doing row-by-row overscan subtraction instead of 2D overscan median scalar subtraction?

profxj:
Have just finished analyzing new gains for SP0.  
This has fixed the Issue with 20190621/00016474 (see image)
and presumably any other SP0 images.  

![new_z0_16474](https://user-images.githubusercontent.com/3998093/60217841-7b82f200-9822-11e9-9901-34df42283497.png)

profxj:
Indeed, 20190625/00016618 now looks 'perfect'

sbailey:
[...]

PR #793, opened 2019-06-25
Modify Preproc Overscan Subtraction

profxj:
Modifies `preproc` to use both overscan regions.  The sequence is now:

- Subtract the column-oriented overscan from all data+overscan_rows
- Collapse and subtract the overscan_rows from the data

This PR also includes a small add-on to `desi_compute_gain` to output
the measured gain values to a Table.

This PR fixes the first item in Issue #792.  You can see
more here:  https://github.com/desihub/desispec/blob/gain_output/doc/nb/Preproc.ipynb

The second issue raised in #792 is partly mitigated by doing a by-row 
removal of the overscan_col region, but I agree with @julienguy that the
extra noise incurred is not worth the slight improvement in appearance.


profxj:
I have modified the code to ignore `overscan_row` if ORSEC keywords are not present.
`orig_over` and `debug` are gone.  
Have added an option to turn off `overscan_row` by
optional argument in `preproc`.   
polished the docs

PR #794, opened 2019-06-28
Timing cleanups

### LINKED ISSUE #726: update desi_pipe job times
Review and update job time allocations for each pipeline step.  Consider adding a minimum of 5 or 10 minutes to every job script.  Examples:

Running `desi_pipe chain --tasktypes traceshift, ...` on edison resulted in a 1 minute batch job for 60 traceshifts.  The job wake up time took longer than that and the job was killed before anything substantive ran.

On the positive side, extractions are now much faster (thanks @lastephey!) and maybe could request less job time.  Don't be too aggressive with job time optimization though -- the impact of under-estimating the time is much more than over-estimating.

tskisner:
The SV data challenge files (2 days of sims) were used to compare run times between Haswell and KNL nodes on cori.  A scale factor of 3 is good for most cases.  I made several very small tweaks to the base per-task runtimes.  The "startup" time estimate is now a machine-dependent setting and is separate from the per-task time requirements.  This closes #726.


py/desispec/pipeline/tasks/psf.py
```diff
@@ -100,7 +100,7 @@ def _run_defaults(self):
         # if that directory does not exist, use a default NERSC
         # location.
         opts["lamplines"] = \
-            "/project/projectdirs/desi/software/edison/specex/specex-0.3.9/data/specex_linelist_desi.txt"
+            "/global/common/software/desi/cori/desiconda/current/code/specex/master/data/specex_linelist_desi.txt"
```
sbailey: For reproducibility, let's tie this to a specific version and not master, e.g. "/global/common/software/desi/cori/desiconda/20180709-1.2.6-spec/code/specex/0.6.2/data/specex_linelist_desi.txt".  Or maybe
`os.path.expandvars('$SPECEX/data/specex_linelist_desi.txt')` and then raise and exception if that doesn't exist.

Would be better if specex itself had a way to query where the linelist file is for whatever version is installed / configured, e.g.
```
import pkg_resources
pkg_resources.resource_filename('specex', 'data/specex_linelist_desi.txt')
```
I haven't checked how the pipeline wrapper finds specex, but perhaps that could be leveraged instead to avoid special environment variables or hardcoding a path (and then forgetting to update it when we move to a new desiconda version).

PR #802, opened 2019-08-02
help numba know the types

sbailey:
This PR enables numba to understand the types used within `desispec.cosmics._reject_cosmic_rays_ala_sdss_single_numba` and `dilate_numba`.  I don't understand why it couldn't handle the previous more obvious `dtype=int` and `dtype=bool`, but that resulted in slews of numba deprecation warnings about falling back to object mode but finding lifted loops blah blah blah.

A secondary benefit of this is that it allows `@numba.jit(nopython=True)` which makes cosmic ray rejection ~2x faster resulting in ~20% faster preprocessing.


lastephey:
That's great! The numba developers at scipy recommended always setting
nopython=True for this reason.

Issue #799, opened 2019-07-21
desi_pipe cleanup --submitted incorrectly moves some "done" jobs back to "ready"


sbailey:
Problem: `desi_pipe cleanup --submitted` moved some "done" jobs back into "ready", even though their outputs were on disk.

Details: I had a spectra-redshifts job that died from an out of memory error during spectral regrouping, though 287/633 of the healpixels had already finished and were marked as "done":
```
[cori12 ~] desi_pipe top --once
----------------+---------+---------+---------+---------+---------+---------+
   Task Type    | waiting | ready   | running | done    | failed  | submit  |
----------------+---------+---------+---------+---------+---------+---------+
preproc         |        0|        0|        0|     5130|        0|     5130|
psf             |        0|        0|        0|      630|        0|      630|
psfnight        |        0|        0|        0|      210|        0|      210|
traceshift      |        0|        0|        0|     4500|        0|     4500|
extract         |        0|        0|        0|     4497|        3|     4500|
fiberflat       |        0|        0|        0|      630|        0|      630|
fiberflatnight  |        0|        0|        0|      210|        0|      210|
sky             |        3|        0|        0|     3867|        0|     3870|
starfit         |        3|        0|        0|     1287|        0|     1290|
fluxcalib       |        9|        0|        0|     3861|        0|     3870|
cframe          |        9|        0|        0|     3861|        0|     3870|
spectra         |        0|        0|      346|      287|        0|       NA|
redshift        |      633|        0|        0|        0|        0|       NA|
----------------+---------+---------+---------+---------+---------+---------+
```

The corresponding output files were on disk, matching the number of "done" spectra processes:
```
[cori12 ~] find $DESI_SPECTRO_REDUX/$SPECPROD/spectra-64/ -name spectra-64\*.fits | wc -l
287
```

But when I ran `desi_pipe cleanup --submitted` it moved all of the spectra "done" tasks back to "ready", instead of only resetting the "running" tasks:
```
[cori12 ~] desi_pipe cleanup --submitted
[cori12 ~] desi_pipe top --once
----------------+---------+---------+---------+---------+---------+---------+
   Task Type    | waiting | ready   | running | done    | failed  | submit  |
----------------+---------+---------+---------+---------+---------+---------+
preproc         |        0|        0|        0|     5130|        0|        0|
psf             |        0|        0|        0|      630|        0|        0|
psfnight        |        0|        0|        0|      210|        0|        0|
[...]

PR #803, opened 2019-08-07
Desi pipe getready fix

julienguy:
Three bug fixes to the pipeline:
 - no more overwriting the same job when multiple jobs of same task type are written
 - db.getready checks  whether 'waiting' spectra or redshift are actually 'done'
 - the spectra and redshift tasks modify the healpix_frame table when setting their state to 'done' in same transaction to avoid  getting an unconsistent state of the db when a job is crashing or is halted. The healpix_frame table modification was previously performed by rank 0 after all tasks are finished. The goal was to reduce overloading the db access, but this was fragile.

One convenient new feature for expert debugging:
  - `desipe_pipe query` to directly query the db with sql syntax.


julienguy:
This fixes issue #799

tskisner:
Looks good @julienguy , thanks for this work!

PR #807, opened 2019-08-09
preproc CCDSEC[1-4] vs. CCDSEC[A-D]

sbailey:
This PR updates desispec to support pixsims that have header keywords more similar to the real raw data.  In particular, it auto-derives whether the sims use CCDSEC[1-4] (old pixsims) or CCDSEC[A-D] (real data & new SV pixsims) without requiring updates in $DESI_SPECTRO_CALIB to define a few simulation config.  This applies to DATASEC* and other amplifier-slice keywords.

I have tested this with old pixsims (redwood), new pixsims (svdc2019d), and real raw data (20190719/00017532/desi-00017532.fits.fz).

This PR also enforces that `desispec.io.write_raw()` use the ABCD convention that the real raw data use.

One other small change is that when looking for specex data (like lamplines), if $LIBSPECEX_DIR is not defined, it falls back to trying $SPECEX, which is a standard variable defined by our installed desispec modules.

I propose merging this and then rebasing the pixeltasks branch in PR #806 before doing final testing of that branch after Cori is back.


julienguy:
ok for me

tskisner:
Thanks for this work @sbailey .  Merge when ready and I'll rebase #806 off that for additional testing on cori tonight.

PR #809, opened 2019-09-04
Move common calculation of runtime distribution to a separate function

tskisner:
Do not merge this yet.  Test jobs are queued.

This change attempts to take the (duplicated in two places) runtime calculation of work distribution and isolate it with better comments and clearer program flow.


sbailey:
Looks good.  OK to merge when you are ready.

Issue #369, opened 2017-04-03
fiber flat-fielding with several lamp exposures


julienguy:
Code has to be developed for the relative calibration of the continuum lamp exposures prior to the fiber flat-fielding (see https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=2761).

julienguy:
Done in branch autocalib-flat

Each calibration lamp gives a non-uniform flat because of genuine plate scale and vignetting radial variations, but also because of its off-axis location. If lamps from the 4 boxes are turned on simultaneously, their variations of intensity would result in variations of flat field.
The solution is to derive a flat field for each lamp at a time and then combined the flats to minimize the spatial variations.
Example: 
The petal/spectrograph 7 has a 1% lower throughput.
lamp1 flat
![flat-lamp1](https://user-images.githubusercontent.com/5192160/65086991-cbc4a880-d968-11e9-95e8-07989966641d.png)
lamp2 flat
![flat-lamp2](https://user-images.githubusercontent.com/5192160/65087017-eac33a80-d968-11e9-85a4-95d2dbdfd357.png)
Auto-calibration based on the output fiberflats for each lamp and each spectrograph :
```desi_autocalib_fiberflat -i mod-fiberflat-r*-lamp*.fits --prefix ./fiberflat-r```
Result 
![flat-autocal](https://user-images.githubusercontent.com/5192160/65087097-38d83e00-d969-11e9-8743-841c2ef67b50.png)
Result after removing a radial pattern
![flat-autcal-corr](https://user-images.githubusercontent.com/5192160/65087103-4392d300-d969-11e9-929a-5fdfaa16f3c7.png)

PR #813, opened 2019-09-17
Auto-calibration of the fiberflats

julienguy:
See motivation and example in #369 .

Added 
* desi_autocalib_fiberflat
* scripts/autocalib_fiberflat.py

Modified
* desispec/io/fiberflat.py save header and fibermap (needed to get the focal plane coordinates of fibers)
* desispec/fiberflat.py save header,fibermap and add function autocalib_fiberflat

We will add this as a new task to the pipeline in another PR.


julienguy:
Ready for merge

PR #821, opened 2019-10-15
Fibermap xy

### LINKED ISSUE #819: DESIGN_X -> FIBERASSIGN_X
In desihub/fiberassign#228, we're renaming "DESIGN_X/Y" -> "FIBERASSIGN_X/Y" because there has been confusion about who was doing the designing (fiberassign vs. platemaker).

Update desispec to support both so that it will work with both old sims and new data.  This appears enough times in the code that we might want to intercept it and rename the column at an initial I/O layer rather than having N>>1 if/then checks scattered about the code.

sbailey:
This PR updates desispec to adapt to the fiberassign/fibermap rename of DESIGN_X/Y -> FIBERASSIGN_X/Y.  Fixes #819 

I attempted to retain backwards compatibility by intercepting and renaming the old column name in `desispec.io.read_frame()`.  I need to test that more but I'm opening a PR to get the ball rolling in the meantime.

In `desispec.io.fibermap.empty_fibermap` I added a few other columns that were missing (REF_EPOCH, PLATEMAKER_X/Y, ...) and removed a few that were removed upstream (DELTA_X/Y, DESIGN_Q/S, ...).  I found a bunch of columns in upstream targeting that are in the real fiberassign files that aren't in empty_fibermap but also aren't used by the pipeline.  I documented them in a comment but didn't go through the process of adding them all here.

### DIFF
--- a/py/desispec/coaddition.py
+++ b/py/desispec/coaddition.py
@@ -65,7 +65,7 @@ def coadd_fibermap(fibermap) :
                 tfmap['FIRST_'+k][i] = np.min(vals)
                 tfmap['LAST_'+k][i] = np.max(vals)
                 tfmap['NUM_'+k][i] = np.unique(vals).size
-        for k in ['DESIGN_X', 'DESIGN_Y','FIBER_RA', 'FIBER_DEC'] :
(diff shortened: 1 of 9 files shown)


sbailey:
Tested with integration minitest notebook and desisim/fiberassign_xy branch for inputs.  Merging now.

PR #825, opened 2019-10-21
Support extractions of data without fibermaps

sbailey:
This PR supports extracting spectra from data without a fibermap (e.g. early CMX arc and flat frames like 20191009/00018082).
  * Include a FIBERMAP HDU in the preproc file so that it becomes standalone input for the extractions without having to look back into the raw data directory for the fibermap.
  * If the fibermap doesn't exist while preprocessing, create a bare minimum fibermap with the `location:fiber:x:y` mapping for positioners at their home position, using `desimodel.io.load_focalplane` instead of the deprecated `desimodel.io.load_fiberpos`
  * extractions continue to support the `--fibermap` argument, but if not provided it will use what it finds in the input preproc file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,12 +56,12 @@ env:
         # - SCIPY_VERSION=0.16
         - ASTROPY_VERSION=2.0.14
         # - SPHINX_VERSION=1.6.6
-        - DESIUTIL_VERSION=master
+        - DESIUTIL_VERSION=2.0.1
         - SPECLITE_VERSION=0.7
         # - SPECTER_VERSION=0.8.1
         - SPECTER_VERSION=0.9.0
-        - DESIMODEL_VERSION=0.9.9
-        - DESIMODEL_DATA=branches/test-0.9.9
+        - DESIMODEL_VERSION=0.10.1
+        - DESIMODEL_DATA=branches/test-0.10
         # - DESIMODEL_DATA=trunk
         - DESITARGET_VERSION=0.29.1
         # - HARP_VERSION=1.0.1
--- a/py/desispec/io/fibermap.py
+++ b/py/desispec/io/fibermap.py
@@ -283,6 +283,11 @@ def read_fibermap(filename):
     #- to update the underlying format, extension name, etc. without having
     #- to change every place that reads a fibermap.
     fibermap = Table.read(filename, 'FIBERMAP')
+    if 'DESIGN_X' in fibermap.colnames:
+        fibermap.rename_column('DESIGN_X', 'FIBERASSIGN_X')
+    if 'DESIGN_Y' in fibermap.colnames:
+        fibermap.rename_column('DESIGN_Y', 'FIBERASSIGN_Y')
+
     return fibermap
 
 def fibermap_new2old(fibermap):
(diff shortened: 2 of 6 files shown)


sbailey:
Thanks for the testing.  It revealed two separate bugs in `desispec.io.empty_fibermap()` (!) which I have fixed and committed.  Mea culpa.

However, there is still a remaining problem for this case where you are trying to extract using the 40-fiber testslit PSF.  When creating the empty fibermap it is using the desimodel focalplane model which thinks that there are 500 fibers/spectrograph, while the PSF thinks there are 40 fibers/spectrograph and the bookkeeping gets messed up.  I'll think more about what to do with that case in the morning.

Even with 500-glorious fibers right around the corner, it seems reasonable to support spectroperfectionism extractions with the 40-fiber test slit too.

sbailey:
I believe this is now fixed and works both with 500 fiber and 40 fiber PSFs.  Please re-review.

julienguy:
Still an issue. It's not related to the number of fibers in the PSF.
```
desi_extract_spectra --wavelength 4000,4100,0.7 -o frame.fits -i preproc-b3-00019822.fits -p /software/datasystems/desi_spectro_calib/trunk/spec/sp3/psf-sm4-b-science-slit-20191015.fits 
Starting preproc-b3-00019822.fits spectra 0:500 at Tue Oct 22 14:34:59 2019
#--- Extraction Parameters ---
input:      preproc-b3-00019822.fits
psf:        /software/datasystems/desi_spectro_calib/trunk/spec/sp3/psf-sm4-b-science-slit-20191015.fits
output:     frame.fits
wavelength: 4000.0 - 4100.0 AA steps 0.7
specmin:    0
nspec:      500
regularize: 0.0
#-----------------------------    
  File "/software/datasystems/users/jguy/desispec/bin/desi_extract_spectra", line 16, in <module>
    extract.main(args)
  File "/software/datasystems/users/jguy/desispec/py/desispec/scripts/extract.py", line 249, in main
    chi2pix=chi2pix)
  File "/software/datasystems/users/jguy/desispec/py/desispec/frame.py", line 147, in __init__
    raise ValueError("len(fibers) != nspec ({} != {})".format(len(fibers), self.nspec))
ValueError: len(fibers) != nspec (0 != 500)
```

sbailey:
Did you re-generate your preproc file?  Part of the problem was the FIBERs assigned to the dummy FIBERMAP added to the preproc file so you need to rerun that step too.  With a re-generated preproc file I am able to extract using this branch, though I now see that it isn't including the dummy fibermap in the frame output.  Chasing that now...

sbailey:
Update: I was on a different branch than I thought I was.  With this preproc-fibermap branch it works for me, including outputing the FIBERMAP into the frame file, *if* you start over again and recreate the preproc file and not just use one that was previously created (including one that was previously created by an earlier version of this branch).

sbailey:
Thanks for the testing; onwards with downstream testing!

sbailey:
I was testing on a different exposure than your exact one.  Use that exact one now I see a complain about "zero" vs. "bias"...  Correcting that (zero as is in the header is correct and valid...)

PR #826, opened 2019-10-22
More flexible task partitioning

sbailey:
This PR implements a new task partitioning scheme across workers, which removes the previous constraint that only contiguous tasks could be grouped together in a worker (a constraint of the original "painter's partition problem" that doesn't apply for us).  This additional flexibility allows better load balancing.  Additionally, it optionally introduces the knowledge of the number of workers per node so that it can put the biggest tasks on different nodes instead of different ranks of the same node.

e.g. consider the case of work weights = [6,5,4,3,2,1] spread across 3 workers.  The previous `dist_discrete_all` code distributed these as
```
Worker  sum(weights)  weights
    0          6      [6]
    1          9      [5 4]
    2          6      [3 2 1]
```
while the new `weighted_partition` code distributes them as
```
Worker  sum(weights)  weights
    0          7      [6 1]
    1          7      [5 2]
    2          7      [4 3]
```

Also consider the case of weights=[2,2,2,1,1,1,1,1,1,] spread across 3 nodes with 2 workers each.  Using weighted_partition(groups, 6, groups_per_node=2):
```
Worker  sum(weights)  weights
    0          2      [2]
    1          2      [1 1]
    2          2      [2]
    3          2      [1 1]
    4          2      [2]
    5          2      [1 1]
``` 
Note that the big (weight=2) tasks are distributed across different nodes instead of putting two of them on different ranks of the same node.

To allow for non-contiguous groupings, the return value is a list of lists of indices to include in each group, instead of a list of (first_index, num_items) tuples.  I left the original dist_discrete_all algorithm for comparison and in case we needed that for some other case, but updated its return values to match so that they are drop in replacements for each other.

I'd like to do more real-world-in-the-pipeline testing at NERSC before merging this, but I'm opening this PR now for review and Travis testing.


sbailey:
For the record and as a starting point of other tests, the code for the examples here:
```python
import numpy as np
from desispec.parallel import weighted_partition, dist_discrete_all

weights = [6,5,4,3,2,1]

groups_old = dist_discrete_all(weights, 3)
groups_new = weighted_partition(weights, 3)

def summarize(weights, groups):
    weights = np.asarray(weights)
    print('Worker  sum(weights)  weights')
    for i in range(len(groups)):
        groupweights = weights[groups[i]]
        sumweights = np.sum(groupweights)
        print('{:5d}   {:8.0f}      {}'.format(i, sumweights, groupweights))

summarize(weights, groups_old)
summarize(weights, groups_new)

#- distributing big tasks across different nodes
weights=[2,2,2,1,1,1,1,1,1,]
groups = weighted_partition(weights, 6, groups_per_node=2)
summarize(weights, groups)
```

tskisner:
These changes look structurally fine to me and I have no problem with changing the distribution algorithm.  If it solves our sporadic OOM conditions then that is even better :-)

PR #838, opened 2019-10-26
Overscan Improvement

profxj:
This implements a new method for overscan subtraction which filters the overscan row with a Savgol.

See this Notebook for more:
https://nbviewer.jupyter.org/github/desihub/desispec/blob/overscan/doc/nb/Overscan.ipynb

Here is a screen shot of the results on the ZERO frame 20655 from 20191024:
zero_20191024

and here is a screen shot for FLAT (camera r3) 20675 from 20191024:
flat_20191024


julienguy:
This looks very nice. Can you add the choice of method (simple clipped mean or Savitsky-Golay filter) and the parameters of the filter as optional parameters of the preprocessing (with your best choice as default values)? If the best choice of parameters vary from CCD to CCD, we can add this  to the calibration yaml files on svn ($DESI_SPECTRO_CALIB).

profxj:
Happy to do so, but you'll need to dumb it down for me (i.e. explain how to do this).
Could be faster for you to do it yourself.  Your choice.

profxj:
Performance on expid 20987, r3
r3_20987

profxj:
Ratio of new/old (where I'm not sure 'old' is exactly the version in `master`).
Am finding it hard to see things in the flat.  Trying a `bias` now.
new_over_current

profxj:
Bias **without** savgol.  Note the striping.
bias_nosavgol

profxj:
Bias with savgol.   No striping.

Not sure what else to say at this point..

bias_savgol

profxj:
@julienguy -- I think this is ready to go (once it passes Travis)

profxj:
gack..
Travis needs power at LBNL for tests to pass.
So I guess we stand by.

sbailey:
if @julienguy agrees with the changes, ok to merge without waiting for LBL power + Travis.  I've added this to 19.10 project to keep it on our radar

julienguy:
I think we need the option to fall back to a simple average per row, or a parameter to tune the correlation length of the filter, because this corrects for trends in the overscans but at the cost of increased statistical noise.

profxj:
I agree with @julienguy 
Almost done fussing.
Will let you know when.

profxj:
Ok, there are now 2 options:

1. savgol -- Default
2. Subtract the median value of the overscan rows, i.e. the method in `master` right now.

Neither does a great job in some amps, see the bottom of this Notebook:
https://nbviewer.jupyter.org/github/desihub/desispec/blob/overscan/doc/nb/Overscan.ipynb

Could be the overscan column routine, could be temporal variations
in the bias.  Not sure.  But I'm not going to address them in this PR.

@julienguy -- I'd give a fresh look at the whole PR.  The masking and
replacement in `_savgol_clipped` is not ideal but will be fine for non edge-cases.

julienguy:
In the example in the jupyter notebook https://nbviewer.jupyter.org/github/desihub/desispec/blob/overscan/doc/nb/Overscan.ipynb, it seems the savgol filter correction introduces noise at a typical scale of ~100 pixels (see wavy pattern for the blue curve in the x range 500:1500) ; and it also has longer range systematic residuals.

julienguy:
I just did at nersc on a zero :
```
desi_preproc -i ~/data/20191029/00022481/desi-00022481.fits.fz --cam b3,r3,z3
```
with the master version and this branch and then
```
   img=fitsio.read(filename)
   prof=np.median(img[3000:3500],axis=0)
   plt.plot(prof,label=filename,alpha=0.6)
```
and I get the following figure that shows an increase fluctuation at a scale of  ~100pix with the 'overscan' branch. It also shows with both versions a global offset of about ~0.2 electron in an amplifier. 

top

profxj:
After a deeper dive into bias frames including analysis of minisv2 data, 
I have generated [this doc](https://docs.google.com/document/d/1ZLgQGxs0kGPKiwG46VKMPNZNL7GXCuO5cHZIVwrAPfo/edit?usp=sharing) summarizing a recommended procedure.

The net change is to have ignoring ORSEC become the default.
This is toggled by `use_overscan_row` in desispec.preproc.preproc()

julienguy:
Getting an error:
```
desi_preproc -i /global/cfs/cdirs/desi/spectro/data/20200315/00055694/desi-00055694.fits.fz
  File "/global/common/software/desi/users/jguy/desispec/bin/desi_preproc", line 8, in <module>
    preproc.main(preproc.parse())
  File "/global/common/software/desi/users/jguy/desispec/py/desispec/scripts/preproc.py", line 110, in main
    if args.flag_savgol is not None:
AttributeError: 'Namespace' object has no attribute 'flag_savgol'

```

profxj:
@julienguy -- Yup, sorry about that.  
Failed to run tests on my machine before pushing.

Now fixed and I ran your command successfully on `cori`

julienguy:
I ran `desi_preproc -i /global/cfs/cdirs/desi/spectro/data/20200315/00055694/desi-00055694.fits.fz
` using the master and overscan branches.
I then compare the two sets of preprocessed zeros  with a mean profile on an horizontal band below the center of the CCD. The overscan branch results are for some CCD better and worse for others. See examples below. 
I'm ok to merge so we have the options in the code but I am not sure we have reached the best solution yet. We should however definitely recompute master bias and darks more frequently.

zero-r3-00055694
zero-r5-00055694
zero-z2-00055694
zero-z5-00055694
zero-z8-00055694

PR #839, opened 2019-10-27
Use OBSTYPE in desi_qproc

julienguy:
Modified desi_qproc to use the new OBSTYPE keyword instead of FLAVOR used by ICS.


julienguy:
Some Travis tests fail because desi SVN is down.

sbailey:
Tested at KPNO with nightwatch.  Works great.  Thanks.

PR #841, opened 2019-10-29
Fixes double application of fiberflat

### LINKED ISSUE #840: desi_proc double-applying fiberflat?
desi_proc uses the fiberflat corrected flux for determining outliers before selecting random sky fibers.  It then updates the fibermap and replaces the frame file.  However, it appears that it is replacing the frame file with the fiberflat corrected flux (not just replacing the fibermap), but sky subtraction doesn't know that and applies the fiberflat again.  Oops.

Confirm that, and correct if needed (probably)

sbailey:
This PR fixes #840 where desi_proc was double applying the fiberflat.  Tested @ KPNO.

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -11,6 +11,7 @@ progress = OrderedDict(start=time.asctime())
 
 import sys, os, argparse, re, time, collections
 import subprocess
+from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


sbailey:
@julienguy says "please merge" on Slack.

PR #842, opened 2019-10-29
Adding a few options to desi_proc

julienguy:
```
  --nofiberflat         Do not apply fiberflat
  --noskysub            Do not subtract the sky
  --psf PSF             use this input psf (trace shifts will still be computed)
  --fiberflat FIBERFLAT use this fiberflat
```


sbailey:
Changes look good; not waiting for Travis tests to fail due to LBL power outage.

PR #843, opened 2019-10-29
qproc use correct fibermap

sbailey:
This PR consolidates how the empty fibermap is created so that qproc and full processing are consistent.  Review carefully.

in particular, the empty fibermap creation is moved into io.read_raw instead of scripts.preproc (because qproc doesn't call scripts.preproc directly).

HACK: in lieu of calibfinder, desimodel, or header keywords actually knowing which petal we're plugged into, this PR adds the mapping.  When calibfinder (or others) know, this could would need to be updated anyway, so I added the mapping in the place that would need to be changed anyway.  Caveat emptor.

Do not merge until:
  * tested with desi_proc
  * tested with nightwatch

But reviews prior to that are welcome too...

### DIFF
--- a/py/desispec/io/raw.py
+++ b/py/desispec/io/raw.py
@@ -92,7 +92,7 @@ def read_raw(filename, camera, fibermapfile=None, **kwargs):
 
     img = desispec.preproc.preproc(rawimage, header, primary_header, **kwargs)
 
-    if fibermapfile is not None:
+    if fibermapfile is not None and os.path.exists(fibermapfile):
         fibermap = desispec.io.read_fibermap(fibermapfile)
     else:
         log.warning('creating blank fibermap')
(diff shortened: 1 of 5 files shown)


sbailey:
Fix before merging: Full extractions fail due to mismatch of fibermap['FIBER'] vs. the fiber list derived based on the camera plus the combination of `--fibermap-index`, `--specmin`, and `--nspec` options.  Previously the fibermap file had a full 5000 fibers, while the extraction was for just one camera, so it used the camera+options to derive the array of fibers it wanted then then filtered the fibermap to match.  In this branch it is filtering down to a null set...

sbailey:
I have tested this at both KPNO and NERSC using ARC, FLAT, SKY, ZERO, and TWILIGHT exposures with desi_proc and nightwatch, e.g. with outputs in kpno:/exposures/nightwatch/redux/sbailey/test.  @julienguy please take a look.

The nightwatch webapp is having trouble finding the spectra in these files (probably another fiber vs. spectrograph bookkeeping issue), but I think that is something to fix on the nightwatch side and we could go ahead and merge this to enable running more exposures with desi_proc.  I'm going to go chase the nightwatch side now.

PR #844, opened 2019-10-30
Database 19.9

weaverba137:
This PR implements changes needed to get the redshift database tutorial working.  However, the tutorial is currently only using `targets-dark.fits` and `truth-dark.fits`, rather than all available target & truth files.  I'd like to incorporate the bright files if possible.


sbailey:
Thanks.  We'll probably be breaking and fixing this several times throughout integration with ICS and into SV and the datamodel converges.

PR #853, opened 2019-12-12
Adapt to SMX naming

julienguy:
All is in the title.
spectro calib SVN repo has also been updated.


sbailey:
Thanks.  Spot tested on new and old data and it found a PSF calib file.

PR #857, opened 2020-01-21
[WIP] running the pipeline on commissioning data

moustakas:
[WIP] Changes needed in order to be able to run the pipeline on (early) commissioning data / spectroscopy.

Related to https://github.com/desihub/desicmx/pull/10.

PR #859, opened 2020-01-28
Scattered light

julienguy:
Code to subtract scattered light at the preprocessing level.
```
desi_preproc --scattered-light (--psf PSFFILE) ...
desi_proc --scattered-light ...
````

The method consists in:
 - computing a "direct light" mask  centered on the spectral traces of a preprocessed CCD image using a PSF file
 - computing a model M = (image * mask ) x K , where K is a tuned convolution kernel adjusted on arc images
 - evaluating a calibration polynomial of CCD rows in the space between fiber bundles where there is only scattered light
 - the final model is interpolated in the fiber bundles

The algorithm is in `desispec/scatteredlight.py`

Example (blue is the profile of a arc lamp image in NIR for the spectro SM4, and orange is the model)
Figure_1

I still need to optimize the kernel for different cameras so this is not yet ready for merging.


sbailey:
Deprecated py3.5 tests failed; fixed in PR #857; merging this one now.

PR #860, opened 2020-01-28
add --calibnight option

julienguy:
Add --calibnight option to desi_proc


sbailey:
Famous last words: it's so obviously right and previously tested that we don't need to wait for Travis...

PR #865, opened 2020-02-04
Calibration

julienguy:
Improved robustness of sky subtraction and flux calibration to poorly constrained wavelength regions and dead fibers, selection of standard stars. Also mask out the region between 4300 and 4500A with the throughput dip in the fit of standard stars.

For illustration, throughput per spectrograph of one exposure on a dither sequence (exposure 44565 from 2020/01/27):

flux-calib-44565


sbailey:
It looks like new fluxcalibration quality cuts are causing the `desi_compute_fluxcalibration` tests to discard all stars and thus fake-fail.  Let's try to find a fix for that, or otherwise disable the test before merging if we have to.

PR #869, opened 2020-02-05
more efficient parallelism packing

sbailey:
This PR improves the parallelism packing of `desi_proc ... --batch` scripts.

Previously it was run as a single srun command tuned for the extraction step, but this resulted in the later multiprocessing-based `desi_fit_stdstars` being every inefficient because most of the cores were tied up by MPI doing nothing instead of being available for multiprocessing.  This new version splits the script into two srun calls: one massively parallel srun for everything through sky subtraction, and then a separate call with less MPI parallelism but more multiprocessing parallelism for flux calibration.

Additionally, this packs the job into 5 nodes instead of 7, to allow two jobs to run simultaneously in the realtime queue.  extractions become 1.5 minutes slower, while flux calibration becomes 8 minutes *faster*.

Bottom line: with this we can run 2.7x faster in wall clock, while being charged 25% less for the queue time.

skysub and fluxcalib could still benefit from MPI integration (combined they take longer than extractions, which used to be the tall pole), but this PR is still a big step forwards for catchup runs using desi_proc and the realtime queue.

I'm using this branch for minisv0c processing.

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -183,7 +183,7 @@ if args.batch:
     elif args.obstype in ('FLAT', 'TESTFLAT'):
         ncores, runtime = 20*nspectro, 25
     elif args.obstype in ('SKY', 'TWILIGHT', 'SCIENCE'):
-        ncores, runtime = 20*nspectro, 30
+        ncores, runtime = 16*nspectro, 30
     elif args.obstype in ('ZERO', 'DARK'):
         ncores, runtime = ncameras, 5
     else:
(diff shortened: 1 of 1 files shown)


sbailey:
Timing notes for the record (in minutes per step, for 20200127 expid 44518):
```
Step        master  thisPR
preproc      1.0    0.7
traceshift   0.6    0.4
extract      4.8    6.2
skysub       4.7    4.8
fluxcalib   11.6    3.6
applycalib   0.7    0.8
```

sbailey:
This worked well for minisv0c processing, so self merging.

PR #870, opened 2020-02-06
Add desi_proc_dashboard.py

zkdtc:
Usage in the description in the beginning part of the code.


py/desispec/desi_proc_dashboard.py
```diff
@@ -0,0 +1,383 @@
...
+from astropy.io import fits
+import subprocess
+import pandas as pd
+import time,datetime
+import numpy as np
+import psycopg2
+import hashlib
+import pdb
+import psutil
+from os import listdir
```
sbailey: Some of these imports aren't used anymore.  Let's remove them to simplify dependencies and load time.  e.g. astropy.io.fits, pandas, psycopg2, hashlib, pdb, subprocess.

py/desispec/desi_proc_dashboard.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def _init_parser(self,parser):
+        parser.add_argument('-n','--nights', type=str, default = None, required = False, help="nights to monitor")
+        parser.add_argument('-nn','--n_night', type=str, default = None, required = False, help="all:all nights. ifdigit: the last n nights.")
+        parser.add_argument('-pd','--prod_dir', type=str, default = None, required = True, help="product base directory")
+        parser.add_argument('-od','--output_dir', type=str, default = None, required = True, help="output portal directory for the html pages ")
+        parser.add_argument('-ou','--output_url', type=str, default = None, required = True, help="output portal directory url ")
+        return parser
+
+    def calculate_one_night(self,night):
```
sbailey: For each function, please add docstrings describing what the function does, what the inputs are, what the function returns, and if there are any side effects (e.g. updating the state of the object).  See https://desi.lbl.gov/trac/wiki/Computing/Software/Guidelines#PythonCodingStyle

py/desispec/desi_proc_dashboard.py
```diff
@@ -0,0 +1,383 @@
...
+        a=psutil.process_iter()
+        running='No'
+        for p in a:
+            if 'desi_dailyproc' in ' '.join(p.cmdline()):
+                running='Yes'
+        return running
+
+        
+if __name__=="__main__":
+    process=DESI_PROC_DASHBOARD()
```
sbailey: Please add a wrapper script in `desispec/bin/desi_proc_dashboard` (no .py on the end) that just loads and calls this code, so that when desispec is installed that script will be in the path.

py/desispec/desi_proc_dashboard.py
```diff
@@ -0,0 +1,383 @@
...
+    def __init__(self):
+        ############
+        ## Input ###
+        ############
+        parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+        parser = self._init_parser(parser)
+        args = parser.parse_args()
+        if not os.getenv('DESI_SPECTRO_REDUX'):
+            os.environ['DESI_SPECTRO_REDUX']='/global/project/projectdirs/desi/spectro/redux/'
+            os.environ['DESI_SPECTRO_DATA']='/global/project/projectdirs/desi/spectro/data/'
```
sbailey: default should be `/global/cfs/cdirs/desi/spectro/redux` and `data` now instead of project.

py/desispec/desi_proc_dashboard.py
```diff
@@ -0,0 +1,383 @@
...
+        if args.nights=='all':
+            nights=listdir(os.getenv('DESI_SPECTRO_REDUX')+'/daily/exposures/')
```
sbailey: Throughout the code, use `args.prod_dir` with that having a default of `desispec.io.specprod_root()` (which currently just returns `$DESI_SPECTRO_REDUX/$SPECPROD), instead of hardcoding "daily" as the only production name that this can be used on.  Similarly use `desispec.io.rawdata_root()` to get the location of the raw data.

Elsewhere it looks like you are using `args.prod_dir` as the equivalent of `$DESI_ROOT`, i.e. `/global/project/projectdirs/desi` -> `/global/cfs/cdirs/desi`.  Let's move the meaning of prod_dir a few levels deeper (i.e. all the way into the spectro/redux/daily/ directory, without hardcoding that name).

py/desispec/desi_proc_dashboard.py
```diff
@@ -0,0 +1,383 @@
...
+            ####################################
+            strTable=strTable+self._add_html_table(stat_night,str(night))
+            
+        timestamp=time.strftime("%Y-%m-%d %H:%M:%S",time.localtime())
+        #print(timestamp)
+        running=self.check_running()
+        strTable=strTable+"<div style='color:#00FF00'>"+timestamp+" "+"desi_dailyproc running: "+running+"</div>"
+        strTable=strTable+self._add_js_script1()
+        strTable=strTable+"</html>"
+        hs=open(self.output_dir+"desi_proc_dashboard.html",'w')
```
sbailey: use `os.path.join` or otherwise include a `/` so that `output_dir` doesn't have to include a `/` as part of the user input.

py/desispec/desi_proc_dashboard.py
```diff
@@ -0,0 +1,383 @@
...
+        ##########################
+        cmd="chmod -R a+xr "+self.output_dir
+        os.system(cmd)
+
+    def _init_parser(self,parser):
+        parser.add_argument('-n','--nights', type=str, default = None, required = False, help="nights to monitor")
+        parser.add_argument('-nn','--n_night', type=str, default = None, required = False, help="all:all nights. ifdigit: the last n nights.")
+        parser.add_argument('-pd','--prod_dir', type=str, default = None, required = True, help="product base directory")
+        parser.add_argument('-od','--output_dir', type=str, default = None, required = True, help="output portal directory for the html pages ")
+        parser.add_argument('-ou','--output_url', type=str, default = None, required = True, help="output portal directory url ")
```
sbailey: It is non-standard python options to have a single-dash options that requires 2 letters, e.g. `-nn`.  It is ok to have options that only have double-dash `--output_url` without having any single-dash option.

Let's make default `args.prod_dir = desispec.io.specprod_root()` and add a new option `args.rawdata_dir` with default `desispec.io.rawdata_root()`.

py/desispec/desi_proc_dashboard.py
```diff
@@ -0,0 +1,383 @@
...
+            except:
+                obstype='None' #h2=fitsio.read_header(filename,2)
+            #print(expid,flavor,obstype)
+            fileglob_psf='/project/projectdirs/desi/spectro/redux/daily/exposures/'+str(night)+'/'+str(expid).zfill(8)+'/psf*.fits'
+            fileglob_fit_psf='/project/projectdirs/desi/spectro/redux/daily/exposures/'+str(night)+'/'+str(expid).zfill(8)+'/fit-psf*.fits'
+            fileglob_fiberflat='/project/projectdirs/desi/spectro/redux/daily/exposures/'+str(night)+'/'+str(expid).zfill(8)+'/fiberflat*.fits'
+            fileglob_frame='/project/projectdirs/desi/spectro/redux/daily/exposures/'+str(night)+'/'+str(expid).zfill(8)+'/frame*.fits'
+            fileglob_sframe='/project/projectdirs/desi/spectro/redux/daily/exposures/'+str(night)+'/'+str(expid).zfill(8)+'/sframe*.fits'
+            fileglob_cframe='/project/projectdirs/desi/spectro/redux/daily/exposures/'+str(night)+'/'+str(expid).zfill(8)+'/cframe*.fits'
+            fileglob_sky='/project/projectdirs/desi/spectro/redux/daily/exposures/'+str(night)+'/'+str(expid).zfill(8)+'/sky*.fits'
```
sbailey: Please use the script options to generate these paths, e.g. `--prod_dir`, with a default of `prod_dir = desispec.io.specproc_root()`.

In the end, the code should not refer to "project" or "cdirs" anywhere, and preferably not `$DESI_SPECTRO_REDUX` either (using `specprod_root` and `rawdata_root` instead).

py/desispec/desi_proc_dashboard.py
zkdtc: Done

py/desispec/desi_proc_dashboard.py
zkdtc: Done

py/desispec/desi_proc_dashboard.py
zkdtc: Fixed

py/desispec/desi_proc_dashboard.py
zkdtc: Realize that prod_dir is not used in fact, thus remove it.

py/desispec/desi_proc_dashboard.py
zkdtc: Fixed.

py/desispec/desi_proc_dashboard.py
zkdtc: prod_dir is in fact useless because the code only checks the redux directory, sorry about the confusion...

zkdtc: I remove all '--xx' options.

py/desispec/desi_proc_dashboard.py
zkdtc: args.prod_dir is not pointing to desispec.io.specprod_root()

py/desispec/desi_proc_dashboard.py
zkdtc: Realize that the code does not use raw data, thus prod_dir is changed to point to desispec.io.specrpod_root() instead of the raw data directory.

zkdtc:
Have finished all requested revisions, please check :-D

sbailey:
What is desi_proc_time_distribution.py?

It looks like you haven't commited the wrapper scripts to bin/ yet.

Please remove the unused pandas import.

`--output_url` is a required option, but appears to be unused.  Please remove it or otherwise clarify what it is used for.

The `--n_nights` option says "ifdigit: the last n nights" but that doesn't seem to work.  I tried this to get the report for the last 5 nights:
```
thon py/desispec/desi_proc_dashboard.py -n 5 --output_dir /global/cfs/cdirs/desi/www/users/sjbailey/dailyprod/ --output_url blat
```
and got
```
Searching /global/cfs/cdirs/desi/spectro/redux/daily
  File "py/desispec/desi_proc_dashboard.py", line 351, in <module>
    process=DESI_PROC_DASHBOARD()
  File "py/desispec/desi_proc_dashboard.py", line 55, in __init__
    if not tonight in nights:
TypeError: argument of type 'int' is not iterable
```
Please fix and/or clarify the correct usage.

sbailey:
nevermind about `--n_nights`.  I was using the wrong shortcut.

sbailey:
Actually, using `--n_nights 5` didn't work either.  Please test and clarify usage.

I also didn't figure out the syntax to specify multiple specific nights with the `--nights` option.  Looking at the code I see you are using `locals` and `exec` which should generally be avoided.  Better to use either:
  * `nargs="*"` in add_argument to support `--nights 20200201 20200202 20200203", or
  * Support comma separated list of nights like `--nights 20200201,20200202,20200203`
```
nights = [int(night) for night in args.nights.split(',')]
```

zkdtc:
--output_url is removed, and now the code can accept '-n all', '-n 20200101,20200103' and '-n 20200101 20200103'. 
desi_proc_time_distribution is my code to make job elapsed time distribution plots. Might not be suitable for here?

sbailey:
It looks like the `--output_url`, `--n_nights`, and `bin/desi_proc_dashboard` changes haven't been pushed to github yet.  Please add those.

desi_proc_time_distribution: In general if something is a single-use analysis, it should go elsewhere (maybe in doc/analysis/, or a different repo or even an attachment to a docdb report).  If it is intended to be run as part of the pipeline operations, then it is ok to have here.  Monitoring job/step runtime is probably something that would be useful to run by default, so let's leave this here for now.

zkdtc:
Shoot, forgot to push desi_proc_dashboard... bin/desi_proc_dashboard claims to be up to date, and it seems OK to me :-)

sbailey:
Thanks.  I made one more change to search the raw data directory for the list of nights, rather than the reduction directory, which might be completely missing a night.  Will merge after tests pass.

PR #872, opened 2020-02-07
Added desi_dailyproc to bin

akremin:
Added desi_dailyproc to bin with file read-in capabilities for catchup nights. Also includes a catchup_list.csv file that can be hardcoded to go anywhere. The list is currently placed in the same /bin/ directory for now.

Also handles environment variables better. Allows user to override default SPECPROD value. If SPECPROD_OVERRIDE exists in the user's environment, then the defaults will not be used and the user's specified SPECPROD env value will be used.


bin/catchup_list.csv
```diff
@@ -0,0 +1 @@
+20200124
```
sbailey: I'm not sure if this was intentional or not, but let's *not* put the catchup list in bin/, and not in github at all, since we want to be able to edit it on the fly without changing the installed version.  It could be useful to have a command line option to override the default "catchup_list.csv" name.

Good that you have it setup so that if catchup_list.csv doesn't exist it just moves on with and empty list instead of crashing.

akremin:
I removed the text file from the git repo. 
I added multiple command line arguments that allow for better flexibility and debugging. This includes options to turn off the cori21 check and the running instances checks. Allows you to force it to use a specific SPECPROD dir. You can also define all relevant filenames and paths. Finally you can set the pause time and number of submissions during each wake cycle.
Code still works by reading in a catchup list, which defaults to catchup_list.csv but can be specified as anything.

I use argparse, so there is a --help option with relevant information on each of these.

PR #875, opened 2020-02-13
more robust to missing columns in coordinates files

sbailey:
This PR makes `assemble_fibermap` more robust to missing input columns from platemaker coordinates files.  e.g. 20200212/00048337/coordinates-00048337.fits doesn't have any FVC-related columns, e.g. FLAGS_CNT_0 or FVC_X_0.


sbailey:
Self merging to be able to catch up on 20200212 data

PR #878, opened 2020-02-16
Cosmics

sybenzvi:
A `repair_mask` utility is available to patch holes in cosmic ray tracks, based on the binary closure algorithm discussed in DESI-5489. At the moment it is only reading the preproc FITS files and producing PNG outputs for visual inspection; some functionality needs to be moved into `desispec.cosmics` to patch the cosmics mask during preprocessing before the pull request can be merged. Meanwhile @sybenzvi and @dylanagreen will keep the branch up to date with the master.


sbailey:
Thanks.  This doesn't alter any of the standard running (good during minisv), so I'll merge it now so you don't get too far out of date with master.

PR #880, opened 2020-02-19
Bug fix in flux calibration

julienguy:
Fix indexation BUG in the standard stars spectra when computing the flux calibration vectors (sorted according to their S/N in desi_fit_stdstars while assumed sorted with increasing fiber number in desi_compute_fluxcalibration). This bug has been introduced recently.


sbailey:
wow.  If I understand correctly, this new code makes it robust to whatever order they are in.  Is that correct?

PR #881, opened 2020-02-20
Dailyproc setcams

akremin:
Can now specify what cameras to process and whether to do scattered light corrections or not at the level of desi_dailyproc. The options are passed down to desi_proc.

### DIFF
--- a/bin/desi_dailyproc
+++ b/bin/desi_dailyproc
@@ -55,35 +55,35 @@ def parse(options=None):
                              "After 'nsubmits,' the script will wait 'pausetime'"+
                              "minutes before submitting another 'nsubmits.'")
 
-    parser.add_argument("--cameras", type=str, required=False, default='None',
+    parser.add_argument("--cameras", type=str, required=False,
                         help="Explicitly define the spectrographs for which you want"+
                              " to reduce the data. Should be a comma separated list."+
                              " Numbers only assumes you want to reduce R, B, and Z "+
                              "for that camera. Otherwise specify separately [BRZ|brz][0-9].")
 
     # File and dir defs
-    parser.add_argument("-c", "--catchupfile", type=str, required=False, default='None',
+    parser.add_argument("-c", "--catchup-file", type=str, required=False,
                         help="Relative path+name for catchup file. Automatically "+
                              "triggers '--catchup' to be true.")
-    parser.add_argument("-s", "--specprod", type=str, required=False, default='None',
+    parser.add_argument("-s", "--specprod", type=str, required=False,
                         help="Directory name where the output files should be saved.")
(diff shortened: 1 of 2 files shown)


bin/desi_dailyproc
```diff
@@ -172,6 +195,18 @@ def main(args):
...
+    ###############################                                                                                                                         
+    ### Define desi_proc params ###                                                                                                                             
+    ############################### 
+    cmd_base = 'desi_proc '
+    cmd_base += '--batch '
+    cmd_base += '--traceshift '
+    if args.scatteredlight:
+        cmd_base += '--scattered-light '
+    if args.cameras.lower() != 'none':
+        cmd_base += '--cameras {} '.format(args.cameras)
```
sbailey: Let's use the default args pattern used elsewhere in desi code:  don't specify the default value, in which case it will be `None` (different than the string `"None"`), and then check with code like:
```python
if args.cameras is not None:
    cmd_base += '--cameras {} '.format(args.cameras) 
```
That way you can also avoid the parsing of the cameras argument on lines 95-110 just to make it back into a string that desi_proc can re-parse.

bin/desi_dailyproc
```diff
@@ -83,16 +92,30 @@ def parse(options=None):
         args = parser.parse_args()
     else:
         args = parser.parse_args(options)
+
+    if args.catchupfile.lower() != 'none':
+        args.catchup = True
+
+    if args.cameras.lower() != 'none':
```
sbailey: See comment below about default `None` vs. `"None"`, which applies to both `args.catchupfile` and `args.cameras`.

bin/desi_dailyproc
```diff
@@ -54,6 +54,13 @@ def parse(options=None):
                         help="Number of submissions to make to cori at a time."+
                              "After 'nsubmits,' the script will wait 'pausetime'"+
                              "minutes before submitting another 'nsubmits.'")
+
+    parser.add_argument("--cameras", type=str, required=False, default='None',
+                        help="Explicitly define the spectrographs for which you want"+
+                             " to reduce the data. Should be a comma separated list."+
+                             " Numbers only assumes you want to reduce R, B, and Z "+
+                             "for that camera. Otherwise specify separately [BRZ|brz][0-9].")
```
sbailey: See comment below about default `None` vs. `"None"` and camera parsing.  I like the idea of allowing a single number to mean b,r,z for that spectrograph (thanks for documenting!).  If that is more flexible parsing than what `desi_proc` supports, let's move that into `desi_proc` itself so that the `--cameras` parsing is only done in one place and is consistent, and `desi_dailyproc` is just passing forward whatever it gets to `desi_proc`.

sbailey:
For the record: ok to self merge after those changes are made to get it in place for tonight.

PR #883, opened 2020-02-21
assemble_fibermap --force option

sbailey:
Adds `assemble_fibermap --force` option to keep going and make a basic fibermap even if missing the input coordinates and guide files.

Tested with:
```
assemble_fibermap -n 20200219 -e 51030 -o $SCRATCH/blat.fits
```
--> fails due to missing coordinates file

```
assemble_fibermap -n 20200219 -e 51030 -o $SCRATCH/blat.fits --force
```
--> complains about missing input but moves on anyway and creates a basic fibermap with some zeroed out columns of missing info.

Intended usage:
* standard desi_proc pipeline should expect the required inputs and will still fail if they are missing so that a human will notice
* if we decide that we really want those exposures anyway
  * rerun `assemble_fibermap --force ...` for that exposure
  * then resubmit desi_proc which will notice that the fibermap now exists, skip that step, and keep going

I tested the rest of the pipeline on 20200219/51030 to confirm that the missing coordinates columns don't cause the pipeline to fail.  I have not, however, tested it on a case of a missing guide file which might cause problems with missing keywords, so if cases like that appear we may need to have a followup PR.

@akremin please review.

### DIFF
--- a/bin/assemble_fibermap
+++ b/bin/assemble_fibermap
@@ -15,10 +15,11 @@ parser.add_argument("-e", "--expid", type=int, help="spectroscopic exposure ID")
 parser.add_argument("-o", "--outfile", type=str, help="output filename")
 parser.add_argument("--debug", action="store_true", help="enter ipython debug mode at end")
 parser.add_argument("--overwrite", action="store_true", help="overwrite pre-existing output file")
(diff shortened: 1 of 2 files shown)


akremin:
Worked with a second example exposure with a missing coordinates ID (51059).

Issue #889, opened 2020-02-24
automate tile grouping, coadds, and redshift fitting


sbailey:
The standard pipeline automates grouping spectra by healpix and running redshifts, but for (mini)SV we're running `desi_group_tileframes` to group by tiles/nights, and then `desi_coadd_spectra` and running `rrdesi` on that.  Automate that process.

What I ran for 20200219 by hand:
```bash
#- Group by tile/night
cd /global/cfs/cdirs/desi/spectro/redux/daily
desi_group_tileframes -i exposures/20200219 -o tiles

#- Generate coadds per-spectrograph per-tile per-night
export NIGHT=20200219
for TILEID in 63075 70003 70004 70005; do
    cd /global/cfs/cdirs/desi/spectro/redux/daily/tiles/$TILEID/$NIGHT
    for SPECTRO in 0 3 6 7 9; do
        coadd=coadd-${SPECTRO}-${TILEID}-${NIGHT}.fits
        if [ ! -f $coadd ]; then
            desi_coadd_spectra --coadd-cameras --nproc 8 \
                -i cframe-?${SPECTRO}-*.fits \
                -o coadd-${SPECTRO}-${TILEID}-${NIGHT}.fits
        else
            echo $coadd exists
        fi
    done
done

#- On a 4 node interactive job, ~1.5 min per coadd:
salloc -N 4 -C haswell -t 04:00:00 -q interactive
export NIGHT=20200219
for TILEID in 63075 70003 70004 70005; do
    cd /global/cfs/cdirs/desi/spectro/redux/daily/tiles/$TILEID/$NIGHT
    for SPECTRO in 0 3 6 7 9; do
        zbest=zbest-${SPECTRO}-${TILEID}-${NIGHT}.fits
        if [ ! -f $zbest ]; then
            echo running redrock for $zbest
            time srun -N 4 -n 128 -c 2 rrdesi_mpi \
                -z $zbest \
                -o redrock-${SPECTRO}-${TILEID}-${NIGHT}.h5 \
                coadd-${SPECTRO}-*.fits
        else
            echo $zbest exists
        fi
    done
done
```

Using a single node (instead of 4) takes ~3 min per coadd, and these could be spawned as N independent single-node jobs per night in the realtime queue.  The previous steps of grouping and coadding per night are fast enough they could be done on the interactive node prior to submitting the redshift jobs.

sbailey:
[...]

PR #892, opened 2020-02-27
add script to run nightly redshifts per tile

sbailey:
Adds a script `desi_nightly_redshifts` to run coadds and redshifts per tile per night (issue #889).  This is intended to be run from a batch job (the script itself runs srun multiple times), or to have it create a batch script and submit it:
```
desi_nightly_redshifts --batch YEARMMDD
```
will submit a batch script to the realtime queue to run:
  * desi_group_tileframes
  * desi_coadd_spectra
  * rrdesi (redrock)

It auto-recognizes what tiles have already been run, so you can delete coadds/zbest/redrock files in $DESI_SPECTRO_REDUX/$SPECPROD/tiles/{TILEID}/{NIGHT}/ and resumit and it will redo whatever is left.

Things that could be better, but let's do that as a future PR:
  * options to specify queue
  * exit with meaningful error message if running without --batch from a login node

@kremin please test on 20200226 when you re-run sp6.

### DIFF
--- a/bin/desi_nightly_redshifts
+++ b/bin/desi_nightly_redshifts
@@ -1,0 +1,133 @@
+#!/bin/bash
+
+# experimental script for automating coadds and redshifts per night
+# Stephen Bailey
+# February 2020
+
+#-----
+#- Parse inputs
+
+print_usage () {
+    echo "desi_nightly_redshifts [--batch] YEARMMDD"
+}
+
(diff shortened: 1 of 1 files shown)


akremin:
I was able to run on night 20200226 without issue. Correctly identified and skipped existing files and only reduced the missing data.

PR #895, opened 2020-02-29
Mostrecentcalib

akremin:
Added --most-recent-calib flag that tells desi_proc to look for most recent psfs instead of immediately using a default. Searches backward 30 days before defaulting to the default.

Also corrected IERS lookup failure mode by using desisurvey.utils.freeze_icrs().

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -83,6 +83,42 @@ parser.add_argument("--runtime", type=int, default=None,  help="batch runtime in
 parser.add_argument("--most-recent-calib",action="store_true",help="If no calibrations exist for the night,"+\
                     " use the most recent calibrations from *past* nights. If not set, uses default calibs instead.")
 
+
+
+def find_most_recent(night, file_type='psfnight', n_nights=30):
+    '''
+       Searches back in time for either psfnight or fiberflatnight (or anything supported by 
+       desispec.calibfinder.findcalibfile. 
+       
+       Inputs:
+         night : str YYYMMDD   night to look back from
+         file_type : str psfnight or fiberflatnight
+         n_nights : int  number of nights to step back before giving up
+       
+      returns: 
+         nightfile : str Full pathname to calibration file of interest.
+                     If none found, None is returned.
+       
+    '''
+    # Set the time as Noon on the day in question
+    today = time.strptime('{} 12'.format(night),'%Y%m%d %H')
+    one_day = 60*60*24 # seconds                                                                              \
+
+    test_night_struct = today
+
+    # Search a month in the past
+    for daysback in range(n_nights) :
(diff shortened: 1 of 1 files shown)


akremin:
I tested with:

`desi_proc --batch --traceshift --most-recent-calib --cameras z7 -n 20200122 -e 43028`

That night doesn't have a nightly calib, which makes it a useful test case. 

Note: the code fails later in the barycentric correction, but that comes well after the new psf search code. I believe it's due to improper formatting of the early data.

bin/desi_proc
```diff
@@ -271,6 +279,7 @@ if args.batch:
         fx.write('fi\n')
     
     print('Wrote {}'.format(scriptfile))
+    print('logfile will be {}/{}-%j.log\n'.format(batchdir, jobname))
```
sbailey: minor: the %j in the slurm batch file gets replaced with the JOBID at runtime.  We don't know that jobid yet, but this printout would be clearer with "-JOBID" instead of "-%j".

bin/desi_proc
```diff
@@ -376,10 +385,27 @@ if rank == 0:
...
+                for daysback in range(30):
+                    test_night_struct = time.strptime(time.ctime(time.mktime(test_night_struct)-one_day))
+                    test_night_str = time.strftime('%Y%m%d', test_night_struct)
+                    psfnightfile = findfile('psfnight', test_night_str, args.expid, camera)
+                    if os.path.isfile(psfnightfile) :
+                        input_psf[camera] = psfnightfile
+                        successful = True
+                        break
+                if not successful:
+                    input_psf[camera] = findcalibfile([hdr, camhdr[camera]], 'PSF')
```
sbailey: This logic looks good but we'll need similar logic starting around line 744 for finding the most recent fiberflatnight too.  Let's pull this lookback code into a function so that it can look back in time for either a psfnight or a fiberflatnight.

akremin:
It currently loops through each of the 30 cameras separately. Is this the
behavior we want to keep?

I could see this being beneficial if there were more recent calibrations
for some cameras compared to others.

In that same line of thinking, do we want to link the nights used for the
flats and the psfs, or search for each of those separately as well?

sbailey:
Since this is a robustness solution for cases (like now) when we haven't yet updated DESI_SPECTRO_CALIB with final calibs, let's keep the per-camera searching so that we are robust to missing cameras on a particular calibnight.  Similarly let's keep the psfnight and fiberflatnight searching independent.  Things are supposed to be stable enough that it shouldn't matter if they pick up a PSF from a short cal one night and a fiberflat from a long cal the night before.  For cases where it does matter we really should be updating DESI_SPECTRO_CALIB instead of algorithmically encoding the correct guesses for whatever special case caused that situation.

PR #898, opened 2020-03-03
Fluxcalib minisv

julienguy:
Several minor changes to improve the flux calibration, the important ones being:
 - use only target bits STD_FAINT|STD_BRIGHT|SV0_STD_FAINT|SV0_STD_BRIGHT
 - do not use any more option --highest-throughput in call to desi_compute_flux_calibration in desi_proc
 - use 3 sigma outlier rejection of stars based on their median throughput by default in compute_flux_calibration


sbailey:
Looks good.  This might "break" flux calibration for very early data with badly miscentered stars, but that was pretty broken anyway.  Looks good for minisv2 which should be our focus.

PR #899, opened 2020-03-03
Joincosmics

sybenzvi:
The cosmics submodule now applies binary closure to patch gaps in the mask when is used in the module (the default option). The repair function is implemented in a class inside joincosmics. A `repair_mask` executable is available for plotting the output of the repair step.


py/desispec/joincosmics.py
```diff
@@ -15,7 +15,6 @@
 from desispec.maskbits import ccdmask
 from desispec.maskbits import specmask
 
-from skimage.transform import probabilistic_hough_line
 from skimage.morphology import binary_dilation, binary_closing
```
sbailey: Is the equivalent functionality available within scipy.ndimage, or are the underlying algorithms simple enough that we could code them ourselves?  I'd like to avoid a new dependency on skimage if at all possible.  This is the reason for the Travis test failures.

sybenzvi: I missed this message, sorry. I'll look into this later today. The function should be simple enough that it's ~an hour of work.

sybenzvi: @sbailey, the dependency issue is fixed and I checked that the output looks ok, so this should be ready for tests. Sorry for not paying attention to the buildbot, I should know better...

sbailey:
Thanks for fixing up the dependency issue.  I would like to test this with a real end-to-end run prior to merging to make sure it doesn't have any unintended consequences (e.g. masking emission lines...).  We'll put together some instructions for that and punt it back to you to run and evaluate.

sbailey:
Hi @sybenzvi .  NERSC is back, and https://desi.lbl.gov/trac/wiki/Pipeline/HowTo/ProcessOneExposure documents how to run exposures end-to-end.  I suggest you use night 20200315 exposures 55380-55382, 55589-55591 of ELG tile 67230 for testing, comparing master with this branch and studying any targets that significantly changed their redshift.  Please take a look and ping us if you need help.

julienguy:
Tested this branch against master for exposure 20200315/00055589 (ELG tile, 900s).

There is a clear improvement in the cosmic ray masking.
Example for z1 (left:master, right:this branch). The fiber at the center of the following figure has its cosmics mask improved. This is fiber 534.
Screenshot from 2020-04-01 15-31-04

There are many cases per frame with similar visual improvement, for instance fiber 555.
Now comparing the sframe (sky subtracted) spectra for those two fibers 534 and 555.

master version (see the two cosmics spikes):
Figure_1_master

joincosmics version (better!)
Figure_1_joincosmics

julienguy:
I compared the redshifts on two exposures 55589 and 55590 and look at the fraction
of "catastrophic failures" |dz| > 0.003*(1+z).

master version :
DELTACHI2>16 selected = 734 , catastrophic failures = 7.7%
DELTACHI2>25 selected = 559 , catastrophic failures = 4.4%

joincosmics version :
DELTACHI2>16 selected = 726 , catastrophic failures = 6.9%
DELTACHI2>25 selected = 562 , catastrophic failures = 4.3%

Unsurprisingly, it's better with improved cosmic ray rejection, but it's true the improvement is not spectacular.

sbailey:
@sybenzvi by email you had mentioned "The patch is causing the full pipeline to take a lot longer" but I'm not seeing that.  Did you perhaps hit a transient I/O slowdown or something, or am I perhaps not running what I think I'm running?  The timing looks pretty good to me.

i.e. last call for objections before I merge.

py/desispec/joincosmics.py
```diff
@@ -15,8 +15,27 @@
...
+        scipy.ndimage.binary_erosion(image, structure=selem, output=out, border_value=True)
+        return out
+
+    def binary_closing(image, selem=None, out=None):
+        dilated = binary_dilation(image, selem)
+        out = binary_erosion(dilated, selem, out=out)
+        return out
 
 import matplotlib as mpl
 import matplotlib.pyplot as plt
```
sbailey, line 41: I didn't catch this until after merging, but since preprocessing now uses joincosmics, these top-level imports of matplotlib broke preprocessing at NERSC in batch mode since X11 $DISPLAY isn't available.  I fixed this with a direct push to master by moving the matplotlib imports into the functions that use them so that it is only imported if actually needed.

@sybenzvi please confirm that I didn't mess up your plotting; I only tested that I did fix the preprocessing.

Side note: Travis tests didn't catch this, but nightly integration tests at NERSC in a cronjob did.

PR #901, opened 2020-03-04
Improve dailyproc dashboard

akremin:
Major refactor of the dailyproc dashboard. Color codes the rows by reduction status, gives summary stats in dropdown buttons, and creates an additional monthly hierarchy layer


akremin:
Only affects daily processing. Self merging

PR #908, opened 2020-03-08
Must subscript infile on call to read_spectra.

sybenzvi:
The `infile` argument is declared with the option `nargs='+'` so it's a list. When `len(args.infile)==1` it needs to be subscripted when calling `read_spectra` on the argument.


sbailey:
Thanks.  It looks like this code never worked for that case.  Merging now.

Issue #910, opened 2020-03-10
interpolate PSF parameters of "broken" fibers (for the psf fit)


julienguy:
The psf fit code ignores some fibers according to the keywork "BROKENFIBERS" in the calibration yaml files. 
 - First, we should change this name because those are not only broken fibers but also fiber traces affected by CCD defects. 
 - Second, we should readjust the "broken" fibers PSF parameters after the fit using an interpolation as done in the script `desi_interpolate_fiber_psf`.

PR #911, opened 2020-03-10
Interpolate bad fibers

julienguy:
Changes to desi_proc to interpolate the PSF parameters of the fibers that have been blacklisted from the PSF fit (because broken, too faint, or with a trace on dead column).
This addresses issue #910 .
The code is also ready for the change of keyword in the calib yaml files: BROKENFIBERS to FIBERBLACKLIST.


akremin:
Worked with a checked example arc and flat:

desi_proc --traceshift --batch -n 20200301 -e 52832

Merging

PR #913, opened 2020-03-11
Propagate yaml blacklist to fibermap bitmask in preproc

akremin:
Now find calibration yaml file in preproc and propagate fiber blacklist to fibermap fiberstatus bitmask.

Currently unsure if using the correct petal matching. Identifying fibers in blacklist with row using fibermap column "DEVICE_LOC" as they are both 0-499. Easy fix to FIBER with logic: FIBER % 500, but I didn't know which was correct.

Tested with a science exposure from March 1st:
```
export SPECPROD=kremin
desi_preproc -i /global/cfs/cdirs/desi/spectro/data/20200306/00053613/desi-00053613.fits.fz -o /global/cfs/cdirs/desi/spectro/redux/kremin/preproc/20200306/00053613/preproc-r4-00053613.fits --outdir /global/cfs/cdirs/desi/spectro/redux/kremin/preproc/20200306/00053613 --cameras r4
```

### DIFF
--- a/py/desispec/io/raw.py
+++ b/py/desispec/io/raw.py
@@ -15,7 +15,7 @@ import desispec.io
 import desispec.io.util
 import desispec.preproc
 from desiutil.log import get_logger
-from desispec.calibfinder import parse_date_obs
+from desispec.calibfinder import parse_date_obs, CalibFinder 
 
 def read_raw(filename, camera, fibermapfile=None, **kwargs):
     '''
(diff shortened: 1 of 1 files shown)


sbailey:
This code is fine and I'll merge now so that we can resume daily processing, with two notes:
  * I updated desihub/fiberassign#192 about cleaning up FIBERSTATUS bit definitions to mention this usage
  * I'll open a new ticket to move the BROKENFIBER/FIBERBLACKLIST logic into calibfinder itself rather than detailed parsing that is similar but different in both preproc and desispec.fiberflat.

julienguy:
We would now need to use this fibermap bitmask in
 - fiberflat
 - skysubtraction
 - stdstar fit
 - flux calibration

Also, we would need to modify desispec.maskbits.fibermask. For now it is
fibermask:
```
    - [BADFIBER,     0, "Broken or otherwise unusable fiber"]
    - [BADTRACE,     1, "Bad trace solution"]
    - [BADFLAT,      2, "Bad fiber flat"]
    - [BADARC,       3, "Bad arc solution"]
    - [MANYBADCOL,   4, ">10% of pixels are bad columns"]
    - [MANYREJECTED, 5, ">10% of pixels rejected in extraction"]
```
Shall we simply add 16 to all those numbers?

PR #920, opened 2020-03-16
Minor fix in sky subtraction when fiber has ivar=0

julienguy:
Use copy of frame instead of applying twice the sky subtraction on the same frame with a correction coefficient; this was a source of a bug when all fiber has ivar=0.

### DIFF
--- a/py/desispec/sky.py
+++ b/py/desispec/sky.py
@@ -1011,7 +1011,7 @@ class SkyModel(object):
         self.nrej = nrej
         self.stat_ivar = stat_ivar
 
-def subtract_sky(frame, skymodel, throughput_correction = False, default_throughput_correction = 1.) :
+def subtract_sky(frame, skymodel, throughput_correction = False) :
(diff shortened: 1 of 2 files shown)


sbailey:
tested on previously bad cframes, looks good.

py/desispec/sky.py
```diff
@@ -1160,7 +1159,6 @@ def subtract_sky(frame, skymodel, throughput_correction = False, default_through
             
             if mcoeferr>np.abs(mcoef-1) :
                 log.warning("throughput corr error = %5.4f is too large compared to the correction value = %5.4f for fiber #%03d, do not apply correction"%(mcoeferr,(mcoef-1),fiber))
-                throughput_correction_value = default_throughput_correction
```
sbailey: turns out this was needed, because otherwise `throughput_correction_value` isn't set in this case and the code crashes a few lines later when it is used.  I updated directly to master to set this to 1.0.  We can chase separately what causes this warning special case and whether we should do something different about it, but for the moment I was looking for the minimal fix to get back and running.

PR #926, opened 2020-03-19
Calib config

julienguy:
Add utility script to write the many yaml spectrograph configuration files.
Tested and used to update the config of all 30 cameras with average PSFs and fiberflats from 20200306 to 20200315.

### DIFF
--- a/bin/desi_calib_config
+++ b/bin/desi_calib_config
@@ -164,7 +164,7 @@ class ConfigEditor(object) :
 
         with open(yaml_file,'r') as ifile :
             for line in ifile.readlines() :
-
+                
                 in_latest_version_header = False
                 for version in versions :
                     if line.find(version)>=0 :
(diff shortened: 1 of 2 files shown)


sbailey:
Looks good, but given that it does some somewhat delicate yaml file parsing and rewriting to preserve comments, we should probably have some unit tests to check that.  Biggest pain would be setting up fake $DESI_SPECTRO_CALIB area for initial testing (I don't want to require the real DESI_SPECTRO_CALIB for testing; too big).

Does `desi_calib_config update` do anything or is it a placeholder?
```
[cori03 desispec] desi_calib_config update -h
usage: desi_calib_config update [options] (use --help for details)

Update an existing configuration

optional arguments:
  --spectro SPECTRO     either SPX or SMY
  --camera-arm CAMERA_ARM
                        b, r or z
```

julienguy:
Ready for final comments before merging.

Implemented the `update` and `test` commands of `desi_calib_config`.
`desi_calib_config test` without arguments goes through all of the yaml files, and tests all configurations with calibfinder and verify calibration files are present. Can also test specific yaml file(s) when called with option `-i ...`.  The same test routine is used at the end of the commands `add` and `update` to verify that the new yaml file is correct.

sbailey:
Thanks for adding the tests (and proving their worth by finding and fixing problems in the svn files).  I'll merge now, but for the record some things we may consider in the future:
  * moving code into the desispec library instead of within the script
  * tests for invalidly overlapping date ranges of the same config
  * tests that comments are preserved in the yaml files (if we weren't trying to preserve comments, this would be much simplier, so we might as well make sure they are preserved correctly)

PR #930, opened 2020-03-23
Set fiberstatus to mask fibers in bad regions of the CCDs

akremin:
Mask ccd2 positioners with bright ccd issue labeling FIBERSTATUS as MANYBADCOL. Lives under "Timebomb" if statement that only corrects for nights before July 1 2020. This should be changed to the date the ccd is replaced, ideally looking in the desi_spectro_calib yaml file.

Also propagates fibermask bits elsewhere in assemble_fibermap where it used to be hardcoded. 

Add new MISSINGPOSITION fibermask bit to coincide with previously hardcoded bit 2**9.


akremin:
Implemented major changes. Completely removed all previous edits in this branch. 

Includes moving parse_fibers out of /qproc/ and changed all references. 
Enhanced parse_fibers functionality and added a unit test. 

also tested with code:
```
from desispec.util import parse_fibers                                                                           
print(parse_fibers('1..5,6,7,8:10,11-14') )
```

Added new calibfinder function to read in EXCLUDEFIBERS (from the DESI_SPECTRO_CALIB yaml files). New function is called *.fibers_to_exclude().

At the moment the updated yaml files are in a separate branch of desi_spectro_calib in my scratch directory: /global/cscratch1/sd/kremin/svn/spec  , but will be checked into that repository after they've been verified by someone else.

I added new bitmask BADAMP to FIBERMASK and updated logic in fiberbitmasking to check against it. 

Updated io/raw.py to use new calibfinder func to mask EXCLUDEFIBERS (from yaml) with BADAMP (bitmask in FIBERMASK). 

Updated scripts/stdstars to take logical OR of all fiberstatus for all three cameras before doing any computations.

Test the stdstars with:
```
export INFILE=$DESI_SPECTRO_REDUX/daily/exposures/20200314/00055478 
export FLATS=/global/cfs/cdirs/desi/spectro/redux/daily/calibnight/20200313
desi_fit_stdstars --frames ${INFILE}/frame-b1-00055478.fits ${INFILE}/frame-r1-00055478.fits ${INFILE}/frame-z1-00055478.fits --skymodels $INFILE/sky-b1-00055478.fits $INFILE/sky-r1-00055478.fits $INFILE/sky-z1-00055478.fits --fiberflats $FLATS/fiberflatnight-b1-20200313.fits $FLATS/fiberflatnight-r1-20200313.fits $FLATS/fiberflatnight-z1-20200313.fits -o ${HOME}/out.fits
```

which stops because it cannot find enough standard stars to continue, but proceeds past all edits made in this PR and doesn't throw any errors or warnings.

julienguy:
Checked it's working fine (new masked fibers from calib. yaml files, propagation in fibermap, veto of masked fibers in standard stars). OK to merge for me.

py/desispec/scripts/stdstars.py
```diff
@@ -81,6 +81,23 @@ def main(args) :
...
     log.info('multiprocess parallelizing with {} processes'.format(args.ncpu))
 
+    # READ DATA
+    ############################################
+    # First loop through and group by exposure and spectrograph
+    frames_by_expid = {}
+    for filename in args.frames :
+        log.info("reading %s"%filename)
+        frame=io.read_frame(filename)
+        header=fits.getheader(filename, 0)
```
sbailey: `frame.meta` has the header from HDU 0 of the frame file read in at the same time as everything else.  Let's use that to avoid another file access.

py/desispec/calibfinder.py
```diff
@@ -276,10 +277,24 @@ def fiberblacklist(self):
...
+    def fibers_to_exclude(self):
+        """                                                                                         
+        Args:                                                                                       
+            None                                                                                  
+        Returns:                                                                                    
+            List of excluded fibers from yaml file as a 1D array of intergers                    
+            If no excluded fibers, returns None                                                  
+        """
+        key = 'EXCLUDEFIBERS'
+        if not self.haskey('EXCLUDEFIBERS') :
```
sbailey: minor: should this be `if not self.haskey(key):`?  otherwise I'm not sure why the key is defined if it is only used one place below.

py/desispec/io/raw.py
```diff
@@ -154,6 +154,14 @@ def read_raw(filename, camera, fibermapfile=None, **kwargs):
             loc = np.where(mod_fibers==fiber)[0]
             fibermap['FIBERSTATUS'][loc] |= maskbits.fibermask.BADFIBER
 
+    # Mask Fibers that are set to be excluded due to CCD/amp/readout issues
+    fibers_to_exclude = cfinder.fibers_to_exclude()
+    if fibers_to_exclude is not None:
```
sbailey: see comment above about `cfinder.fibers_to_exclude()` returning an empty list instead of None to slightly simplify logic here.

py/desispec/scripts/stdstars.py
akremin: Moving offline discussion here for the future:  this header read in is also done for sky and flat frames, which have similar sky.header and flat.header  attributes. So I'm removing all three of these calls to fits.getheader() and using those attributes that are already in memory.

Issue #896, opened 2020-03-02
Barycentric Corrections fail when both TARGTRA and RA are unassigned


akremin:
Barycentric Correction Fails for some old exposures where the header keys aren't properly labeled. This is a fringe case that hasn't appeared in the most recent month of data.

```
RUNNING: desi_extract_spectra -w 7520.0,9824.0,0.8 -i /global/cfs/cdirs/desi/spectro/redux/daily/preproc/20200122/00043028\
/preproc-z7-00043028.fits -p /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20200122/00043028/psf-z7-00043028.fits -\
o /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20200122/00043028/frame-z7-00043028.fits --psferr 0.1 --barycentric\
-correction
Traceback (most recent call last):
  File "/global/homes/k/kremin//workspace/most_recent_calib/desispec/bin/desi_proc", line 664, in <module>
    desispec.scripts.extract.main_mpi(extract_args, comm=comm_extract)
  File "/global/homes/k/kremin/workspace/most_recent_calib/desispec/py/desispec/scripts/extract.py", line 337, in main_mpi
    barycentric_correction_factor = barycentric_correction_multiplicative_factor(img.meta)
  File "/global/homes/k/kremin/workspace/most_recent_calib/desispec/py/desispec/scripts/extract.py", line 95, in barycentr\
ic_correction_multiplicative_factor
    raise KeyError("no TARGTRA nor RA in header")
KeyError: 'no TARGTRA nor RA in header'
```

sbailey:
For this exposure, TARGTRA is in the FIBERMAP HDU but not in the primary HDU.

In the raw data on 20200122, TARGTRA is in the SPS HDU (now called SPEC) but not in the per-camera HDUs.  More recent data has it in the per-camera HDUs too.

desispec could try using the fibermap header (fibermap.met) if the img.meta header fails.

Issue #827, opened 2019-10-23
consolidate scripts.extract.main and main_mpi


sbailey:
Technical debt strikes back!  `desispec.scripts.extract.main` and `main_mpi` are nearly identical but different.  PR #825 is the latest example of updating one while forgetting to update the other.  I'll submit a separate PR to make the emergency hack to get MPI extractions working on CMX data without fibermaps, but slightly longer term let's clean this up and consolidate those two mains into a single maintainable code.

tskisner:
Mea culpa.  The extract script was the first one I converted to use MPI years ago, and did not want to stomp all over the serial version.  A poor choice in hindsight.

sbailey:
main and main_mpi were merged by @akremin in PR #943.  Closing this ticket.

PR #934, opened 2020-03-26
try fibermap header if (TARGT)RA isn't in primary header

sbailey:
This PR fixes the crash in PR #896 by trying the fibermap header if it can't find RA or TARGTRA in the primary HDU (as was the case in early commissioning data) for the purposes of barycentric corrections.

Tested on the command originally reported in #896 (20200122/00043028), and also on 20200315/00055611 (new data)

### DIFF
--- a/py/desispec/scripts/extract.py
+++ b/py/desispec/scripts/extract.py
@@ -170,11 +170,18 @@ def main(args):
         dw = 0.7
 
     if args.barycentric_correction :
-        try:
-            barycentric_correction_factor = barycentric_correction_multiplicative_factor(img.meta)
-        except KeyError:
-            #- Early commissioning has RA/TARGTRA in fibermap but not HDU 0
-            barycentric_correction_factor = barycentric_correction_multiplicative_factor(fibermap.meta)
+        if ('RA' in img.meta) or ('TARGTRA' in img.meta):
+            barycentric_correction_factor = \
+                    barycentric_correction_multiplicative_factor(img.meta)
+        #- Early commissioning has RA/TARGTRA in fibermap but not HDU 0
+        elif fibermap is not None and \
+                (('RA' in fibermap.meta) or ('TARGTRA' in fibermap.meta)):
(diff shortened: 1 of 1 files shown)


py/desispec/scripts/extract.py
```diff
@@ -170,7 +170,12 @@ def main(args):
         dw = 0.7
 
     if args.barycentric_correction :
-        barycentric_correction_factor = barycentric_correction_multiplicative_factor(img.meta)
+        try:
```
akremin: Personally I'd prefer an "if" statement here. If the key is in img.meta, else use fibermap.meta. Better yet, check both with if-elif-else where we return a more informative error in the else.

Try's are great for robustness but since we know the failure point here I think it's more efficient/readable to address it directly

py/desispec/scripts/extract.py
```diff
@@ -334,7 +339,12 @@ def main_mpi(args, comm=None, timing=None):
         dw = 0.7
 
     if args.barycentric_correction :
-        barycentric_correction_factor = barycentric_correction_multiplicative_factor(img.meta)        
+        try:
```
akremin: The same comment as above. Assuming these are the only two places this is done I think it's fine that they're copy-pasted.

py/desispec/scripts/extract.py
sbailey: I originally used try/except because it was fewer lines of code, but agree that if/else is *clearer* code, and that is valuable.  I switched it.

py/desispec/scripts/extract.py
sbailey: There is a ton of replicated code in `extract.main` vs. `extract.main_mpi` which is indeed a maintenance risk; issue #827 is about resolving that overall but in the meantime we continue to accrete more copied code.  Related: I'm working with @lastephey and @ziyaointl on an extraction refactor to make it easier to port to GPUs, and that's where we'll reconcile this MPI vs. non-MPI code duplication.

PR #936, opened 2020-03-27
Force assemble_fibermap for nights before or during 20200310

akremin:
" --force" is an optional keyword to assemble_fibermap meant to be run if the code fails because of a missing coordinates file. We want the code to fail by default to indicate that ICS didn't provide coordinate information. However, early commissioning data sometimes didn't include this and for reprocessing of minisv2 and sv0 it is a known issue we would like to look past.

For future data we still want to know when coord* files aren't supplied, so the code still fails by default in those instances.

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -371,7 +371,16 @@ if rank == 0 and args.obstype == 'SCIENCE':
 
     fibermap_ok = os.path.exists(fibermap)
 
-#- if assemble_fibermap failed and obstype is SCIENCE, exit now
+    #- Some commissioning files didn't have coords* files that caused assemble_fibermap to fail
+    #- these are well known failures with no other solution, so for those, just force creation
+    #- of a fibermap with null coordinate information
+    if not fibermap_ok and int(args.night) <	20200310:
+        log.info("Since night is before 20200310, trying to force fibermap creation without coords file")
+        cmd += ' --force'
+        runcmd(cmd, inputs=[], outputs=[fibermap])
+        fibermap_ok = os.path.exists(fibermap)   
+    
+#- If assemble_fibermap failed and obstype is SCIENCE, exit now
 if comm is not None:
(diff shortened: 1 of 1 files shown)


bin/desi_proc
```diff
@@ -371,7 +371,16 @@ if rank == 0 and args.obstype == 'SCIENCE':
...
     fibermap_ok = os.path.exists(fibermap)
 
-#- if assemble_fibermap failed and obstype is SCIENCE, exit now
+    #- Some commissioning files didn't have coords* files that caused assemble_fibermap to fail
+    #- these are well known failures with no other solution, so for those, just force creation
+    #- of a fibermap with null coordinate information
+    if not fibermap_ok and int(args.night) <	20200310:
+        log.info("Since night is before 20200310, trying to force fibermap creation without coords file")
+        cmd += ' --force'
+        runcmd(cmd, inputs=[], outputs=[fibermap])
```
sbailey: Because this is still within the indented block, it will only run for data prior to 20200310.  I think you want to add `--force` if before 20200310, but then do the `if not fibermap_ok` and run for any date.

akremin:
Tested with:
desi_proc --batch -n 20200219 -e 51059
(where there is no coordinates file)

and:
desi_proc --batch -n 20200219 -e 51060
(where this is a coordinates file).

Works in both instances

bin/desi_proc
akremin: Hi Stephen, the intention was to only have it run if before March 10th. If we want to run for future dates as well, I can change it to what you mention

akremin: If the goal is to do this everytime if a failure occurs, then I think the correct solution is to have --force be the default

sbailey: OK, I was missing the context before this where it tries to run without --force and then here tries again with --force if the first call failed and if before 20200310.  This is fine.

Issue #938, opened 2020-03-30
calculate and save sky throughput corr terms with sky model


sbailey:
Currently the sky throughput corrections are optional (though we always use them), and are calculated when applying sky instead of when calculating the sky model.  Switch that to calculating and saving during the sky model, and then just using them when applying the model.  This is partially motivated by the usefulness of plotting these corrections as a function of focalplane position.

From Julien on Slack:

we should move what is there:  https://github.com/desihub/desispec/blob/2e427b4fd8c7771c333a46d891c8a2dc8902dda1/py/desispec/sky.py#L1039

to a separate function that would be called at the end or even after https://github.com/desihub/desispec/blob/2e427b4fd8c7771c333a46d891c8a2dc8902dda1/py/desispec/sky.py#L21

probably the best would be to compute the correction at the end of compute_sky.
The correction would be read (instead of computed) and applied in subtract_sky.

sbailey:
...and make applying the throughput correction the default, while retaining an option to *not* apply the correction for debugging.

sbailey:
implemented in PR #939.

PR #939, opened 2020-03-31
calculate throughput corrections in a separate function and save the them to disk. change default to use sky correction

akremin:
Moved calculation of throughput corrections into a separate function and save the factors to `skymodel` files.

This includes the creation of a new function in `sky.py` called `calculate_throuhput_corrections()` that is run at the end of `compute_sky()`. 

`subtract_sky()` uses the values, which are saved in a 1D array in `skymodel`. The index corresponds to the fiber number. Its value is 1 by default.  The array is stored in `skymodel.throughput_corrections`

The array is saved in a new HDU within the `skymodel` fits file labeled `THRPUTCORR`. If that exists in a loaded file, it is assigned to  `skymodel.throughput_corrections` otherwise `None` is assigned.

This PR also changes the default of the pipeline to use sky corrections in fiber throughput. It changes the input flag of `procexp` (`desi_process_exposure`) to `--no-sky-throughput-correction`   and removes the use of the previous flag from `desi_proc`.

Checked with:
`srun -N 1 -n 2 -c 32 /global/homes/k/kremin/workspace/default_skycorrect/desispec/bin/desi_proc --cameras 1 -n 20200314 -e 55444 --mpi --traceshift`

It runs to completion.

The results are in:
`/global/cfs/cdirs/desi/spectro/redux/kremin/exposures/20200314/00055444`

### DIFF
--- a/py/desispec/scripts/procexp.py
+++ b/py/desispec/scripts/procexp.py
@@ -83,7 +83,7 @@ def main(args):
             copied_frame = copy.deepcopy(frame)
             
             # first subtract sky without throughput correction
-            subtract_sky(copied_frame, skymodel, throughput_correction = False)
+            subtract_sky(copied_frame, skymodel, apply_throughput_correction = False)
 
             # then find cosmics
             log.info("cosmics ray 1D rejection after sky subtraction")
@@ -92,13 +92,12 @@ def main(args):
             # copy mask
             frame.mask = copied_frame.mask
             
-            if not args.no_sky_throughput_correction :
-                # and (re-)subtract sky, but just the correction term
-                subtract_sky(frame, skymodel, throughput_correction = True)
+            # and (re-)subtract sky, but just the correction term
+            subtract_sky(frame, skymodel, apply_throughput_correction = (not args.no_sky_throughput_correction) )
 
         else :
             # subtract sky
-            subtract_sky(frame, skymodel, throughput_correction = (not args.no_sky_throughput_correction) )
+            subtract_sky(frame, skymodel, apply_throughput_correction = (not args.no_sky_throughput_correction) )
 
         compute_and_append_frame_scores(frame,suffix="SKYSUB")
 
--- a/py/desispec/scripts/procexp.py
+++ b/py/desispec/scripts/procexp.py
@@ -34,8 +34,8 @@ def parse(options=None):
                         help = 'path of DESI sky fits file')
     parser.add_argument('--cosmics-nsig', type = float, default = 0, required=False,
                         help = 'n sigma rejection for cosmics in 1D (default, no rejection)')
-    parser.add_argument('--sky-throughput-correction', action='store_true',
-                        help = 'apply a throughput correction when subtraction the sky')
+    parser.add_argument('--no-sky-throughput-correction', action='store_true',
+                        help = 'Do NOT apply a throughput correction when subtraction the sky')
 
     args = None
     if options is None:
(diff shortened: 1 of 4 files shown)


julienguy:
Good for me.

akremin:
This addresses issue #938

py/desispec/scripts/procexp.py
```diff
@@ -92,17 +92,16 @@ def main(args):
             # copy mask
             frame.mask = copied_frame.mask
             
-            if args.sky_throughput_correction :
+            if not args.no_sky_throughput_correction :
                 # and (re-)subtract sky, but just the correction term
                 subtract_sky(frame, skymodel, throughput_correction = True)
```
sbailey: This predates this PR, but I'm confused about the logic here.  If cosmics_nsig>0 and args.no_sky_throughput_correction is True, then doesn't this mean that the skymodel is never subtracted from the original frame?  I don't understand the comment "and (re-)subtract sky, but just the correction term".

py/desispec/sky.py
```diff
@@ -1010,8 +1013,9 @@ def __init__(self, wave, flux, ivar, mask, header=None, nrej=0, stat_ivar=None):
         self.header = header
         self.nrej = nrej
         self.stat_ivar = stat_ivar
+        self.throughput_corrections = throughput_corrections
 
-def subtract_sky(frame, skymodel, throughput_correction = False) :
+def subtract_sky(frame, skymodel, throughput_correction = True) :
```
sbailey: Stylistic: it is confusing to me reading the code that "throughput_correction" is a boolean about whether to apply the throughtput correction, while "skymodel.throughput_corrections" is the correction itself.  Consider making these names differ by more than "s", e.g. with "apply_throughput_correction" or "apply_thrucorr" for the boolean.

py/desispec/sky.py
```diff
@@ -1036,142 +1040,167 @@ def subtract_sky(frame, skymodel, throughput_correction = False) :
...
+    """
+    Calculate the throughput corrections for each fiber based on the skymodel.
+
+    Args:
+        frame (Frame object): frame containing the data that may need to be corrected
+        skymodel (SkyModel object): skymodel object that contains the information about the sky for the given exposure/frame
+
+    Output:
+        corrections (dict):  keys are the FIBER number and the values are the multiplicative corrections that would 
+                             be applied to the fluxes in frame.flux to correct them based on the input skymodel
```
sbailey: docstring says this returns a dict keyed by fiber, but it looks like it is returning an array indexed by fiber%500.

py/desispec/sky.py
akremin: That also predates my changes as well, but it's a great idea. I'll go with apply_throughput_correction

py/desispec/sky.py
akremin: This was left over from my first incarnation of the function. I've corrected this

py/desispec/scripts/procexp.py
akremin: I think you're right.

Line 80 has
```
           # first subtract sky without throughput correction                                                                                             
            subtract_sky(copied_frame, skymodel, apply_throughput_correction = False)
```

But that is only on the copied frame, which only affects the mask. The actual frame is only handed to `subtract_sky` *if* `sky_throughput_correction` is `True`  (`no_sky_throughput_correction` is `False`).

I think we can turn lines 95-97  to a single line identical to line 100:
`subtract_sky(frame, skymodel, apply_throughput_correction = (not args.no_sky_throughput_correction) )`

@julienguy -- comments?

julienguy: It's indeed a bug I introduced recently when fixing another one.

akremin: I changed this to what I mentioned in the comment above. removing the if statement and directly setting apply_throughput_correction using the flag

akremin:
If the above is due to the wavelength miscalibration seen in test reductions, those are due to a separate issue already in master: #941 .

New test exposures where that issue has been negated can be found in the same data directory (note that it's a different exposure than the original post):

`/global/cfs/cdirs/desi/spectro/redux/kremin/preproc/20200314/0005545`
`/global/cfs/cdirs/desi/spectro/redux/kremin/exposures/20200314/0005545`

PR #940, opened 2020-03-31
Average fiberflat

julienguy:
Several improvements to the averaging and merging of fiber flats.
 - fix one bug in propagation of inverse variance in autocalib
 - make sure we use the same list of exposures for each camera in autocalib 
 - select valid fibers to compute the coefficient of each flat in the average
 - select fewer fiber mask bits to mask out fibers in fiber flat computation
 - improve averaging of mean spectrum when combining several flats.

This has been tested on the 2020/03 flat field data.


julienguy:
Note I also include here a bug fix to desi_qproc introduced by the recent refactor of the parse_fibers function.

sbailey:
Looks good to me, but not merging yet so that we're not changing too many things at once while debugging @akremin #939.  OK to merge whenever that is ready (or otherwise needs this).

Issue #941, opened 2020-04-02
Barycentric Correction performed mutliple times in mpi if a rank processes multiple fiber bundles


akremin:
We originally found this in a branch during a pull request but identified the same behavior in master. The issue first seen in bad wavelength calibrations in fibers run for a test exposure. 

The command run was:

`srun -N 1 -n 2 -c 32 desi_proc --cameras 1 -n 20200314 -e 55445 --mpi --traceshift`

For running an entire test science exposure rather than using the typical `-n 20 -c 2` for the spectral extractions with the `--nofluxcalib` flag before switching over to using the above to run the later steps (which skip over the early steps because of the existence of files).

Here is an example in the z camera (it is visible in all cameras), comparing fibers 249 and 250:

![image](https://user-images.githubusercontent.com/8572603/78218301-6a7ea580-7472-11ea-8bc0-0fec8289a5b2.png)

This is compared to the same fibers when run through desi_proc via a batch submission which uses the correct, default `-n` and `-c` values (it also worked on an interactive node when using the proper `-n` and `-c` values):

![image](https://user-images.githubusercontent.com/8572603/78218905-6e5ef780-7473-11ea-9a21-64165c3aabf5.png)

  
  
  

The Issue was traced to:

https://github.com/desihub/desispec/blob/2e6ee86aca8b22ffc4490f9b2a56062d84176a87/py/desispec/scripts/extract.py#L450-L486

akremin:
Corrected with PR #943

PR #943, opened 2020-04-03
Fixed bary_corr issue in and refactor of extract.py to remove repeated code

akremin:
Addressed issue #941 .  Also goes ahead and refactors extract.py to remove repeated code.

Tested using different numbers of tasks and cpus-per-task ( `srun ... -n  -c` ).

Tested with night 20200314, exposure 55445. 

Barycentric correction is applied appropriately now. Below I show examples in cam=z1.

Normal `-n 20 -c 2`:
image

Using `-n 2 -c 32`:
image

No mpi:
image

The two mpi versions have bitwise-identical data.
The non-mpi version does differ to a minor extent, I believe this is due to the precision loss in the read/write operations in the mpi version which writes temp files before consolidation.

Below I show the sum of all differences in a given HDU (over all values in the HDU.data). 

> **b1**
> FLUX    -0.012568751
> IVAR    9.1393304e-08
> RESOLUTION     2.3883758e-05
> CHI2PIX    0.0015733168
> 
> **r1**
> FLUX    0.004627799
> IVAR    7.579832e-07
> RESOLUTION    1.715213e-05
> CHI2PIX    -0.000250794
> 
> **z1**
> FLUX    -0.046244144
> IVAR    -3.4323136e-08
> RESOLUTION    1.4052641e-05
> CHI2PIX    0.0011709929

### DIFF
--- a/py/desispec/scripts/extract.py
+++ b/py/desispec/scripts/extract.py
@@ -12,6 +12,7 @@ import os.path
 import time
 import argparse
 import numpy as np
+from astropy.io import fits
 
 import specter
 from specter.psf import load_psf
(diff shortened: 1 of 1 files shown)


akremin:
I showed the straight sum of differenced pixels, here I show the more useful sum of the abs for all pixels:

Results are unsurprisingly higher:
b1
FLUX 2.7046037
IVAR 1.382468e-05
RESOLUTION 0.0042085135
CHI2PIX 0.036702715

r1
FLUX 4.6476836
IVAR 1.9532863e-05
RESOLUTION 0.007019518
CHI2PIX 0.036098786

z1
FLUX 22.699293
IVAR 3.2151926e-05
RESOLUTION 0.0072929785
CHI2PIX 0.07679814


but keep in mind this is a sum over 500 * 2881 = 1440500 differences.

sbailey:
I'm still a bit mystified about why there are differences at all, but I agree that they are very small.  I also checked the flux difference normalized by the errors, and the max differences are 0.004 sigma.  I also ran just 2 bundles and got identical fluxes.  So it's a bit odd, but I think this is good enough to merge.

Side note: there is a bug in the MPI extractions that predates this PR, where it writes out 500 spectra regardless of the --nspec parameter.  Laurie, Mark, and I are working on another extraction refactor in prep for GPUs so we can leave this for now.

PR #944, opened 2020-04-03
propagate MJD to spectra fibermap

### LINKED ISSUE #852: propagate DATE-OBS and/or MJD to output spectra
Input raw data have DATE-OBS and MJD, but we don't propagate those through to the final output spectra files, thus requiring inconvenient bookkeeping to match EXPID back to headers of individual files to find what exact time a spectrum was taken.  Fix that by including some form of a high-resolution time marker in the final spectra.

sbailey:
This PR fixes #852 by propagating the MJD-OBS header keyword to a MJD column in the output fibermap of spectra files (similar to propagating NIGHT, EXPID, and TILEID).

Tested with:
```
cd /global/cfs/cdirs/desi/spectro/redux/minisv2/tiles/70002/20200304
desi_group_spectra --inframes cframe-?0-*.fits --outfile $SCRATCH/spectra.fits
```
and verifying that the output file has an MJD column filled into the FIBERMAP table.

### DIFF
--- a/py/desispec/pixgroup.py
+++ b/py/desispec/pixgroup.py
@@ -181,8 +181,14 @@ class FrameLite(object):
         nspec = len(fibermap)
         night = np.tile(header['NIGHT'], nspec).astype('i4')
         expid = np.tile(header['EXPID'], nspec).astype('i4')
-        mjd = np.tile(header['MJD-OBS'], nspec).astype('f8')
         tileid = np.tile(header['TILEID'], nspec).astype('i4')
+        if 'MJD-OBS' in header:
+            mjd = np.tile(header['MJD-OBS'], nspec).astype('f8')
+        elif 'MJD' in header:
+            mjd = np.tile(header['MJD'], nspec).astype('f8')
+        else:
+            mjd = np.zeros(nspec, dtype='f8')-1
+
         fibermap = np.lib.recfunctions.append_fields(
             fibermap, ['NIGHT', 'EXPID', 'MJD', 'TILEID'],
             [night, expid, mjd, tileid],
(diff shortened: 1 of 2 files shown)


py/desispec/pixgroup.py
```diff
@@ -181,9 +181,11 @@ def read(cls, filename):
         nspec = len(fibermap)
         night = np.tile(header['NIGHT'], nspec).astype('i4')
         expid = np.tile(header['EXPID'], nspec).astype('i4')
+        mjd = np.tile(header['MJD-OBS'], nspec).astype('f8')
```
akremin, line 184: MJD-OBS is a pretty fundamental keyword that should be in every exposure, but are we confident enough of this card's existence that we want the code to fail if it isn't there?

sbailey: Good point.  I updated code to support "MJD-OBS" (current data), "MJD" (older sims), and not crash if both are missing (still propagates an MJD=-1 column to keep the data model consistent).

py/desispec/pixgroup.py
```diff
@@ -182,8 +182,16 @@ def read(cls, filename):
         night = np.tile(header['NIGHT'], nspec).astype('i4')
         expid = np.tile(header['EXPID'], nspec).astype('i4')
         tileid = np.tile(header['TILEID'], nspec).astype('i4')
+        if 'MJD-OBS' in header:
+            mjd = np.tile(header['MJD-OBS'], nspec).astype('f8')
+        elif 'MJD' in header:
+            mjd = np.tile(header['MJD'], nspec).astype('f8')
+        else:
+            mjd = np.zeros(nspec, dtype='f8')-1
```
akremin, line 190: This 'else' version was tested and worked?

sbailey: I tested just that one line to confirm that I had my np.zeros and dtype syntax correct, but I didn't try generating files without MJD at all to confirm that it would make the MJD=-1 column.  I hope that two brains (yours and mine) are sufficient to confirm that without actually running...

Issue #914, opened 2020-03-11
spectra files with separated cameras


julienguy:
Add to the automated pipeline the spectra files , not coadded, with the cameras separated.
This is a request of the MWS group.

segasai:
I  guess here there are two requests. 
1) We would like to see spectra-64 like files for individual exposures, so I wouldn't have to fake them myself, by gluing together the cframes.
2) The second one is a bit of a wish, is that we'd prefer to have coadd as 3 separate arms rather than glued together. That makes more sense from modeling point of view and also matches exactly individual exposure format, so we don't have to have a separate code path, separate set of spectral templates generated etc.

moustakas:
Coadded spectra separated by camera would also be useful from the perspective of the GQP working group.

callendeprieto:
Just to add that there are sound reasons to avoid merging the arms, and those led to the original format in the spectra-64 files. We should try to stick to the same format in the coadds, for consistency and simplicity. If required the merged array could be included in an additional HDU?

akremin:
All requests have been completed in PR #945

PR #945, opened 2020-04-05
Generate spectra-64 per exposure and also create coadds not across cameras

akremin:
This addresses issue #914 .

Added lines to the bash script `desi_nightly_redshifts` to generate `spectra-64` files for each exposure and create coadds across exposures that are *not* across cameras (in addition to across cameras).

This PR also fixes bugs in `/scripts/coadd_spectra.py` that removed exposures without all three cameras even when `coadd-cameras` was not set. The removal code now sits under that flag.

It also looks in input frames for the relevant cameras (`bands`) in `frames2spectra` instead of assuming all three are present. That is in `pixgroup.py.`

Files were tested in my `SPECPROD` using:
```
salloc -N 10 -t 1:00:00 -C haswell -q interactive
export SPECPROD=kremin
 desi_nightly_redshifts 20200314
```

outputs at nersc are in:
`/global/cfs/cdirs/desi/spectro/redux/kremin/tiles/66000/20200314`

### DIFF
--- a/bin/desi_nightly_redshifts
+++ b/bin/desi_nightly_redshifts
@@ -71,7 +71,6 @@ desi_group_tileframes -i exposures/$NIGHT -o tiles
 #- Find what tiles we have0
 TILEIDS=$(ls tiles | sed 's/\///g')
 SPECTROGRAPHS=$(seq 0 9)
-CAMS="b r z"
 
 #- Create Spectra-64 files
 echo -- Spectral Grouping --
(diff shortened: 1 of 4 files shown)


sbailey:
> Added lines to the bash script desi_nightly_redshifts to generate spectra-64 files for each exposure and create coadds across exposures that are not across cameras (in addition to across cameras).

I suggest that we switch to only providing the "not across cameras" coadds and drop the across cameras version.  In the cases of coadding across exposures but not across cameras, all the cameras should still be in the same output file, i.e. having the same format as the spectra files while in each HDU having only one row per target (coadds) instead of potentially multiple rows per targets (spectra files, with one row per individual exposure per target).

Regarding the spectra files: the "-64" part of the name originally referred to the healpix nside by which they were grouped (in addition to that info being in the headers), but that doesn't apply here since they are grouped by tile instead of healpix.  At minimum let's drop the "-64" part of the name here.

It also appears that this makes one spectra file per exposure, when I think we want one spectra file per tile-spectrograph, same as the original coadd and redrock grouping.  i.e. instead of "spectra-64-{expid}-{tileid}-{night}.fits" -> "spectra-{spectronum}-{tileid}-{night}.fits".

For consideration/discussion: This creates the potentially confusing situation that under the tiles/ directories the files would be called spectra-{spectronum}-{tileid}-{night}.fits grouped by tile-spectrograph, and for the main survey under the spectra-64/ directories they would be called spectra-{nside}-{healpix}.fits grouped by healpix.  I think this is ok in the spirit that the spectra*.fits format is intended to be a generic format for holding an arbitrary grouping of spectra, regardless of how they were selected (by tile, by healpix, by target class, by visual inspection failure mode, whatever), so the "spectra" prefix implies a particular format rather than a particular grouping.  But if people think that is going to cause an endless amount of confusion, we could call these "tilespec-*.fits" or something like that.

It looks like you had some other good catches in putting this together (dropped exposures, looking for frames instead of assuming), but let's converge on exactly what is grouped into files of what names before merging this.

akremin:
All changes requested have been made.

In addition; the coadds and rrdesi are given the spectra-{spectro}-{tileid}-{night}.fits file as input instead of cframes and coadds respectively.

sbailey:
Looks good; thanks.  Merging now.

PR #946, opened 2020-04-06
Scripts for creating pixmasks

biprateep:
The following changes were made:

- The name of one of the ccd mask bit has been changed (saturated->highvar) 
- A script to generate pix masks from darks had been added to /bin
@julienguy

### DIFF
--- a/bin/desi_compute_mask_dark
+++ b/bin/desi_compute_mask_dark
@@ -36,6 +36,7 @@ parser.add_argument('--miniqr', type = float, default = 0, required = True,
                     help = 'Minimum percentile threshold for good IQR')
 parser.add_argument('--maxiqr', type = float, default = 99.999, required = True,
                     help = 'Maximum percentile threshold for good IQR')
+parser.add_argument('--colfrac', type = float, default=0.4, required=False,                       help='If more than this fraction of pixels are blocked in a column, the whole column in that amplifier is blocked.')
 parser.add_argument('--mask', type = str, default = None, required = False,
                     help = 'Path to a previous mask file. If given, the output is the bitwise OR of the input mask and the mask created by this script.')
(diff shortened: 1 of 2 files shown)


julienguy:
Thanks for this. Two requests of changes:
 * add a new bit HIGHVAR instead of replacing the SATURATED one. we still need it.
 * add default values for all the input parameter arguments of the script `desi_compute_mask_dark`

biprateep:
I have made the requested changes. One thing to note is that the default values are very indicative and need to be tuned on a per ccd basis. The real masks were generated using tuned values. I'll probably store these exact values on a wikipage

sbailey:
@biprateep what was the method used to tune the values on a per CCD basis?  Is there supporting code or analysis that could be included here, e.g. to facilitate determining the values to use when a CCD is replaced?  Have you generated new pixmasks for all 30 CCDs yet (if so, please point to where they are), or is this laying the groundwork for how you would do that?  A wiki page with instructions under https://desi.lbl.gov/trac/wiki/Pipeline/HowTo would be helpful.  Thanks.

biprateep:
Hi @julienguy @sbailey @janewman-pitt-edu,
As per the suggestions the following updates have been made:

- The thresholds are set in terms of percentiles rather than absolute values of median and IQT
- Header info added to the mask files about the parameters
- incompletely blocked regions are closed
- columns in an amplifier which has more than a certain fraction of bad pixels (default 0.4) are blocked
- The scripts takes an optional old mask an input and generates the output as the bitwise OR of the new mask and the old mask

Masks for all the 30 CCDs have been generated and are available here: `/global/cscratch1/sd/bid13/desi_masks_dark`
the filenames are `pixmask-CAMERA.fits`. These also masks incorporate the handmade masks.

I have created a page on the wiki how to section and will be updating a detailed documentation about the algorithms and thresholds.

Issue #919, opened 2020-03-16
Add ability to coadd coadds


sbailey:
e.g. starting from a coadd across exposures but not across cameras, it would be nice to be able to then coadd across cameras.  But:
```
%> desi_coadd_spectra -i $SCRATCH/coadd-0.fits -o $SCRATCH/coaddcams-0.fits --coadd-cameras
INFO:coadd_spectra.py:54:main: reading spectra ...
INFO:coadd_spectra.py:74:main: coadding cameras ...
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/code/desispec/master/bin/desi_coadd_spectra", line 16, in <module>
    coadd_spectra.main(args)
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/code/desispec/master/py/desispec/scripts/coadd_spectra.py", line 75, in main
    spectra = coadd_cameras(spectra,cosmics_nsig=args.nsig)
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/code/desispec/master/py/desispec/coaddition.py", line 342, in coadd_cameras
    fibermap=coadd_fibermap(spectra.fibermap)
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/code/desispec/master/py/desispec/coaddition.py", line 53, in coadd_fibermap
    tfmap.add_column(xx,name='FIRST_'+k)
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/conda/lib/python3.6/site-packages/astropy/table/table.py", line 1569, in add_column
    self.add_columns([col], [index], name, rename_duplicate=rename_duplicate)
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/conda/lib/python3.6/site-packages/astropy/table/table.py", line 1708, in add_columns
    self._init_from_cols(newcols)
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/conda/lib/python3.6/site-packages/astropy/table/table.py", line 758, in _init_from_cols
    self._make_table_from_cols(self, newcols)
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/conda/lib/python3.6/site-packages/astropy/table/table.py", line 803, in _make_table_from_cols
    raise ValueError('Duplicate column names')
ValueError: Duplicate column names
```

[...]

PR #948, opened 2020-04-07
Allow coadding across cameras of coadds

akremin:
Addresses issue  #919 .

This only allows the combination across cameras for coadds that had previously been combined across exposures/frames. It still *does not* support coaddition of exposures from coadds across cameras.

This was tested using the `cframes` and coadd files in:
`/global/cfs/cdirs/desi/spectro/redux/kremin/tiles/66000/20200314`

For co-adding coadds, I ran the following which I'll refer to throughout as "CAMS":
`desi_coadd_spectra -i coadd-0-66000-20200314.fits --coadd-cameras -o coadd-0-66000-20200314.cams.fits`

For comparison I generated an analogous coadd from the cframes (which I'll refer to as "FULL"):
`desi_coadd_spectra -i cframe-[brz]0-*.fits --coadd-cameras -o coadd-0-66000-20200314.full.fits`

The numeric values are identical:
image

The `fibermaps` differ slightly because the outputs of `coadd() `includes additional columns that `coadd_cameras() `does not. So the coadd of the `coadds` (i.e. "CAMS") includes several additional columns:
{'NUM_TILEID', 'LAST_EXPID', 'FIRST_EXPID', 'TILEID', 'FIRST_TILEID', 'LAST_TILEID', 'NUM_NIGHT', 'LAST_NIGHT', 'EXPID', 'MJD', 'NUM_EXPID', 'FIRST_NIGHT', 'NIGHT'}

### DIFF
--- a/py/desispec/coaddition.py
+++ b/py/desispec/coaddition.py
@@ -338,9 +338,12 @@ def coadd_cameras(spectra,cosmics_nsig=0.) :
         ok=(ivar_unmasked[i]>0)
         if np.sum(ok)>0 :
             rdata[i][:,ok] /= ivar_unmasked[i][ok]
-    
-    fibermap=coadd_fibermap(spectra.fibermap)
-    
+
+    if 'COADD_NUMEXP' in spectra.fibermap.colnames:
+        fibermap = spectra.fibermap
+    else:
+        fibermap = coadd_fibermap(spectra.fibermap)
+        
     bands=""
     for b in sbands :
         bands+=b


sbailey:
The difference in fibermaps is unfortunate, but that is a problem with coadd_cameras that pre-dated this PR.  The fix here is remarkably simple and fixes #948 even if it doesn't fix the related problem in coadd_cameras.  We can save that for another day.  Merging now.

PR #949, opened 2020-04-09
Fibermask by camera

akremin:
Fibermap contains a `FIBERSTATUS` column that is camera dependent. This PR adds camera-specific bits to the `fibermask` bits, propagates them into the `FIBERSTATUS` column, and correctly bitwise OR's the three cameras (or however many bands are given to the function) to produce the combined `fibermap` for the spectra file. The bitwise OR takes place in the generation of the spectra files in `frames2spectra() `within `pixgroup.py `.

Test cases were run on two nights on two different tiles. This includes spectrographs SP2 and SP4 that have camera specific `FIBERSTATUS` bits. 

All fluxes, ivars, etc. are identical to files produced with the old code. 
Fibermaps are identical except for the `FIBERSTATUS` column in SP2 and SP4, and are completely identical in SP0 and SP9.

The data can be found on nersc at:
`/global/cfs/cdirs/desi/spectro/redux/kremin/tiles/68001/20200315`

The newly generated spectra files are prefixed with `newcode-spectra`*.
Where the first has multiple spectrographs and is the most useful or checks.

The second night/exposure is at `/global/cfs/cdirs/desi/spectro/redux/kremin/tiles/66000/20200314`, but is less useful because it doesn't have multiple spectrographs.

### DIFF
--- a/py/desispec/coaddition.py
+++ b/py/desispec/coaddition.py
@@ -23,6 +23,7 @@ from desiutil.log import get_logger
 from desispec.interpolation import resample_flux
 from desispec.spectra import Spectra
 from desispec.resolution import Resolution
+from desispec.fiberbitmasking import get_all_fiberbitmask_with_amp, get_all_nonamp_fiberbitmask_val, get_justamps_fiberbitmask
 
 def coadd_fibermap(fibermap) :
 
@@ -63,7 +64,13 @@ def coadd_fibermap(fibermap) :
         tfmap['FIBERSTATUS'][i] = np.bitwise_and.reduce(fibermap['FIBERSTATUS'][jj])
 
         #- Only FIBERSTATUS=0 were included in the coadd
-        tfmap['COADD_NUMEXP'][i] = np.count_nonzero(fibermap['FIBERSTATUS'][jj] == 0)
+        fiberstatus_nonamp_bits = get_all_nonamp_fiberbitmask_val()
+        fiberstatus_amp_bits = get_justamps_fiberbitmask()
+        targ_fibstatuses = fibermap['FIBERSTATUS'][jj]
+        nonamp_fiberstatus_flagged = ( (targ_fibstatuses & fiberstatus_nonamp_bits) > 0 )
+        allamps_flagged = ( (targ_fibstatuses & fiberstatus_amp_bits) == fiberstatus_amp_bits )
+        good_coadds = np.bitwise_not( nonamp_fiberstatus_flagged | allamps_flagged )
+        tfmap['COADD_NUMEXP'][i] = np.count_nonzero(good_coadds)
 
         for k in ['DELTA_X','DELTA_Y'] :
             if k in fibermap.colnames :
@@ -110,10 +117,11 @@ def coadd(spectra, cosmics_nsig=0.) :
         else :
             tmask=None
         trdata=np.zeros((ntarget,spectra.resolution_data[b].shape[1],nwave),dtype=spectra.resolution_data[b].dtype)
-        
+
+        fiberstatus_bits = get_all_fiberbitmask_with_amp(b)
+        good_fiberstatus = ( (spectra.fibermap["FIBERSTATUS"] & fiberstatus_bits) == 0 )
         for i,tid in enumerate(targets) :
-            jj=np.where((spectra.fibermap["TARGETID"]==tid) & \
-                        (spectra.fibermap["FIBERSTATUS"]==0))[0]
+            jj=np.where( (spectra.fibermap["TARGETID"]==tid) & good_fiberstatus )[0]
 
             #- if all spectra were flagged as bad (FIBERSTATUS != 0), contine
             #- to next target, leaving tflux and tivar=0 for this target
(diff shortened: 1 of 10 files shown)


sbailey:
This may be the same problem identified by the failing unit tests, or it may have pre-dated this PR, but let's fix it here too: `desi_group_spectra` should work even with a missing frame for one exposure.  e.g.
```
cd /global/cfs/cdirs/desi/spectro/redux/kremin/tiles/68002/20200315

# WORKS when 6 frames are included as input (3 per exposure)
desi_group_spectra --inframes cframe-[brz]0-00055626.fits cframe-[brz]0-00055627.fits --outfile $SCRATCH/blat.fits

# FAILS when cframe-r0-00055626.fits isn't included
desi_group_spectra --inframes cframe-[bz]0-00055626.fits cframe-[brz]0-00055627.fits --outfile $SCRATCH/blat.fits

mpi4py not found, using only one process
INFO:group_spectra.py:62:main: Starting at Thu Apr  9 15:17:13 2020
INFO:group_spectra.py:63:main: Reading 5 frame files
INFO:group_spectra.py:72:main: Combining into spectra
  File "/global/common/software/desi/users/sjbailey/desispec/bin/desi_group_spectra", line 24, in <module>
    group_spectra.main(args, comm=comm)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/scripts/group_spectra.py", line 73, in main
    spectra = frames2spectra(frames)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/pixgroup.py", line 543, in frames2spectra
    if banddict[band] is None:
KeyError: 'r'
```

Other cases to include in unit tests:
  * when combining frames into spectra: spectra.fibermap['FIBERSTATUS'][ispec] == logical OR of the individual camera cframe.fibermap['FIBERSTATUS'][ispec] when merging 3 cameras into one fibermap
  * when coadding spectra: if spectra.fibermap['FIBERSTATUS'][ispec] has BADAMPB set, the coadd should ignore spectra.flux['b'][ispec] but still use spectra.flux['r'] and spectra.flux['z']
    * i.e. simply checking using entries with FIBERSTATUS==0 is not sufficient because it could discard R and Z spectra just because BADAMPB was set
    * if we have the individual spectra.mask[b/r/z] bits set with corresponding spectra.ivar[b/r/z]=0 for bad camfibers, then that should work without needing to check FIBERSTATUS.  Double check that.
  * for the purposes of the output flux, ivar, and fibermap['FIBERSTATUS'], coadding frames should be equivalent to combining those frames into a spectra file and then coadding the spectra file.
  * when coadding across multiple exposures, coadd.fibermap['FIBERSTATUS'] == logical AND of the input spectra fibermap['FIBERSTATUS'].

sbailey:
Unit tests aren't catching this (bad!) but our "header" vs. "meta" inconsistencies are coming back to haunt us (or you specifically @akremin even though this wasn't your fault...):

When coadding with all cameras present, it works, but when coadding with a missing frame it fails:
```
cd /global/cfs/cdirs/desi/spectro/redux/kremin/tiles/68002/20200315
desi_coadd_spectra --infile cframe-[bz]0-00055626.fits cframe-[brz]0-00055627.fits --outfile $SCRATCH/foo.fits
INFO:coadd_spectra.py:58:main: reading spectra ...
WARNING:pixgroup.py:409:add_missing_frames: Creating blank data for missing frame ('20200315', 55626, 'r0')
  File "/global/common/software/desi/users/sjbailey/desispec/bin/desi_coadd_spectra", line 16, in <module>
    coadd_spectra.main(args)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/scripts/coadd_spectra.py", line 92, in main
    spectra = frames2spectra(frames)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/pixgroup.py", line 455, in frames2spectra
    add_missing_frames(frames)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/pixgroup.py", line 418, in add_missing_frames
    header = fitsio.FITSHDR(frame.header)
AttributeError: 'Frame' object has no attribute 'header'
```
When frames are present, they use the FrameLite object, which supports both ".header" and ".meta", but the full Frame object only supports ".meta".  Ugh.

Please add a unit test for coadding with a missing frame (which should fail), and then fix this by changing ".header" to ".meta" or otherwise padding the missing data with a FrameLite instead of a Frame, and seeing if anything else fails about coadding when there is a missing camera.

akremin:
It's even more haunting than we thought, switching to *.meta doesn't work because it's an astropy header and that piece of code wants a fitsio header... which apparently don't work together. If I use fitsio.FITSHDR( dict( frame.meta ) ), then it works.

sbailey:
Without having dug deeply in this, some observations:
  * the coadd code does work when all the cameras exist, and they are loaded as FrameLite objects
  * it doesn't work when there is a missing camera, I think because the missing camera is filled in with a Frame (not FrameLite) object, and the two aren't playing nicely together
  * could you fill in the missing frame with a FrameLite object instead of a full Frame object?
  * the other option seems to be making FrameLite and Frame more compatible with each other.  FrameLite is purposefully lighter weight for faster manipulation, but it isn't purposefully incompatible with Frame  (e.g. "header" vs. "meta" was an oversight from implementing them years apart...)

akremin:
The issues cited are due to the coadds using full Frame files, which are needed because they need proper data precision, endianness, etc., but it's calling a grouping function meant specifically for io and only written to work on FrameLite files. I have tweaked the code so that it now appears to be compatible with both. The issue was in adding new frames. 

The issues stem from:
Frame uses astropy.io.fits
FrameLite uses fitsio

But I believe in the identified use cases we can now use frames2spectra with either Frame or FrameLite objects. (About to push this to github).

sbailey:
I confirm that the previous failing cases now work.  However, it appears that a bad amp FIBERSTATUS on one frame still excludes other frames of the same expid from the coadd even though they are fine and could be used in the coadd.

Example data are in `/global/cscratch1/sd/sjbailey/desi/debug/fiberstatus`, created with `make_test_frames` to create two exposures with b,r,z each.  test-r0-00055626.fits has fibermap['FIBERSTATUS'][5] = BADAMPR set, but b0 and z0 for that exposure do not, and the other exposure has FIBERSTATUS=0 for everything.  I coadd these with:
```
desi_coadd_spectra --infile test*.fits -o coadd-frames.fits
```
Note that the coadd is ordered by TARGETID instead of FIBER, so FIBER=5 is now at index=429 (ok).

All input exposures except r0-00055626  fiber 5 have ivar=1 such that coadding two exposures should result in ivar=2, while for fiber 5 r-band only should stay at ivar=1.  Instead, coadd.ivar['z'][429] and coadd.ivar['b'][429] are also = 1, even though they could have used two exposures.

Converting the frames into a spectra file appears correct: the individual camera ivars remain as-is, and the single fibermap picks up the BADAMPR bit from the r0-00055626.  Coadding the spectra file (coadd-spectra.fits) has the same behavior as coadding the frames, so they are at least consistent.

akremin:
The reported issue when using desi_coadd_spectra --infile test*.fits -o coadd-frames.fits has been solved and some of the unit tests implemented, but the coadd unit tests have not been added yet

PR #957, opened 2020-04-16
mask high readnoise CCD amps

sbailey:
This PR adds a new CCD mask bit BADREADNOISE and masks pixels with readnoise>100.  The value is hardcoded, but so are some of our other mask bit thresholds in the same part of preprocessing and I don't think it is worth calling that out into a separate config file (at least not right now).

Tested along with the branch in desihub/specter#78:
```
desi_preproc --cameras z4 \
    -i /global/cfs/cdirs/desi/spectro/data/20200306/00053610/desi-00053610.fits.fz \
    --outfile $SCRATCH/preproc-z4-00053610.fits

desi_extract_spectra -w 7520.0,9824.0,0.8 --psferr 0.1 --barycentric-correction \
  -i $SCRATCH/preproc-z4-00053610.fits \
  -p /global/cfs/cdirs/desi/spectro/redux/andes/exposures/20200306/00053610/psf-z4-00053610.fits \
  --specmin 150 --nspec 25 \
  -o $SCRATCH/frame-z4-00053610.fits
```
With resulting `plot_frame`:
image
i.e. the wavelengths>8700 A are all on the bad amp and get masked and thus don't appear here (instead of previously sometimes slipping through with tiny-ivar huge-flux values).

Also added `freeze_iers` to `desispec.scripts.extract.main` so that `desi_extract_spectra` can be run by itself without triggering an IERS download.  astropy grumble mumble.

### DIFF
--- a/py/desispec/maskbits.py
+++ b/py/desispec/maskbits.py
@@ -40,14 +40,15 @@ from desiutil.bitmask import BitMask
 _bitdefs = yaml.safe_load("""
 #- CCD pixel mask
 ccdmask:
-    - [BAD,         0, "Pre-determined bad pixel (any reason)"]
-    - [HOT,         1, "Hot pixel"]
-    - [DEAD,        2, "Dead pixel"]
-    - [SATURATED,   3, "Saturated pixel from object"]
-    - [COSMIC,      4, "Cosmic ray"]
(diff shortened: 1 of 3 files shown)


akremin:
Test run with 53628 (on 0306, z4) works.

Travis issue appears to be travis related, not due to this PR.

PR #971, opened 2020-05-27
use the center of the exposure for the barycentric correction

segasai:
Currently the start of the exposure is used for the barycentric correction, but for 20 min exposure the difference in the correction between center of exposure and beginning can reach 0.4 km/s.

The commit fixes that.


sbailey:
Looks good, thanks.

PR #993, opened 2020-07-24
[WIP] add support for gpu_specter in desi_extract_spectra

dmargala:
**Note that this PR is currently WIP and should not be merged**

This PR adds support for using [gpu_specter](https://github.com/sbailey/gpu_specter) in place of specter in desi_extract_spectra. gpu_specter is essentially a refactor of `desispec.extract.main_mpi`+`specter.extract.ex2d` with support for running on GPUs. 

desi-related todo:
- [x] `--barycentric-corrections`. 
- [ ] `--fiber-map`. not sure how to support this yet. gpu_specter has pretty strict assumptions regarding `specmin`, `nspec`, `bundlesize`, etc. can it be read from the image file or inferred from `specmin` and `nspec`?
- [x] `--no-scores`. note that this option is currently required to use gpu_specter in desi_extract_spectra. need wrap extraction output in a `desispec.Frame` object and make sure relevant data is all there
- [x] use `desispec.io.write_frame()` to write results

extraction-related todo (implemention needed in gpu_specter):
- [x] `--model`.  compute the model spectra during extraction
- [x] `--regularize`.  use regularization during extraction
- [x] `--psferr`. note that the command line option does not seem to be used by desi_extract_spectra so specter falls back to a value in the psf object.
- [x] compute `chi2pix` during extraction. needed to compute mask for frame output
- [x] compute `pixmask_fraction` during extraction. needed to compute mask for frame output

### DIFF
--- a/py/desispec/scripts/extract.py
+++ b/py/desispec/scripts/extract.py
@@ -65,6 +65,8 @@ def parse(options=None):
     # parser.add_argument("--fibermap-index", type=int, default=None, required=False,
     #                     help="start at this index in the fibermap table instead of using the spectro id from the camera")
     parser.add_argument("--barycentric-correction", action="store_true", help="apply barycentric correction to wavelength")
+    parser.add_argument("--gpu-specter", action="store_true", help="use gpu_specter instead of specter")
+    parser.add_argument("--gpu", action="store_true", help="use gpu device for extraction when using gpu_specter")
     
     args = None
     if options is None:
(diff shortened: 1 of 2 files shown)


sbailey:
@dmargala please update this to use your new ...IOCoordinator classes in gpu_specter.  Thanks.

sbailey:
Thanks for your patience with this PR.  I've verified that it has no effect if *not* using `--gpu-specter`, and when using `--gpu-specter` the results are consistent to approximately the same level (~5% of a sigma) that specter is consistent with itself when changing nuisance parameters like `--nsubbundles`.  Merging now (will deal with test coverage later...)

Issue #1001, opened 2020-09-24
bug in dark current subtraction during image preprocessing


julienguy:
- dark (in e/s) subtracted before applying gain to current counts.
- dark current noise not in inverse variance.

sbailey:
Fixed by PR #1002.

PR #1002, opened 2020-09-24
subtract dark after applying gain and take into account dark poisson noise

julienguy:
Bug fix issue #1001 .
It looked bad but it had a very minor effect because the dark current is low.


sbailey (approved):
Good catches, even if the end effect is small.  Thanks.

PR #1003, opened 2020-09-25
update desi_proc timing logging

sbailey:
This PR uses the new `desiutil.timer.Timer` class to standardize timing in `bin/desi_proc`.  It uses some features in desihub/desiutil#152, so don't merge this until that PR is finalized.

New options to `desi_proc`:
  * `--starttime $(date +%s)`: log the time between this starttime and the beginning of the python script (srun & python interpreter launch overhead).
  * `--timingfile TIMINGFILE`: save timing information to a json file for later analysis.  When running in batch, this file is saved for each exposure.

New script `bin/plot_timing` uses the timing information saved with `--timingfile` to make plots like this:
science-20200315-00055611

@akremin please take a look to see if (or how badly) this will conflict with your changes in the desi_proc_mod branch.

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -24,21 +24,15 @@ time srun -n 20 -N 1 -C haswell -t 15:00 --qos realtime desi_proc --mpi -n 20191
 import time
 start_imports = time.time()
 
-from collections import OrderedDict
-import sys, os, argparse, re, time, collections
+import sys, os, argparse, re
 import subprocess
 from copy import deepcopy
-import glob
-import json
 
 import numpy as np
 import fitsio
 from astropy.io import fits
-
-from desiutil.log import get_logger, DEBUG, INFO
+import glob
 import desiutil.timer
-from desitarget.targetmask import desi_mask
-
 import desispec.io
 from desispec.io import findfile
 from desispec.io.util import create_camword
@@ -48,6 +42,10 @@ from desispec.sky import subtract_sky
 from desispec.util import runcmd
 import desispec.scripts.extract
 import desispec.scripts.specex
+
+from desitarget.targetmask import desi_mask
+
+from desiutil.log import get_logger, DEBUG, INFO
 stop_imports = time.time()
 
 parser = argparse.ArgumentParser(usage = "{prog} [options]")
@@ -314,7 +312,10 @@ if args.batch:
         cmd = ' '.join(inparams)
         cmd = cmd.replace(' --batch', ' ').replace(' --nosubmit', ' ')
 
-        timingfile = f'{jobname}-timing-$SLURM_JOBID.json'
+        if args.timingfile is None:
+            timingfile = f'{jobname}-timing-$SLURM_JOBID.json'
+        else:
+            timingfile = args.timingfile
 
         if not args.mpi:
             cmd += ' --mpi'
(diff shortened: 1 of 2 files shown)


sbailey:
desihub/desiutil#152 has been merged, so this PR is now ready for full review and can be merged when ready

bin/desi_proc
```diff
@@ -296,24 +314,30 @@ if args.batch:
         cmd = ' '.join(inparams)
         cmd = cmd.replace(' --batch', ' ').replace(' --nosubmit', ' ')
 
+        timingfile = f'{jobname}-timing-$SLURM_JOBID.json'
```
akremin, line 317: Should this be using the arg.timingfile from the parser or at least checking for its existence before defining to default?

bin/desi_proc
```diff
@@ -22,18 +22,23 @@ time srun -n 20 -N 1 -C haswell -t 15:00 --qos realtime desi_proc --mpi -n 20191
 """
 
 import time
-from collections import OrderedDict
-
-progress = OrderedDict(start=time.asctime())
+start_imports = time.time()
 
+from collections import OrderedDict
```
akremin, line 27: Some of these moved imports aren't picked up as the same. Is there a motivation behind the moves or was this a result of adding and removing code during development?

It isn't a blocking factor but it does increase the differences with master that will impact the upcoming rebase of the desi_proc_mod branch.

akremin (commented):
Two small comments. Both are considerations to take, but neither is required for merge. I will wait for comment before clicking approve, but these changes look good.

bin/desi_proc
sbailey: Good point; I added a check to use `args.timingfile` here.

Default usage with `--batch` is to *not* specify `--timingfile`, and then let `desi_proc` pick a timing file name based upon the job name and slurm jobid; but if the user specifies `--batch` and a specific `--timingfile`, that will be used instead.

bin/desi_proc
sbailey: It was mostly a side effect of rearranging imports to be able to time them properly (`import time` and grab `time.time()` ASAP; wait on MPI until after argparsing; wait on `freeze_iers` until we've got the timers going...)  For now I reverted the order to close to the original with only the minimally necessary rearrangement.

At the same time, our import order has gotten a bit messy compared to our (perhaps undocumented?) typical order:
  * python standard libraries (os, sys, glob, ...)
  * 3rd party libraries (numpy, scipy, astropy, ...)
  * other desi packages (desiutil, desitarget, ...)
  * other modules from within this package (desispec.io, desispec.sky, ...)

We can standardize that in a later update.

sbailey:
Thanks for review @akremin; I made updates in response to your comments.  Just in case I introduced a typo, I'd like to double check this at NERSC when Cori is back before merging.

bin/desi_proc
akremin: Great. I think this is the best approach even if its not meant to be used by end users. It's either this or documenting the use cases, and this is simpler and more flexible.

bin/desi_proc
akremin: Thank you for reverting. We can refactor right after rebase so you can capture the timing appropriately.

akremin (approved):
Feel free to merge after testing when Cori comes back online

PR #1007, opened 2020-10-15
set MPICH_GNI_FORK_MODE=FULLCOPY

sbailey:
This PR adds MPICH_GNI_FORK_MODE=FULLCOPY to the desispec.module file, and adds checks for that in `desi_proc` when running.  This is required for MPI+multiprocessing to work together at NERSC.  In standard pipeline running, we were getting this environment variable via the redrock module file, but this PR makes sure that we also get it for desispec itself whether or not redrock is loaded, and provides some robustness when running development copies outside of the pre-installed modules.

Other changes that came along for the ride:
  * also added KMP_AFFINITY=disabled, which we also set it redrock modules
  * turns off an error message that wasn't really an error, so that greping for ^ERROR will focus on more genuine errors.

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -131,6 +131,18 @@ else:
 if 'PMI_MMAP_SYNC_WAIT_TIME' not in os.environ:
     os.environ['PMI_MMAP_SYNC_WAIT_TIME'] = '3600'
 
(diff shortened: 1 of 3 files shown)


julienguy (approved):
Tested with and without the MPICH_GNI_FORK_MODE env. variable. Both working fine.

PR #1010, opened 2020-10-24
Newbias

zkdtc:
New preproc code to use new time-dependent darks to reduce raw image.

PR #1011, opened 2020-10-26
New bias+dark model (rebased version)

sbailey:
This PR is a replacement for PR #1010 by providing the changes based on a more recent branch of master (the original newbias branch was branched from master in late June, and there have been other desi_proc and preproc changes since then).  This will hopefully simplify coordination with other work in progress like the pipev4 branch.

In a brief test of:
```
desi_preproc -i /global/cfs/cdirs/desi/spectro/data/20200315/00055589/desi-00055589.fits.fz \
   --cam b0 --outdir test/
```
this branch works (i.e. doesn't crash) with both old calibs ($DESI_SPECTRO_CALIB=/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk) and new calibs ($DESI_SPECTRO_CALIB=/global/cfs/cdirs/desi/users/zhangkai/desi/code/desi_spectro_calib/trunk).

Note that this branch (newbias2) does produce different answers than newbias due to other changes to master, so evaluate carefully, but I believe this branch is representative of what we would get if we merged the original newbias into master.

Original code and model evaluation from @zkdtc .

For the record, the rebasing was:
```
git checkout master
git pull
git checkout newbias
git checkout -b newbias2
git rebase master
```

### DIFF
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -502,7 +502,6 @@ def preproc(rawimage, header, primary_header, bias=True, dark=True, pixflat=True
     else:
         dark=False
 
-
     if bias is not False : #- it's an array
         if bias.shape == rawimage.shape  :
             log.info("subtracting bias")
@@ -866,7 +865,11 @@ def read_2d_dark(filename=None, exptime=0):
             # Run interpolation
             image1=recover_2d_bias_dark(hdus,exptime_arr[ind1],ext_arr[str(int(exptime_arr[ind1]))])
             image2=recover_2d_bias_dark(hdus,exptime_arr[ind2],ext_arr[str(int(exptime_arr[ind2]))])
+<<<<<<< HEAD
             img=image1*(1-precision)+image2*precision #interp_shape(image1,image2,precision)
+=======
+            img=image1*precision+image2*(1-precision) #interp_shape(image1,image2,precision)
+>>>>>>> 982963f14803cdf61f63c407e0760a88fabd0e72
 
             return img
 
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -961,6 +961,19 @@ def interp_shape(top, bottom, precision):
 
     return out
 
+def recover_2d_bias_dark(hdus,exptime):
+    nx=len(hdus[0].data) #4162
+    ny=len(hdus[0].data[0]) #4232
+    profileLeft=hdus[str(int(exptime))].data[0]
+    profileRight=hdus[str(int(exptime))].data[1]
+    profile_2d_Left=np.transpose(np.tile(profileLeft,(int(ny/2),1)))
+    profile_2d_Right=np.transpose(np.tile(profileRight,(int(ny/2),1)))
+    profile_2d=np.concatenate((profile_2d_Left,profile_2d_Right),axis=1)
+    #import pdb;pdb.set_trace()
+    image=hdus['0'].data+profile_2d+hdus['DARK'].data*float(exptime)
+    return image
+
+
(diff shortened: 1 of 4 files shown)


py/desispec/preproc.py
```diff
@@ -469,6 +469,38 @@ def preproc(rawimage, header, primary_header, bias=True, dark=True, pixflat=True
 
     readnoise = np.zeros_like(image)
 
+    #- Load dark
+    dark_info=cfinder.data['DARK'].lower().replace('/','-').replace('_','-').split('-')
```
julienguy, line 473: Test first if cfinder.data has a 'DARK' entry. If not and the argument dark==None, skip dark subtraction.
Also I find fragile to rely on the formatting of the  filename to decide on the next action. 
I suggest we add instead a new keyword in the yaml file, like DARKARRAY or any other keyword that would differentiate this type of file from a traditional DARK.

py/desispec/preproc.py
```diff
@@ -317,7 +320,11 @@ def get_calibration_image(cfinder,keyword,entry) :
     elif keyword == "MASK" :
         return read_mask(filename=filename)
     elif keyword == "DARK" :
-        return read_dark(filename=filename)
+        if header is not None: # Use master bias+dark file if header is set to pass exptime
```
julienguy, line 323: Same as other comment. Having a non null header argument to trigger a different read routine seem fragile. I would consider another keyword for this (like 'DARKARRAY').

py/desispec/preproc.py
```diff
@@ -840,6 +853,79 @@ def read_bias(filename=None, camera=None, dateobs=None):
...
+            exptime_arr=np.sort(np.array(exptime_arr))
+            ind=np.where(exptime_arr>int(exptime))
+            ind1=ind[0][0]-1
+            ind2=ind[0][0]
+            log.info('Interpolate between '+str(exptime_arr[ind1])+' and '+str(exptime_arr[ind2]))
+            precision=(float(exptime)-exptime_arr[ind1])/(exptime_arr[ind2]-exptime_arr[ind1])
+            # Run interpolation
+            image1=recover_2d_bias_dark(hdus,exptime_arr[ind1],ext_arr[str(int(exptime_arr[ind1]))])
+            image2=recover_2d_bias_dark(hdus,exptime_arr[ind2],ext_arr[str(int(exptime_arr[ind2]))])
+            img=image1*precision+image2*(1-precision) #interp_shape(image1,image2,precision)
```
julienguy, line 925: Isn't it the opposite? because here precision=0 if exptime=exptime_arr[ind1].

zkdtc: You are absolutely right

py/desispec/preproc.py
zkdtc: I have added a new keyword "DARKNL" which means non-linear

julienguy:
The merge with master didn't go well. All the changes with model_variance are not in your version of py/desispec/preproc.py . There are also remaining git conflict markers in the file.

zkdtc:
Hi Julien, 
    Maybe it is faster that I edit the preproc.py from the master branch and commit it?

sbailey:
@zkdtc good idea, yes, let's start again from master to clean this up.  To be clear:
  * start a new branch from current master
  * re-apply the edits you want to make from preproc.py
  * open a new PR

Then let's discuss how to proceed.  I have some ideas for a more maintainable way to deal with DARK vs. DARKNL but it will be easier to evaluate once we are back to having a clean branch with a clean PR.  Thanks.

sbailey:
closing as superseded by #1013

PR #1012, opened 2020-10-27
Refactoring desi_proc

akremin:
This PR focuses on `desi_proc` and turning most of the early setup code into function calls. The functions are placed under `/py/desispec/workflow/` where a number of new files will eventually go in the ongoing updates. 

The new file is `desi_proc_funcs.py`. It has function definitions for creating the parser object, reading header information to infer undefined arguments based on arguments that are defined, determining the required resources for running a job, and writing the batch script to submit desi_proc to an HPC. It also moves a function that previously existed at the start of `desi_proc`.

The new parser functions make the old code generalized to be used in both desi_proc and a future joint fitting script `desi_proc_joint_fit`. The goal is to reduce the repeated code by putting arguments common to both in a single location that can be used to generate both parsers. 

The code in `update_args_with_headers()` that defines unspecified arguments already existed in desi_proc but is now a function that can be called from the joint fitting script as well.

The joint fitting code itself is to be added and vetted separately in another PR. This is to push the functional form of `desi_proc` to master.

This was tested with a science exposure: `desi_proc -n 20200315 -e 55594 --traceshift --batch`  
which ran successfully to completion through all steps from raw data to cframes.

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -872,6 +872,8 @@ if args.obstype in ['SCIENCE'] and \
             (not args.noskysub) and \
             (not args.nofluxcalib):
     timer.start('fluxcalib')
+
+    night, expid = args.night, args.expid #- shorter  
     #- Compute flux calibration vectors per camera
     for camera in args.cameras[rank::size]:
         framefile = findfile('frame', night, expid, camera)
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -134,12 +134,16 @@ if args.batch:
     #exp_str = '{:08d}'.format(args.expid)
     jobdesc = args.obstype.lower()
     if args.obstype.lower() == 'science':
+        # if not doing pre-stdstar fitting or stdstar fitting and if there is
+        # no flag stopping flux calibration, set job to poststdstar
         if args.noprestdstarfit and args.nostdstarfit and (not args.nofluxcalib):
             jobdesc = 'poststdstar'
+        # elif told not to do std or post stdstar but the flag for prestdstar isn't set,
+        # then perform prestdstar
         elif (not args.noprestdstarfit) and args.nostdstarfit and args.nofluxcalib:
             jobdesc = 'prestdstar'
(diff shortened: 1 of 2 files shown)


akremin:
Tested each of the following:
`desi_proc -n 20200315 -e 55592 --traceshift --noprestdstar --nostdstarfit --nofluxcalib --batch`
Result: Correct behavior. Script ran very quickly without running any of the processing code or creating any outputs.

`desi_proc -n 20200315 -e 55592 --traceshift --noprestdstar --nofluxcalib --batch`
Result: Correct behavior. Threw error saying that the inputs to stdstar fitting were unavailable. The script exited "Successfully," however.

`desi_proc -n 20200315 -e 55592 --traceshift  --nostdstarfit --nofluxcalib --batch`
Result: Correct behavior. Ran and successfully created sframes before stopping.

Then ran:
`desi_proc -n 20200315 -e 55592 --traceshift  --noprestdstar --nofluxcalib --batch`
Result: Correct behavior. Started at stdstar fitting and stopped before flux calib.

Then ran:
`desi_proc -n 20200315 -e 55592 --traceshift  --noprestdstar --nostdstarfit --batch`
Result: Eventually correct behavior.  Initially failed because of shortcut terms night, expid not being defined. Fixed (pushed new commit) and then it ran through to cframes.

moustakas (commented):
Nice to see this refactor, but just in terms of using the existing organization couldn't `workflow/desi_proc_funcs.py` go under `scripts/desiproc.py`? Or will `workflow` get additional modules?

sbailey:
@moustakas indeed more workflow modules are coming.  After discussing this with @akremin we decided to get the minimal desi_proc related changes in so that it doesn't become a blocking factor for other desi_proc changes that might have merge conflicts.

You make a good point that normally this argparse stuff would live in scripts/ .  Let's wait to see what else comes in workflow/ in a future PR and how they inter-operate, and then consider moving pieces over into scripts/ for maintainability standardization with other scripts.

I've also tested this with desi_dailyproc and confirm that still works, so this is backwards compatible with the current daily processing method.

sbailey (approved):

PR #1013, opened 2020-10-28
Codes to deal with new non-linear darks added

zkdtc:
This code uses the new non-linear dark image generated using a series of darks. 

Major changes to the code:
1) Add read_2d_dark.py and recover_2d_dark.py to preproc.py to read new non-linear dark file and reconstruct the dark. 
2) Revised get_calibration_image function to read the correct dark file. 
3) Revised preproc.py to make sure both old and new dark files can be used. 
4) Add 'DARKNL' keyword to yaml files to use non-linear dark.

### DIFF
--- a/py/desispec/desi_bias_dark_1d_model.py
+++ b/py/desispec/desi_bias_dark_1d_model.py
@@ -1,0 +1,181 @@
+from astropy.io import fits
+import pdb
+import matplotlib.pyplot as plt
+import numpy as np
+import copy
+import desispec.preproc
+import argparse
+"""
+Usage: python3 desi_bias_dark_1d_model.py
+"""
+parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+description="Search all bias and dark images for a series of night, Compute a master bias from a set of raw images, and combine them according to exptime",
+epilog='''Cool.'''
+)
+parser.add_argument('-e','--exp', type = str, default = [300,450,700,900,1200], required = False, nargs="*",
+                    help = 'liner')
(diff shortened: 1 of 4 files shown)


julienguy:
Still need to work on extrapolating the actual dark current to longer exposure times.

julienguy:
ready for review

zkdtc:
Wow!

sbailey (changes requested):
@zkdtc please add your code that generates the new bias+dark calib files to this PR, along with some documentation about how to run it, since we will need to run it again in the future, e.g. for the soon-to-be-replaced CCDs.

There appear to be some mismatches between old and new calibrations that leads to some fragile code in how they are used here.  I think we want darks that are:
  * overscan subtracted and trimmed to remove the overscan regions
  * gain corrected
  * include a BUNIT keyword defining the units, i.e. "electron/s" for DARK, and electron for the ZERO and T* HDUs, if I understand them correctly.  The old calibs don't have this keyword either, but we will save ourselves some confusion if we add units to all these calib files.  I'll open a separate ticket about that for the old calib files.

But let's start with getting that code in this PR too, even before making any of those changes.  Thanks.

zkdtc:
Hi Stephen, 
   I have fixed desi_bias_dark_1d_model.py and tested it. Very sorry for the inconvenience...
   Best~
                   Kai

sbailey:
Thanks for the fixes.  I will make some further changes to enable calculating the model using daily biases since our reference darks were taken over 2 calendar days spanning 3 DESI YEARMMDD "nights", and calculating the dark model in electron/s using only the active region instead of including the overscan.

sbailey:
superseded by PR #1029 which combined code from this branch and PR #1023.  Closing this one.

PR #1014, opened 2020-10-29
Add desi_proc_joint_fit

akremin:
## Overview
This adds a joint fitting script that acts very similarly to desi_proc except that it operates on multiple exposure ID's and is meant for joint fitting of: psf's, flat's, and standard stars within science's. These generate psfnight's, fiberflatnight's, and stdstar's.

This does not break backward compatibility, as it just adds a new script. It reimplements joint psf and flat fitting which has been commented out in desi_proc for quite a while.

The script is `desi_proc_joint_fit`. 

It takes many of the same arguments and flags as desi_proc. A previous refactor (PR #1012) allows the reuse of the same parser object, with a few specialized flags to account for the plural nature of the exposure ids (--expids rather than --expid). Tests were done to ensure the same results if using a list of input files vs. exposure ids and night. Both were consistent.

## Test 1
This was tested at NERSC under a test production: ` /global/cfs/cdirs/desi/spectro/redux/test_joint_fit`.

Data from the minisv0 night of 20200219 was selected as it had a long-cal and science data on the same night.

Note: Preproc and exposure data from `andes` were copied over for the following exposures:
Note: As noted, only 5 spectrographs were tested as those were the ones with clean data.

All calibration data was deleted, as was `stdstar-?-*.fits` files prior to testing.

### The Data
**Exposures**:
ARCS: 50936,50937,50938,50939,50940
FLATS:  50943,50944,50945,50948,50949,50950,50953,50954,50955,50958,50959,50960
SCIENCES (from TILE 70003): 51073,51072,51060,51059,51058,51057,51056,51055,51054,51053

cleaned spectrographs: 0,3,6,7,9

### The Commands

The following tests were run successfully:

**psf**:
command: `desi_proc_joint_fit -n 20200219 -e 50936,50937,50938,50939,50940 --cameras=0,3,6,7,9 --obstype=arc --batch`

with log: `psfnight-20200219-00050936-00050937-00050938-00050939-00050940-a03679-35613547.log`

**flat**:
command: `desi_proc_joint_fit -n 20200219 -e 50943,50944,50945,50948,50949,50950,50953,50954,50955,50958,50959,50960 --cameras=0,3,6,7,9 --obstype=flat --batch`

with log: `nightlyflat-20200219-00050943-00050944-00050945-00050948-00050949-00050950-00050953-00050954-00050955-00050958-00050959-00050960-a03679-35646247.log`

**science**:
command: `desi_proc_joint_fit -n 20200219 -e 51073,51072,51060,51059,51058,51057,51056,51055,51054,51053 --cameras=0,3,6,7,9 --obstype=science --batch`

with log: `stdstarfit-20200219-00051073-00051072-00051060-00051059-00051058-00051057-00051056-00051055-00051054-00051053-a03679-35652372.log`

## Test 2
Run the joint standard star fitting routine on exposures that failed in andes. Selected low transparency night of 20200314. Found 3 tiles to test. The first tile was successful in all 4x450s exposures. The second had two failing exposures out of 4x450s exposures in standard star fitting. The third had two failing exposures, but in flux calibration. All three ran successfully using the new joint fitting. All three saw flux calibrations being successful.

### Data under:
/global/cfs/cdirs/desi/spectro/redux/test_joint_fit/exposures/20200314

In andes, standard star files were missing for the following spectrographs:
&nbsp; &nbsp; &nbsp; &nbsp; EXP 55458: missing specs = 0,2,3,5,6
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; 55459:  &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;  &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;  0,2,3,6

In new test,
&nbsp; &nbsp; &nbsp; &nbsp; NONE

In andes, cframe files were missing for the following exposures:
&nbsp; &nbsp; &nbsp; &nbsp; Tile 65008:  Exp = 55458, 55459
&nbsp; &nbsp; &nbsp; &nbsp; Tile 66019:  &nbsp; &nbsp;   &nbsp; &nbsp; &nbsp; &nbsp;   55476, 55478

In new test,
&nbsp; &nbsp; &nbsp; &nbsp; None

### Commands
**Tile 66000**: 
4x450s exposures. Previously Successful, **Still successful.** 
command: `desi_proc_joint_fit -n 20200314 -e 55442,55443,55444,55445 --batch`

logfile: `/global/cfs/cdirs/desi/spectro/redux/test_joint_fit/run/scripts/night/20200314/stdstarfit-20200314-00055442-00055443-00055444-00055445-a0123456789- 35697483.log`

**Tile 65008**: 
4x450s exposures. Previously failed stdstars (in exps 55458,55459). **Newly successful.** 
command: `desi_proc_joint_fit -n 20200314 -e 55456,55457,55458,55459 --batch`

logfile: `/global/cfs/cdirs/desi/spectro/redux/test_joint_fit/run/scripts/night/20200314/stdstarfit-20200314-00055456-00055457-00055458-00055459-a0123456789-35697597.log`

**Tile 66019**:  
4x450s exposures. Previously Successful stdstars but failed fluxcalib, **Newly successful.** 
command: `desi_proc_joint_fit -n 20200314 -e 55475,55476,55477,55478 --batch`

logfile: `/global/cfs/cdirs/desi/spectro/redux/test_joint_fit/run/scripts/night/20200314/stdstarfit-20200314-00055475-00055476-00055477-00055478-a0123456789-35697746.log`

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -783,7 +783,7 @@ if args.obstype in ['SCIENCE', 'SKY'] and (not args.noskysub ) and (not args.nop
         comm.barrier()
 
 #-------------------------------------------------------------------------
-#- Flux calibration
+#- Standard Star Fitting
 
 if args.obstype in ['SCIENCE',] and \
         (not args.noskysub ) and \
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -49,7 +49,7 @@ from desitarget.targetmask import desi_mask
 from desiutil.log import get_logger, DEBUG, INFO
 stop_imports = time.time()
 
-from desispec.workflow.desi_proc_funcs import get_desi_proc_parser, update_args_with_headers
+from desispec.workflow.desi_proc_funcs import assign_mpi, get_desi_proc_parser, update_args_with_headers
 from desispec.workflow.desi_proc_funcs import determine_resources, create_desi_proc_batch_script
 
 
@@ -61,31 +61,8 @@ parser = get_desi_proc_parser()
 args = parser.parse_args()
 log = get_logger()
 
-if args.mpi and not args.batch:
-    from mpi4py import MPI
-    comm = MPI.COMM_WORLD
-    rank = comm.rank
-    size = comm.size
-else:
-    comm = None
-    rank = 0
-    size = 1
-
-#- Prevent MPI from killing off spawned processes
-if 'PMI_MMAP_SYNC_WAIT_TIME' not in os.environ:
-    os.environ['PMI_MMAP_SYNC_WAIT_TIME'] = '3600'
-
-#- Double check env for MPI+multiprocessing at NERSC
-if 'MPICH_GNI_FORK_MODE' not in os.environ:
-    os.environ['MPICH_GNI_FORK_MODE'] = 'FULLCOPY'
-    if rank == 0:
-        log.info('Setting MPICH_GNI_FORK_MODE=FULLCOPY for MPI+multiprocessing')
-elif os.environ['MPICH_GNI_FORK_MODE'] != 'FULLCOPY':
-    gnifork = os.environ['MPICH_GNI_FORK_MODE']
-    if rank == 0:
-        log.error(f'MPICH_GNI_FORK_MODE={gnifork} is not "FULLCOPY"; this might not work with MPI+multiprocessing, but not overriding')
-elif rank == 0:
-    log.debug('MPICH_GNI_FORK_MODE={}'.format(os.environ['MPICH_GNI_FORK_MODE']))
+#- Check MPI flags and determine the comm, rank, and size given the arguments
+comm, rank, size = assign_mpi(do_mpi=args.mpi, do_batch=args.batch, log=log)
 
 #- Start timer; only print log messages from rank 0 (others are silent)
 timer = desiutil.timer.Timer(silent=(rank>0))
@@ -109,11 +86,6 @@ if rank > 0:
     args, hdr, camhdr = None, None, None
 else:
     args, hdr, camhdr = update_args_with_headers(args)
-    #- Update args to be in consistent format
-    if args.batch_opts is not None:
-        args.batch_opts = args.batch_opts.strip('"\'')
-    args.cameras = sorted(args.cameras)
-    args.obstype = args.obstype.upper()
 
 if comm is not None:
     args = comm.bcast(args, root=0)
--- a/bin/desi_proc_joint_fit
+++ b/bin/desi_proc_joint_fit
@@ -28,41 +28,15 @@ from desitarget.targetmask import desi_mask
 from desiutil.log import get_logger, DEBUG, INFO
 stop_imports = time.time()
 
-from desispec.workflow.desi_proc_funcs import get_desi_proc_joint_fit_parser, create_desi_proc_batch_script
-from desispec.workflow.desi_proc_funcs import update_args_with_headers
+from desispec.workflow.desi_proc_funcs import assign_mpi, get_desi_proc_joint_fit_parser, create_desi_proc_batch_script
+from desispec.workflow.desi_proc_funcs import load_raw_data_header, update_args_with_headers
 
 parser = get_desi_proc_joint_fit_parser()
 args = parser.parse_args()
 log = get_logger()
 
-
-
-if args.mpi and not args.batch:
-    from mpi4py import MPI
-
-    comm = MPI.COMM_WORLD
-    rank = comm.rank
-    size = comm.size
-else:
-    comm = None
-    rank = 0
-    size = 1
-
-#- Prevent MPI from killing off spawned processes
-if 'PMI_MMAP_SYNC_WAIT_TIME' not in os.environ:
-    os.environ['PMI_MMAP_SYNC_WAIT_TIME'] = '3600'
-
-#- Double check env for MPI+multiprocessing at NERSC
-if 'MPICH_GNI_FORK_MODE' not in os.environ:
-    os.environ['MPICH_GNI_FORK_MODE'] = 'FULLCOPY'
-    if rank == 0:
-        log.info('Setting MPICH_GNI_FORK_MODE=FULLCOPY for MPI+multiprocessing')
-elif os.environ['MPICH_GNI_FORK_MODE'] != 'FULLCOPY':
-    gnifork = os.environ['MPICH_GNI_FORK_MODE']
-    if rank == 0:
-        log.error(f'MPICH_GNI_FORK_MODE={gnifork} is not "FULLCOPY"; this might not work with MPI+multiprocessing, but not overriding')
-elif rank == 0:
-    log.debug('MPICH_GNI_FORK_MODE={}'.format(os.environ['MPICH_GNI_FORK_MODE']))
+#- Check MPI flags and determine the comm, rank, and size given the arguments
+comm, rank, size = assign_mpi(do_mpi=args.mpi, do_batch=args.batch, log=log)
 
 #- Start timer; only print log messages from rank 0 (others are silent)
 timer = desiutil.timer.Timer(silent=(rank>0))
@@ -108,14 +82,7 @@ else:
         #- NOTE: inputs has priority. Overwriting expids if they existed.
         args.expids = []
         for infile in args.inputs:
-            # - Fill in values from raw data header if not overridden by command line
-            fx = fitsio.FITS(infile)
-            if 'SPEC' in fx:  # - 20200225 onwards
-                hdr = fx['SPEC'].read_header()
-            elif 'SPS' in fx:  # - 20200224 and before
-                hdr = fx['SPS'].read_header()
-            else:
-                hdr = fx[0].read_header()
+            hdr = load_raw_data_header(pathname=infile, return_filehandle=False)
             args.expids.append(int(hdr['EXPID']))
             fx.close()
 
@@ -124,18 +91,13 @@ else:
     args.expid = args.expids[0]
     args.input = args.inputs[0]
 
+    #- Use header information to fill in missing information in the arguments object
     args, hdr, camhdr = update_args_with_headers(args)
-    args.obstype = args.obstype.upper()
+
+    #- If not a science observation, we don't need the hdr or camhdr objects,
+    #- So let's not broadcast them to all the ranks
     if args.obstype != 'SCIENCE':
         hdr, camhdr = None, None
-        
-    args.night = int(args.night)
-
-    # - Update args to be in consistent format
-    if args.batch_opts is not None:
-        args.batch_opts = args.batch_opts.strip('"\'')
-    args.cameras = sorted(args.cameras)
-    args.obstype = args.obstype.upper()
     
 if comm is not None:
     args = comm.bcast(args, root=0)
--- a/bin/desi_proc_joint_fit
+++ b/bin/desi_proc_joint_fit
@@ -90,30 +90,60 @@ if rank > 0:
     # - Let rank 0 fetch these, and then broadcast
     args = None
 else:
-    cam_str = args.cameras.strip(' \t').lower()
-    cameras = cam_str.split(',')
-    if cam_str[0] not in ['b', 'r', 'z'] and cameras[0].isnumeric():
-        args.cameras = []
-        for camnum in cameras:
-            for ccd in ['b', 'r', 'z']:
-                args.cameras.append('{}{}'.format(ccd, camnum))
+    if args.night is None or args.expids is None:
+        if args.inputs is None:
+            raise RuntimeError('Must specify --inputs or --night AND --expids')
+        else:
+            args.inputs = np.array(args.inputs.strip(' \t').split(','))
+            #- args.night will be defined in update_args_with_headers,
+            #- but let's define the expids here
+            args.expids = []
+            for infile in args.inputs:
+                # - Fill in values from raw data header if not overridden by command line
+                fx = fitsio.FITS(infile)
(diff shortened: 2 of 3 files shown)


moustakas:
Great to see the new joint fitting @akremin.

Is there any QA output we can review? For example, how do the jointly fit PSF models compare with the individually fit ones?

akremin:
That's a good question, John. I wasn't sure how to give relevant comparisons for each of these in a plot. 

The psf and flats have always been jointly fit, so I am not sure if justification is needed? But I should make some QA plots to show that they look like the ones produced for andes, or something similar.

For the stdstar joint fitting. I just tested that using jobs that had originally failed in andes due to low signal to noise. I'm adding that test to the original text of the PR above. Again, there isn't a plot for that, just the fact that things that had previously failed now run. 

I'll work on getting you some comparison plots next.

sbailey:
Thanks for the update and the details of the tests.  Initial reactions but not a deep review yet:

These filenames are getting out of out control.  :)  Let's brainstorm other options.

The case you are testing here is mimicking nightly processing where we would have run `desi_proc ... --nostdstarfit` on every exposure as it arrived, and then pick up with running `desi_proc_joint_fit ...` after we've noticed that the exposures have moved on to a new tile?

You've already factored out some things like argparsing, but let's look for other cases of duplicated code between desi_proc and desi_proc_joint_fit and consider whether that can be put into library functions, e.g. extname SPEC vs. SPS parsing when reading the headers.

The psfnight-20200219-*.slurm script specifies a timing file output, and the log looks like it ran successfully, but I don't see a timing json file for it.  Check for a logic bug where it exits before writing the timing file.

akremin:
Agreed the names are too long but there was no obvious way to losslessly include the exposure information for globbing. But I did think of an idea similar to the camera shortening where you determine the greatest common overlap and only list the trailing digits:

E.g 00059456-00059457-00059458-00059460.ext

Becomes 000594-56_57_58_60.ext

Something to that effect.


Yes for the sciences this tests joint standard star fitting as it would be done in nightly running or during full production reruns for a given set of exposures of a single tile.


I've also factored out more of the intro code than just arg parsong, but there is more that could be done with, e.g. common handling of mpi keywords and variables and the timing object. 

The header code you mention is also split out into update_args_with_headers() in desi_proc_funcs.py. However to make the --inputs work I had to reuse that code again, so it was reintroduced back into the body. I had considered making another function that can be called in all these cases, but had chosen to leave it for the time being. Based on your comment, I'll move forward with that and turn it into a function. For now I'll put it under desi_proc_funcs.py until we can determine a proper home. 

For the timing information, I'll take a look tomorrow morning and figure out what happened there. It might well be an outstanding bug. I hadn't checked that.

weaverba137:
If I may interject: this sounds like you're trying to pack a lot of metadata into a file name. There's no possibility of using a database here is there? Or even just using a keyword-value store, which could be implemented with FITS headers if necessary?

akremin:
We could do something more sophisticated like that, but these are just names for the slurm scripts and resulting logs. The names under discusduon aren't used for any of the data products. 

The scripts and logs are simple text files and contain all the relevant information inside them, so we could potentially omit things as well. The key goals are to (1) make the script names unique and (2) make it easy for someone to identify what script was used to produce what data. 

I'm not sure that having to open fits files or reference a database are as ideal for (2), if we can reasonably do it using another method.

akremin:
@sbailey I just reran the exact same slurm job for the psf calculation and it produced a timing file. It's most likely that I accidentally deleted the original one while cleaning up the directory for review with this PR.

To get back to the naming things, I'll provide a few brainstorming ideas below for comment. Keep in mind most tiles do not contain such a large number of exposures, so this is an extreme case:

Original: stdstarfit-20200219-00051073-00051072-00051060-00051059-00051058-00051057-00051056-00051055-00051054-00051053-a03679.slurm

Proposal 1, remove night and leading zeros: stdstarfit-51073-51072-51060-51059-51058-51057-51056-51055-51054-51053-a03679.slurm

Proposal 2, all of prop 1 with only unique digits: stdstarfit-510-73_72_60_59_58_57_56_55_54_53-a03679.slurm

weaverba137:
I see. Maybe the way to go would be to use `uuid` or a SHA on the various exposures (like a git commit), and just record the mapping in a text file.

sbailey:
Thinking aloud about filenames (at the risk of bikeshedding, but we've gotta do something shorter...):
  1. we want them to be unique so that a debugging or cleanup run doesn't overwrite a previous one
  2. storing metadata in the filename is useful for helping humans find the file they want quickly, but the filename itself doesn't have to be a complete representation of that metadata, and we shouldn't rely on the filename for how to get that metadata.

uuid or sha would be good for (1) but a pain for (2).  The current scheme is unique and you can find the filename, but ouch they are long, and even @akremin suggestions are still quite long.  Another idea for consideration:

  * psfnight-{night}-{sha(expids)}
  * fiberflatnight-{night}-{sha(expids)}
  * stdstarfit-{night}-{tiled}-{sha(expids)}
    - normally there would only be one of these per tile and you can ignore the sha
    - if we do observe a tile twice non-contiguously we could end up with 2 of these for the nightly processing, and the sha would keep them unique even if it isn't particularly useful for knowing which is which.

Maybe include {first_expid} in the name?  I'm not sure how helpful that would be.

consider using short SHAs (like 6 or 8 characters, or maybe even 4?).

Regardless of filesnames for the scripts, having yaml/json/something that maps which exposures are combined now will be useful.  That may be more clear with Anthony's next round of updates with the speclog exposures/processing tables.

weaverba137:
Indeed, I was thinking about how git commits have a full SHA, but are usually abbreviated to 6-8 "digits".  Over the course of the survey, we should be careful about collisions, but I doubt that would happen in a single run.

akremin:
I moved more shared code between desi_proc and desi_proc_joint_fit into commonly called functions. All the functions are currently dumped in the desi_proc_funcs.py. Most can probably stay there or a similar-in-spirit file but some might be useful to move to more relevant homes.

For the filenames, the sha proposal is an interesting one that I could potentially get on board with. I might be preferential to just including the first exposure id rather than the SHA. As a SHA is still unreadable to humans.

The other challenge is that occasionally a job is rerun with a smaller set of cameras because, e.g. only some failed on a prior job, so camera uniqueness is a nice feature (though not a requirement).

I notice you (@sbailey) retain the night as the only metadata, even though they are housed under a nightly directory. Is there strong reasoning in that?


Another avenue is that in the upcoming pipeline refactor each job has a unique internal id, which is stored and is therefore directly mapped to these files and the metadata. However, it doesn't help with instant human recognition without reference a table. Maybe that can't be avoided, however.

sbailey:
> I notice you (@sbailey) retain the night as the only metadata, even though they are housed under a nightly directory. Is there strong reasoning in that?

Not necessarily a strong reason, but I do like the principle of keeping filenames unique and meaningful even across directories.  If we drop the {night} part it's still unique due to the SHA, but even less readily interpretable if taken out of the directory context.  I'm more sympathetic to dropping {night} if {first_expid} is put back in, though I think {night} is the more context-relevant piece anyway (it is the psf*night*, fiberflat*night*...)

I had also considered {night}-{first_expid} or {night}-{first_expid}-{num_exp} but those aren't completely guaranteed to be unique under special case reruns.  I'm ambivalent about how much we should care about that.

Good point about the cameras.  OK to add that back in.

akremin:
I suggest using e.g. psfnight-{night}-{first_expid}-{camword}.slurm then.

This makes it analogous to the desi_proc scripts which are e.g. science-{night}-{expid}-{camword}.slurm

The use of only the first exposure ID does allow for non-uniqueness if different exposure sets are run, but that shouldn't happen unless an exposure is deemed bad after the fact and a new run is done without it. In which case, the old data products will be overwritten anyway, so I don't think there is necessarily harm in having the script that generates them also being overwritten.

We can certainly add a SHA for uniqueness, but then we're getting back into increasingly long filenames.

sbailey:
@akremin sounds good, let's try `psfnight-{night}-{first_expid}-{camword}.slurm`, and if there are unanticipated cases where the potential non-uniqueness is a problem, we can revisit that.

akremin:
Okay I made that change. I also sorted the expids so that the smallest one is now the one listed in the filename.

As of now both the single desi_proc scripts and the joint desi_proc_joint_fit scripts are still zero-padded exposure id's (consistent with what desi_proc has always done). But those could be changed to reduce the lengths by another 3 (or eventually 2 characters).

akremin:
## Comparing cframes from two productions:
blue - andes
orange - test_joint_fit

Trying to give a rough graphical comparison of spectra created using the old single exposure determination of standard stars versus the new joint fits. One exposure (55442) is from a tile that had sufficient signal-to-noise during its observations that the standard star fits were obtained without issue. The second exposure (55457) comes from a set of exposures where stdstar-* files were produced for half the exposures, but the other half failed. While 55457 failed, the idea is that it was likely a marginal case given that the next exposure of the same targets failed.

### exposure 00055442, z camera

From a set of exposures with sufficient signal to noise, where the andes reductions didn't have trouble. There is no discernible differences between the two reductions.

**spec 0, fiber 270, cframes**:
plot_frame -i /global/cfs/cdirs/desi/spectro/redux/{andes,test_joint_fit}/exposures/20200314/00055442/cframe-z0-00055442.fits --fiber 270 --legend
image

**spec 5, fiber 2780, cframes**:
plot_frame -i /global/cfs/cdirs/desi/spectro/redux/{andes,test_joint_fit}/exposures/20200314/00055442/cframe-z5-00055442.fits --fiber 2780 --legend
image

### exposure 00055457, z camera
From a tile where 2 of 4 exposures failed to determine standard star fits. This was one of the two successful exposures of the tile in andes (because I can't compare to a cframe that doesn't exist in a failed job). But my assumption is that if 2/4 failed, the others were likely challenging fits for single-exposure std star fitting. We see either smaller fluxes using the joint fit and/or reduced variance. Also, sky subtraction artifacts that differ between the two productions.

**spec 0, fiber 270, cframes**:
plot_frame -i /global/cfs/cdirs/desi/spectro/redux/{andes,test_joint_fit}/exposures/20200314/00055457/cframe-z0-00055457.fits --fiber 270 --legend
image
image

**spec 5, fiber 2780, cframes**:
plot_frame -i /global/cfs/cdirs/desi/spectro/redux/{andes,test_joint_fit}/exposures/20200314/00055457/cframe-z5-00055457.fits --fiber 2780 --legend
image
image

julienguy:
For tile 66000 (exposures 55442 ... 55445), the calibration is ok (consistent flux for bright star within ~5%). See figures.
```
plot_frame -i /global/homes/j/jguy/redux/test_joint_fit/exposures/20200314/0005544{2,3,4,5}/sframe-b0-0005544* --fiber 20 --legend --rebin 8
plot_frame -i /global/homes/j/jguy/redux/test_joint_fit/exposures/20200314/0005544{2,3,4,5}/cframe-b0-0005544* --fiber 20 --legend --rebin 8
```
sframe-20200314-00055442-00055445-fiber-20
cframe-20200314-00055442-00055445-fiber-20

Predicted colors of std stars are also good with a g-r rms of 0.028.
stars-55442

However, for tile 65008 (exposures 55456 ... 55459), it seems something went wrong with the flux calib of SP0 (see sframe and
 cframe figures).

```
plot_frame -i /global/homes/j/jguy/redux/test_joint_fit/exposures/20200314/0005545{6,7,8,9}/sframe-b0-0005545* --fiber 453 --legend --rebin 8
plot_frame -i /global/homes/j/jguy/redux/test_joint_fit/exposures/20200314/0005545{6,7,8,9}/cframe-b0-0005545* --fiber 453 --legend --rebin 8
```

sframe-20200314-00055456-00055459-fiber-453
cframe-20200314-00055456-00055459-fiber-453

julienguy:
This points to a potential problem when combining good and very poor data. 
 - For tile 65008, the fit of standard stars for exposure 55456 in Andes gives a G-R rms of 0.029 whereas the combined fit in test_joint_fit gives an rms of 0.043. Combining data degraded the precision of the stellar model fit.
 - In contrast, for tile 66000, the fit is improved with the combination of exposures (rms=0.032 in Andes and rms=0.028 in test_joint_fit)

akremin:
FYI: My tests used the andes files as inputs for standard star joint fitting, so the sframe-* files are directly from the andes production run. 

Very interesting results. I think the standard star rms values you quote make sense to me naively. Though correct me if either of my inferences below are incorrect:

The 66000 tile had 4 exposures that could all be fit individually, so the combination improved things. 

For the 65008 tile, two exposures had previously failed on individual fitting, indicating poorer data quality for half of the tile exposures. So I'm not surprised that for exposure 55456 (which did successfully run individually in andes) it would actually see worse RMS under joint fit. It was an above average exposure and was therefore degraded by the poorer exposures in the tile.

julienguy:
I checked the part of the code addressing the standard star fit. It looks fine. Maybe propagate option --delta-color to desi_proc_joint_fit or consider adding hardcoded value '--delta-color 0.1' to the call to 'desi_fit_std_stars' because this allows the code to run faster (and we now know we don't have a significant bias on the g-r color of the models and the rms of the predicted colors is about 0.03).
This PR is good for merging for me.

akremin:
Okay, I added it as a hardcoded value. In general I hate doing that, but I don't think we'd ever use it as a desi_proc_joint_fit parameter, so it seems excessive to add it as one. We'd also just be shifting the problem as the pipeline calling desi_proc_joint_fit would then just get it written as a hardcoded value anyway, so that would just pass the problem up one level without resolution.  At least it is an input to the underlying code.

PR #1020, opened 2020-11-05
Newrdnoise

zkdtc:
1) Previous OBSRDN[A-D] are now OVSRDN[A-D]
2) Current OBSRDN[A-D] are measured using slices defined in desispec_param.yml 'preproc' section. 
3) The regions are shown in the plot below
b


sbailey:
Please make a new pull request that has just the readnoise update, based upon a new branch off of current master.  This PR branch includes dark/bias changes that are also in #1013, which will create merge conflicts when that PR gets more updates.

Unit tests are failing; please check with "python setup.py test".  If they work for your environment but not Travis, let us know.

Is there a specific reason for not using a larger range of rows to get larger patches for better statistical accuracy on the read noise?

zkdtc:
Oops, just saw your comments @sbailey 
The width is to make sure the region is far from fiber traces enough, while the location and length are chosen to avoid the bad residual we saw when the work began. Since the residuals are removed now,  we are good to expand it to larger length. Do you want the full length?
I made the newrdnoise branch out of newbias4 branch because the master branch is not removing the CCD bias+dark correctly yet. Maybe I should wait for the merge of newbias4 then revise and make a new PR? It will not take more than  10 minutes :-D

sbailey:
@zkdtc please add the plot that you showed at group meeting on Monday, documenting why it doesn't work to extend the regions all the way to the middle of the CCD (scattered light reaching into the edge columns).

Also, I think you had identified one camera where the traces were getting uncomfortably close to the RDNOISE regions chosen here.  Which camera was that?  Do you have a fix, e.g. a per-camera override for just that camera to use a small patch?

py/desispec/preproc.py
```diff
@@ -609,7 +612,17 @@ def preproc(rawimage, header, primary_header, bias=True, dark=True, pixflat=True
...
             gain_message    = 'gain not applied to image'
-        header['OBSRDN'+amp] = (median_rdnoise,rdnoise_message)
+
+        if 'preproc' in desispec_params and len(rawimage)>4000: # There are 1100*900 b0 image in the test script
+            header['OVSRDN'+amp] = (median_rdnoise,rdnoise_message)
+            ind=desispec_params['preproc']['rdnoise']['PARAMS']['OBSRDN'+amp+'_'+camera.upper()[0]]
+            ind_sec = parse_sec_keyword(ind)
+            d=rawimage[ind_sec]
+            d = d[~np.isnan(d)].ravel()
+            header['OBSRDN'+amp]=(gain*(np.percentile(d,84)-np.percentile(d,16))/2.,rdnoise_message)
```
sbailey, line 622: Please use the *same* statistical method for calculating the readnoise in the overscan vs. the data region.  Since the data are quantized, the method here (difference of percentiles) will result in quantized readnoise measurements.  Please see how the readnoise is calculated in the `_overscan()` function (std of sigma-clipped pixels, with a correction for the bias from the sigma-clipping), pull that out into a separate function that both can call, and then use that function for calculating the readnoise in both cases.

Additionally, that function should accept a mask so that known bad pixels can be pre-flagged and ignored without potentially confusing the sigma-clipping.

Ideally cosmics would also be ignored (i.e. part of the mask), but that creates a chicken-and-the-egg problem: cosmic finding uses the estimated variance, but the variance depends upon the readnoise, but we want to mask cosmics when calculating readnoise.  Maybe the sigma clipping will be sufficient for that, but we'll have to watch out for outlier readnoise measurements and perhaps adjust (e.g. using a fast approximate cosmic finder first, or using the overscan readnoise for the cosmics, and then updating with the final data-region readnoise).

zkdtc:
Why we select the area shown to calculate rdnoise
It is because there is significant scattered light halo at the center of the CCD, as shown in the first figure below. Camera b4 has the area edge very close to the traces. Will implement a special case parameter for this camera in the yaml file. 

Screen Shot 2020-11-23 at 12 51 38 PM
Screen Shot 2020-11-23 at 12 52 04 PM

zkdtc:
Add _sigma_clip to calculate sigma clipped mean and std as suggested, enabling mask. The new function is used in _overscan and rdnoise measurement.

julienguy (changes requested):
Looks good, but 
* please change the header keyword of the regions so that they have a 8 character length or less. You do not need to keep the band name (_B,_R,_Z) in the header keyword because it is written in a different header for each camera. But, you need to change the first part of the name 'OBSRDNA' ... because it is the same as the keyword for the readnoise value. TO be constructive, what about: 'RDNSECA' (B,C,D) ? 
* also please merge master into this branch because there has been some refactoring of the preprocessing that could generate conflicts.

zkdtc:
Thank you @julienguy  I have revised the codes and create a new branch 'newrdnoise2' to include all changes. The  'OBSRDNA's are kept as is because we intentionally use it to replace the previous values measured in overscan regions. The RDNOISE measured in overscan regions are stored as 'OVSRDN'+amp. Would be happy to change the name if we agree on a better one.

sbailey:
This PR is superseded by #1041, so closing this one.

PR #1026, opened 2020-11-13
write timing file to batch script dir with other logs

sbailey:
Minor cleanup:  previously the default timing*.json file written by `desi_proc ... --batch` was dumped into whatever directory you were in when running `desi_proc`, instead of writing it into the batch script directory along with the other batch logs.  This PR adds an explicit path for the timing file so that it always ends up in the same place regardless of where `desi_proc ...` is run from.

@akremin please check (can use `desi_proc ... --batch --nosubmit`) and inspect the written script if you don't want to run a full exposure.  This might also explain that missing arc timing file from one of your runs.

### DIFF
--- a/py/desispec/workflow/desi_proc_funcs.py
+++ b/py/desispec/workflow/desi_proc_funcs.py
@@ -366,6 +366,7 @@ def create_desi_proc_batch_script(night, exp, cameras, jobdesc, queue, runtime=N
 
     if timingfile is None:
         timingfile = f'{jobname}-timing-$SLURM_JOBID.json'
+        timingfile = os.path.join(batchdir, timingfile)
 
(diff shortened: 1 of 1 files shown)


akremin:
Thanks @sbailey. I noticed this issue and put it on my personal to-do to resolve. Great to see it was as simple as this to fix. I tested calling from a random directory with both desi_proc and desi_proc_joint_fit (which both use this underlying function). It worked for both. 

desi_proc -n 20200219 -e 50940 --cameras=0,3,6,7,9 --obstype=arc --batch --nosubmit
desi_proc_joint_fit -n 20200219 -e 50936,50937,50938,50939,50940 --cameras=0,3,6,7,9 --obstype=arc --batch --nosubmit

akremin (approved):

PR #1028, opened 2020-11-16
Nightwatch issue162 warning

jalasker:
Nightwatch is unable to handle images with !=500 fibers and this branch issues specific warnings about that.


sbailey:
Concept is good, but please run `python setup.py test` which will reveal a syntax error in your latest commit (an errant "\" on line 67 causing a syntax error on line 68).

Also, on lines 66 and 69, typically "&" and "|" are reserved for bitwise-and/or operations, while "and" and "or" are used for boolean and/or like here.  In this case it is equivalent given your parentheses, but they do have different operator precedence that can have some surprising side effects, so when you are doing operations on booleans (like here) it is generally better to use "and" and "or", and reserve "&" and "|" for cases where you really do mean to be doing bitwise operations on integers.

PR #1029, opened 2020-11-18
New bias+dark model

sbailey:
This PR supersedes PR #1020 since I wanted to combine some changes from that branch (newbias4) with the dark/bias units updates from PR #1023 that had been merged into master.  Creates and uses new dark model:

model(x,y,t) = bias(x,y) + dark(x,y)*t + nonlinear(y,t)

Compared to master:
  * moved `bin/desi_compute_bias` and `bin/desi_compute_dark` functionality into functions in `desispec.ccdcalib` so that they can be independently called from other scripts
  * fixed units bug so that darks are consistently calculated and used in electron/s (not ADU/s).

Compared to newbias4 (PR #1020):
  * combined `desi_create_bias_dark.py` and `desi_bias_dark_1d_model.py` into `bin/desi_compute_dark_nonlinear`.
  * calculate dark model using overscan trimmed DARKs in electron/s instead of non-trimmed ADU/s
  * generate nightly (daily) BIASes while evaluating the darks
  * use `desispec.ccdcalib.compute_dark_file` (formerly in `bin/desi_compute_dark`) to combine darks, including masking cosmics, taking a masked median, and then doing a sigma-clipped average about that median.

Example updated biases + darks are in /global/cfs/cdirs/desi/users/sjbailey/spectro/desi_spectro_calib/trunk/ccd/bias-*-20201115.fits and dark-*-20201115.fits (named after date they were generated, using DARKs and ZEROs from 20200729 and 0730).  In the `spec/sm*/sm*.yaml` files, I updated every config that had the same DETECTOR, CCDCFG, and CCDTMING as the DARKs taken on 20200729 & 0730.  This means that these are *not* updated:
  * sm7=[brz]6: no data from 0729 and 0730
  * sm5-r2 V20201113 (latest config): new DETECTOR
  * sm1-r4 V20201112 (latest config): new CCDCFG

## How to run ##

Generate a new model (takes 30-40 min per camera):
```
desi_compute_dark_nonlinear --days 20200729 20200730 --camera b0 \
        --darkfile dark-sm4-b0-20201115.fits.gz \
        --biasfile bias-sm4-b0-20201115.fits.gz 
```
This uses a `temp/` directory in the output directory of the darkfile to collect the daily biases and per-exptime combined DARKs, so that if you need to rerun it can reuse any pieces that were previously successfully calculated.

Convenience wrapper to generate one batch script per camera:
```
from desispec.ccdcalib import make_dark_scripts
days = [20200729, 20200730]
outdir = os.path.expandvars('$SCRATCH/desi/ccdcalib')
make_dark_scripts(days, outdir=outdir, nosubmit=True)
```
if `nosubmit=False` it will submit the batch scripts to the realtime queue.

Example outputs in `/global/cscratch1/sd/sjbailey/desi/ccdcalib/`, including the intermediate files in `ccdcalib/temp`.  The final `bias-*` and `dark-*` files were put in `/global/cfs/cdirs/desi/users/sjbailey/spectro/desi_spectro_calib/trunk/ccd/`.
Also `ccdcalib/test/` contains directories with several example exposures processed with the current model (master), the new model with default bias (new), and the new model using a nightly bias (newbias).

## Performance ##

Comparing the row profiles for an individual dark exposure that contributed to the overall model, it is pretty good though not perfect:
y1d-2020730-60755-300s-L
Notes:
  * r2 is the bad CCD, thus the borked profiles
  * I don't know what's up with r7 (TODO)
  * several others are noisier than ideal, but not worse than current model (e.g. r9, z7, z9)
  * for tiny plots, I only showed every 4th row, thus the axis range to ~1000 instead of ~4000

Applying this model to current data is less encouraging, though arguably still better than current model (202011/62806 300s dark using nightly bias build from ZEROs on same night):
y1d-20201114-62806-300s-nightlybias-L
Notes:
  * r2 is new chip which doesn't have a new model yet, and apparently using a nightly bias with the old model (orange) is actually worse than using the old default bias with the old default dark (blue, with offset but not step at amp boundary).
  * r4, and [brz]6 are other cases of not having a new dark model
  * overall these y-dependent wings are partially absorbed by the bias (ZERO) model, and partially absorbed by the time-dependent nonlinear dark model, and sometimes this model overcorrects (e.g. b5, b7, b9)

@julienguy please review; let's chat.

### DIFF
--- a/bin/desi_compute_bias
+++ b/bin/desi_compute_bias
@@ -1,16 +1,8 @@
 #!/usr/bin/env python
 
 
-import os,sys
-import astropy.io.fits as pyfits
 import argparse
-import numpy as np
-
-from desiutil.log import get_logger
-from desispec.preproc import parse_sec_keyword, _overscan
-from desispec.calibfinder import CalibFinder
-
-
+from desispec.ccdcalib import compute_bias_file
 
 parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter,
 description="Compute a master bias from a set of raw data bias images",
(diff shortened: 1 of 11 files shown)


julienguy:
I preprocessed 20201114/00062859 (DARK of 300.0 sec) using this dark model + a master bias from the 10 previous zero exposures. I see residual backgrounds of ~0.4 electrons for most blue CCDs, an offset + a step of about ~0.5 electron in r7 and r8 between amplifiers and large negative values or the first and last rows of r5. So unfortunately there are still things to figure out here ... possibly related to the change of dark with time since VCCD=ON. For this dark VCCD has been on for  4992.3sec.

julienguy:
It is difficult to reduce further biases at a level < 0.5 electron. So we may merge this PR and work on observing conditions that would be more repeatable, for instance with leaving the voltages ON on the CCDs.
As an example the following figure shows in orange the median pixel values per quadrant row in a 600sec dark after a preprocessing that includes the dark subtraction of the above model. 

The residuals are of 0.3 electrons , up to 1 electron on the edges of the CCD. In order to try and reduce further the residuals, we measure the CCD level in side bands of width 50 pixels on the corners of the CCD (the blue curves), and then use the dark model to fit a correction using the side band (the red curves). It improves a bit the tails on the edges but not the average level where we still have 0.3 electrons offsets (the green curves show the residuals to this last correction). 
The result is not improved because there are residual spatial variations as a function of CCD column at a level of a few 0.1 electron in the CCD. And we cannot measure those easily on science frames.

preproc-b7-00063146
 
We will continue to investigate possible improvements, for instance a stitching of the background level at the amplifier boundary using pixels between fiber bundles.

sbailey:
I want to check what happened with the edge of r5 model before merging.  That might involve an update to the robustness of the fits.

sbailey:
I checked the edge of r5: there is genuinely a huge amount of dark current in row 0 coming from a bright feature leaking charge around column 1260.  This is reflected in the model, but since the charge is so large the differential to new data is significant compared to the normal dark current in other pixels.  This row is masked in the r5 pixmask, so I don't think it actually matters here (i.e. our eyes are being draw by an excursion in a masked row).

I will proceed with committing the svn files (one by one!) before merging the code.

PR #1030, opened 2020-11-21
New Daily Processing Manager and Accompanying Functions

akremin:
This PR is still a work in progress. Docstrings are still missing in a fair number of the underlying functions (under /workflow/). Those will be added before this is merged.

#### This adds three new scripts to the /bin/ directory:
`desi_daily_proc_manager`: The new daily pipeline manager that submits jobs to the queue and monitors the results.
`create_exposure_tables`: A helper script that can be used to generate summary tables of raw exposure data on a nightly basis.
`create_processing_tables`: A helper script that can be used to generate a summary table of an entire processing run of the data. This includes jobtypes, dependencies, uniquely defined internal job ids, queue job ids, and more.

#### Three files are added to the python /py/desispec/scripts/ subdirectory:
`daily_processing.py`: The main calling function associated with desi_daily_proc_manager
`exposuretable.py`: The main calling function associated with create_exposure_tables
`processingtable.py`: The main calling function associated with create_processing_tables

#### A number of additional files are added to a relatively new subdirectory /py/desispec/workflow/:
`exptable.py`: Contains the code that defines the columns of an exposure table, creates the tables, and names the tables, as well as other helper functions.
`proctable.py`: Contains the code that defines the columns of a processing table, creates the tables, and names the tables, as well as other helper functions. Functions to transform an exposure table (or exposure table row) to a processing table (or row) is also provided.
`utils.py`: Various convenience functions for listing files in a path, getting environment variables, etc.
`procfuncs.py`: Where the majority of the daily_processing code lives. It has the function calls to interface with desi_proc, as well as use and manipulate processing tables rows. Other helper functions are also in there if they didn't have a more obvious home within /workflow/.
`queue.py`: Functions that can call and ingest SLURM scheduler information.
`tableio.py`: Reading and writing functions for exposure tables and processing tables.  Functions required for e.g. transforming arrays into strings for storage in a single csv column cell are also here.
`timing.py`: Anything that deals with timing; which is used for querying the SLURM schedule, generating timestamps, and identifying the current night and time.

### Testing
Tests were done at NERSC under the `SPECPROD` directory "workflow_test". The manager was able to identify data in a specified raw data directory and submit jobs to the queue with appropriate dependencies. This includes joint fitting jobs for arcs, flats, and standard stars. It also successfully monitored the SLURM queue and identified when jobs succeeded, updating them with "COMPLETE" in the processing table.

Test command: 
`export SPECPROD=workflow_test`
`desi_daily_proc_manager --cameras=0,3,6,7,9 --override-night=20200219 --raw-data-path=/global/cfs/cdirs/desi/spectro/redux/kremin/raw_dir --ignore-cori-node`

Output in: `/global/cfs/cdirs/desi/spectro/redux/workflow_test`

Also tested the other two scripts with:
`create_exposure_tables -n 20200314`
`create_processing_tables -n 20200314`

both were successful.


sbailey:
Thanks for this significant update to the job processing organization.  I expect this review will take several iterations, so starting with big picture items:

As we discussed offlist: It doesn't yet work if desi_daily_proc_manager is started before any data has arrived for the night.  It recognizes that the exposures table is empty, won't write it, and then exists.  If it finds exposures that it won't process (like ZEROs), it notes those and then sleeps, but if it wakes up an not arc/flat/science exposures have arrived yet, it exits.  This is a blocking factor that needs to be fixed before we can use this in operations.

The sleep timeouts of 300 sec should be tunable.  I see the speed_modifier for dry-run, but for testing I wanted it to check frequently and actually submit jobs.

There is one code path that sleeps for 1800 sec, which is pretty long time.  Let's discuss if the logic can just be "if no new data, sleep for N seconds before checking again".

I tested with an arcs and flat calibration sequence from 20201119.  It appeared to submit the jobs with the proper dependencies, with some notes/caveats:
  * psfnight waits on "afterany" on the arc jobs, i.e. an individual arc job can fail and psfnight will still run, which I think is what we want for daily processing robustness
  * flats wait on "afterok" for psfnight, i.e. the psfnight has to succeed for them to run.  I think that's ok as a default, but we'll need a way to say "oops, psfnight is messed up tonight but proceed anyway with default PSF for now".
  * nightlyflat waits on "afterok" for the first flat, but "afterany" for the others, which seemed odd.  Our nightly flat algorithm is more strictly dependent upon the full 4x3 flats succeeding, so I think we want "afterok" on all of them. And we'll also want a way to say "oops, flats are messed up tonight, proceed anyway with a default flat"

After getting a full suite of arcs+flats in the queue, I then tried to proceed with the non-standard operations of 20201119 where we took additional arcs.  desi_daily_proc_manager didn't seem to find them, and was stuck in the "Starting iteration N" loop with 1800 sec waits.  I don't know whether it is waiting for the existing arcs+flats to finish before doing anything with the new data, or if it is genuinely missing them.  What is the expected behavior here?

It doesn't yet process ZEROs and make a nightly bias, but neither does the existing `desi_daily_proc`, so I won't complain too loudly, but we will want to get those in really soon.

In the very short term, let's address the issue of it exiting if no data has arrived yet, and then get setup to use this branch for actual daily processing in the upcoming days while still working on other issues before merging.  Anything that blocks actual daily processing gets top priority, and other things (like docstrings or code structure) is second priority.

I'll submit other comments inline as part of a review, or add additional structural comments later.

py/desispec/scripts/daily_processing.py
```diff
@@ -0,0 +1,239 @@
...
+        night = int(override_night)
+        print(f"True night is {true_night}, but running for night={night}")
+    else:
+        night = true_night
+
+    ## Get default values for input variables
+    if procobstypes is None:
+        procobstypes = default_exptypes_for_proctable()
+    if camword is not None:
+        print(f"Over-riding camword in data with user provided value: {camword}")
```
sbailey, line 66: let's convert `print` statements to `desiutil.log.get_logger` logging with info, debug, warning, etc.

py/desispec/scripts/daily_processing.py
```diff
@@ -0,0 +1,239 @@
...
+    ## Now we resubmit failed jobs and their dependencies until all jobs have un-submittable end state
+    ## e.g. they either succeeded or failed with a code-related issue
+    ii,nsubmits = 0, 0
+    while ii < 4 and continue_looping(ptable['STATUS']):
+        print(f"Starting iteration {ii}")
+        ptable, nsubmits = update_and_recurvsively_submit(ptable, submits=nsubmits, start_time=nersc_start,end_time=nersc_end,
+                                                          ptab_name=proc_table_pathname, dry_run=dry_run)
+        write_table(ptable, tablename=proc_table_pathname)
+        if continue_looping(ptable['STATUS']):
+            time.sleep(1800*speed_modifier)
```
sbailey, line 229: This is a really long sleep; could it just be the standard sleep time?

py/desispec/scripts/daily_processing.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+    ## All jobs now submitted, update information from job queue and save
+    ptable = update_from_queue(ptable,start_time=nersc_start,end_time=nersc_end, dry_run=dry_run)
+    write_table(ptable, tablename=proc_table_pathname)
+
+    ## Now we resubmit failed jobs and their dependencies until all jobs have un-submittable end state
+    ## e.g. they either succeeded or failed with a code-related issue
+    ii,nsubmits = 0, 0
+    while ii < 4 and continue_looping(ptable['STATUS']):
```
sbailey, line 223: hardcoded 4 iterations?  What's the core stopping condition that it is trying to achieve here?

py/desispec/scripts/exposuretable.py
```diff
@@ -0,0 +1,105 @@
...
+                                       get_surveynum, get_exposure_table_path, get_exposure_table_name, night_to_month, \
+                                       get_exposure_table_column_defs
+from desispec.workflow.os import define_variable_from_environment, listpath, pathjoin
+from desispec.workflow.timing import get_night_banner
+from desispec.workflow.tableio import write_table
+
+
+
+def create_exposure_tables(nights, path_to_data=None, exp_table_path=None, science_types=None, \
+                           exp_filetype='csv', verbose=False, overwrite_files=False):
```
sbailey, line 19: As part of documenting this, we'll need documentation about what all of the columns in the exposure table are.  I think this is the table that a human would edit to override things like marking a specific exposure as bad (or even fixing an incorrect keyword?), so that procedure will need to be clearly documented too.

py/desispec/scripts/exposuretable.py
```diff
@@ -0,0 +1,105 @@
...
+            exptab_name = get_exposure_table_name(night, extension=exp_filetype)
+            exptab_name = pathjoin(exptab_path, exptab_name)
+            write_table(nightly_tabs[night], exptab_name, overwrite=overwrite_files)
+        else:
+            print('No rows to write to a file.')
+
+        return nightly_tabs
+
+
+if __name__ == '__main__':
```
sbailey, line 85: Is this `if __name__ == '__main__':` section now deprecated by `bin/create_exposure_scripts`?  Though see comment there about whether that needs to be a command line script.  If running this from the command line is a ninja operation, it might be ok to leave this here as the secret place to `python py/desispec/scripts/exposuretable.py ...`, but in that case it should be commented here why this exists.

py/desispec/scripts/exposuretable.py
```diff
@@ -0,0 +1,105 @@
...
+    overwrite_files = False
+    verbose = True
+    science_types = ['arc', 'flat', 'twilight', 'science', 'sci', 'dither']
+
+    if 'DESI_SPECTRO_DATA' not in os.environ:
+        os.environ['DESI_SPECTRO_DATA'] = pathjoin(os.path.curdir, 'test_raw_data')
+    if 'DESI_SPECTRO_REDUX' not in os.environ:
+        os.environ['DESI_SPECTRO_REDUX'] = os.path.curdir
```
sbailey, line 93: Having defaults for these is understandable for testing, but let's remove this for production code, since normally not having those defined means you've done something wrong in the setup.

py/desispec/scripts/exposuretable.py
```diff
@@ -0,0 +1,105 @@
...
+        else:
+            print('No rows to write to a file.')
+
+        return nightly_tabs
+
+
+if __name__ == '__main__':
+    overwrite_files = False
+    verbose = True
+    science_types = ['arc', 'flat', 'twilight', 'science', 'sci', 'dither']
```
sbailey, line 88: Do we really have a science_type "sci"?  And should this be obstype?  Are you trying to distinguish obstypes that we process vs. ones we don't?  "zero" and "dark" are the notable ones not on this list, and in future operations we should process those, and get Klaus to use a different "test" or similar obstype for genuine test junk exposures.

py/desispec/scripts/exposuretable.py
```diff
@@ -0,0 +1,105 @@
...
+    if 'DESI_SPECTRO_REDUX' not in os.environ:
+        os.environ['DESI_SPECTRO_REDUX'] = os.path.curdir
+
+    ## Define where to find the data
+    path_to_data = os.environ['DESI_SPECTRO_DATA']
+    ## Define where to save the data
+    exptable_base_path = pathjoin(os.environ['DESI_SPECTRO_REDUX'], 'exposure_tables')
+
+    ## Define the nights of interest
+    nights = list(range(20200219, 20200230)) + list(range(20200301, 20200316))
```
sbailey, line 101: commenting linearly before reading all the code... more hardcoded testing stuff, indicating more and more that this is deprecated testing code.

py/desispec/scripts/processingtable.py
```diff
@@ -0,0 +1,148 @@
...
+    unproc_name = proc_name.replace('processing', 'unprocessed')
+    for tab, name in zip([processing_table, unprocessed_table], [proc_name, unproc_name]):
+        if len(tab) > 0:
+            pathname = pathjoin(proc_table_path, name)
+            write_table(tab, pathname, overwrite=overwrite_files)
+            print(f'Wrote file: {name}')
+
+
+
+if __name__ == '__main__':
```
sbailey, line 128: see comments in scripts/exposuretables.py about whether this `if __name__ == '__main__':` block is deprecated testing/development code.

py/desispec/workflow/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+"""
+desispec.workflow
+===========
+
+Tools for workflow management and running the pipeline
+"""
+from __future__ import absolute_import
```
sbailey, line 9: FWIW, this `from __future__ ...` stuff was for the python 2.x -> 3.x transition, and we no longer need it.

py/desispec/workflow/exptable.py
```diff
@@ -0,0 +1,239 @@
...
+    if return_default_values:
+        return colnames, coldtypes, coldeflts
+    else:
+        return colnames, coldtypes
+
+def default_exptypes_for_exptable():
+    ## Define the science types to be included in the exposure table (case insensitive)
+    return ['arc','flat','twilight','science','sci','dither','dark','bias','zero']
+
+def get_survey_definitions():
```
sbailey, line 40: This survey definition and survey number seems a new concept.  What is it used for at the spectro processing level?

py/desispec/workflow/os.py
```diff
@@ -0,0 +1,117 @@
+
```
sbailey, line 1: Let's pick a different name for this, to not confuse it with the standard library `os` module.  I think python3 has worked out non-ambiguous rules about if a parallel module does "import os" whether it means the system `os` or this `os.py` in the same directory, but let's not make ourselves have to remember that.  Maybe `desispec.workflow.util`?

py/desispec/workflow/os.py
```diff
@@ -0,0 +1,117 @@
...
+    return set(newexp)
+
+def get_json_dict(reqpath):
+    req_dict = {}
+    if os.path.isfile(reqpath):
+        with open(reqpath, 'r') as req:
+            req_dict = json.load(req)
+    return req_dict
+
+def give_relevant_details(verbose_output, non_verbose_output=None, verbosely=False):
```
sbailey, line 71: Just guessing from the name, but it smells like this is better handled by log.DEBUG vs. log.INFO and setting the verbosity by the logger loglevel.

py/desispec/workflow/os.py
```diff
@@ -0,0 +1,117 @@
...
+        pass
+
+def define_variable_from_environment(env_name, var_descr):
+    if env_name in os.environ:
+        return os.environ[env_name]
+    else:
+        print(f'{var_descr} needs to be given explicitly or set using environment variable {env_name}')
+        exit(1)
+
+def verify_variable_with_environment(var, var_name, env_name, output_mechanism=print):
```
sbailey, line 86: `verify_variable_with_environment` and `define_variable_from_environment` feel overly specific for how infrequently they are used, which I think makes the client code less readable since it isn't clear to the naive user what these functions are doing, vs. just letting them have the slightly replicated code of e.g. checking os.environ['DESI_SPECTRO_DATA'] and letting it raise an exception if it isn't defined.

akremin:
Thank you for the review, Stephen.

The comments/complaints on timing are reasonable. I agree that the numerology can be tweaked. I had imagined that would adjust the numbers as we ran the code over a few nights where we got a sense of what good timing is. But we can try to do some of that now as well. The dependencies statements are concerning because I always submit with the same "afterok". This occurs in exactly one place and that is the only possibility, so I will have to investigate if there are SLURM nuances I am misinterpreting.

For the context on the timing issue:
The code when running normally will continue to wait for new data. The problem is in testing. The code runs using information from the data itself, which means it can't be easily tricked by pointing to a directory with the current night's name and old data. Therefore we need to explicitly tell it to run on an old night, which you correctly do using `--override-night`. The problem is that this was intended to be used to run on old data, where data will never continue to trickle in. We need to be able to test it though, so this lapse is a problem as you pointed out. I will add another flag that can be set to tell it that we're testing and to therefore continue watching the directory for new data.

A few other clarifications on the timing: 
When the code is running over data it waits 5 minutes before checking for new exposures.

The 1800s is only when monitoring the queue. I didn't want to clobber the queue frequently, and a single job takes O(10-30 minutes).

py/desispec/workflow/proctable.py
```diff
@@ -0,0 +1,241 @@
...
+
+    #     colnames1 = ['INTERNAL_ID', 'EXPID'    , 'NIGHT', 'INT_DEP_IDS', 'JOBNAME', 'TYPE']
+    #     coltypes1 = [ int         ,  list      ,  int    , list        , 'S30'    , 'S10']
+
+    #     colnames2 = ['LATEST_QID', 'LAST_SUBMIT_DATE', 'STATUS', 'LATEST_DEP_QID', 'ALL_QIDS']
+    #     coltypes2 = [ int        , 'S20'             , 'S10'   ,  list           ,  list      ]
+
+    colnames = colnames1 + colnames2 + colnames3
+    coldtypes = coltypes1 + coltypes2 + coltypes3
+    coldefaults = coldefault1 + coldefault2 + coldefault3
```
sbailey, line 40: I think it is a topic for a future update given that this is nominally working, but I'm concerned about the maintainability of these variable length arrays within a row of a column for tracking variable numbers of dependencies.  I think it will be more maintainable in the long run to keep one table for jobids, and a separate table or json structure for tracking the dependencies between the jobs.  Keeping the dependencies as a separate table would also make this more amenable to loading into an SQL database if needed for further scaling.  Future work.

py/desispec/workflow/tableio.py
```diff
@@ -0,0 +1,304 @@
+import os
+import numpy as np
+from astropy.table import Table
```
sbailey, line 3: See earlier comment about maintainability of these table structures.  The sheer amount of custom code here for reading/writing/parsing these tables is a maintainability red flag.  If you find yourself having to make any more updates here to get basic functionality working, let's pause and reconsider whether further updates or restructuring is the right path.

py/desispec/workflow/timing.py
```diff
@@ -0,0 +1,70 @@
...
+
+    yester_timetup = time.strptime(f'{obsnight}{endhour:02d}', '%Y%m%d%H')
+    yester_sec = time.mktime(yester_timetup)
+
+    today_sec = yester_sec + one_day_in_seconds
+    today_timetup = time.localtime(today_sec)
+    return nersc_format_datetime(today_timetup)
+
+
+def during_operating_hours(dry_run=False, start_hour=get_nightly_start_time(), end_hour=get_nightly_end_time()):
```
sbailey, line 60: I know you are still adding documentation, but please include a description of the concepts of "operating hours", "nersc start/end times", and "nightly start/end time".  How are these used?  Are they fundamentally necessary?  Would taking calibrations at non-standard times break these concepts?

sbailey (commented):
I added a bunch of inline comments, but I'll reiterate the basic idea of getting this working for the case of "no data has arrived yet tonight" and then actually putting this branch into use while addressing other things, since it actually working is most important.

There's enough subtle moving parts here that we'll need a organized way of testing this going forward.  e.g. units tests plus a standardized fake night with zero+arc+flat+science+gotcha observations that we could run as part of vetting the pipeline for a new release prior to a rerun.

bin/create_exposure_tables
```diff
@@ -0,0 +1,37 @@
...
+
+
+def get_parser():
+    """
+    Creates an arguments parser for the create_exposure_tables script
+    """
+    parser = argparse.ArgumentParser(usage = "{prog} [options]")
+    parser.add_argument("-n", "--nights", type=str,  help="nights as comma separated string")
+    parser.add_argument("--science-types", type=str, default=None, help="comma separated list of exposure types to include in "+\
+                                                           "exposure table, e.g. science,arc,flat,dark,zero, ...")
```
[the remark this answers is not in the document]
akremin: This was back when only SCIENCE flavors were going to be saved into the exposure tables. But now that zeros and darks are going to be included that has changed. obstypes sounds like a better name. I'll make that change.

bin/create_exposure_tables
```diff
@@ -0,0 +1,37 @@
...
+
+
+
+if __name__ == '__main__':
+    parser = get_parser()
+    args = parser.parse_args()
+
+    args.nights = [ int(val) for val in args.nights.strip("\n\t ").split(",") ]
+    if args.science_types is not None:
+        args.science_types = [ val.strip('\t ') for val in args.science_types.split(",") ]
```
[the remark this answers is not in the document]
akremin: These are just overly preventative (based on years of programming gotchas in other situations), but likely unnecessary here. I never encountered it here, but the code was also always written that way.

bin/create_exposure_tables
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+# coding: utf-8
```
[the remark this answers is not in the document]
akremin: Creating exposure tables and production tables for a custom production run would be useful. In addition to a future script still in development to use the exposure and production tables to run a full-scale production, with comment over-rides from the exposure tables and information from the production tables as well.

Personally, I think they're useful as stand-alone scripts. But we can certainly iterate on this. Not something I'm adamant about.

py/desispec/scripts/daily_processing.py
```diff
@@ -0,0 +1,239 @@
...
+    ## Get default values for input variables
+    if procobstypes is None:
+        procobstypes = default_exptypes_for_proctable()
+    if camword is not None:
+        print(f"Over-riding camword in data with user provided value: {camword}")
+
+    ## Adjust wait times if simulating things
+    speed_modifier = 1
+    if dry_run:
+        speed_modifier = 0.1
```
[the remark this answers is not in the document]
akremin: Understood. The wait times are different in different loops, but we can use a multiplier of some user specified value perhaps, or completely rethink how it's done.

py/desispec/scripts/daily_processing.py
akremin: Okay. I followed the precedent of desi_dailyproc where print statements were used in the manager and the underlying code was the only ones using the official logger. But I will make that change

py/desispec/scripts/daily_processing.py
akremin: This is monitoring and resubmitting queue jobs. Arc jobs take 30+ minutes. Maybe two or three flats/sciences could run in that time, so I don't think it's unrealistic.

py/desispec/scripts/daily_processing.py
akremin: As with many of the hardcoded numbers, it is a bit arbitrary, but with some thought put in. This code here is executed at the end of the night, looping over and resubmitting failed jobs. If a job has failed 4 times, odds are it will fail again on resubmission even if it seems like it should pass (based on exit codes).

py/desispec/scripts/exposuretable.py
akremin: I fully agree. In exptable there is a function that gives the names, types, and defaults. This is my way of documenting temporarily while I get feedback. Once it solidifies, then a data model should probably be made. The same is true with proctable.py.

sbailey:
> For the context on the timing issue:
The code when running normally will continue to wait for new data. The problem is in testing. The code runs using information from the data itself, which means it can't be easily tricked by pointing to a directory with the current night's name and old data. Therefore we need to explicitly tell it to run on an old night, which you correctly do using --override-night. The problem is that this was intended to be used to run on old data, where data will never continue to trickle in. We need to be able to test it though, so this lapse is a problem as you pointed out. I will add another flag that can be set to tell it that we're testing and to therefore continue watching the directory for new data.

I'm not sure I understand that.  For my testing, I created a completely different blank `$DESI_SPECTRO_DATA=/global/cfs/cdirs/desi/users/sjbailey/spectro/data` location, started `desi_daily_proc_manager` monitoring that directory, and then by hand started adding links to real data for the requested night.  I don't think `desi_daily_proc_manager` should be able to distinguish the difference between me creating those links vs. real new data actually arriving from KPNO at whatever cadence I was creating those links.

In my test case, it eventually got though 4 iterations of checking the queue every 30 minutes, and then exited while jobs were still in the queue without ever finding that last non-standard arc that I had linked.  I don't think `desi_daily_proc_manager` should ever exit on its own, vs. always patiently waiting for new data (or exit upon request, e.g. if a sentinel file is created to request it to exit cleanly at the next check).

py/desispec/scripts/exposuretable.py
akremin: These were all for testing, but left as a way of documenting a set of commands that could be used to run the functions. The official scripts use args and become more obscure in terms of "what minimal set of commands can I type in a python interpreter to run the function?" 

But I am happy to delete all these or comment them out

py/desispec/scripts/exposuretable.py
akremin: There is no type 'sci'. That is something I through in preventatively (and rather uselessly). This hasn't been updated since I started including darks and zeros as default types to load in. 

As mentioned in a prior comment, these were meant as examples more than actual means of running the code from the command line, so it is likely best if I just remove these.

py/desispec/workflow/__init__.py
akremin: I assumed as much, but it still existed in another `__init__.py `within desisepc, so I included it out of an abundance of caution.

py/desispec/workflow/exptable.py
akremin: It is something you and I brainstormed over the summer, but it's still half-formed. Currently it's not used for anything but could be useful in re-processing multiyear data to delineate between e.g. minisv2, SV0, SV, Y1, etc.

akremin:
> I'm not sure I understand that. For my testing, I created a completely different blank `$DESI_SPECTRO_DATA=/global/cfs/cdirs/desi/users/sjbailey/spectro/data` location, started `desi_daily_proc_manager` monitoring that directory, and then by hand started adding links to real data for the requested night. I don't think `desi_daily_proc_manager` should be able to distinguish the difference between me creating those links vs. real new data actually arriving from KPNO at whatever cadence I was creating those links.
> 
> In my test case, it eventually got though 4 iterations of checking the queue every 30 minutes, and then exited while jobs were still in the queue without ever finding that last non-standard arc that I had linked. I don't think `desi_daily_proc_manager` should ever exit on its own, vs. always patiently waiting for new data (or exit upon request, e.g. if a sentinel file is created to request it to exit cleanly at the next check).

When running "for real" it waits. It's when you use --override-night that it no longer waits. The code doesn't have a configuration to do what you want. I should have anticipated it, as it is an important thing to test. I tested it manually by commenting out a line of code. The fix will take only a few lines of code (which I just changed and am about to push).

`--override-night` was intended to be used to reduce prior night's data, not to pretend that the script itself is running during a previous night. New data doesn't show up in historic data and the script would continue looking until the end of time without a halting mechanism. Normally the halting mechanism is triggered by time of day, but that doesn't work for historical runs. Therefore I added an explicitly break statement when the script is called with `--override-night`. I'm adding a new flag `--continue-looping-debug` that will allow us to do the testing you propose.

If you still have questions we can talk offline. I think there is a miscommunication about what the script is doing at various stages.

And to clarify further:
That 30-minute cadence loop is meant for end-of-night cleanup of failed jobs AFTER all jobs have been submitted. In that loop it never checks for new data, because all data should already have been submitted.

akremin:
In the new implementation, running with:
desi_daily_proc_manager --override-night <NIGHT> --continue-looping-debug

should produce the desired ability to have the code watch for data trickling in. The downside is that it will never terminate on its own and needs to be killed by the user. I don't see a way of doing both unless you arbitrarily kill it after a certain number of hours. Because it continues to search for new data, it doesn't enter into the final "cleanup" phase where it tries to resubmit jobs. But I think that it okay for the purposes of the proposed test. A second test on a fixed size data set can be done to test these features.

bin/create_exposure_tables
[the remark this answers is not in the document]
sbailey: ok, let's leave as command line scripts.

py/desispec/scripts/daily_processing.py
sbailey: You are good at catching me requesting a change to something that you copied from me :)

Retroactively justifying my inconsistency: when all the outputs come from a single script and they occur in linear order, I like "print" for the extra readability of not having all the prefix stuff.  But once code gets spread over multiple functions in multiple modules, and admittedly when I am less familiar with the code, I find it useful to have the extra module names and line numbers to assist with understanding where a message is coming from, even if it is uglier to read.  And especially if the code starts supporting verbose/non-verbose/quiet stuff, different log levels start looking more attractive.

I think I was also confused by messages like
```
Outfile is: 
[] <class 'numpy.ndarray'>
```
and
```
Starting iteration 1
```
and had a knee-jerk reaction to want module names and line numbers.  It might still be ok to have readable prints at the very top level script, and only underlying modules use logging.

bin/create_exposure_tables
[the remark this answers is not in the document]
akremin: Changed to obstypes as suggested

py/desispec/scripts/daily_processing.py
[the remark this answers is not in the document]
akremin: There are now two command line arguments to specify wait time between checks for data and for checks on the queue.

py/desispec/scripts/daily_processing.py
akremin: Can now be user specified, and I think there was a mixup in the use of this sleep that is hopefully now more clear.

py/desispec/scripts/daily_processing.py
akremin: Still left with the thought-out but ultimately arbitrary number of iterations being 4. I t should be studied and optimized with test runs. For now this is commented out and punted to a future PR anyway, so I will resolve this here.

py/desispec/scripts/exposuretable.py
akremin: All if __name__ == '__main__' statements were removed from scripts, including this one.

py/desispec/scripts/exposuretable.py
akremin: Removed

py/desispec/scripts/processingtable.py
akremin: Removed

py/desispec/workflow/__init__.py
akremin: I'll leave this in and let a future refactor that cleans up all of these __future__ items clean it up

py/desispec/workflow/os.py
akremin: changed to utils

py/desispec/workflow/os.py
akremin: This was created out of the mindset of "repeated code should be put into a function" and this was repeated numerous times in the workflow. Since then the uses has decreased, and maybe it is no longer worth using. But for now I'm going to leave it.

It is different because it gives different output depending on if the verbose flag is set or not. The two different outputs could be sent to each of those respectively, but it doesn't eliminate the need for this function.

py/desispec/workflow/proctable.py
akremin: The concern is certainly warranted, but the scale of these things, especially for a daily production alone, is very small and a database seems like overkill.

As you say, it is something to investigate in the future. Especially if robustness starts to become questioned, then we will certainly want to look at shifting to a database.

py/desispec/workflow/tableio.py
akremin: The code is required for users to be able to include multiple comments into the exposure tables. That code can then be utilized by the other multi-value cells like ALL_QIDS, DEP_IDS, etc.

The code here is also with regards to guessing default values and overcoming masked columns. Those things are not changed by having scalar-valued elements.

Agreed that it is something to be revisited regularly, especially if bugs/issues/backward compatibility issues start to arise.

py/desispec/workflow/timing.py
akremin: Doc strings are now there. The operating hours tells the script when it should be running and when it should stop running. Offline Stephen and I discussed moving to a duration based model rather than a wallclock based model. I don't think it simplifies issues, however, and makes the idea of restarted failed jobs overnight with a cron job more problematic because then you don't know how long to tell it to run for. 

I think this is something that can be discussed more and potentially changed in the future, but doesn't need to be a blocking factor here.

py/desispec/scripts/daily_processing.py
akremin: Cleaned up some debug print statements that aren't needed and changed all the print statements under the /workflow/ subdir to be desi logger calls. I retain the print statements in the scripts themselves for what I think is better readability, but I don't feel strongly about that and could be convinced to change those as well.

py/desispec/workflow/os.py
akremin: I think it cleans up the calling scripts while making the output warning/errors more uniform, and the function names and doc strings make it clear what is being done. Otherwise, the start of all the scripts have 2-4 environment variable checks, all of which require 5-10 lines of code to nicely describe what is happening. This lets us conveniently print what we are doing so the user knows if environment variables are being used or, even more significantly, altered while not getting bogged in boilerplate code.

I am happy to discuss this further, but I don't think this is a blocking factor for this PR, so I will consider this resolved for the purposes of this PR.

py/desispec/scripts/exposuretable.py
akremin: I've added a rudimentary explanation of each table's columns in the exptable.py and proctable.py files. Hopefully, these can serve as pseudo datamodels until the things are ironed out enough to justify creating more permanent documentation.

py/desispec/workflow/exptable.py
akremin: As mentioned in my last comment, this is still evolving and the definitions are rather arbitrary at this point. The values aren't used for anything at the moment. They could eventually be useful for parsing a large table for data from specific time periods, e.g. commissioning, SV0, SV1, Y1, etc. But I expect this to evolve as we get more data and determine the best use of such a flag.
I added a brief description of the motivation in the exptable.py "datamodel" at the beginning of that file. For now I will let that resolve this issue for the time being and we can revisit it when it becomes more relevant.

sbailey (approved):
Travis tests are completely bogged down but I have verified that tests pass elsewhere, including sphinx docstrings.  Please commit any final tweaks and remove "draft" flag then I'll merge and punt any future requests to separate tickets.  Thanks.

PR #1031, opened 2020-11-21
Fix bug in specex.py when trying to create directory with empty path

julienguy:
Fixing a trivial bug which is unfortunately difficult to debug inside a mpi script.
`srun -N 1 -n 20 desi_compute_psf_mpi -o psf.fits .... ` was hanging, and never running the psf for bundle 0 because rank 0 was trying to create a directory with an empty path. 

I fixed this, but we should find a way to have the mpi run aborted instead of hanging when python throws an exception in a rank (here `FileNotFoundError: [Errno 2] No such file or directory: ''`)

### DIFF
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -180,8 +180,9 @@ def main(args, comm=None):
 
     outdir = os.path.dirname(outroot)
     if rank == 0:
-        if not os.path.isdir(outdir):
-            os.makedirs(outdir)
+        if outdir != "" :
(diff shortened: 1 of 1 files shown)


sbailey:
Fine to merge when Travis catches up (or without, if Travis is stuck).  Agreed about making the MPI handling more robust, but that can be a separate PR to this specific fix.

FWIW, another way around this is
`outdir = os.path.abspath(os.path.dirname(outroot))`, which will turn a blank dirname into a real path.

PR #1032, opened 2020-11-22
add parallelism to desi_preproc

sbailey:
This PR adds multiprocessing parallelism to `bin/desi_preproc`, using one process per camera, enabling all 30 cameras to be processed in ~2.5 minutes on a cori batch node.  If called with a single camera or `--ncpu 1` then multiprocessing will not be used and the cameras will be iterated over serially.  If there is a failure on a single camera a log error message will be printed but the others can proceed to finish.

This also adds convenient options `--night` and `--expid` which can be used instead of `--infile`.

### DIFF
--- a/bin/desi_preproc
+++ b/bin/desi_preproc
@@ -4,5 +4,6 @@
 Preprocess a DESI raw data exposure, outputing individual pix files
 '''
 
+import sys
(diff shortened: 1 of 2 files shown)


julienguy (approved):
great

PR #1033, opened 2020-11-22
more robust desi_proc spexec+MPI

sbailey:
This PR improves the robustness of `bin/desi_proc` calling `desispec.scripts.specex.main` by wrapping it in try/except.  Previously, a PSF fit failure due to bad data would throw an uncaught exception, causing those MPI ranks to hang and never reach the post-PSF MPI barrier, causing the entire job to hang.

It also fixes a bookkeeping bug when "psf" appeared in `$DESI_SPECTRO_REDUX/$SPECPROD`, due to use of code like `outpsf = inpsf.replace('psf', 'shifted-input-psf')`.  A new utility function `desispec.io.util.replace_prefix(filepath, oldprefix, newprefix)` will replace the prefix exactly once at the beginning of the filename, even if that same prefix appears elsewhere in the path or filename.

Tested with `$SPECPROD=psfmpi` and `desi_proc -n 20201114 -e 62809 --cameras b1,r1,z1 --batch`, which includes an r1 camera exposure with an FEE problem that causes PSF fitting to fail.

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -35,7 +35,7 @@ from astropy.io import fits
 import glob
 import desiutil.timer
 import desispec.io
-from desispec.io import findfile
+from desispec.io import findfile, replace_prefix
 from desispec.io.util import create_camword
 from desispec.calibfinder import findcalibfile,CalibFinder
 from desispec.fiberflat import apply_fiberflat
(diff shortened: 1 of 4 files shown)


julienguy (approved):
Great. I doubled-checked we now get z7 fit for this exposure where r1 had a bad amplifier init.
I also see the error message in the log.

PR #1034, opened 2020-11-23
Paste

julienguy:
Add a script to copy part of a preproc image into another one.
This is to use a combination of lamps and exposure times together to fit the NIR wavelength solution and PSF.

The use case is 
```
desi_paste_preproc --from-image preproc-z0-00062868.fits --into-image preproc-z0-00062862.fits -o preproc-z0-00062862-00062868.fits --wrange 8840:9900
```
producing this spectrum with higher S/N lines >9000A

Figure_1

### DIFF
--- a/bin/desi_paste_preproc
+++ b/bin/desi_paste_preproc
@@ -66,11 +66,6 @@ for tmp in args.wrange.split(",") :
     yy=np.tile(np.arange(n0),(n1,1)).T
     mask=(yy>=ymin[None,:])&(yy<ymax[None,:])
 
-    import matplotlib.pyplot as plt
-    plt.figure()
-    plt.plot(x,ymin)
-    plt.show()
-
     if args.replace :
         into_image.pix[mask]  = scale*from_image.pix[mask]
         into_image.ivar[mask] = scale_ivar*from_image.ivar[mask]
--- a/bin/desi_paste_preproc
+++ b/bin/desi_paste_preproc
@@ -1,0 +1,87 @@
+#!/usr/bin/env python
+
+import os,sys
+import numpy as np
+import argparse
+
+from desispec.io import read_image,write_image,read_xytraceset
+from desispec.calibfinder import findcalibfile
+
+parser = argparse.ArgumentParser(description="copy part of an image and paste it into another")
+parser.add_argument("--from-image", type=str, required=True, help="input preproc image from which part of the image is copied")
+parser.add_argument("--into-image", type=str, required=True, help="input preproc image into which part of the other image is paste")
+parser.add_argument("-o","--outfile", type=str, required=True, help="output preproc image")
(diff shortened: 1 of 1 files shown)


bin/desi_paste_preproc
```diff
@@ -0,0 +1,87 @@
...
+    x=np.arange(n1)
+    ymin=np.interp(x,xmin,ymin).astype(int) # interpolate across fibers , and convert to int = pixel index
+    ymax=np.interp(x,xmax,ymax).astype(int) # interpolate across fibers , and convert to int = pixel index
+    yy=np.tile(np.arange(n0),(n1,1)).T
+    mask=(yy>=ymin[None,:])&(yy<ymax[None,:])
+
+    import matplotlib.pyplot as plt
+    plt.figure()
+    plt.plot(x,ymin)
+    plt.show()
```
sbailey, line 72: Leftover debugging  plotting to remove, or otherwise wrap in `if args.plot: ...`

bin/desi_paste_preproc
```diff
@@ -0,0 +1,87 @@
...
+    if args.replace :
+        into_image.pix[mask]  = scale*from_image.pix[mask]
+        into_image.ivar[mask] = scale_ivar*from_image.ivar[mask]
+        into_image.mask[mask] = from_image.mask[mask]
+    else :
+        into_image.pix[mask]  += scale*from_image.pix[mask]
+        into_image.ivar[mask] *= (from_image.ivar[mask]>0)
+        into_image.mask[mask] |= from_image.mask[mask]
+        mask2=mask&(into_image.ivar>0)&(from_image.ivar>0)
+        into_image.ivar[mask2] = 1/(1./into_image.ivar[mask2]+1./from_image.ivar[mask2]/scale_ivar)
```
sbailey, line 83: Clarifying intention: if a pixel in the mask has into_image.ivar=0 *or* from_image.ivar=0, then the final output has ivar=0 for that pixel.  i.e. output ivar>0 only occurs when both inputs have ivar>0.  I think that is what this code does, and I think that is correct for adding images, but it took me a minute to sort that out.  Right?

sbailey (commented):
Looks good, with a cross check question for clarification, and let's remove or make optional the debugging plot so that it can be used in the pipeline.

bin/desi_paste_preproc
julienguy: yes that is the intention when we add the values. there is also the option to replace the content.

sbailey (approved):

sbailey:
Looks good, fine to merge.  New code only so I don't think we need to wait for Travis to repeat tests of old code, but I don't see a way to cancel Travis tests without it flagging the PR as failed tests.  Merge whenever we need it; let's add this to the pipeline separately.

PR #1035, opened 2020-11-26
RE: recent testing of this code

jalasker:
When running setup.py test on my laptop's VM, I ran into an edge case for test_util.py in test_utils_default_nproc. When using a single processor machine (or presumably if you request one processor on NERSC, etc) the test fails. 

This pull request fixes that test failure by ensuring that in the case there is one CPU, the automatic calculation involving multiprocessing.cpu_count()//2 is floored at 1 rather than 0.


sbailey (approved):
Looks good, thanks.  OK to merge when Travis finishes.

PR #1041, opened 2020-12-02
Newrdnoise2

zkdtc:
Create a new branch from master and add the codes to estimate RDNOISE in active regions. For cameras listed in desispec_param.yml, the code uses the regions provides. For other cameras, just use the generic values in it. 

Why we select the area shown to calculate rdnoise
It is because there is significant scattered light halo at the center of the CCD, as shown in the first figure below. Camera b4 has the area edge very close to the traces.

### DIFF
--- a/py/desispec/data/params/desispec_param.yml
+++ b/py/desispec/data/params/desispec_param.yml
@@ -33,3 +33,27 @@ qa:
        FIDSNR_WARN_RANGE: [6.0, 8.0]
        FIDMAG: 22.
 
+preproc:
+  rdnoise:
+    PARAMS:
+       B:
+          OBSRDNA: '[10:69,250:750]' # CCD amplifier A regions for estimating readnoise,select region[ with [a:b,c:d] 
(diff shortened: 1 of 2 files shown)


sbailey:
brief updates:

I pushed a fix to how the mask is used in `_sigma_clip` (0 is good, not bad).

Although this branch updates the OBSRND[A-D] keyword with the readnoise calculated from the data region, that readnoise isn't actually being used for generating the IVAR and READNOISE HDUs.  I thought it would be a simple fix, but when trying that I got bogged down in all the options for doing row-by-row vs. not, using overscan columns vs. not, etc.  I'm still working on doing that cleanly, since our preproc code has gotten surprisingly messy with all the various options and I don't want to make it any worse.

I was concerned that measuring the readnoise from the data region prior to having the cosmics mask could lead to outliers or biases from the simple sigma clipped.  @julienguy suggested studying patches of the same size on a dark image and comparing with and without cosmic masking.  Although there is a measurable bias in both the offset level and the readnoise measured when compared over many patches, it is small compared to the statistical error of an individual patch and I didn't see any large outliers in any of my test cases.

PR #1046, opened 2020-12-11
Saturlev202012

julienguy:
Minor change to preprocessing
 - default saturation level = 2**16-1 (but all set in calib SVN repo)
 - write header keywords SATULEVA,B,C,D (in ADU, prior to bias subtraction to have a meaningful numeric saturation value)
 - and SATUELEA,B,C,D (in electrons, after bias subtraction and gain multiplication, for information)

### DIFF
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -557,9 +557,9 @@ def preproc(rawimage, header, primary_header, bias=True, dark=True, pixflat=True
                 saturlev_adu = float(cfinder.value('SATURLEV'+amp))
                 log.info('Using SATURLEV{}={} from calibration data'.format(amp,saturlev_adu))
             else :
-                saturlev_adu = 2**16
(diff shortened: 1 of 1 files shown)


sbailey (approved):
Looks good.  Please post link to documentation or code + exposure IDs that were used to determine these saturation levels.  i.e. what do do in N years when we replace another CCD and need to re-measure its saturation in a consistent manner?  That doesn't need to be a blocking factor for merging this PR to use the new saturation values.

sbailey:
I confirmed that tests pass locally so ok to merge without waiting for travis if this is needed sooner.  Related: in desihub/desiutil#156 @weaverba137 demonstrates desi unit testing using GitHub actions instead of travis as a template for converting our other repos.

julienguy:
The linearity study is still work in progress. I will merge when travis is happy.

PR #1048, opened 2020-12-14
use extinction_total_to_selective_ratio func to get extinction from EBV

julienguy:
Use extinction_total_to_selective_ratio func to get extinction from EBV for the fit of standard stars.

### DIFF
--- a/py/desispec/scripts/stdstars.py
+++ b/py/desispec/scripts/stdstars.py
@@ -20,7 +20,7 @@ from desispec.interpolation import resample_flux
 from desiutil.log import get_logger
 from desispec.parallel import default_nproc
 from desispec.io.filters import load_legacy_survey_filter
-from desiutil.dust import ext_odonnell
+from desiutil.dust import ext_odonnell,extinction_total_to_selective_ratio
 from desispec.fiberbitmasking import get_fiberbitmasked_frame
 
 def parse(options=None):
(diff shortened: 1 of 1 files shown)


julienguy:
(needs first to merge the PR with same name in desiutil)

akremin (commented):
It's nice that this has the added bonus of removing some trailing whitespace. 

A few comments below, but none are blocking factors for a PR as the code works based on a test exposure: 67289 from Dec 12th, 2020 done in SPECPROD=daily at NERSC. outputs: `/global/cfs/cdirs/desi/spectro/redux/daily/exposures/20201212/00067289/`
log: `/global/cfs/cdirs/desi/spectro/redux/daily/run/scripts/night/20201212/science-20201212-00067289-a0123456789-37239928.log`

Comment 1:
You could add comments to describe what the r_band and a_band refer to/mean. 

Comment 2:
Is there ever a chance that these inputs will be lists that don't understand boolean indexing? It currently works, so as long as they are always arrays, the logic will work.

Comment 3:
You could also add a short comment on the dimensions of each. I figured it out but it took a minute to parse together the dimensions of everything.

julienguy:
About your 'Comment 2', the fibermap type is a table, being an attribute of the class Frame. And star_unextincted_mags[band] is a numpy array declared in the function so they always accept boolean indexing.

akremin (approved):
Looks good, thanks. The added comments are informative. I was more asking about the shapes of the input/output arrays than in physical units when I referred to "dimensions," but it's not an important issue. I'm okay with merging if Stephen gives a thumbs up

sbailey:
Thanks for cross checks and additional comments.  Merging now, not waiting for Travis.

PR #1050, opened 2020-12-15
fallback to using FA_TYPE if no stdstars in (SVn_)DESI_TARGET

sbailey:
This is a hack to work around a problem on 20201214 where the SV1_DESI_TARGET bits were zeroed out in the fiberassign files, but the FA_TYPE bits were still ok for determining what was a standard star.  I'm running this via a PR for a post-facto review but want to merge and run tonight before NERSC is out.  If people think this hack is just way too hacky we can undo it.

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -41,6 +41,7 @@ def isStdStar(fibermap, bright=None):
 
     TODO: move out of scripts/stdstars.py
     """
+    log = get_logger()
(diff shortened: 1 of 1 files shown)


sbailey:
Actually, nevermind.  Those files also have a bunch of other columns zeroed out that we need, e.g. FLUX_R.  We do have the original good uncorrupted fiberassign files so it will be possible to recover the data, but not with this hack.  Good night.

PR #1053, opened 2020-12-15
Migrate to GitHub Actions

weaverba137:
This PR moves unit tests from Travis to GitHub actions.

* Required packages are installed via `requirements.txt`.
  - Note however that `desiutil` must be installed before other DESI packages.
* Tests with Python 3.6 are dropped due to compatibility issues with `fitsio`.
* A few module-level calculations had to be wrapped in `try:` clauses to catch exceptions when building documentation.


weaverba137:
PS, I haven't yet removed the `.travis.yml` file.  That will be the last step before merging this PR.

sbailey:
What's the specific problem with fitsio on py 3.6?  I'm using fitsio 1.1.2 on python 3.6.7, and it looks like fitsio still thinks it supported 3.6.  I'd rather not drop desi support of 3.6 yet if we don't have to.

Installing desiutil first: is this because `setup.py` includes `from desiutil.setup import DesiTest, DesiVersion, get_version` and thus running `python setup.py install` requires desiutil even though that specific action doesn't use desiutil?  If that's the case, I suggest wrapping it in a try/except ImportError block so that desiutil is only required for functions that specifically need desiutil, but not for basic installation.  Then desiutil can be installed via requirements.txt just like the others without having to be special cases in the yaml file.

weaverba137:
For fitsio, I have no idea.  Here is the traceback: https://github.com/desihub/desispec/runs/1558867066?check_suite_focus=true#step:4:68

It is correct that desiutil is needed by the setup.py file, but it might be more difficult to remove that than just wrapping it in a `try:` block.  We could probably dispense with `desiutil.DesiTest` (migrate to `pytest`) `desiutil.DesiVersion` (used to *set* the version, which is not done during unit tests), but `desiutil.get_version()` sets the official version of the package during installation, so that's still very much needed.

weaverba137:
PS, I will also note that it is not necessarily `desispec` that needs `desiutil` installed first, but rather `desimodel`, `desitarget` and `redrock`, which need to be built and installed via `pip` (I don't think `specter` needs `desiutil`).

So maybe the answer is, "remove that dependency ordering eventually, but it will require refactoring many, if not most DESI/desihub packages".

sbailey:
@weaverba137 thanks for the explanations.  I also experimented a bit with getting fitsio to install on py3.6 but don't see a solution (despite having an existence proof of fitsio/1.1.3 installed under py3.6, albeit perhaps under a different cfitsio version).  OK to drop py3.6 testing.

I see your point about desiutil requirements of lower level dependencies like desimodel; it looks like we'll need to work our way up from the bottom on that one, or live with desiutil being installed separately.  The main thing I was trying to avoid was the replicated DESIUTIL_VERSION and DESIMODEL_DATA in both the "tests" and the "coverage" section which seems like maintenance hassle.  Too bad `pip ... -r requirements.txt` doesn't install in the specified order.

sbailey (approved):
OK to remove .travis.yml and merge when ready.  Thanks for adding this.  It will be great to have timely unit tests again.

weaverba137:
I'll keep that in mind as part of the general Python ecosystem move away from using setup.py at all.  See desihub/desiutil#146 for discussion of that.

Issue #1051, opened 2020-12-15
Absolute symlinks in reductions are invalid on other filesystems


weaverba137:
The daily reductions (desi/spectro/redux/daily) contain absolute symlinks such as:
```
exposures/20201214/00067768/stdstars-0-00067768.fits -> /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20201214/00067744/stdstars-0-00067744.fits
```
These links are not valid on any other filesystem, such as the NOIRLab Mirror.  These should be changed to relative links.  In this example:
```
exposures/20201214/00067768/stdstars-0-00067768.fits -> ../00067744/stdstars-0-00067744.fits
```
would be valid anywhere.

sbailey:
Fixed in PR #1056.  Closing.

weaverba137:
I have to reopen this issue, because the daily reductions *still* contain absolute symlinks that are invalid on the Tucson mirror.  Here's an example:

```
dashboard/links/202101/arc-20210106-00071034-a0123456789-37715282.log -> /global/cfs/cdirs/desi/spectro/redux/daily/run/scripts/night/20210106/arc-20210106-00071034-a0123456789-37715282.log
```

Please thoroughly review *all* symlink creation code and make sure *all* symlinks are relative.  Please also fix all existing absolute symlinks.

sbailey:
@zkdtc could you update the dashboard code to only use relative links instead of absolute links, and search all desispec code for `os.symlink` calls for other places that may need absolute -> relative fixes?  Thanks.

sbailey:
@zkdtc can you take care of converting the absolute links in the dashboard into relative links (or otherwise confirm that it no longer generates absolute links)?  From an earlier comment, e.g.

```
dashboard/links/202101/arc-20210106-00071034-a0123456789-37715282.log -> /global/cfs/cdirs/desi/spectro/redux/daily/run/scripts/night/20210106/arc-20210106-00071034-a0123456789-37715282.log
```

Thanks

zkdtc:
Sure~

zkdtc:
In desi_proc_dashboard, the ln command uses absolute directory now. Below is an example: 
" ln -s /global/cfs/cdirs/desi/spectro/redux/cascades/run/scripts/night/20210116/poststdstar-20210116-00072716-a0123456789-39626671.log /global/cfs/cdirs/desi/www/users/zhangkai/new/links/202101/poststdstar-20210116-00072716-a0123456789-39626671.log "

[...]

Issue #1049, opened 2020-12-15
make fiberflat jobs 5 nodes instead of 7


sbailey:
Adjust desi_proc fiberflat node allocation logic to be like science exposures that use 5 nodes instead of 7.  This should allow better throughput in the realtime queue by processing two exposures simultaneously.

sbailey:
implemented in PR #1056.  Closing.

PR #1056, opened 2020-12-18
Desiproc fixes to add relative sym links and realtime flat cal node reduction

akremin:
Note I have *not* tested this yet, but can do so once dirac/NERSC is back. The changes are basic and should "just work." I tested the changes with toy code, but not within the desi_proc framework.

Cleans a few things in the code, but primarily this resolves issues #1051 and #1049. We now symbolic link with relative paths so that it works on mirrors outside NERSC. Also run with a max of 5 nodes in the realtime queue for flat cals.

### DIFF
--- a/bin/desi_proc_joint_fit
+++ b/bin/desi_proc_joint_fit
@@ -28,7 +28,8 @@ from desitarget.targetmask import desi_mask
 from desiutil.log import get_logger, DEBUG, INFO
 stop_imports = time.time()
 
-from desispec.workflow.desi_proc_funcs import assign_mpi, get_desi_proc_joint_fit_parser, create_desi_proc_batch_script
+from desispec.workflow.desi_proc_funcs import assign_mpi, get_desi_proc_joint_fit_parser, create_desi_proc_batch_script, \
+                                              find_most_recent
 from desispec.workflow.desi_proc_funcs import load_raw_data_header, update_args_with_headers
 
 parser = get_desi_proc_joint_fit_parser()
@@ -366,13 +367,30 @@ if args.obstype in ['SCIENCE']:
         comm.barrier()
 
     if rank==0 and len(args.expids) > 1:
+        start_directory = os.path.abspath(os.curdir)
         for sp in spectro_nums:
             saved_stdfile = findfile('stdstars', args.night, args.expids[0], spectrograph=sp)
+            existing_dirname, existing_fname = os.path.split(saved_stdfile)
             for expid in args.expids[1:]:
                 new_stdfile = findfile('stdstars', args.night, expid, spectrograph=sp)
-                cmd = f'ln -s {saved_stdfile} {new_stdfile}'
-                log.info(f'Sym Linking jointly fitted stdstar file: {saved_stdfile}  to  {new_stdfile}')
+                new_dirname, new_fname = os.path.split(new_stdfile)
+                log.debug("Path exists: {}, file exists: {}, link exists: {}".format(os.path.exists(new_stdfile),
+                                                                                     os.path.isfile(new_stdfile),
+                                                                                     os.path.islink(new_stdfile)))
+                os.chdir(new_dirname)
+                rel_path_to_orig = os.path.relpath(existing_dirname,new_dirname)
+                rel_path_to_orig_name = os.path.join(rel_path_to_orig,existing_fname)
+                log.info(f'Sym Linking jointly fitted stdstar file: {new_stdfile} '+\
+                         f'to existing file at rel. path {rel_path_to_orig_name}')
+                cmd = f'ln -s {rel_path_to_orig_name} {new_fname}'
+                cmdlist = cmd.split(' ')
+                log.info(str(cmdlist))
                 runcmd(cmd, inputs=[saved_stdfile], outputs=[new_stdfile])
+                log.debug("Path exists: {}, file exists: {}, link exists: {}".format(os.path.exists(new_stdfile),
+                                                                                     os.path.isfile(new_stdfile),
+                                                                                     os.path.islink(new_stdfile)))
+        os.chdir(start_directory)
+
 # -------------------------------------------------------------------------
 # - Wrap up
 
--- a/bin/desi_proc_joint_fit
+++ b/bin/desi_proc_joint_fit
@@ -378,7 +378,7 @@ if args.obstype in ['SCIENCE']:
                 log.debug(f'Sym Linking jointly fitted stdstar file: {new_stdfile} '+\
                           f'to existing file at rel. path {relpath_saved_std}')
                 runcmd(os.symlink, args=(relpath_saved_std, new_stdfile), \
-                       inputs=[relpath_saved_std, ], outputs=[new_stdfile, ])
+                       inputs=[saved_stdfile, ], outputs=[new_stdfile, ])
                 log.debug("Path exists: {}, file exists: {}, link exists: {}".format(os.path.exists(new_stdfile),
                                                                                      os.path.isfile(new_stdfile),
                                                                                      os.path.islink(new_stdfile)))
--- a/bin/desi_proc_joint_fit
+++ b/bin/desi_proc_joint_fit
@@ -367,29 +367,24 @@ if args.obstype in ['SCIENCE']:
         comm.barrier()
 
     if rank==0 and len(args.expids) > 1:
-        start_directory = os.path.abspath(os.curdir)
         for sp in spectro_nums:
             saved_stdfile = findfile('stdstars', args.night, args.expids[0], spectrograph=sp)
-            existing_dirname, existing_fname = os.path.split(saved_stdfile)
             for expid in args.expids[1:]:
                 new_stdfile = findfile('stdstars', args.night, expid, spectrograph=sp)
                 new_dirname, new_fname = os.path.split(new_stdfile)
                 log.debug("Path exists: {}, file exists: {}, link exists: {}".format(os.path.exists(new_stdfile),
                                                                                      os.path.isfile(new_stdfile),
                                                                                      os.path.islink(new_stdfile)))
-                os.chdir(new_dirname)
-                rel_path_to_orig = os.path.relpath(existing_dirname,new_dirname)
-                rel_path_to_orig_name = os.path.join(rel_path_to_orig,existing_fname)
-                log.info(f'Sym Linking jointly fitted stdstar file: {new_stdfile} '+\
-                         f'to existing file at rel. path {rel_path_to_orig_name}')
-                cmd = f'ln -s {rel_path_to_orig_name} {new_fname}'
+                relpath_saved_std = os.path.relpath(saved_stdfile, new_dirname)
+                log.debug(f'Sym Linking jointly fitted stdstar file: {new_stdfile} '+\
+                         f'to existing file at rel. path {relpath_saved_std}')
+                cmd = f'ln -s {relpath_saved_std} {new_stdfile}'
                 cmdlist = cmd.split(' ')
                 log.info(str(cmdlist))
                 runcmd(cmd, inputs=[saved_stdfile], outputs=[new_stdfile])
                 log.debug("Path exists: {}, file exists: {}, link exists: {}".format(os.path.exists(new_stdfile),
                                                                                      os.path.isfile(new_stdfile),
                                                                                      os.path.islink(new_stdfile)))
-        os.chdir(start_directory)
 
 # -------------------------------------------------------------------------
 # - Wrap up
--- a/bin/desi_proc_joint_fit
+++ b/bin/desi_proc_joint_fit
@@ -85,7 +85,6 @@ else:
         for infile in args.inputs:
             hdr = load_raw_data_header(pathname=infile, return_filehandle=False)
             args.expids.append(int(hdr['EXPID']))
-            fx.close()
 
     args.expids = np.sort(args.expids)
     args.inputs = np.sort(args.inputs)
--- a/bin/desi_proc_joint_fit
+++ b/bin/desi_proc_joint_fit
@@ -377,11 +377,9 @@ if args.obstype in ['SCIENCE']:
                                                                                      os.path.islink(new_stdfile)))
                 relpath_saved_std = os.path.relpath(saved_stdfile, new_dirname)
                 log.debug(f'Sym Linking jointly fitted stdstar file: {new_stdfile} '+\
-                         f'to existing file at rel. path {relpath_saved_std}')
-                cmd = f'ln -s {relpath_saved_std} {new_stdfile}'
-                cmdlist = cmd.split(' ')
-                log.info(str(cmdlist))
-                runcmd(cmd, inputs=[saved_stdfile], outputs=[new_stdfile])
+                          f'to existing file at rel. path {relpath_saved_std}')
+                runcmd(os.symlink, args=(relpath_saved_std, new_stdfile), \
+                       inputs=[relpath_saved_std, ], outputs=[new_stdfile, ])
                 log.debug("Path exists: {}, file exists: {}, link exists: {}".format(os.path.exists(new_stdfile),
                                                                                      os.path.isfile(new_stdfile),
                                                                                      os.path.islink(new_stdfile)))
--- a/py/desispec/workflow/desi_proc_funcs.py
+++ b/py/desispec/workflow/desi_proc_funcs.py
@@ -559,8 +559,7 @@ def find_most_recent(night, file_type='psfnight', cam='r', n_nights=30):
     for daysback in range(n_nights) :
         test_night_struct = time.strptime(time.ctime(time.mktime(test_night_struct)-one_day))
         test_night_str = time.strftime('%Y%m%d', test_night_struct)
-        ## Give Expid of 0, because we don't use that anyway
-        nightfile = findfile(file_type, test_night_str, 0, cam)
+        nightfile = findfile(file_type, test_night_str, camera=cam)
         if os.path.isfile(nightfile) :
             return nightfile
 
--- a/py/desispec/workflow/desi_proc_funcs.py
+++ b/py/desispec/workflow/desi_proc_funcs.py
@@ -300,7 +300,7 @@ def determine_resources(ncameras, jobdesc, queue, nexps=1, forced_runtime=None):
     # - exposures to 5 nodes to enable two to run together in the 10-node
     # - realtime queue, since their wallclock is dominated by less
     # - efficient sky and fluxcalib steps
-    if jobdesc in ('ARC', 'TESTARC', 'FLAT', 'TESTFLAT'):
+    if jobdesc in ('ARC', 'TESTARC'):#, 'FLAT', 'TESTFLAT'):
         max_realtime_nodes = 10
     else:
         max_realtime_nodes = 5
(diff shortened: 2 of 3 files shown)


bin/desi_proc_joint_fit
```diff
@@ -366,13 +367,30 @@ if args.obstype in ['SCIENCE']:
...
+                                                                                     os.path.isfile(new_stdfile),
+                                                                                     os.path.islink(new_stdfile)))
+                os.chdir(new_dirname)
+                rel_path_to_orig = os.path.relpath(existing_dirname,new_dirname)
+                rel_path_to_orig_name = os.path.join(rel_path_to_orig,existing_fname)
+                log.info(f'Sym Linking jointly fitted stdstar file: {new_stdfile} '+\
+                         f'to existing file at rel. path {rel_path_to_orig_name}')
+                cmd = f'ln -s {rel_path_to_orig_name} {new_fname}'
+                cmdlist = cmd.split(' ')
+                log.info(str(cmdlist))
```
sbailey, line 387: I think this logic is right, but you might be doing more directory parsing and command spawning than needed.  i.e. would this be equivalent but simpler for future maintenance?
```
relpath_saved_std = os.path.relpath(saved_stdfile, new_stdfile)
os.symlink(relpath_saved_std, new_stdfile)
```
(double check my src/dst order, but I think I go that right)

Related: what should code do if `saved_stdfile` doesn't exist (RuntimeError?), or if `new_stdfile` link already exists (replace it? log and move on?)

py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -525,18 +532,20 @@ def create_desi_proc_batch_script(night, exp, cameras, jobdesc, queue, runtime=N
 
     return scriptfile
 
-def find_most_recent(night, file_type='psfnight', n_nights=30):
+def find_most_recent(night, file_type='psfnight', exp='00000000', cam='r', n_nights=30):
```
sbailey, line 535: Elsewhere we've been standardizing on expid being an integer not a string.  In this case, it appears that this function only works for night-based `file_type` anyway (e.g. psfnight, fiberflatnight, but not psf or fiberflat) since it doesn't actually do anything with `exp` except pass it forward to `findfile`, which crashes if given a str expid for file formats that actually need an expid.

Suggestion: drop `exp` from the arguments and make it clear in the docstring that this searches back by nights, not by expids.  The findfile below could be `nightfile = findfile(file_type, test_night_str, camera=cam)` (i.e. it doesn't need an expid option for filetypes that are night-based).

py/desispec/workflow/procfuncs.py
```diff
@@ -779,6 +779,8 @@ def checkfor_and_submit_joint_job(ptable, arcs, flats, sciences, arcjob, flatjob
     elif lasttype == 'arc' and arcjob is None and len(arcs) > 4:
         ptable, arcjob, internal_id = arc_joint_fit(ptable, arcs, internal_id, dry_run=dry_run, queue=queue)
         internal_id += 1
+    if dry_run:
+        time.sleep(1)
```
sbailey, line 783: Minor: Consider removing this or making it 0.1 sec.

sbailey (approved):
I think everything is correct as-is, but I put in some comments related to long term maintainability for consideration.  Please test at NERSC when it is back and then ok to merge.

bin/desi_proc_joint_fit
akremin: I'm okay with making your change. I personally felt the opposite way about less code vs maintainable in this particular case, which is why I went with the longer solution. Your method works but I felt that the choices of keeping the first filename extension and throwing the second was rather arbitrary and I worried about the persistence of that choice in an external dependency. Maybe that was overly cautious. To illustrate my point:

Example:
```
os.path.relpath("./test1/foo.fits","./test2/bar.fits")
Out: '../../test1/foo.fits'
```

So it gives what we want, which is a path from the second directory to the first file. But I don't find it obvious that this should return the filename for the first entry and ignore the filename of the second entry. Because of that, I didn't think it was readably as clear or free from future changes in the os module.

Again, happy to change it though.


For the related point:
I purposefully give it to desispec's `runcmd` so that it is handled like any other function call within desi_proc. If the outputs exist, I believe that it skips them with a message stating that. If the input doesn't exist, it raises an error stating that. I personally think that is appropriate, but I'm happy to hear arguments to the contrary.

py/desispec/workflow/desi_proc_funcs.py
akremin: Sounds good. I just noticed that the current master version doesn't work because cam is undefined, so I made this as a quick change. I didn't put much thought into the expid choice. I will remove expid and update the docstring.

py/desispec/workflow/procfuncs.py
akremin: This is only run a handful of times (it's in the joint fit code) and I added it because the mock queue id's are generated using the cpu clock time in seconds. It, therefore, needs to wait at least a second. This prevents two mock submissions from having the same mock Slurm id.

I can move it to within just the arc and flat, where the problem arises. That will reduce the number of total waits in a night, but added more code. So I just went with this.

bin/desi_proc_joint_fit
akremin: Actually, your snippet doesn't exactly work. Now that I'm testing, I realize why I didn't choose to move forward with this. If you look at my toy example, it gives one too many ".." because it thinks that the file in the second argument is another directory to be traversed. So to make it work, you'd need to split off the directory for the second argument.

Similarly, I think we still need to change into the proper directory and use just the filename for the second argument in sym link, otherwise I don't think the relative path's will evaluate properly. I'm still testing this statement, but that's my naive thinking.

So to;dr this could be simplified slightly, but only a line or two and I think those save lines would hurt readability. 

The discussion on using runcmd vs os.symlink is still an open one, however. I'll wait for you to make an executive decision on that

sbailey: Thanks for the explanations.  Let's split the difference with
```
rel_path_to_orig_name = os.path.relpath(saved_stdfile, new_stdfile)
cmd = f'ln -s {rel_path_to_orig_name} {new_stdfile}'
runcmd(cmd, inputs=[saved_stdfile], outputs=[new_stdfile])
```
which minimizes path manipulation logic while providing the runcmd checks for what to do if the output already exists or if the input is missing.

py/desispec/workflow/procfuncs.py
sbailey: ok, that makes sense.  I thought you were just trying to watch the progress like a movie while debugging, but now I see that it has a specific purpose.  Please add a comment about why it needs to be a second so that someone clever like me doesn't try to remove it in the future.

bin/desi_proc_joint_fit
weaverba137: Hold the presses here.  Why is it necessary to call out to an external command to create the symlink?  I did not follow the logic of the discussion above.

bin/desi_proc_joint_fit
```diff
@@ -370,9 +371,21 @@ if args.obstype in ['SCIENCE']:
...
+                new_dirname, new_fname = os.path.split(new_stdfile)
+                log.debug("Path exists: {}, file exists: {}, link exists: {}".format(os.path.exists(new_stdfile),
+                                                                                     os.path.isfile(new_stdfile),
+                                                                                     os.path.islink(new_stdfile)))
+                relpath_saved_std = os.path.relpath(saved_stdfile, new_dirname)
+                log.debug(f'Sym Linking jointly fitted stdstar file: {new_stdfile} '+\
+                         f'to existing file at rel. path {relpath_saved_std}')
+                cmd = f'ln -s {relpath_saved_std} {new_stdfile}'
+                cmdlist = cmd.split(' ')
+                log.info(str(cmdlist))
```
weaverba137, line 383: What is the advantage to splitting the command and then converting *back* into a string in order to log it?

py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -463,14 +463,21 @@ def create_desi_proc_batch_script(night, exp, cameras, jobdesc, queue, runtime=N
                     inparams.append(subparam)
         else:
             inparams = list(cmdline)        
-        for parameter in ['-q', '--queue', '--batch-opts']:
+        for parameter in ['--queue', '-q', '--batch-opts']:
```
weaverba137, line 466: It looks like you're doing some complex argument processing here.  Are you quite certain that it is impossible for `argparse` to handle this?

bin/desi_proc_joint_fit
akremin: It's not necessary, runcmd just provides useful I/O handling which I argued makes it more convenient than os.symlink. I appreciate the fact that using os.symlink is more pythonic and would more smoothly work with the file system, but do you have reason to think this is a terrible idea?

If it will crash and burn, I can write my own I/O checks and use the os module.

bin/desi_proc_joint_fit
weaverba137: Also `cmd.split(' ')` is *not* the same as splitting on space characters:
```python
>>> foo = 'a  b'
>>> foo.split()
['a', 'b']
>>> foo.split(' ')
['a', '', 'b']
```
If you accidentally introduce an extra space character, `.split(' ')` does not behave as expected.

If you have to construct a command as a string `shutil.split()` is probably the safest way to go.  I strongly recommend constructing the command explicitly as a string.  You have absolutely nothing to lose here:
```python
cmdlist = ['ln', '-s', relpath_saved_std, new_stdfile]
```
Or you could use `os.symlink`.  I really, really cannot imagine why you wouldn't.

bin/desi_proc_joint_fit
weaverba137: Yes, it is always a terrible idea to call out to the shell to do something you can do within the language itself.  I am not at all convinced that you can't do the IO handling in Python itself.  Could you describe in detail what you mean by "IO handling"?

bin/desi_proc_joint_fit
akremin: It's a good question. str(cmdlist) gives a str form of the list being given to the following function. I wanted to ensure that the split worked appropriately and verify that we're giving the runcmd function the proper thing it is expecting. It is not the same as rejoining the list back into the original.

weaverba137: Sorry:

>  I strongly recommend constructing the command explicitly as a string.

I meant "as a `list()`".

bin/desi_proc_joint_fit
weaverba137: Also, how does this work with your unit tests?  I think it would be easier to construct a unit test around `os.symlink` instead of a system call.

py/desispec/workflow/desi_proc_funcs.py
akremin: Long story short, I am not aware of `argparse`' ability to deal with this. I'll describe below and you can let me know if you are aware of such capabilities. These are nuissances of the desi_proc code writing a slurm script for itself, which means there are inputs given that define the script generation but should be kept for the version run on the compute nodes.

This is a special situation. Here we're trying to remove things from the input command before writing it into the script.   For example, the queue can be specified in the slurm file and is meaningless when running on the HPC node, so we are removing it. The same with batch options, which are things that the slurm script is defined with and the desi_proc running in the slurm script shouldn't try to specify. 

`argparse` does lots of things with parsing but I don't know of a way for it to remove some of its own items after it has already parsed. I also don't know a way to only get the arguments defined by the user on the command line. We can't just list the whole argparse object because we'd be providing 10-20 default value flags and make the srun commands quite long and chaotic.

bin/desi_proc_joint_fit
sbailey: The motivation for using `runcmd` is that it provides some standardized checks and logging for what to do if a requested input is missing, what to do if the output already exists, and what to do if the command is run and returns but the output still doesn't exist.

However, @weaverba137 question reminded me that `runcmd` also accepts a callable with args as input instead of a string command to spawn.  I think that's the right solution here for this simple command.  Something like:
```
runcmd(os.symlink, args=(rel_path_to_orig_name, new_fname), \
    inputs=[rel_path_to_orig_name,], outputs=[new_fname,])
```

At the same time, we're spinning off to bikeshed territory in the name of maintainability (apologies for starting that...), so I'll leave it to @akremin to wrap this up with something that actually produces relative symlinks by whatever method, and we can revisit that implementation only if it becomes a problem in the future.

For the record: in other cases we preferred to make an actual command string that could be run from the command line so there is a super-direct connection between what is printed in the logging and what to run for debugging if it fails and so that any issues like memory leaks or leftover file handles are isolated in a process that runs and ends, but on some systems MPI+fork() may not be so happy about that so we may have to revisit it (in a different PR!)

akremin: I will implement those checks in python and switch to using os.symlink.

For I/O, I am just referring to checking if the input files exist and output file exists, and raising the appropriate warnings/errors if they do/don't. It isn't hard to implement in python, I was just avoiding writing some code that existed in a useful plug-and-play way within the desispec ecosystem. It also had the benefit of being the same way we check all other files on disk when running our various functions, and how we execute the other function calls.

But you're right, using the native tool is certainly smarter.

py/desispec/workflow/desi_proc_funcs.py
weaverba137: OK, here's how I would do it.

1. argparse parses the arguments.
2. You already know what the default values are.
3. Construct a completely new command line from scratch, rather than manipulating an existing command line, omitting default values (which you know what they are) and only including non-default values.

weaverba137:
Fussiness aside, there is one very important point that is not addressed at all by this PR.  How are *existing* absolute symlinks being converted into relative symlinks?  If these absolute symlinks exist in the daily reductions, they could exist in other reductions as well, and those need to be fixed to keep the mirror functional.

akremin:
> Fussiness aside, there is one very important point that is not addressed at all by this PR. How are _existing_ absolute symlinks being converted into relative symlinks? If these absolute symlinks exist in the daily reductions, they could exist in other reductions as well, and those need to be fixed to keep the mirror functional.

There have only been test production runs done using this version of the code that contains absolute links (at least due to my stdstar fitting code). Those I can simply remove once they've served their usefulness in testing (very soon). The other daily's are only impacted for nights with science data in Dec 2020 with NERSC up. Ie only 20201214. I will deal with those manually.

The pressing concern is how to not make the problem much worse when I start running on 5+ nights worth of data Sunday night. Which is why any implementation here is of great use

py/desispec/workflow/desi_proc_funcs.py
akremin: I've considered your approach in the past. It is much more elaborate to implement than checking the inputs for a few commands and removing matching items.

What you're suggesting is what I eventually want to do, because it allows for more flexibility in the command line options we can implement. When we have things running smoothly, I am happy to revisit this.

bin/desi_proc_joint_fit
akremin: These are removed using the new implementation. Thanks for the comments

bin/desi_proc_joint_fit
akremin: I've have used Stephen's compromise of using runcmd for consistent handling of the data while using the os version of symbolic linking

akremin: To be clear, it now uses: 
```
runcmd(os.symlink, args=(relpath_saved_std, new_stdfile), \
       inputs=[relpath_saved_std, ], outputs=[new_stdfile, ])
```
Which means we are using the built-in os.symlink

akremin:
Tested on cori today using:

The daily had absolute links:
```
desi@cori12 [20201214]$ ls -l ./*/std*-0-*.fits
-rw-r----- 1 desi desi 3412800 Dec 14 20:25 ./00067678/stdstars-0-00067678.fits
-rw-r----- 1 desi desi 9241920 Dec 15 01:47 ./00067744/stdstars-0-00067744.fits
lrwxrwxrwx 1 desi desi      95 Dec 15 01:48 ./00067765/stdstars-0-00067765.fits -> /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20201214/00067744/stdstars-0-00067744.fits
lrwxrwxrwx 1 desi desi      95 Dec 15 01:48 ./00067766/stdstars-0-00067766.fits -> /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20201214/00067744/stdstars-0-00067744.fits
lrwxrwxrwx 1 desi desi      95 Dec 15 01:48 ./00067767/stdstars-0-00067767.fits -> /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20201214/00067744/stdstars-0-00067744.fits
lrwxrwxrwx 1 desi desi      95 Dec 15 01:48 ./00067768/stdstars-0-00067768.fits -> /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20201214/00067744/stdstars-0-00067744.fits
```

I stashed the links away for testing. Then ran:

`desi_proc_joint_fit --batch -n 20201214 -e 67744,67765,67766,67767,67768`


The actual fits still existed, so the code to generate them was skipped. Only the sym-linking was run. Which produced the correct outcomes:
```
desi@cori12 [20201214]$ ls -l ./*/std*-0-*.fits
-rw-r----- 1 desi desi 3412800 Dec 14 20:25 ./00067678/stdstars-0-00067678.fits
-rw-r----- 1 desi desi 9241920 Dec 15 01:47 ./00067744/stdstars-0-00067744.fits
lrwxrwxrwx 1 desi desi      36 Dec 19 08:59 ./00067765/stdstars-0-00067765.fits -> ../00067744/stdstars-0-00067744.fits
lrwxrwxrwx 1 desi desi      36 Dec 19 08:59 ./00067766/stdstars-0-00067766.fits -> ../00067744/stdstars-0-00067744.fits
lrwxrwxrwx 1 desi desi      36 Dec 19 08:59 ./00067767/stdstars-0-00067767.fits -> ../00067744/stdstars-0-00067744.fits
lrwxrwxrwx 1 desi desi      36 Dec 19 08:59 ./00067768/stdstars-0-00067768.fits -> ../00067744/stdstars-0-00067744.fits
```

Next I tested the new flat node settings with:

`desi_proc -n 20201214 -e 67628 --batch --nosubmit`

Which gave the desired outcome:
```
#SBATCH -C haswell                                                                                                                                                                                                                                          
#SBATCH -N 5                                                                                                                                                                                                                                                
#SBATCH -n 160  
```

Running in the regular queue still gives the proper nodes:

`desi_proc -n 20201214 -e 67628 --batch --nosubmit -q regular`

```
#SBATCH -C haswell                                                                                                                                                                                                                                          
#SBATCH -N 7                                                                                                                                                                                                                                                
#SBATCH -n 200   
```

The actual `srun` commands agree with the slurm headers as well in both instances.

sbailey:
Thanks, @akremin.  Ben is raising valid maintenance and stylistic concerns, but let's save that for future updates, including deciding what to do about absolute links in past runs that are broken on the mirror.

akremin:
> Thanks, @akremin. Ben is raising valid maintenance and stylistic concerns, but let's save that for future updates, including deciding what to do about absolute links in past runs that are broken on the mirror.

I fully agree with both of you. Let's revisit in a future PR.

Note that in testing here I was able to show a quick way of changing the sym links. My testing this morning removed all absolute sym links from the Dec 14th daily. 

There are *no* absolute links left because of stdstars in the daily SPECPROD. The others are testing and we can remove those once we're done with them.

weaverba137:
Is there a reason the branch `desiproc_tweaks` has not been deleted?

akremin:
No, I must have forgotten to click the button. Doing that now.

PR #1057, opened 2020-12-19
Workflow mods to avoid specified exposures and use new end-of-cal flags

akremin:
This PR includes two new command line options --ignore-expid-list and --ignore-expid-file. The file should be a text file with one exposure per line. The list should be given as a comma separated string. 

I tested this on Dec 15th data.

Tested with zero-padded and not (e.g. 67890 vs. 00067890), both work. Tested that with zero-padded and without, both worked. Lastly, I tested with both specified. It correctly used the superset of expids given in both options.

It skips all exposures specified, including tested arc, science, and flats. It doesn't use them in the joint fitting either. It catches the new flags for both arcs and flats.

### DIFF
--- a/bin/desi_daily_proc_manager
+++ b/bin/desi_daily_proc_manager
@@ -5,9 +5,11 @@ import argparse
 import socket
 import sys
 import numpy as np
+import os
 
 from desispec.scripts.daily_processing import daily_processing_manager
 from desispec.io.util import create_camword
+from desispec.desi_proc_dashboard import get_skipped_expids
 
 from desispec.workflow.timing import during_operating_hours
 from desispec.workflow.utils import check_running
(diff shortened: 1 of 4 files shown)


sbailey:
Looks good, but getting greedy: on the SV1 data we also have cases of needing to exclude individual cameras while letting the rest of the expid proceed.  Do you want to try to add support for this here?  Otherwise ok to merge.

akremin:
It's an important issue, but I think that would be a challenge to implement on the short timescale. I do define the "camword" for each job (whether exposure or joint fit). That could be modified for each flagged exposure, but it would require some thought as to how we want to specify and read in the idea of "for exposure xxx, avoid cams yi,yj,yk". 

And the joint fits in the current implementation would need a consistent choice of cameras (though we could make it the minimal set of overlapping cams).

sbailey:
ok, let's merge this as is and then expand in the future.

PR #1068, opened 2020-12-23
adjust number of nodes requested per job step

sbailey:
This PR makes the number of nodes requested per job in the regular queue the same as those in the realtime queue.  Originally the regular queue tried to go wider and faster, but led to significant inefficiencies, e.g. PSFs with 19 nodes are only slightly faster than PSFs with 10 nodes due to how fast b vs. slow r,z cameras are distributed.  Pending better optimization, this PR makes the regular jobs the same as the realtime jobs which we know works fairly well.

Along for the ride:
  * more logging around the PSF fitting
  * drop unnecessary search for the default fiberflat when doing a PSF fit
  * parallelize over cameras interpolation of PSF over bad fibers, saving ~1 min runtime

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -386,7 +386,7 @@ if args.obstype in ['ARC', 'TESTARC']:
         comm_group = comm.Split(color=group)
 
         if rank == 0:
-            log.info('Fitting PSFs with {num_groups} sub-communicators of size {group_size}')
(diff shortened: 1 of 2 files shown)


julienguy:
ok for merging (trusting Stephen the changes to the MPI process work).

julienguy:
(double checked PSF run fine)

weaverba137:
Is there a reason the branch `numnode-testing` has not been deleted?

PR #1071, opened 2020-12-23
fix desi_proc_joint_fit fiberflatnight indentation bug

sbailey:
In `desi_proc_joint_fit` a set of nested if/else had an indentation bug such that previously successful `fiberflatnight-camera-{camera}-lamp-{n}.fits` files were not being added to the list of files to pass to `desi_autocalib_fiberflat`.

This was discovered, tested, and fixed on a genuine recovery case for 20201222, where the previous partially successful job had timed out, but the recovery job was skipping then ignoring the previously successful outputs.

Could you double check this @julienguy?  In particular, what is the minimum number of inputs to `desi_autocalib_fiberflat` for it to be worth trying?  If the answer is messy bookkeeping we might punt that to the future if/when we actually have a problem of missing lamps or similar.

### DIFF
--- a/bin/desi_proc_joint_fit
+++ b/bin/desi_proc_joint_fit
@@ -237,8 +237,10 @@ if args.obstype in ['FLAT']:
                             if os.path.isfile(ofile):
                                 average_flats[camera].append(ofile)
                             else:
-                                log.info("Will use existing {}".format(ofile))
-                                average_flats[camera].append(ofile)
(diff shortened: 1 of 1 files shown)


sbailey:
Discussed with @julienguy; will merge now in prep for 20.12 release.  If needed, open a separate issue about minimum number of files needed by `desi_autocalib_fiberflat` for a future update.

PR #1079, opened 2021-01-04
EXPTIME and COADD_EXPTIME in fibermap 

segasai:
and COADD_EXPTIME to coadded fibermap
add EXPTIME column to fibermap of single exposures


sbailey:
Thanks.  Now that blanc is finished I'll merge this (didn't want to change master mid-production, just in case we needed to make any other emergency updates to finish blanc).

PR #1083, opened 2021-01-06
standardize getting NIGHT from raw data headers

sbailey:
This PR provides `desispec.util.header2night` which standardizes how to get the YEARMMDD NIGHT from the raw data fits headers, including handling real-world problems that have come up:
  * NIGHT keyword is completely missing
  * NIGHT keyword is present but blank (i.e. None)
  * NIGHT keyword is present but not an integer (e.g. '        ')
in these cases it falls back to using `DATE-OBS`, and then falls back to `MJD-OBS`, including handling the definition of "NIGHT" rolling over at noon MST (KPNO local time).

`dateobs2night` and `mjd2night` are also added as utility functions used by `header2night`.  These might be useful by other code too, but if you are deriving a YEARMMDD from a header, please use `header2night(header)` to handle the various bogus-header cases.

I have tested this with `desi_proc`, `desi_qproc` (via nightwatch), and the unit tests, but I have *not* updated the `desi_daily_proc_manager` bookkeeping code.

@akremin please add the necessary fixes to this branch for the nightly processing bookkeeping to work too.  Thanks.

### DIFF
--- a/py/desispec/io/raw.py
+++ b/py/desispec/io/raw.py
@@ -56,15 +56,14 @@ def read_raw(filename, camera, fibermapfile=None, **kwargs):
             log.error("Did not find header keyword EXPTIME in any HDU of {}".format(filename))
             raise KeyError("Did not find header keyword EXPTIME in any HDU of {}".format(filename))
 
-    #- Fix occasional missing or None NIGHT (e.g. on 20210105/70934)
-    #- update header as needed so the correction propagates downstream
+    #- Check if NIGHT keyword is present and valid; fix if needed
     try:
-        night = int(primary_header['NIGHT'])
+        tmp = int(primary_header['NIGHT'])
     except (KeyError, ValueError, TypeError):
         primary_header['NIGHT'] = header2night(primary_header)
 
     try:
-        night = int(header['NIGHT'])
+        tmp = int(header['NIGHT'])
     except (KeyError, ValueError, TypeError):
         header['NIGHT'] = header2night(header)
 
--- a/py/desispec/io/raw.py
+++ b/py/desispec/io/raw.py
@@ -13,6 +13,7 @@ from desiutil.depend import add_dependencies
 
 import desispec.io
 import desispec.io.util
+from desispec.util import header2night
 import desispec.preproc
 from desiutil.log import get_logger
 from desispec.calibfinder import parse_date_obs, CalibFinder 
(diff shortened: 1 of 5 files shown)


py/desispec/io/raw.py
```diff
@@ -55,6 +56,24 @@ def read_raw(filename, camera, fibermapfile=None, **kwargs):
             log.error("Did not find header keyword EXPTIME in any HDU of {}".format(filename))
             raise KeyError("Did not find header keyword EXPTIME in any HDU of {}".format(filename))
 
+    #- Fix occasional missing or None NIGHT (e.g. on 20210105/70934)
+    #- update header as needed so the correction propagates downstream
+    try:
+        night = int(primary_header['NIGHT'])
+    except (KeyError, ValueError, TypeError):
```
akremin, line 63: If this fails, I don't think the variable "night" gets defined using this try statement. 

Also, is there a motivation for using try rather than just checking that the key exists and value isn't None? I appreciate that the net result is equivalent, so I'm fine with this.

akremin: If you don't use the variable night, I think your function call already has this logic, so it might be redundant as well. You could just call it for everything and it will return without assigning anything of the header keyword is valid.

py/desispec/util.py
```diff
@@ -299,6 +299,44 @@ def ymd2night(year, month, day):
...
+    except (KeyError, ValueError, TypeError): # i.e. missing, not ISO, or None
+        pass
+
+    try:
+        return mjd2night(header['MJD-OBS'])
+    except (KeyError, ValueError, TypeError): # i.e. missing, not float, or None
+        pass
+
+    raise ValueError('Unable to derive YEARMMDD from header NIGHT,DATE-OBS,MJD')
+
```
akremin, line 339: Same comment here as before about try's vs if's. Fine either way

py/desispec/io/raw.py
sbailey: The point here is to test if the keyword "NIGHT" exists and whether it can be converted into an int, and if not replace it with a valid NIGHT.  Although "and if not None" logic would work for the problem last night (20210105), on 20190626 the keyword existed, wasn't None, but was a blank string instead.  Rather than try to individually test for each way it could be wrong in the past and future, I opted for try/except for whether you can succeed to hopefully be robust to new and novel ways in which that keyword could be wrong in the future...

You make a good point about `night` not being set if the try fails.  I didn't actually need that variable but the meaningful name implied that maybe you could rely upon it, so I renamed it to `tmp` instead.

sbailey:
Thanks for updating offline pipeline code.  Merging now.

PR #1084, opened 2021-01-07
use acquisition guide file if full guide file isn't there

sbailey:
This PR provides robustness for `assemble_fibermap` for exposures like 20210105/00070965 and 20210105/00070969 where the full guide-EXPID.fits.fz file isn't present, but the acquisition guide-EXPID-0000.fits.fz file is (not sure why...).  Unfortunately they keep their full set of header keywords in different HDUs, but this catches that too.

@akremin please try for 20200105.


akremin:
Ran without issue for exposures 70965 and 70969 on Jan 5th 2021 that didn't have guide* files.

PR #1085, opened 2021-01-08
Average flux calibration

julienguy:
Minor change to the AverageFluxCalib class to use DESIMODEL for the geometric area and scipy.constants instead of hard coded numbers. This class is used to produce the average flux calibration used in night watch, and was also used for the throughput analysis in DESI-6043. There is the project to use this average calibration to constrain the calibration of each exposure.

The fitting function in the code has a provision to fit for an airmass term and a seeing term. This was tested in sims but has not yet been used with real data. We will use it when we have more data covering a representative range of airmass and seeing.

### DIFF
--- a/py/desispec/averagefluxcalibration.py
+++ b/py/desispec/averagefluxcalibration.py
@@ -1,4 +1,6 @@
 import numpy as np
+import scipy.constants
+from desimodel.io import load_desiparams
 
 class AverageFluxCalib(object):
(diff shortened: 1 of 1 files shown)


julienguy:
Eddie can you remove 'import pdb' from the code before I merge this.
For context, this has been a very useful tool when we disentangled petal 0 in Dec.

sbailey:

Issue #1086, opened 2021-01-11
New targeting bits for standard stars


geordie666:
In https://github.com/desihub/desitarget/pull/669 I added three new standard star classes:
```
    - [GAIA_STD_FAINT,      33, "Standard stars for dark/gray conditions",   {obsconditions: DARK|GRAY}]
    - [GAIA_STD_WD,         34, "White Dwarf stars",                         {obsconditions: DARK|GRAY|BRIGHT}]
    - [GAIA_STD_BRIGHT,     35, "Standard stars for BRIGHT conditions",      {obsconditions: BRIGHT}]
```
Crucially, these are added to the `mws_mask` (and `sv1_mws_mask`), not to `desi_mask`:

https://github.com/desihub/desitarget/blob/22229d03115ec1804fa6e5464d221c1bb97bf22d/py/desitarget/sv1/data/sv1_targetmask.yaml#L140-L142

These are analogues of the already-defined standards, but based _only_ on Gaia to facilitate observations outside of the Legacy Surveys imaging footprint

sbailey:
`desispec.fluxcalibration.isStdStar` is the place to add these bits.  Since this takes the fibermap as input, and the fibermap has (SV1_)MWS_TARGET I don't think this is a fundamental problem here, but it is the first time that core pipeline "what should I do with this target" leaks out of the (SV1_)DESI_TARGET column, so I'm highlighting that, e.g. are there other places (qproc? nightwatch?) that might expect to be able to derive ELG/LRG/QSO/MWS/BGS/STD/SKY only from (SV1_)DESI_TARGET?

@geordie666 if these bits are set but no other MWS bits are set, is the (sv1_)desi_mask MWS_ANY bit set?  We tend to view that bit as meaning "some MWS science target bit is set" rather than "some bit in MWS_TARGET is set".  Maybe these are all MWS science targets too, but we'll need to consider if this has impact on bookkeeping.

geordie666:
I currently _do_ set the ``MWS_ANY`` bit if the bits for the Gaia standards are set: 

https://github.com/desihub/desitarget/blob/22229d03115ec1804fa6e5464d221c1bb97bf22d/py/desitarget/cuts.py#L2395-L2407

I don't have to, though, I can change that back. We've only run a very limited set of tiles for the M31 dedicated program with this set-up so far.

Let me know if you'd prefer ``MWS_ANY`` _not_ to be set purely for Gaia standards.

sbailey:
[...]

PR #1105, opened 2021-01-16
add a new gaia standards

segasai:
Add new Gaia standards for the flux calibration ( see https://github.com/desihub/fiberassign/issues/304 )

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -49,25 +49,31 @@ def isStdStar(fibermap, bright=None):
     mws_target = fibermap[target_colnames[2]]
     desi_mask = target_masks[0]                 # (sv1) desi_mask
     mws_mask =  target_masks[2]
-    if bright is None:
-        yes = (((desi_target & desi_mask.mask('STD_FAINT|STD_BRIGHT')) != 0)|
-         ((mws_target & mws_mask.mask('GAIA_STD_FAINT|GAIA_STD_WD|GAIA_STD_BRIGHT')) != 0))
-    elif bright:
-        yes = (((desi_target & desi_mask.mask('STD_BRIGHT')) != 0)|
-         ((mws_target & mws_mask.mask('GAIA_STD_WD|GAIA_STD_BRIGHT')) != 0))
-    else:
-        yes = (((desi_target & desi_mask.mask('STD_FAINT')) != 0)|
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks for jumping in on this, but simply adding the names of these bits to the list to check won't work because the `GAIA_*` standards are in `(SV1_)MWS_TARGETS` column while the normal standards are in the `(sv1_)desi_target` column.  So this needs to be expanded to check both.  See the `main_cmx_or_sv` call a few lines earlier that is grabbing `(SV1_)DESI_TARGET` and `(sv1_)desi_mask` and do the same for MWS (index 2 instead of 0).  As a test of whether you are finding standards on the M31 tile:

```python
import numpy as np
from desispec.fluxcalibration import isStdStar

fibermap = Table.read('/global/cfs/cdirs/desi/spectro/redux/daily/preproc/20210115/00072498/fibermap-00072498.fits')
nstd = np.count_nonzero(isStdStar(fibermap)) > 0
if nstd > 0:
    print(f'Good, you found {nstd} standard stars')
else:
    print('Oops; no standard stars found')
```

Unrelated, it also looks like our interface to coveralls is broken in the github tests, but don't worry about that.

segasai:
Thank you for the advice Stephen. I updated the branch and the test that you suggested works now. 
I wasn't quite sure whether the GAIA_WD_STD should be in bright/or faint, so I assigned them to both.

sbailey:
Thanks.  For the record I removed GAIA_STD_WD from the list for now because we aren't fitting WD standards yet.  Merging now.  I'll rerun the failed tile 80715 exposures 20210115/72498-72500.

segasai:
Thanks!

sbailey:
Alas, we're not done yet.  I updated ticket #1086 with the next problem for processing these GAIA-only standards (no FLUX_G/R/Z for quality and color cuts and absolute normalization).  In the meantime, no M31 tile 80715 flux calibrations...

segasai:
Yes, thanks. I'll take a look. It may take me a bit longer, as I dont' quite know that area of the code. 
But regarding M31, since my rvspecfit code works fine without flux calibration, we got the measurements from it (although without coadds).

geordie666:
Obviously, the Gaia-only standards don't have fluxes from the Legacy Surveys, by definition. But, we do pass forward:
```
'GAIA_PHOT_G_MEAN_MAG', 'GAIA_PHOT_BP_MEAN_MAG', 'GAIA_PHOT_RP_MEAN_MAG'
```
(at least in the targeting files).

PR #1109, opened 2021-01-18
Allow gaia-only standards for flux calibration

segasai:
This is a zero-th order PR.
It works on M31 data

```
$ cd /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20210115/ ; desi_fit_stdstars --color 'GAIA-BP-RP' --delta-color 0.1     --frames 00072498/frame-?2-*.fits 00072499/frame-?2-*.fits 00072500/frame-?2-*.fits     --skymodels 00072498/sky-?2-*.fits 00072499/sky-?2-*.fits 00072500/sky-?2-*.fits     --fiberflats /global/cfs/cdirs/desi/spectro/redux/daily/calibnight/20210115/fiberflatnight-?2-20210115.fits     --starmodels /global/cfs/cdirs/desi/spectro/templates/basis_templates/v3.2/stdstar_templates_v2.2.fits     --outfile $SCRATCH/stdstars-2-00072498.fits 
INFO:stdstars.py:82:main: mag delta GAIA-BP-RP = 0.100000 (for the pre-selection of stellar models)
INFO:stdstars.py:190:main: found 30 STD stars
INFO:stdstars.py:243:main: coadding 3 exposures in cam b2, w=[1.19306681 1.08994043 0.71699277]
INFO:stdstars.py:243:main: coadding 3 exposures in cam r2, w=[1.14960636 1.07593574 0.7744579 ]
INFO:stdstars.py:243:main: coadding 3 exposures in cam z2, w=[1.1462321  1.07336732 0.78040057]
INFO:stdstars.py:322:main: reading star models in /global/cfs/cdirs/desi/spectro/templates/basis_templates/v3.2/stdstar_templates_v2.2.fits
INFO:stdstars.py:342:main: No Legacy survey standards
INFO:stdstars.py:349:main: computing model mags for ['GAIA-BP', 'GAIA-G', 'GAIA-RP']
INFO:stdstars.py:354:main: done computing model mags
INFO:stdstars.py:403:main: finding best model for observed star #0
INFO:stdstars.py:444:main: star#0 fiber #1406, GAIA-BP-RP = 0.814451, number of pre-selected models = 142/1491
INFO:stdstars.py:459:main: Star Fiber: 1406; TEFF: 5000.000; LOGG: 3.041; FEH: -1.803; Redshift: -0.000299952; Chisq/dof: 0.130
INFO:stdstars.py:476:main: Applying MW dust extinction to star 0 with EBV = 0.0
INFO:stdstars.py:518:main: scaling G mag 7.568 to 16.161 using scale 0.0003656185640443011
```


sbailey:
WIP: I hand-installed a module for speclite/0.13 that can shadow the copy installed in desiconda, available via
```
source /global/cfs/cdirs/desi/software/desi_environment.sh master
module load speclite/0.13
```
We can use that plus this branch for testing.  I'd like to confirm that we can run end-to-end a CMX exposure, an on-footprint SV1 exposure, and an off-footprint M31 exposure before merging.

segasai:
@sbailey 
One thing that I need advice/change on this patch, is how to make decisions when to use gaia selected standards and delta(gaia_colors) vs delta(legacy_colors).
I.e. what I coded is 
basically options that would be either 
--color 'GAIA-BP-RP' --delta-color 0.1
or 
--color 'G-R' --delta-color 0.1
depending on which colours will be used for filtering. 
but obviously that means that the script itself needs to know whether it's a Gaia-only field or not.
One thing I could do is basically hardcode the logic that:  if --color is not specified then we use Gaia color and delta when no Legacy survey stds, and legacy colors and delta if they are specified.

sbailey:
@segasai that sounds good; thanks for working on this.  Repeating to make sure I understand and we agree:

if `--color` is given, use that.  If the user specifies colors that don't exist in the data that's their problem, ok to raise exception.

if `--color` is not given, default to trying to use use Legacy Survey standards first with color G-R, and if there are no Legacy Survey standards, use Gaia standards with color BP-RP .

It sounds like it would be useful to separate out isStdStar (regular on-footprint Legacy Survey based standards) from isGaiaStdStar (Gaia-only standard stars), rather than isStdStar combining both without specifying which is which.

segasai:
Yes, @sbailey that's exactly the code logic I had in mind.

segasai:
I've implemented the logic we discussed, so now code should work without any aditional --color flags on gaia-only or legacy fields (tested). 
I didn't try to implement the usage of Legacy and Gaia standards simultaneously. So now if legacies are available they'll be used and the gaias won't.

segasai:
I also decided against splitting isStdStar() myself as it's called from multiple places + the benefits of splitting are not clear to me and whether there could be cases when both gaia and standard DESI stds are present in one field.

sbailey:
Hi @segasai .  Please test with the following commands to check that SV1 off-footprint, SV1 on-footprint, and CMX tiles all work.  Currently 2/3 fail, including the original example with which you opened this PR.

Since we seem to be playing whack-a-mole with the various flavors of tiles and ways that this could fail, please investigate whether additional unit tests could catch failures like this without requiring us to rerun full stdstar fitting, while also not bogging down unit tests.  Note that the last failure already has a fix + unit tests in master from PR #1114.

### Environment setup

Setup your environment however you like, but for my later cut-and-paste convenience I used:
```
source /global/cfs/cdirs/desi/software/desi_environment.sh master
module load speclite/0.13
module swap desispec/sjb
```
(others will need to replace the last line with getting their own copy of desispec into PATH and PYTHONPATH).

### M31 tile 80715 night 20210115 exposures 72498-72500
```
cd /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20210115/
desi_fit_stdstars --frames 00072498/frame-?2-*.fits 00072499/frame-?2-*.fits 00072500/frame-?2-*.fits \
  --skymodels 00072498/sky-?2-*.fits 00072499/sky-?2-*.fits 00072500/sky-?2-*.fits \
  --fiberflats ../../calibnight/20210115/fiberflatnight-?2-20210115.fits \
  --outfile $SCRATCH/stdstars-2-00072498.fits 
```
fails with
```
  File "/global/common/software/desi/users/sjbailey/desispec/bin/desi_fit_stdstars", line 20, in <module>
    stdstars.main(args)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/scripts/stdstars.py", line 218, in main
    if n_legacy_std == 0 and legacy_color:
UnboundLocalError: local variable 'legacy_color' referenced before assignment
```

### SV1 tile 80605 night 20201215 exposure 67972
```
cd /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20201215
desi_fit_stdstars --frames 00067972/frame-?2-*.fits \
  --skymodels 00067972/sky-?2-*.fits \
  --fiberflats ../../calibnight/20201215/fiberflatnight-?2-*.fits \
  --outfile $SCRATCH/stdstars-2-00067972.fits 
```
Works, but please re-test after fixing the others to make sure it still works

### CMX tile 70003 night 20200219 expid 51053
```
cd /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20200219
desi_fit_stdstars --frames 00051053/frame-?3-*.fits \
  --skymodels 00051053/sky-?3-*.fits \
  --fiberflats ../../calibnight/20200219/fiberflatnight-?3-*.fits \
  --outfile $SCRATCH/stdstars-3-00051053.fits 
```
Fails with
```
  File "/global/common/software/desi/users/sjbailey/desispec/bin/desi_fit_stdstars", line 20, in <module>
    stdstars.main(args)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/scripts/stdstars.py", line 143, in main
    frame_starindices = np.where(isStdStar(frame_fibermap))[0]
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/fluxcalibration.py", line 47, in isStdStar
    mws_target = fibermap[target_colnames[2]]   # (SV1_)MWS_TARGET
IndexError: list index out of range
```
This was already fixed in PR #1114 include unit tests that would have caught it, but it looks like you branched before that fix was in master.

segasai:
Sorry about that @sbailey , I had a couple of tests, but it didn't have a right permutations of flags/tiles to see the issue.
I've fixed the issues so all three example tests pass.
I'll now look into putting some tests into test_binscripts

segasai:
I've added the tests as well, so as far as I can see it's ready. 
The tests do not go through all the permutations of options, but they do test the cases of with legacy and without legacy standards.

sbailey:
I pushed a bug fix to correct the output stdstar files having HDU METADATA column "MODEL_None" instead of "MODEL_G-R" or "MODEL_GAIA-BP-RP".  Now the on-footprint exposures work end-to-end, but there are still some hiccups with the M31 tile:

  * Testing on 20210115/00072498 (the M31 example above), only petals/spectrographs 2,4, and 5 found Gaia standards that passed cuts.  Please check whether the other petals had standards that should have passed.
  * code now writes stdstar files for M31, but now desi_compute_fluxcalibration is failing since it is looking for MODEL_G-R.  Please update that to support Gaia colors too.

```
cd /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/gaiastds/exposures/20210115/00072498/
desi_compute_fluxcalibration --delta-color-cut 0.1 \
    --infile frame-b2-00072498.fits --sky sky-b2-00072498.fits \
    --models stdstars-2-00072498.fits \
    --fiberflat /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm5/fiberflatnight-b2-20201214.fits \
    --outfile $SCRATCH/fluxcalib-b2-00072498.fits 
```
fails with 
```
...
INFO:fluxcalibration.py:112:main: apply cut |delta color|<0.1
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/astropy/io/fits/column.py", line 2032, in _get_index
    indx = names.index(key.rstrip())
ValueError: 'MODEL_G-R' is not in list

During handling of the above exception, another exception occurred:

  File "/global/common/software/desi/users/sjbailey/desispec/bin/desi_compute_fluxcalibration", line 16, in <module>
    fluxcal.main(args)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/scripts/fluxcalibration.py", line 113, in main
    good = (np.abs(model_metadata["MODEL_G-R"]-model_metadata["DATA_G-R"])<args.delta_color_cut)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/astropy/io/fits/fitsrec.py", line 507, in __getitem__
    return self.field(key)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/astropy/io/fits/fitsrec.py", line 694, in field
    column = self.columns[key]
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/astropy/io/fits/column.py", line 1639, in __getitem__
    key = _get_index(self.names, key)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/astropy/io/fits/column.py", line 2041, in _get_index
    raise KeyError(f"Key '{key}' does not exist.")
KeyError: "Key 'MODEL_G-R' does not exist."

**Comment by segasai** [comment]
Thanks, I was expecting/worried that something would fail down the line. I'll check...

**Comment by segasai** [comment]
The current version now works, 
but another issue is that for some reason all the predicted Bp_Rp colors are off by ~ 0.3 mag. And I can confirm this is present irrespective of the patch.
For all stars in Blanc the synthetic Gaia Rp (computed from stdstars*) is systematically off from the Gaia ones.  
See below
image
I don't quite know what's the cause of that is...
So I've had to increase the color-cut to 0.3 to have more than 1 standard.

Also, since the gaia only standards do not have E(B-V) in  the fibermap I've had to calculate it from SFD, which in turn leads to test failure, because SFD maps are currently not installed in CI. I wasn't quite sure whether you are okay in fetching SFD there or not, or it is better to just remove that test...

**Comment by geordie666** [comment]
@segasai: As I'm now calculating the dust corrections for Gaia myself, using `E(B-V)` from the SFD maps with the Schlafly scaling, I'll soon be passing that forward.

So, my suggestion is to not do that twice, but get the calculation correct in `desitarget` and pass forward the appropriate column.

It's likely that we won't close the loop on that before I need to run fresh sets of targets tonight/tomorrow, but we should probably try to coordinate in future (i.e. you can correct my dust calculations!) for consistency.

**Comment by segasai** [comment]
@geordie666  But presumably that won't apply for already observed M31 tiles, right ? so we'll probably need to have some code anyway in stdstars.
And I agree to coordination.

**Comment by geordie666** [comment]
Ah, true, yes. That won't work for existing tiles unless we hack the fibermaps.

**Comment by sbailey** [comment]
See [desiutil.dust.extinction_total_to_selective_ratio](https://github.com/desihub/desiutil/blob/master/py/desiutil/dust.py); I think that's where we would add GAIA coefficients too to handle E(B-V) extinction.  Or maybe I'm missing the point, but I don't think we need to recalculate from SFD maps every time.

SynthRP - GaiaRP: is this some side effect of the Gaia filters extending beyond the wavelength range of DESI and how speclite handles that extrapolation?  Or is the Synth offset mystery when applied to models that do extend to the full wavelength region covered by the Gaia filters?

**Comment by segasai** [comment]
The point is that EBV is set 0 for gaia only objects in fibermap. 
The gaia exctinction coefficients were already added in stdstars (currently).
Regarding the 0.3 offset, I'm looking currently at it...  The templates for sure extend way past 1 micron. What I'm looking also is the alternative set of filter curves for GDR2 https://www.aanda.org/articles/aa/full_html/2018/11/aa34051-18/aa34051-18.html I wasn't expecting these curves to make that much difference, but maybe I was wrong.

**Comment by segasai** [comment]
Different throughput curves didn't make a difference, and from testing done here https://github.com/desihub/desispec/issues/1110 with different templates (PHOENIX) (where RP magnitudes show the same offset) I can conclude that templates are not at fault either. So it's something else. Weird...

**Comment by segasai** [comment]
Sorry for so many iterations/traffic. 
The 0.3 offset was caused by Vega vs AB...  After that fix,
desi_compute_fluxcalibration works fine as well and doesn't require adjustment of the delta-color-cut.

**Comment by segasai** [comment]
@sbailey This PR is ready for review. For sure there are possible refactors that could be done here, but to my knowledge everything works like it should, including the processing of those test fields.

**Comment by sbailey** [comment]
@segasai checking: with this branch the M31 tile 80715 night 20210115 exposure 72498 still only finds standards on petals 2, 3, 4, and 5 (and alas, petal 3 was unpositioned).

Please confirm whether it should be finding standards on the other petals.  e.g.
```
import fitsio
from desispec.fluxcalibration import isStdStar
fm = fitsio.read('/global/cfs/cdirs/desi/spectro/redux/daily/preproc/20210115/00072498/fibermap-00072498.fits')
ii = isStdStar(fm)                                                                                      
set(np.where(ii)[0] // 500)
```
results in {2, 3, 4, 5}.  Are there targets on petals 0, 1, 6, 7, 8, 9 that `isStdStar` should be finding, or is this as well as we can do for this tile?  Otherwise looks good for merging, but I wanted to double check that the missing standards is a fundamental feature of this tile, not a missing bit or something in isStdStar.  Thanks.

segasai:
Thanks @sbailey The answer about that tile is I don't know, but let me ask who constructed that tile.

geordie666:
Yes, I believe there were _only_ standards on this tile for the petals 2, 3, 4 and 5. See:

https://desi.lbl.gov/svn/data/tiles/trunk/080/fiberassign-080715.png

In general, for answers to these sorts of questions, see @araichoor's QA pages at:

https://data.desi.lbl.gov/desi/survey/observations/SV1/sv1-html/index.html

segasai:
Yes, I've found that page (see my -milkyway email). I just wasn't sure whether the accounting of what is STD for that page was up to date (as new types were introduced). Hopefully Anand or Arjun will confirm.

sbailey:
OK, thanks, that's sufficient for me — fiberassign didn't think there were standards on those other petals, and so desispec shouldn't expect to find any either.  Thanks for the check.  It is a topic for future tiles to make sure that they have sufficient standards on every petal.

Merging now.  Thanks for adding this Gaia standard functionality.

segasai:
Thanks for merging/help.

Issue #817, opened 2019-10-09
BUG in coaddition.coadd(spectra, cosmics_nsig=0)


julienguy:
There is a bug in coaddition.coadd(spectra, cosmics_nsig) for cosmics_nsig > 0.

sbailey:
@julienguy do you remember what the bug was or at least its symptoms?

Our recent reductions (minisv2, daily) have *not* used the --nsig option.

julienguy:
solved in PR #1113 .

PR #1113, opened 2021-01-19
Coaddition

julienguy:
- Fix bug in coaddition with nsig>0 for cosmic ray rejection (a reference instead of a copy of the flux and ivar were used in the outlier identification routine)/
 - Set nsig=4 to new default
 - Add scripts `desi_coadd_frames` and `desi_coadd_preproc` that are useful for some studies.

I would like to add to the PR the possibility to coadd several spectra files (to be discussed). For now the code `desi_coadd_spectra` accepts either a single spectra file (with internally several exposures) or several frame files.

### DIFF
--- a/py/desispec/scripts/coadd_frames.py
+++ b/py/desispec/scripts/coadd_frames.py
@@ -1,0 +1,76 @@
+"""
+Coadd spectra
+"""
+
+from __future__ import absolute_import, division, print_function
+
+import os
+import numpy as np
+from astropy.table import Table
+
+from desiutil.log import get_logger
+from desispec.io import read_frame,write_frame
+from desispec.coaddition import coadd,coadd_cameras,resample_spectra_lin_or_log
+
+
+def parse(options=None):
+    import argparse
+
+    parser = argparse.ArgumentParser("Coadd frames")
+    parser.add_argument("-i","--infile", type=str, nargs='+', help="input frame files")
+    parser.add_argument("-o","--outfile", type=str,  help="output frame file")
+
+    if options is None:
+        args = parser.parse_args()
+    else:
+        args = parser.parse_args(options)
+
+    return args
+
+def main(args=None):
+
+    log = get_logger()
+
+    if args is None:
+        args = parse()
+
+
+    log.info("reading frames ...")
+
+    frames = []
+    for filename in args.infile :
+        frame = read_frame(filename)
+        if len(frames)==0 :
+            frames.append(frame)
+            continue
+
+        if "CAMERA" in frames[0].meta and (frame.meta["CAMERA"] != frames[0].meta["CAMERA"]) :
+            log.error("ignore {} because not same camera".format(filename))
+            continue
+        if frame.wave.size != frames[0].wave.size or np.max(np.abs(frame.wave-frames[0].wave))>0.01 :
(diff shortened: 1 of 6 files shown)


julienguy:
This solves issue #817

julienguy:
Added possibility to coadd several spectra files. For instance, 
`desi_coadd_spectra -i $DESI_SPECTRO_REDUX/blanc/tiles/80608/202012*/coadd-0-80608-*.fits -o coadd-0-80608.fits`

py/desispec/scripts/coadd_frames.py
```diff
@@ -0,0 +1,76 @@
...
+    mask = frames[0].mask
+    for frame in frames[1:] :
+        ivar += frame.ivar * (frame.mask == 0)
+        ivarflux += frame.ivar * (frame.mask == 0) * frame.flux
+        mask &= frame.mask # and mask
+
+    coadd = frames[0]
+    coadd.ivar = ivar
+    coadd.flux = ivarflux/(ivar+(ivar==0))
+    coadd.mask = mask
```
sbailey, line 71: What about other `Frame` quantities: fibermap, chi2pix, resolution_data, scores?

I'm concerned about the maintainability of coading spectra vs. coadding frames, which are doing the same basic operations in different files, and possibly making different choices about things like how/if to coadd scores, what to do with the resolution matrix, what to do with masked ivars, etc.

py/desispec/scripts/coadd_spectra.py
```diff
@@ -54,12 +55,46 @@ def main(args=None):
...
+        log.error("cannot combine input spectra and frames")
+        sys.exit(1)
+
+
+    if input_is_spectra :
         spectra = read_spectra(args.infile[0])
-    else:
+        for filename in args.infile[1:] :
+            log.info("append {}".format(filename))
+            spectra.update(read_spectra(filename))
```
sbailey, line 96: We should consider merging #1107 and then using `desispec.spectra.stack` instead which is way faster than `Spectra.update`, albeit with the restriction that the input spectra have to be homogeneous (e.g. they all have the same cameras).

sbailey (commented):
Thanks for fixing nsig, and adding more coadding flexibility.  Two issues for consideration:
  * it looks like you are coadding some frame quantities (flux, ivar) but not others (resolution matrix, scores).
  * we're building up multiple coadd functions for the same basic operations on frames vs. spectra.  Can this be consolidated for maintainability and consistent choices?

py/desispec/scripts/coadd_frames.py
julienguy: The use case for this script is debugging the spectral processing. 

For this we probably won't use chi2pix and scores. So I will remove chi2pix and scores, and update resolution_data. I am not sure how to "average" fibermaps. I keep the first one for now and check the target ids are the same.

py/desispec/scripts/coadd_spectra.py
julienguy: ok

sbailey: to avoid rebasing confusion between two different open PRs, we can address this optimization later.

sbailey (approved):

PR #1115, opened 2021-01-19
Coadd fix compatibility

alxogm:
coadd_cameras.py is not working for mocks previous to this recent PR https://github.com/desihub/desispec/pull/1079, since the 'EXPTIME' column is not present in the fibermap.  Is it ok to fix it here, for back compatibility? or should I try to solve it in the code where we use it. It is one line.


julienguy:
I think it is now fixed in master, can you please verify?

julienguy:
I checked it is fixed in master, so I am closing this PR.

alxogm:
Thanks @julienguy, yes it is working.

alxogm:
No, I've deleted it. thanks

Issue #1100, opened 2021-01-12
spectra with flux ~0 and extremely small uncertainties (blanc)


segasai:
Hi, 

While looking at blanc data, I came upon this file 
/global/project/projectdirs/desi/spectro/redux/blanc/tiles/65008/20200314/coadd-8-65008-20200314.fits

and it's zero-th fiber for example. 

For some reasone the spectrum has a plenty of parts where both the spectrum and the uncertainty are extremely small. 
See for example the error-spectrum in the r-arm for 0th fiber (I only plot notmasked pixels):
![qq](https://user-images.githubusercontent.com/109513/104384372-41269600-5529-11eb-92a6-603dee54fca9.png)

This problem seems to persist across many spectra as far as I can see, which suggests to me, flux-calibration as a cause ? Or am I doing something wrong ? 

      S

segasai:
![qq_117148](https://user-images.githubusercontent.com/109513/104514507-45b28380-55e9-11eb-807f-f97f55e98d7d.png)
![qq_226405](https://user-images.githubusercontent.com/109513/104515682-084ef580-55eb-11eb-80d2-f08a2e1e1e4f.png)
another couple of examples from the coadd (the exact image info is in the title)

julienguy:
I am working on it. There are 2 exposures with very low transmission.

julienguy:
I think it is solved in a PR to be submitted.
I reran the flux calibration for SP8 for this tile for 2020/03/14 and redid the coadd. It's there: 
`/project/projectdirs/desi/spectro/redux/jguy/tiles/65008/20200314/coadd-8-65008-20200314.fits`
@segasai  can you please check it?

julienguy:
I am also going to rerun the 2 other examples.

segasai:
Yes, it seems fine. Before to be able to still go through this file, I've had to clip the indivudal pixel uncertainties not be too small comparing to the median, and I was reporting if I had to clip too many pixels. Now none of that seems happening.

segasai:
Thank you!

julienguy:
After code fix, tile-80618-20201220-fiber-3396
![coadd-6-80618-20201220-fiber-3396](https://user-images.githubusercontent.com/5192160/105115246-4df13f80-5a7d-11eb-920a-279a84f726e4.png)

julienguy:
And tile-70003-20200224-fiber-3215
[...]

PR #1116, opened 2021-01-19
Flux calibration for exposures with very low transmission

julienguy:
If the median S/N in the standard stars is <2 , use a scaled version of the average DESI throughput, fitting a single scalar instead of a calibration coefficient per wavelength bin.

Addresses issue #1100

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -13,6 +13,8 @@ from desiutil.log import get_logger
 from .io.filters import load_legacy_survey_filter
 from desispec import util
 from desispec.frame import Frame
+from desispec.io.fluxcalibration import read_average_flux_calibration
+from desispec.calibfinder import findcalibfile
 from desitarget.targets import main_cmx_or_sv
 import scipy, scipy.sparse, scipy.ndimage
 import sys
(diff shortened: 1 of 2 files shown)


julienguy:
Test failing because DESI_SPECTRO_CALIB not set in the tests.

sbailey:
@julienguy for consideration: see test_fibermap.py which has hooks for running some tests only at NERSC where the full environment is available, e.g. in this case a real $DESI_SPECTRO_CALIB.  GitHub would skip the tests, but the nightly update would catch them.

sbailey:
@julienguy the test that you disabled for github is also failing at NERSC, albeit for a different reason.  I suspect that this code works for real data headers but we need to update our simulated test data headers for the tests.  Did you have the tests working on Dirac or your laptop, i.e. is this already fixed somewhere but not committed?
```
$> python py/desispec/test/test_binscripts.py 
...
WARNING:fluxcalibration.py:955:compute_flux_calibration: Very low S/N, use simple scaled version of the average throughput
  File "/global/common/software/desi/users/sjbailey/desispec/bin/desi_compute_fluxcalibration", line 16, in <module>
    fluxcal.main(args)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/scripts/fluxcalibration.py", line 172, in main
    fluxcalib = compute_flux_calibration(frame, model_wave, model_flux, model_fibers%500, highest_throughput_nstars = args.highest_throughput)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/fluxcalibration.py", line 956, in compute_flux_calibration
    fluxcalib_filename = findcalibfile([frame.meta],"FLUXCALIB")
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/calibfinder.py", line 105, in findcalibfile
    cfinder = CalibFinder(headers,yaml_file)
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/calibfinder.py", line 175, in __init__
    detector=header["DETECTOR"].strip()
KeyError: 'DETECTOR'
```

sbailey:
For the record I fixed this test failure in commit e7d139e direct to master by adding DETECTOR='SIM' to the test frame header.

PR #1112, opened 2021-01-19
Add token to coverage test

weaverba137:
This PR fixes #1106.

PR #1119, opened 2021-01-22
Dark robustness for desi_spectro_calib 0.2.x

sbailey:
This PR includes the code used for building the new bias+dark models for desi_spectro_calib 0.2.1, 0.2.2, and 0.2.3 used in blanc.  The substantive change is an option `--mindarks` to filter by number of darks for a given EXPTIME so that it won't try and fail on a single test dark.


sbailey:
Coverage tests are failing because this was branched before PR #1112, but other tests pass.  Since this is a PR for the record of what was actually done for blanc calibs I'm going to merge now, and if there are any suggestions/comments/complaints they can apply to future PRs for what we should do in the future.

PR #1121, opened 2021-01-30
Calib config

julienguy:
Add option --rm-entries to desi_calib_config which the editor of calibration configurations for the spectrographs.
Example (used inside a loop over spectrographs and cameras) to update to the new FEE configuration:
```
desi_calib_config add --spectro SP0 --camera-arm b --name V20210128 --date-obs-begin 20210128 --entries CCDCFG=default_sta_20210128.cfg,CCDTMING=flatdark_sta_timing.txt --comments "FEE configuration with 400 prescan" --rm-entries DARK,BIAS -o test.yaml
```


sbailey:
Thanks, it's helpful to have a standardized way to temporarily remove entries when making a new config.  Merging now.

PR #1122, opened 2021-01-30
Tile depth

michaelJwilson:
I think everything is in order so here's a PR, in case you want it for next week.  Happy to hear further suggestions.


py/desispec/scripts/master_nea.py
```diff
@@ -0,0 +1,125 @@
...
+import sys
+import copy
+import astropy.io.fits as fits
+
+# https://github.com/desihub/desispec/issues/1006
+sampling           = 500
+
+wmin, wmax, wdelta = 3600., 9824., 0.8
+fullwave           = np.round(np.arange(wmin, wmax + wdelta, wdelta), 1)
+cslice             = {"b": slice(0, 2751, sampling), "r": slice(2700, 5026, sampling), "z": slice(4900, 7781, sampling)}
```
sbailey, line 20: Are the NEA files sensitive to the exact details of the wavelength grid?  This is likely to change again in the future, e.g. to extend the wavelength range further blue.  If it isn't sensitive to the sampling but perhaps is sensitive to the edges (e.g. due to being able to interpolate but not extrapolate) we should make sure this is a superset of what we might need in the future, e.g. 3520 - 9950.

py/desispec/scripts/master_nea.py
```diff
@@ -0,0 +1,125 @@
...
+    parser.add_argument('--outdir', type = str, default = None, required=True,
+			help = 'dir. of output maser nea file for a given camera')
+    args = None
+    if options is None:
+        args = parser.parse_args()
+    else:
+        args = parser.parse_args(options)
+    return args
+
+def process_one(w, psf, ifiber):
```
sbailey, line 35: please add docstring

py/desispec/scripts/master_nea.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # http://articles.adsabs.harvard.edu/pdf/1983PASP...95..163K                                                                                                                                                                
+        nea       = 1. / np.sum(psf_1d**2.)  # [pixel units].                                                                                                                                                                           
+        angperpix = psf.angstroms_per_pixel(ifiber, w)
+
+        return  [nea, angperpix]
+
+    except:
+        print('Failed on fiber {} and wavelength {} [{} to {} limit] with norm {}.'.format(ifiber, w, psf._wmin_spec[ifiber], psf._wmax_spec[ifiber], norm))        
+        return [-99., -99.]
```
sbailey, line 56: Is this try/except needed, or leftover from testing.  If still needed, let's use the logger instead of a bare print.

py/desispec/scripts/procexp.py
```diff
@@ -1,21 +1,15 @@
-
 """
-This script processes an exposure by applying fiberflat, sky subtraction,
-spectro-photometric calibration depending on input.
+This script computes the template SNR
```
sbailey, line 2: Something has gone wrong here.  This script originally was about processing the exposure to apply all of the calibration steps and write the final cframe files.  The comment here implies that it now only computes the template SNR, and many of the other imports have been removed, e.g. `from desispec.fluxcalibration import apply_flux_calibration`, but downstream in this code `apply_flux_calibration` is still called.  i.e. this code will not run in the pipeline as-is.

py/desispec/scripts/procexp.py
```diff
@@ -30,7 +24,7 @@ def parse(options=None):
     parser.add_argument('--sky', type = str, default = None,
                         help = 'path of DESI sky fits file')
     parser.add_argument('--calib', type = str, default = None,
-                        help = 'path of DESI calibration fits file')
+                        help = 'path of DESI calibration fits file')xs
```
sbailey, line 27: typo that prevents this script from even being imported; i.e. this branch needs more testing.

py/desispec/scripts/tsnr_afterburner.py
```diff
@@ -0,0 +1,197 @@
...
+            night  = parts[9]
+            expid  = np.int(parts[10])
+
+            name   = parts[-1]
+
+            parts  = name.split('-')
+            camera = parts[1]
+            
+            calib  = findfile('fluxcalib', night=night, expid=expid, camera=camera, specprod_dir=None)
+            calib  = calib.replace('SPCALIB', calibdir)
```
sbailey, line 121: What is the case where this SPCALIB replacement is needed?  I recognize that from header value shortening, but findfile should find the full correct path to the flux calibration file without a 'SPCALIB' in it, and `calibdir = os.environ['DESI_SPECTRO_CALIB']` only has a default fluxcalib used by nightwatch, not the nightly fluxcalib files for this night, expid.  Since `calib` shouldn't have 'SPCALIB` in it, this is a no-op, but could lead to confusion about what the code is trying to do.

py/desispec/scripts/tsnr_afterburner.py
```diff
@@ -0,0 +1,197 @@
...
+            
+            cframe = fits.open(x)
+            hdr    = cframe[0].header['FIBERFLT']
+            flat   = hdr.replace('SPECPROD', args.prod)
+            flat   = flat.replace('SPCALIB', calibdir)
```
sbailey, line 126: This case of replacement makes sense to me, since if the night didn't get good fiberflats, it falls back to default fiberflats in desi_spectro_calib.

py/desispec/scripts/tsnr_afterburner.py
```diff
@@ -0,0 +1,197 @@
...
+            
+            tileid = cframe[0].header['TILEID']
+        
+            iin = x.replace('cframe', 'frame')
+            sky = x.replace('cframe', 'sky')
+            psf = sky.replace('sky', 'psf')
+            nea = '/project/projectdirs/desi/users/mjwilson/master_nea/masternea_{}.fits'.format(camera)  
+            ens = '/project/projectdirs/desi/users/mjwilson/tsnr-ensemble/'
```
sbailey, line 134: these hardcoded paths appears to be unused leftovers.  Please remove.

py/desispec/scripts/tsnr_afterburner.py
```diff
@@ -0,0 +1,197 @@
...
+
+            else:
+                summary = vstack((summary, entry))
+                
+            log.info('{:08d}  {}: Reduced {} of {}.'.format(expid, cam, kk, len(sci_frames[cam])))
+
+        hdul.close()
+        
+    # 
+    summary.write('/global/cscratch1/sd/mjwilson/trash/afterburner/tsnr/summary_{}.fits'.format(cam), format='fits', overwrite=True)
```
sbailey, line 197: This hardcoded path, however, appears to be used.  Please scrub all hardcoded paths from this code and use options like `--outdir` instead.

py/desispec/scripts/tsnr_afterburner.py
```diff
@@ -0,0 +1,197 @@
...
+    parser.add_argument('--expids', type = str, default = None, required=False,
+                        help = 'Comma separated list of exp ids to be reduced.')
+    args = None
+    if options is None:
+        args = parser.parse_args()
+    else:
+        args = parser.parse_args(options)
+    return args
+
+args=parse()
```
sbailey, line 39: All of the code from here on down in this file should be wrapped in a function so that `import desispec.scripts.tsnr_afterburner` doesn't trigger it trying to run a bunch of code.  That could happen e.g. by a test runner looking for tests, or a documentation scraper looking for every function to auto-generate docs.  It's ok to have
```
if __name__ == '__main__':
    main()
```
so that this script can be called with `python tsnr_afterburner.py` without putting in the default installed $PATH.

py/desispec/scripts/tsnr_afterburner.py
```diff
@@ -0,0 +1,197 @@
...
+
+#
+log = get_logger()
+
+# Set SPEC_PROD also.
+calibdir = os.environ['DESI_SPECTRO_CALIB']
```
sbailey, line 45: I think this is used to replace `SPCALIB` in headers with a full file path, but in that case the headers themselves should also have kept the value of `$DESI_SPECTRO_CALIB` in some DEPNAMnn/DEPVERnn keyword pair.  `desiutil.depend.getdep(header, 'DESI_SPECTRO_CALIB')` parses those for you to get the DESI_SPECTRO_CALIB value.   Using that would prevent the case of having the file generated pointing to one version of `$DESI_SEPCTRO_CALIB`.

py/desispec/scripts/tsnr_ensemble.py
```diff
@@ -0,0 +1,147 @@
...
+    Generate an ensemble of templates to sample tSNR for a range of points in                                                                                                                                                             
+    (z, m, OII, etc.) space.                                                                                                                                                                                                                                                                                                                                                                                                                                                       
+    If conditioned, uses deepfield redshifts and (currently r) magnitudes to condition simulated templates.                                                                                                                               
+    '''
+    def __init__(self, outdir, tracer='ELG', nmodel=5):
```
sbailey, line 54: please add a docstring defining inputs and e.g. what options for "tracer" are allowed, and noting that the act of creating this object writes a file to disk.

py/desispec/scripts/tsnr_ensemble.py
```diff
@@ -0,0 +1,147 @@
...
+    Generate an ensemble of templates to sample tSNR for a range of points in                                                                                                                                                             
+    (z, m, OII, etc.) space.                                                                                                                                                                                                                                                                                                                                                                                                                                                       
+    If conditioned, uses deepfield redshifts and (currently r) magnitudes to condition simulated templates.                                                                                                                               
+    '''
+    def __init__(self, outdir, tracer='ELG', nmodel=5):        
+        def tracer_maker(wave, tracer=tracer, nmodel=nmodel, redshifts=None, mags=None):
```
sbailey, line 55: please add docstring

py/desispec/scripts/tsnr_ensemble.py
```diff
@@ -0,0 +1,147 @@
...
+            # https://arxiv.org/pdf/1611.00036.pdf
+            if tracer == 'elg':
```
sbailey, line 57: The default `template_ensemble` constructor `tracer` is "ELG" (uppercase), while this `tracer_maker` function appears to assume all tracers are lowercase.  Let's be consistent throughout, or otherwise include code like `tracer = tracer.lower()` to make it input case agnostic.

py/desispec/scripts/tsnr_ensemble.py
```diff
@@ -0,0 +1,147 @@
...
+                                    
+def main():
+    log = get_logger()
+
+    args = parse()
+    
+    rads = template_ensemble(args.outdir, tracer=args.tracer, nmodel=args.nmodel)
+
+if __name__ == '__main__':
+    main()
```
sbailey, line 147: This is good, and is the method I was suggesting for `tsnr_afterburner.py` too.

py/desispec/tsnr.py
```diff
@@ -0,0 +1,269 @@
...
+
+    else:
+        return alpha * rdnoise_variance + fiberflat.fiberflat * skymodel.flux
+
+def calc_alpha(frame, fibermap, rdnoise_sigma, npix_1d, angperpix, angperspecbin, fiberflat, skymodel):
+    '''
+    Model Var = alpha * rdnoise component + sky.
+
+    Calcualte the best-fit alpha using the sky fibers
+    available to the frame.
```
sbailey, line 166: Please expand docstring to define all inputs and what is returned (or written?)

sbailey (changes requested):
Major (but hopefully easy to fix): procexp.py is broken for the spectro pipeline.  See comments inline.

Plus a bunch of cleanup and maintainability requests like removing hardcoded personal paths and adding docstrings.

py/desispec/scripts/master_nea.py
michaelJwilson: I'm requesting the wavelengths in the current spectro-perf. reductions.  I can update the nea files to the larger range, but I'm already seeing some breakdown at the current range and patching those with a median.  We argued the wavelength dependence was smooth so extrapolation may not be so bad.

py/desispec/scripts/master_nea.py
michaelJwilson: This is catching the un-normalized psf values as erroneous and patching.  Will change to logger.

py/desispec/scripts/procexp.py
michaelJwilson: Yes, agreed.  Sorry this has crept in by accident.  It should be a feature added to procexp, and was running as such recently.  I'll fix that up.

sbailey (approved):
Looks good, thanks.  I have some minor cleanup changes in my copy, but from Slack it sounds like @michaelJwilson you have some other proposed algorithmic adjustments to make, so I'll hold off on pushing my changes until we sort out potential merge conflicts.

sbailey:
I made a bunch of changes; example outputs in /global/cfs/cdirs/desi/users/sjbailey/tsnr/blanc-tsnr, from
```
cd /global/cfs/cdirs/desi/users/sjbailey/tsnr/blanc-tsnr
desi_tsnr_afterburner --prod /global/cfs/cdirs/desi/spectro/redux/blanc \
    --night 20201215 --cameras b0,r0,z0,b1,r1,z1,b2,r2,z2 \
    --details-dir . --outfile tsnr-blanc-20201215.fits
```

Punting back to Julien for whether these outputs look like the way he'd like to use them.

A few changes to note:
  * columns / HDUS now have "TSNR2" to emphasize that these are squared quantities
  * you now have to give the name of the summary file via `--outfile`, which will enable running different nights in parallel producing different summaries
  * The summary file now has one HDU for per-frame medians (TSNR2_FRAME) and another HDU for per-expid summed over cameras and averaged over petals (TSNR2_EXPID).  Also TSNR2_EXPID is derivable from TSNR2_FRAME, it is a primary end-user "what S/N have we achieved on this tile?" result and a bit tedious to calculate.
  * I renamed `--outdir` to `--details-dir` and dropped the 'tsnr/' forced prefix to give more flexibility.  I'm not sure I like the name, but I didn't want to confuse with `--outfile` which I view as the more primary output than the per-frame quantities that going forward are also going to be included in the frame SCORES table HDU anyway.

julienguy:
Looks good. I tested this on the night 20201216.
```
desi_tsnr_afterburner --night 20201216 --outfile summary.fits --details-dir . --prod /global/cfs/cdirs/desi/spectro/redux/blanc/
```
Some plots for the record.
Histograms of TSNR ( sqrt(TSNR2) ):
tsnr-20201216

As a function of exposure number
tsnr-vs-expid-20201216

As a function of camera, per target class
tsnr-ELG-vs-camera-20201216
tsnr-LRG-vs-camera-20201216
tsnr-QSO-vs-camera-20201216


Merging now.

PR #1124, opened 2021-02-01
I/O time logging

sbailey:
This PR adds additional log statements for timing I/O operations to help debug how much fluctuations in job runtimes are due to I/O fluctuations.  Messages appear like
```
INFO:frame.py:118:write_frame: iotime 0.237 sec to write frame-b5-00069400_10.fits at 2021-01-24T21:44:12.045713
```
These follow a standardized format parsable by `desispec.qa.qa_iotime.read_iotimes(logfiles)` which returns a table with rows for timing of every I/O operation.  This Table can be plotted with `desispec.qa.qa_iotime.plot_iotimes` and `hist_iotimes`, e.g.
image
which shows that there is quite a bit of variation in `write_frame` timing, to be investigated.

Getting this into the daily productions will allow us to build up statistics on I/O timing and variations for further study and optimization.

Example logfile output when running with this branch is in `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/iotime/run/scripts/night/20201222`.

Any comments, @akremin, @dmargala, or others?

### DIFF
--- a/py/desispec/io/fiberflat.py
+++ b/py/desispec/io/fiberflat.py
@@ -9,10 +9,12 @@ IO routines for fiberflat.
 from __future__ import absolute_import
 # The line above will help with 2to3 support.
 import os
+import time
 from astropy.io import fits
 
 from desiutil.io import encode_table
 from desiutil.depend import add_dependencies
+from desiutil.log import get_logger
 
 from ..fiberflat import FiberFlat
 from .meta import findfile
@@ -32,6 +34,7 @@ def write_fiberflat(outfile,fiberflat,header=None, fibermap=None):
     Returns:
         filepath of file that was written
     """
+    log = get_logger()
     outfile = makepath(outfile, 'fiberflat')
 
     if header is None:
@@ -66,9 +69,12 @@ def write_fiberflat(outfile,fiberflat,header=None, fibermap=None):
     hdus["MEANSPEC"].header['BUNIT'] = ("electron/Angstrom")
     hdus["WAVELENGTH"].header['BUNIT'] = 'Angstrom'
 
-
+    t0 = time.time()
     hdus.writeto(outfile+'.tmp', overwrite=True, checksum=True)
     os.rename(outfile+'.tmp', outfile)
+    iotime = time.time() - t0
+    log.info(f'iotime {iotime:.3f} sec to write {outfile}')
+
     return outfile
 
 
@@ -91,6 +97,8 @@ def read_fiberflat(filename):
         night, expid, camera = filename
         filename = findfile('fiberflat', night, expid, camera)
 
+    log = get_logger()
+    t0 = time.time()
     with fits.open(filename, uint=True, memmap=False) as fx:
         header    = fx[0].header
         fiberflat = native_endian(fx[0].data.astype('f8'))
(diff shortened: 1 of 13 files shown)


dmargala:
LGTM.

At the risk of over-engineering this, I wonder if it would be useful to codify the connection between the standardized log message and the parsing pattern. This might come in handy if there were multiple consumers of the log messages or if/when changes are made to the structured message format.

One way to accomplish this would be to use the following class in place of `iotime_message(...)`:

```python
class IOTimeMessage:

    message = "iotime {iotime:.3f} sec to {readwrite} {basename} at {timestamp}"
    #- Note that the log prefix ".*:(.*): " is not included in pattern since that is out of context
    pattern = r"iotime ([\d.]+) sec to (\b(?:read|write)) (.*) at (.*)"

    def __init__(self, readwrite, filename, iotime):
        self.readwrite = readwrite
        self.iotime = iotime
        self.basename = os.path.basename(filename)
        self.timestamp = datetime.datetime.now().isoformat()
        
    def __str__(self):
        return self.message.format(iotime=self.iotime, readwrite=self.readwrite, 
            basename=self.basename, timestamp=self.timestamp)
```

```python
#- Logging usage
log.info(IOTimeMessage("read", filename, iotime))

#- Log parsing usage
iolog = re.compile(r".*:(.*): " + io.util.IOTimeMessage.pattern)
```

```python
#- Example test
readwrite = "read"
filename, basename = "/path/to/file.ext", "file.ext"
iotime = 123.45

string = str(IOTimeMessage(readwrite, filename, iotime))
m = re.match(IOTimeMessage.pattern, string)
parsed_iotime, parsed_readwrite, parsed_filename, parsed_isotime = m.groups()

assert parsed_readwrite == readwrite, "Failed to parse readwrite"
assert parsed_filename == basename, "Failed to parse filename"
try:
    assert float(parsed_iotime) == iotime, "Failed to parse iotime"
except ValueError:
    assert False, "Failed to parse iotime"
try:
    datetime.datetime.fromisoformat(parsed_isotime)
except ValueError:
    assert False, "Failed to parse isotime"
```

You could also push the parsing logic into this class:

```python
class IOTimeMessage:
    ...
    @classmethod
    def parse(cls, string, prefix=""):
        m = re.match(prefix + cls.pattern, string)
        if m is not None:
            groups = m.groups()
            return groups
            #- Or, could convert to desired types here, passing potential prefix groups through:
            # iotime, readwrite, filename, isotime = groups[-4:]
            # iotime = float(iotime)
            # isotime = datetime.datetime.fromisoformat(isotime)
            # return groups[:-4] + (iotime, readwrite, filename, isotime, )
        else:
            return None

#- Example usage
string = "INFO:frame.py:118:write_frame: iotime 0.237 sec to write frame-b5-00069400_10.fits at 2021-01-24T21:44:12.045713"
result = IOTimeMessage.parse(string, prefix=r".*:(.*): ")
if result is not None:
    func, iotime, readwrite, filename, isotime = result
```

sbailey:
@dmargala thanks for the comments.  Good point about the fragility of having the formatting separated from the parsing.  I'm not going to go full object-oriented on this, but I did at least consolidate the functions down into a single module plus a unit test that the formatted iotime string can be parsed either bare or when passed to a desi default logger.

PR #1127, opened 2021-02-04
Desi proc dashboard

zkdtc:
Major revisions:
1) Read exposure table to retrieve header keywords. This makes the code 20 times faster. If no exposure table is present, use old way of scanning files and read headers. 
2) Color code the progress columns. 0/0 is blue background gray font. m/n is orange background black font. n/n is blue background black font. 0/n is pink background black font. 
3) Add tileid column, linked to tile information page. 
4) Swap the Fflat and frame columns positions to reflect processing orders. 

Screen Shot 2021-02-03 at 9 05 51 PM


sbailey (approved):
Looks good to me, thanks, but I will wait for @akremin to have an opportunity to comment before merging.

akremin (approved):
I have not tested the code, but the screenshot looks excellent and the code revisions look good as well. Thanks!

PR #1131, opened 2021-02-08
Backward compatibility

julienguy:
Allows to preprocess and extract early test stand data with sparse fiber slit, and with different or missing header keywords.
Hopefully harmless PR for recent data.

### DIFF
--- a/py/desispec/io/raw.py
+++ b/py/desispec/io/raw.py
@@ -16,7 +16,7 @@ import desispec.io.util
 from desispec.util import header2night
 import desispec.preproc
 from desiutil.log import get_logger
-from desispec.calibfinder import parse_date_obs, CalibFinder 
+from desispec.calibfinder import parse_date_obs, CalibFinder
 import desispec.maskbits as maskbits
 
 def read_raw(filename, camera, fibermapfile=None, **kwargs):
@@ -34,9 +34,9 @@ def read_raw(filename, camera, fibermapfile=None, **kwargs):
 
     Returns Image object with member variables pix, ivar, mask, readnoise
     '''
-    
+
     log = get_logger()
-    
+
     fx = fits.open(filename, memmap=False)
     if camera.upper() not in fx:
         raise IOError('Camera {} not in {}'.format(camera, filename))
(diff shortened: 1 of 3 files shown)


py/desispec/io/raw.py
```diff
@@ -68,7 +68,7 @@ def read_raw(filename, camera, fibermapfile=None, **kwargs):
     try:
         tmp = int(header['NIGHT'])
     except (KeyError, ValueError, TypeError):
-        header['NIGHT'] = header2night(header)
+        header['NIGHT'] = primary_header['NIGHT']
```
sbailey, line 71: This need to go back to `header2night(header)` or maybe `header2night(primary_header)` or something using `header2night` to be able to catch all the special cases of missing HEADER, str vs. int, ...

sbailey (changes requested):
OK for restoring support for <500 fibers, but we need to keep the use of `header2night` to not break several special-case nights of more recent data.  Let's sort out offline what the problem was that caused you to revert that, and then get this to support that case while still using `header2night`.

py/desispec/io/raw.py
julienguy: I would like to be able to preprocess `/global/cfs/cdirs/desi/spectro/teststand/rawdata/SM01/2018/20180403/WINLIGHT_00007873.fits`  camera `z1` , but hdu `Z1` does not have date or mjd info. It's only in the primary header.

sbailey: Cascading chains of try/except: I updated it so that if header2night(header) also fails, it will fall back to using the primary header.  This supports both 20210105 with corrupted NIGHT keywords, normal nights with good NIGHT keywords, and early teststand data which only have NIGHT / timestamp keywords in the primary header.  I tested your case with
```
desi_preproc -i /global/cfs/cdirs/desi/spectro/teststand/rawdata/SM01/2018/20180403/WINLIGHT_00007873.fits --cameras z1 --outfile $SCRATCH/blat.fits
```

sbailey (approved):

Issue #1132, opened 2021-02-10
fix PHOTSYS='G'


sbailey:
yet another Gaia-target case that wasn't caught by #1105, #1109, and #1114.  As reported by @akremin
```
desi_fit_stdstars --delta-color 0.1 \
--frames \
  /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20210208/00075147/frame-b6-00075147.fits \
  /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20210208/00075147/frame-r6-00075147.fits \
  /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20210208/00075147/frame-z6-00075147.fits \
--skymodels \
  /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20210208/00075147/sky-b6-00075147.fits \
  /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20210208/00075147/sky-r6-00075147.fits \
  /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20210208/00075147/sky-z6-00075147.fits \
--fiberflats \
  /global/cfs/cdirs/desi/spectro/redux/daily/calibnight/20210208/fiberflatnight-b6-20210208.fits \
  /global/cfs/cdirs/desi/spectro/redux/daily/calibnight/20210208/fiberflatnight-r6-20210208.fits \
  /global/cfs/cdirs/desi/spectro/redux/daily/calibnight/20210208/fiberflatnight-z6-20210208.fits \
--starmodels \
  /global/cfs/cdirs/desi/spectro/templates/basis_templates/v3.2/stdstar_templates_v2.2.fits \
--outfile $SCRATCH/stdstars-6-00075147.fits
...
Traceback (most recent call last):
 File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/bin/desi_fit_stdstars", line 20, in <module>
  stdstars.main(args)
 File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/scripts/stdstars.py", line 449, in main
  model_filters[band+photsys] = load_legacy_survey_filter(band=band,photsys=photsys)
 File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/io/filters.py", line 50, in load_legacy_survey_filter
[...]

PR #1133, opened 2021-02-10
Photsys fix

segasai:
Fix of stdstars to support PHOTSYS='G' (see #1132 )


sbailey:
Thanks for the quick fix; looks good.

PR #1138, opened 2021-02-16
Fiber crosstalk

julienguy:
Code to compute and correct for fiber cross-talk at the spectral level
===================================================
Fiber cross-talk (or fiber to fiber contamination) is induced by extended Lorentzian PSF tails on the spectrographs CCDs. It is a small effect in most of the wavelength range but becomes large (reaching 1%) in the NIR for wavelength > 9000A. It causes a visible spurious increase at the largest NIR wavelength in the spectra of faint targets which fiber is next to a bright one in the spectrographs.

 2D modeling
----------------------
The PSF tails can be modeled accurately using data from the parse test fiber slit. The following figures show the PSF tails measured with continuum lamp data from SM1 NIR camera (solid curves), along with the model (dashed curves) obtained by adding a Lorentzian tail to the 2D PSF model used for extraction ( in specter ).

psf-cross-profile-sm1-z
psf-tails-vs-wave-sm1-z-continuum

A precise measurement shows that the PSF tails are asymmetric and depend on the position of the fiber in the CCD.
On the following figure the specter PSF tails (dashed curve) are also asymmetric,
psf-tails-precise-profile-sm1-z

The 2D PSF tail model fitted on the sparse fiber data is found consistent with the signal measured on the edge fibers for exposures with the continuum lamps on the Mayall dome white spot.

SM1-z-profile

1D correction
----------------------
Despite having a rather good modeling of the 2D PSF tails, we resorted not to use this approach to correct for the fiber cross-talk. 
Some cross-talk is visible on fiber F+2 (next to next fiber). Correcting for this with a 2D PSF would require to extend its size/footprint to more than 2x15 pixels on each dimension which would make the extraction prohibitively slow and would also make the extraction even more sensitive to unmasked cosmic ray pixels. 

For this PR we resort to use an empirical 1D correction, meaning at the spectral level, after extraction.

* `desi_compute_fiber_crosstalk` , measure the crosstalk
  * runs on a very large set of sframe files for a given camera (or multiple cameras of the  same arm (b,r, or z) but different spectrographs)
  * computes for each fiber bundle (group of 25, to detect variations across the CCD) the contamination from fiber F-2,-1,+1+2, as a function of wavelength, after convolving the contaminant spectra in 1D by the correct kernel derived from the known 2D PSF tails (and taking into account the fiber separation in the CCD), accounting for the offset in the wavelength solution between the neighboring fibers.
[...]

PR #1139, opened 2021-02-16
Create script to submit an entire desi production

akremin:
This is a first attempt at a script to submit a full set of DESI nights for reprocessing. It's based on an input yaml file, which I chose so that the production is self-documenting in what data was run and what wasn't. A complete example file is provided later on below.

The script is called `desi_run_prod` and takes the following:

```
  --production-yaml PRODUCTION_YAML.  Relative or absolute pathname to the yaml file summarizing the production.
  --dry-run, bool. Default is False. Perform a dry run where no jobs are actually created or submitted.
  --error-if-not-available, bool. Default is False. Raise an error if an exposure table doesn't exist instead of reporting and moving on to the next night.
```

`desi_run_prod`  requires that the exposure tables already exist in the specprod. These can either be copied from the `/exptabs/` specprod or they can be generated using `create_exposure_tables`.

An example yaml file at NERSC: `$DESI_SPECTRO_REDUX/prodruntest/prodruntest.yaml`
```
# Configuration File Defining a Spectroscopic Production
name: prodruntest

# Not actually used anywhere, but document what tag you're using
codetag: 21.02

# Queue to run with
queue: realtime

# Specify any nights you don't want to process,
# even though they have some sort of data
# (nights with no data are never considered)
SkipNights:
    MiniSV2: [20200221, 20200222, 20200306, 20200307]
    SV0: [20200310, 20200311, 20200312]
    SV1: []

# Specify whether or not to process given datasets
ProcessData:
    ExtraCals: True
    MiniSV: True
    SV0: True
    SV1: True

# Define the date ranges for the surveys
DateRanges:
    MiniSV: [20200219, 20200311]
    SV0: [20200312, 20200315]
    SV1: [20201214, 20210214]
```

Tested with:
```
export SPECPROD=prodruntest
desi_run_prod --production-yaml=$DESI_SPECTRO_REDUX/prodruntest/prodruntest.yaml --dry-run
```

run all the way through all nights. This produced all the processing tables, which passed spot checks.

Also tested with:
`desi_run_prod --production-yaml=$DESI_SPECTRO_REDUX/prodruntest/prodruntest.yaml `

which actually submitted jobs. The dependencies were given to slurm and the jobs were submitted appropriately. Note that I terminated this after two nights and canceled the jobs from actually running through the queue.


akremin:
It's worth adding that this does pass "python setup.py build_sphinx" locally on NERSC, even though the tests on github are not passing for documentation.

sbailey:
I haven't reviewed this in detail, and knowing that you have N>>1 things you'd like to improve on this too and ran out of time for Cascades, I won't dig into quibble on details.  If it pragmatically works to launch a production, let's go for it.

Merge when ready and update desispec/master at NERSC, and then run a miniprod with
```
TILEID TARGETS  NIGHT     EXPIDS
       ARC      20201216  68217-68221
       FLAT     20201216  68232-68249
80608  ELG      20201216  68317,68327,68328
80609  QSO+LRG  20201216  68334,68336,68337,68338,68339,68340
       ARC      20201221  69151-69155
       FLAT     20201221  69166-69183
80608  ELG      20201221  69249,69251,69252,69253
80613  BGS      20201221  69225,69226,69227,69228,69229,69230
```
We won't wait for crosstalk PR #1138.

sbailey (approved):

akremin:
This also creates `desi_run_night` which can submit a single night for reprocessing. It also requires the exposure table to exist in the current specprod. 

Tested on multiple nights with different `--queue`'s and with `--dry-run` set and not set. Also verified that `desi_run_prod` still works and verified that none of this affected the daily processing code.

`desi_run_night` arguments:

  ```
    "-n","--night", type=str, The night you want to be processed.
    "-q", "--queue", type=str, The queue to submit jobs to. Default is realtime
```

And I once more verified that documentation still passes at NERSC running `python setup.py build_sphinx`.

PR #1140, opened 2021-02-16
fail faster with error for joint fits

sbailey:
This PR updates `desi_proc_joint_fit` to fail harder with an exit code if the joint fitting fails for all cameras, so that the job itself can exit with an error code and alter dependent jobs that they shouldn't run.  This is for psfnight, fiberflatnight, and stdstars.  In principle the same logic could/should be added to `desi_proc` for other steps but I'm trying to minimize the last-minute organizational changes for Cascades, and these joint fit steps are our most critical barrier steps where we don't want the pipeline to proceed if they fail.

It also fixes a bug where if rank 0 `desispec.scripts.specex.mean_psf` failed, all other ranks would be waiting at a barrier instead of letting the job exit.  We had fixed the equivalent problem in desi_proc individual arc fits, but not yet in the joint fit.  This now has a generic try/except that will print the exception and then move on to let all ranks exit.

I tested this by moving minicascades files out of the way and then running the pipeline step as my own account, which fails to write the output files due to write permission problems, and now correctly exits with an error.

This also pins sphinx to an older version for unit tests; docs build locally and with this change hopefully will build on github actions too.

@akremin any comments / requests / questions?

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -97,7 +97,7 @@ jobs:
               with:
                 python-version: ${{ matrix.python-version }}
             - name: Install Python dependencies
-              run: python -m pip install --upgrade pip wheel Sphinx
+              run: python -m pip install --upgrade pip wheel Sphinx<3.5
             - name: Test the documentation
               run: sphinx-build -W --keep-going -b html doc doc/_build/html
 
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -97,7 +97,7 @@ jobs:
               with:
                 python-version: ${{ matrix.python-version }}
             - name: Install Python dependencies
-              run: python -m pip install --upgrade pip wheel Sphinx<3.5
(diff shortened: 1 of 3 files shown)


akremin:
I'm happy with this PR (for the record I didn't test it, so I'm trusting your tests)

weaverba137:
`'Sphinx<3.5'` or `Sphinx\<3.5`.

sbailey:
@akremin yes, I tested it with all 3 cases of missing files: psfnight, fiberflatnight, and stdstars (actually, for stdstars I tested the case where speclite<0.13 doesn't know about gaia filters and will crash differently).  i.e. I didn't trust that I edited all three cases correctly. :)

@weaverba137 thanks for the syntax correction.  Since Sphinx==3.1.2 worked I'm going to leave it at that so that I can merge without rerunning, but if there is a reason to test sphinx newer than 3.1.2 but older than 3.5, that could be tested separately.

PR #1146, opened 2021-02-23
Restore desi process exposure

julienguy:
Minor fixes so that the following does not crash and does not set ivar=0 for entries where mask!=0.
```
desi_process_exposure -i /project/projectdirs/desi/spectro/redux/daily/exposures/20210101/00070317/frame-z1-00070317.fits --sky /project/projectdirs/desi/spectro/redux/daily/exposures/20210101/00070317/sky-z1-00070317.fits --fiberflat /project/projectdirs/desi/users/jguy/teststand/desi_spectro_calib/spec/sm10/fiberflatnight-z1-20201214.fits --no-zero-ivar --no-tsnr -o 20210101/sframe-z1-00070317.fits
```

### DIFF
--- a/py/desispec/fiberbitmasking.py
+++ b/py/desispec/fiberbitmasking.py
@@ -17,9 +17,9 @@ from desispec.maskbits import specmask
 def get_fiberbitmasked_frame(frame,bitmask=None,ivar_framemask=True):
     """
     Wrapper script of get_fiberbitmasked_frame_arrays that will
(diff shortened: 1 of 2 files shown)


sbailey:
Thanks.  I confirmed that this fixes the intended behavior of the `--no-zero-ivar` _option_, i.e. that the default behavior as called by desi_proc is unchanged, so I'll merge now to include in the next tag for Cascades/21.2 take-2.

PR #1154, opened 2021-02-25
undo heliocentric correction

julienguy:
Undo the heliocentric correction when averaging the throughput of many exposures in desi_average_flux_calibration.
This is a minor bug fix that has no influence on the pipeline production (which does not use desi_average_flux_calibration).

As expected, the telluric features appear sharper after this applying this correction.

desi-throughput


sbailey:
ok to merge after tests finish

Issue #1156, opened 2021-02-26
fluxcalib crash: null ivar


sbailey:
Using desispec/0.39.1:
```
desi_compute_fluxcalibration --delta-color-cut 0.1 \
  --infile /global/cfs/cdirs/desi/spectro/redux/cascades/exposures/20210106/00071101/frame-r1-00071101.fits \
  --sky /global/cfs/cdirs/desi/spectro/redux/cascades/exposures/20210106/00071101/sky-r1-00071101.fits \
  --fiberflat /global/cfs/cdirs/desi/spectro/redux/cascades/calibnight/20210106/fiberflatnight-r1-20210106.fits \
  --models /global/cfs/cdirs/desi/spectro/redux/cascades/exposures/20210106/00071101/stdstars-1-00071101.fits \
  --outfile $SCRATCH/fluxcalib-r1-00071101.fits 
...
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/0.39.1/bin/desi_compute_fluxcalibration", line 4, in <module>
    __import__('pkg_resources').run_script('desispec==0.39.1', 'desi_compute_fluxcalibration')
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/pkg_resources/__init__.py", line 667, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/pkg_resources/__init__.py", line 1464, in run_script
    exec(code, namespace, namespace)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/0.39.1/lib/python3.8/site-packages/desispec-0.39.1-py3.8.egg/EGG-INFO/scripts/desi_compute_fluxcalibration", line 16, in <module>
    fluxcal.main(args)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/0.39.1/lib/python3.8/site-packages/desispec-0.39.1-py3.8.egg/desispec/scripts/fluxcalibration.py", line 196, in main
    fluxcalib = compute_flux_calibration(frame, model_wave, model_flux, model_fibers%500, highest_throughput_nstars = args.highest_throughput)
[...]

PR #1157, opened 2021-02-26
Simplify fluxcalib

julienguy:
Simplification of the flux calibration now that we have good fiber positioning and standard star fit. A single scale factor is used per std star instead of a polynomial function of wavelength. This makes the code more robust. This normalization is used to discard problematic stars and to detect outlier flux values.  The solves issue #1156 . The resulting flux calibration vectors have the same average than before.

Here average of exposures 68265,68266,68272,68273 from 20201216.

desi-throughput
desi-throughput-zoom

### DIFF
--- a/bin/plot_frame
+++ b/bin/plot_frame
@@ -12,6 +12,7 @@ from desispec.util import parse_fibers
 from desispec.qproc.io import read_qframe
 from desispec.io import read_fibermap,read_frame
 from desispec.interpolation import resample_flux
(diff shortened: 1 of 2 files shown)


sbailey:
Thanks.  I confirmed that previously successful calibrations are mostly unchanged and don't have a significant impact on redshift success, so this is a net win to recovery previously crashing flux calibrations.  Merging.

Issue #1151, opened 2021-02-25
Gaia standards failing color cuts


sbailey:
sv1rosette tile 80718 has 186 standards on petal 5 according to https://data.desi.lbl.gov/desi/target/fiberassign/tiles/trunk/080/fiberassign-080718.png .  The spectro pipeline apparently considered 50 of these and discarded all of them due to the `|delta color|<0.1` cut, which smells suspicious.  Please investigate why the pipeline didn't find any good standards for this petal.
```
desi_compute_fluxcalibration --delta-color-cut 0.1 \
--infile /global/cfs/cdirs/desi/spectro/redux/cascades/exposures/20210130/00073720/frame-b5-00073720.fits \
--sky /global/cfs/cdirs/desi/spectro/redux/cascades/exposures/20210130/00073720/sky-b5-00073720.fits \
--fiberflat /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm9/fiberflatnight-b5-20201214.fits \
--models /global/cfs/cdirs/desi/spectro/redux/cascades/exposures/20210130/00073720/stdstars-5-00073720.fits \
--outfile $SCRATCH/fluxcalib-b5-00073720.fits 
```
Other petals just barely made it, e.g. sp0 discards 49/50:
```
desi_compute_fluxcalibration  --delta-color-cut 0.1 \
--infile /global/cfs/cdirs/desi/spectro/redux/cascades/exposures/20210130/00073720/frame-b0-00073720.fits \
--sky /global/cfs/cdirs/desi/spectro/redux/cascades/exposures/20210130/00073720/sky-b0-00073720.fits \
--fiberflat /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm4/fiberflatnight-b0-20201214.fits \
--models /global/cfs/cdirs/desi/spectro/redux/cascades/exposures/20210130/00073720/stdstars-0-00073720.fits \
--outfile $SCRATCH/fluxcalib-b0-00073720.fits
```
@segasai please investigate this as a followup to your implementation of supporting Gaia standards.

segasai:
Thanks, I'll take a look in next couple of days. A quick check shows that extinction in the field is EBV=0.5, So I'd guess issues must be connected to that.

segasai:
[...]

PR #1159, opened 2021-03-01
do the correction from sfd EBV when correcting model spectrum

segasai:
When making a model spectrum in stdstars the EBV needs to be transformed from SFD scale using coefficient from Schlafly+11. Fixes the issue #1151

### DIFF
--- a/py/desispec/scripts/stdstars.py
+++ b/py/desispec/scripts/stdstars.py
@@ -68,7 +68,8 @@ def safe_read_key(header,key) :
         value=header[key.ljust(8).upper()]
     return value
 
-def dust_transmission(wave,ebv) :
+def dust_transmission(wave,ebv_sfd) :
+    ebv = .86 * ebv_sfd
     Rv = 3.1
     extinction = ext_odonnell(wave,Rv=Rv)
     return 10**(-Rv*extinction*ebv/2.5)


julienguy:
This PR consists in changing the normalization of the E(B-V) from the SFD map (ebv = .86 * ebv_sfd).
This recalibration of the map has to be done elsewhere, maybe in desiutil.dust , to make sure we do not in the future apply twice the correction because of some misunderstanding. It would also be needed to add references in the comments to the code.

segasai:
I don't think I agree. Essentially E(B-V) exists in different scales. Some routines have been calibrated to take things in SFD scale, some in 'correct' scale. This patch makes all the functions in stdstars take things in the SFD scale. I.e. the gaia/decam color corrections already take EBV in SFD  scale, so this fixes the last one.
(I agree though that some functions may be better placed in desiutil.dust though)

geordie666:
Please don't change the default rescaling in `desiutil.dust`. There are multiple places in `desitarget` where `desiutil.dust` is used, and we explicitly pass the scaling in `desitarget`, where needed. It would be better for `desiutil.dust` to handle querying the maps (with a scaling passed as desired) and for other routines to use whatever scaling they wish.

If you only want to apply a scaling once for all of `desispec`, due to a concern about double-correcting, then you could create a `desispec.dust` function that wraps `desituil.dust` and automatically applies the chosen scaling.

julienguy:
@segasai , please add a single entry point for dust. I suggest we add functions in desiutil.dust instead of modifying the existing ones so it doesn't break desitarget but at least we have everything in one place and understand what is going on.
We should avoid scaling factors hidden in the middle of a specific code.

segasai:
I can create a function  like 
correct_sfd_ebv_to_true_ebv() in desiutil.dust to encapsulate the .86 factor
And use that in stdstars. Does that sound reasonable ? 

Another (long term) possibility is to force all the functions that take EBV to accept an argument like 'SFD' or 'true' to force the caller explicitly specify what kind of extinction it is.

julienguy:
yes that would be great

segasai:
Given the discussion in https://github.com/desihub/desiutil/pull/166 and https://github.com/desihub/desiutil/issues/165#issuecomment-800475188 and planned refactors of the dust code,
should we merge this for the time being (as it is a bug fix and without it some fields can't be calibrated ). @sbailey @julienguy  ?

sbailey:
Let's not merge this until after the "0.1% survey" = sv2 is completed (in a few days).

Although this fix is simple, it impacts the standard star selection of all tiles (not just off-footprint Gaia-stdstar tiles).  Realistically we don't have time to cross check its impact on normal tiles prior to sv2, nor do we have the time to benefit from it by re-running the previously failing tiles anyway.  sv2 has taken priority for the remainder of this week.

If you want to get a head start on those cross checks, it would be useful to re-run standard star selection with this branch on a variety of tiles from cascades (especially any on-footprint but low galactic latitude) to make sure that we're not moving a bunch of Gaia standards into the selection window while moving a bunch of normal standards out of the selection window due to SFD vs. corrected EBV inconsistencies.

segasai:
I've just did a crude run on a random subset of exposures (5th spectrograph) with EBV>.1 and compared the number of stars with |MODEL_color-DATA_color|<0.1 in the stdstars.fits for the original (cascades) vs patched.

EXPID NEW OLD
00069016 8 9
00069021 9 9
00069580 2 8
00070340 18 17
00070560 26 26
00070561 26 26
00070683 8 8
00070731 24 24
00070933 33 33
00070936 23 24
00071100 32 33
00071102 22 23
00071103 21 23
00071237 8 8
00071239 8 8
00071260 33 33
00071263 33 33
00071277 24 24
00071732 33 37
00071734 35 37
00071737 11 12
00071738 10 12
00071876 39 39
00071877 39 39
00071878 37 39
00071881 12 12
00072681 39 39
00073727 28 0
00073728 27 0
00076297 7 8

The rosetta tiles are obviously fixed, across all other exposures it's mostly the same number of stars +/-1 with the exception of exposure   69580.  I don't know what's going on there...

julienguy:
Superseded by PR #1269

PR #1160, opened 2021-03-01
Interpolate flux calibration over Ca and Na ISM lines

julienguy:
At the request of the MWS group, the flux calibration is now interpolated over the Ca and Na ISM lines.
Ca and Na lines appeared as residuals to the calibration with WDs (from Chris Manser):
Screenshot from 2021-03-01 11-52-26

Now interpolated (see following plots). A second systematic correction using WD flux residuals or QSO flux residuals could/will be applied in future versions.

desi-throughput-2
desi-throughput-1


sbailey:
Looks good, thanks.  I confirmed that this doesn't have unintended consequences on ELG redshifts using tile 80606 on nights 20201215, 20201216, 20201218, 20201219.

Issue #1153, opened 2021-02-25
coadd crash


sbailey:
using desispec/0.39.1:
```
desi_coadd_spectra \
  -i /global/cfs/cdirs/desi/spectro/redux/cascades/tiles/80617/20201222/spectra-4-80617-20201222.fits \
  -o $SCRATCH/coadd-4-80617-20201222.fits
...
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/0.39.1/bin/desi_coadd_spectra", line 4, in <module>
    __import__('pkg_resources').run_script('desispec==0.39.1', 'desi_coadd_spectra')
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/pkg_resources/__init__.py", line 667, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/pkg_resources/__init__.py", line 1464, in run_script
    exec(code, namespace, namespace)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/0.39.1/lib/python3.8/site-packages/desispec-0.39.1-py3.8.egg/EGG-INFO/scripts/desi_coadd_spectra", line 16, in <module>
    coadd_spectra.main(args)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/0.39.1/lib/python3.8/site-packages/desispec-0.39.1-py3.8.egg/desispec/scripts/coadd_spectra.py", line 139, in main
    coadd(spectra,cosmics_nsig=args.nsig)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/0.39.1/lib/python3.8/site-packages/desispec-0.39.1-py3.8.egg/desispec/coaddition.py", line 153, in coadd
    ttflux[bad] = np.interp(spectra.wave[b][bad],spectra.wave[b][good],ttflux[good])
  File "<__array_function__ internals>", line 5, in interp
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/numpy/lib/function_base.py", line 1423, in interp
    return interp_func(x, xp, fp, left, right)
ValueError: array of sample points is empty
```

[...]

PR #1163, opened 2021-03-02
Check number of good and bad pixels before trying interpolation

julienguy:
Solves issue #1153.


julienguy:
This fix affects only the spectra that were failing so we can push this change and  fix the spectra in the cascades run without losing consistency,

PR #1164, opened 2021-03-02
TSNR fix

julienguy:
Force a minimum value of alpha=0.8. Do not raise an exception for values lower than 0.8. Noisy values of alpha can occur for observations dominated by sky noise where it is not possible to calibrated the read noise. For those exposures, the precise value of alpha does not impact the SNR estimation.


sbailey:
I added comment from this PR about why setting a floor of alpha=0.8 is ok, and removed the try/except because calc_alpha doesn't purposefully raise a ValueError anymore anyway.

Issue #1174, opened 2021-03-06
TSNR: apply Galactic dust extinction


julienguy:
Need to apply the Galactic dust extinction to the signal.

michaelJwilson:
This seems a simple one, I can handle in dustytsnr.  Do you want to add the spectral arm coefficients to desiutil.dust, or approximate them by the imaging bands?

julienguy:
you can apply the extinction to the delta_f as a function of wavelength as in https://github.com/desihub/desispec/blob/3699f70a57afff58d2c843f2741af16ef91d70da/py/desispec/scripts/stdstars.py#L71

michaelJwilson:
I pushed this to dustytsnr.  It seems simple enough to vet by itself, otherwise happy to run some test tiles for the effect.  We can do one PR per issue, or try to clean up multiple of the ticket in one.  Whatever your preference.

sbailey:
Thanks @michaelJwilson let's do one PR per issue so that questions about one set of changes (e.g. the ongoing conversation over in #1175 about TSNR and fiberfrac/seeing) don't block another set of useful changes from getting merged.  It also distinguishes what changes come from dust updates vs. seeing updates, etc.

michaelJwilson:
Sounds good. I'll wait for @julienguy to say whether this is simple enough to go in, or needs vetting with some test tiles. If the latter, we should probably discuss how to best set up some development suite or similar. It seems there's enough material for a discussion for the data call, if the agenda isn't otherwise full.

julienguy:
I think this is simple enough to go in. Please open a PR.

julienguy:
solved

PR #1176, opened 2021-03-08
Dust transmission inclusion.

michaelJwilson:
Addresses #1174.


julienguy:
Thanks. (I checked with a test run of desi_tsnr_afterburner that the TSNR2 values are smaller because of the dust extinction term).

PR #1178, opened 2021-03-09
update copyprod

sbailey:
This PR rewrites `bin/copyprod` to copy subsets of a production into a symlink farm directory structure of a new production.  You can then delete files at will in the new production and reprocess.  The intended uses are:
  * making mini test production starting points, e.g. to redo cframes on all truth table exposures without having to start from scratch
  * fast-track productions like Denali that will start by "copying" Cascades

The default is to make relative symlinks, but there are options to make absolute links or fully copy the files.  The `--explist` option can provide a file with NIGHT and EXPID columns to filter the subset of nights and expids that are linked.

What is does:
  * directory structure copied, links created (or files copied), filtered by NIGHT and EXPID:
    * `calibnight/NIGHT/*.*`
    * `exposures/NIGHT/EXPID/*.*`
    * `preproc/NIGHT/EXPID/*.*`
  * input files read, filtered by NIGHT/EXPID, and rewritten to new prod:
    * `exposure_tables/`
    * `processing_tables/`
  * not copied or linked: anything else not mentioned above, e.g.
    * `run/`
    * `tiles/`
    * `dashboard/`
    * top level files like `zcatalog*.fits`

e.g.
```
cd /global/cfs/cdirs/desi/users/$USER/
copyprod /global/cfs/cdirs/desi/spectro/redux/cascades minicascades \
    --explist /global/cfs/cdirs/desi/spectro/redux/cascades/run/scripts/tiles/explist-deep-dark.txt
```

@akremin please review, especially if I got the exposures_tables and processing_tables filtering correct.  I'm pretty sure we do want to copy+filter the input exposure_tables; it is less clear to me if we also want to copy+filter the processing_tables.

### DIFF
--- a/bin/copyprod
+++ b/bin/copyprod
@@ -9,6 +9,7 @@ March 2021
 
 import os, sys, glob
 import shutil
+from collections import Counter
 import optparse
 
 import numpy as np
@@ -23,6 +24,8 @@ parser.add_option("--fullcopy", action="store_true",
         help="Copy files instead of linking")
 parser.add_option("--abspath", action="store_true",
         help="Link to absolute path instead of relative path")
+parser.add_option("--filter-exptables", action="store_true",
+        help="Filter exposure_tables to ignore entries not in input explist")
 
 opts, args = parser.parse_args()
 inroot, outroot = args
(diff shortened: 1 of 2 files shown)


akremin:
This looks good and I fully support having this as an official and documented piece of code that shows how we got from one production to the other. I have a number of questions so that our thoughts can be captured for future memory and a few of comments pertaining mainly pertaining to the tables. Mostly just questions to get clarification on.

Is this meant to be a general-purpose tool for future use or specific to the cascades->denali transition?

There are some features of copying or symbolic-linking that won't perfectly map, which include:
1. Complete file paths in fit's headers with old path locations,
2. `SPECPROD` specific header info.

I think these are fine, but wanted to point them out. Certainly, the dependencies are correct in what was used to process that data (since the data itself was taken from the original location).

For the exposure tables I see you are searching for just the exposures you want to copy data for. If there are circumstances where you'll want to process exposures that you don't copy over, then this will break that ability since the pipeline looks to these tables to know what exposures are available to ever be processed.  I think it general we just want to copy these over. If you never want to process the other exposures, you could alternatively set their `LASTSTEP` to 'ignore' rather than deleting them.

I could also imagine the "official" exposure tables (presently defined in SPECPROD=exptable) may change and improve compared with the older production being copied. Theoretically, you could take the rows from the prod-specific exposure table for the processed exposures and the global truth rows from the "official" for the exposures that haven't been processed. I think this is the rigorously correct thing to do but is also a huge pain and likely not worth the effort.

For the production tables, these are meant to explicitly say what was processed. Carrying over the processing rows corresponding to precisely the files on disk you are copying sounds reasonable, but your current implementation keeps any row that contains an exposure ID, which will include e.g. flux calibration steps where those files weren't copied. That information will therefore be incorrect to the files on disk for the new production.

Also, if we're copying the production table, shouldn't we also copy the scripts and logs (only those that generated copied files)? These serve a similar purpose of documenting what, when, where it was run; and what the outcomes were.

For the exposure table and processing table you should use the load_table() function rather than parsing the strings of an ascii.csv table yourself. The purpose of `desispec.workflow.tablio.load_table()` is to keep all of that messiness in one maintainable place. There is an analogous `desispec.workflow.tablio.write_table()` function as well.

sbailey:
> Is this meant to be a general-purpose tool for future use or specific to the cascades->denali transition?

Primary purpose is to setup mini test productions, e.g. I revived this while setting up productions for re-processing cframes for evaluating fiber cross talk.  It's potential utility for Denali is a extra benefit, but not the original purpose.

> There are some features of copying or symbolic-linking that won't perfectly map, which include:
> * Complete file paths in fit's headers with old path locations,
> * SPECPROD specific header info.
> I think these are fine but wanted to point them out. Certainly, the dependencies are correct in what was used to process that data (since the data itself was taken from the original location).

Good point; I think for Denali we just live with that as a consequence of a Frankenprod.

> For the exposure tables I see you are searching for just the exposures you want to copy data for. If there are circumstances where you'll want to process exposures that you don't copy over, then this will break that ability since the pipeline looks to these tables to know what exposures are available to ever be processed. I think it general we just want to copy these over. If you never want to process the other exposures, you could alternatively set their LASTSTEP to 'ignore' rather than deleting them.

Copying and setting LASTSTEP to 'ignore' sounds good.

> I could also imagine the "official" exposure tables (presently defined in SPECPROD=exptable) may change and improve compared with the older production being copied. Theoretically, you could take the rows from the prod-specific exposure table for the processed exposures and the global truth rows from the "official" for the exposures that haven't been processed. I think this is the rigorously correct thing to do but is also a huge pain and likely not worth the effort.

Let's add that if/when it actually becomes necessary.  I don't think it applies for mini test prods or Denali.

> For the production tables, these are meant to explicitly say what was processed. Carrying over the processing rows corresponding to precisely the files on disk you are copying sounds reasonable, but your current implementation keeps any row that contains an exposure ID, which will include e.g. flux calibration steps where those files weren't copied. That information will therefore be incorrect to the files on disk for the new production.

> Also, if we're copying the production table, shouldn't we also copy the scripts and logs (only those that generated copied files)? These serve a similar purpose of documenting what, when, where it was run; and what the outcomes were.

It's unclear to me whether we need to copy the production table as a starting point for mini test prods at all.  I'm open to dropping this or otherwise updating it to be whatever is useful for Denali.  However, I think we should *not* copy the scripts since they have hardcoded paths to the previous prod, and rerunning them as the same user risks accidentally updating the previous prod.  I'm not trying to achieve exact self-contained documentation of what was run, but rather achieving a clean starting point for continuing with a new run.

> For the exposure table and processing table you should use the load_table() function rather than parsing the strings of an ascii.csv table yourself. The purpose of desispec.workflow.tablio.load_table() is to keep all of that messiness in one maintainable place. There is an analogous desispec.workflow.tablio.write_table() function as well.

Noted

sbailey:
@akremin I updated to code to:
  * in exposure tables, EXPIDs that are not copied are set to LASTSTEP='ignore' rather than deleting them from the table entirely
    * but I do still delete them from the proctable, which I think is equivalent to what would happen if they were LASTSTEP='ignore' in the exptable to start with.
  * use `load_table` / `write_table` (which also simplified the proctable EXPID parsing, thanks)

For my mini testprod case I don't actually need either of these tables, but let's try to get them to be close to what we think we'd want for Denali.  e.g. I'm not going to worry about cases like "a tile was observed on two exposures but I'm only copying one of them so what to do about the stdstar proctable entry that references one exposure I'm copying and one exposure that I'm not copying".

Do you have other suggestions/requests for how to properly handle the proctable?  An alternative would be to drop it from copyprod for now, and re-add it when we have a clearer idea of what we really want for Denali.

akremin:
Thanks for improving the docstring on load_table. Eyes from end-users are helpful for that. 

The updates look great.

It sounds like you're primarily approaching this from the minitest perspective where you want a small subset of data, and this logic excises that for you?

For the purpose of a production I'm a bit nervous about the idea of quietly ignoring a bunch of exposures that were deemed by the official tables as being okay to process. For denali I'd prefer it to either copy the exposure tables with no modifications or raise big loud warnings for each exposure it flags so that I can investigate why it wasn't in cascades to start with.

Maybe a remedy is to set this curation step as an optional flag (you can choose the default polarity)?

For the processing tables, I think we should choose one of the following. Given that (2) takes effort and (1) doesn't, I'm perfectly okay with you choosing (1).  
1) Don't copy them
2) Only keep the rows corresponding to things with data products in the new production. Ie. keep the table['JOBDESC']=='prestdstar' rows if you are copying over sframes and earlier. Keep 'prestdstar' or 'stdstarfit' if you're also copying the stdstar\*fits files. Keep all rows if you're copying over everything up to and including cframes.

As you say-- given that the re-processing script is evolving as we speak, and given the uniqueness of denali -- we may want to go the simple route of not copying the processing tables until we know that we would use the information in (2).

akremin (approved):
Thank you for adding the flag so that we have to be explicit to alter the exposure tables. The code all looks good and it was able to correctly copy a mini-prod I setup. It (by default) copied the full exposure tables and made the links. The linking all worked and I was able to open and work with the data as if it were local.

PR #1181, opened 2021-03-11
Updates to TSNR

julienguy:
- improved read noise normalization (masking sky lines)
- averaging per exposure: use median across the valid fibers of a camera, but mean across spectrographs
- compute ELG_EFFTIME_DARK and BGS_EFFTIME_BRIGHT based on the analysis reported in https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=6110
- add multiprocessing option to `desi_tsnr_afterburner`
- add `--append` option to `desi_tsnr_afterburner` so that it can be run daily and it's output be used for the afternoon planning.

Example:
```
desi_tsnr_afterburner -o tsnr-exposures-daily.fits --update --prod daily --nights 20210304 --nproc 16
```

Figures from DESI-6110

calibration-ELG

completeness-ELG

### DIFF
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -32,8 +32,8 @@ def parse(options=None):
                 description="Calculate template S/N ratio for exposures")
     parser.add_argument('-o','--outfile', type=str, default=None, required=False,
                         help = 'Output summary file')
-    parser.add_argument('--append', action = 'store_true',
-                        help = 'Append pre-existing output summary file')
+    parser.add_argument('--update', action = 'store_true',
+                        help = 'Update pre-existing output summary file (replace or append)')
     parser.add_argument('--details-dir', type=str, default = None, required=False,
                         help = 'Dir. to write per-exposure per-camera files with per-fiber tSNR details')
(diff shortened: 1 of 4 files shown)


bin/desi_tsnr_afterburner
```diff
@@ -210,99 +343,44 @@ def main():
...
+
+
+        # write result after every other night
+        if len(summary_rows)>0 :
+
+            print("summary_rows = ",summary_rows)
+
+            tmpfilename=args.outfile.replace(".fits","_tmp.fits")
+            write_summary(summary_rows,tmpfilename,efftime_config,preexisting_tsnr2_expid_table,preexisting_tsnr2_frame_table)
+            log.info("wrote {} entries in tmp file {}".format(len(summary_rows),tmpfilename))
```
sbailey, line 376: This `tmpfilename` is written, but I don't see it ever renamed to a non-temp filename, or otherwise removed after the final file is written.  Am I missing something?

sbailey:
Looks good, thanks.  I tested that --update works when adding completely new exposures, and also when adding a night that had been previously partially added and now we're adding more exposures.

PR #1182, opened 2021-03-12
Use sparse matrix types in compute_uniform_sky

ziyaointl:
This change should not affect the end results at all - fitsdiff shows that only timestamps are different.
However, the performance improvement is drastic. For the exposure I tested, desi_compute_sky goes from ~5m40s to ~1m30s.

### DIFF
--- a/py/desispec/sky.py
+++ b/py/desispec/sky.py
@@ -234,11 +234,11 @@ def compute_uniform_sky(frame, nsig_clipping=4.,max_iterations=100,model_ivar=Fa
         # B = sum_fiber sum_wave_w sqrt(ivar)[fiber,w]*flux[fiber,w] sqrtwR[fiber,wave]
 
         #A=scipy.sparse.lil_matrix((nwave,nwave)).tocsr()
-        A=np.zeros((nwave,nwave))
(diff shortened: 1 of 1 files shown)


ziyaointl:
One question remain: should I make the same changes for compute_polynomial_times_sky and compute_non_uniform_sky?

sbailey:
Thanks!  I confirm that this runs much faster without changing the outputs, tested on exposure 20201215/00068065 cameras r0 and z0.  For the record, which night/expid/camera were you using for testing?  I didn't immediately find one that was 5m slow for testing.

Indeed it looks like the same flavor of changes would significantly help `compute_polynomial_times_sky` and `compute_non_uniform_sky`, but I'm going to be greedy and merge this PR to get the quick improvement now.  Please then make those additional changes and continue to test that the outputs are unchanged and open a new PR.

ziyaointl:
I was using 20200227/00052426 on camera z5 for this one.
I still need to do some digging to figure out which cameras/exposures/nights use either `compute_polynomial_times_sky` or `compute_non_uniform_sky`, but if you have an example handy that would be great as well.

PR #1184, opened 2021-03-12
MPI-ify stdstar fit

dmargala:
This PR adds an option to `desi_fit_stdstars` to use MPI for parallelism instead of multiprocessing. 

This could allow the stdstar fitting step in `desi_proc` to run with the same node/task/cpu configuration as previous steps in the pipeline (`srun -N 5 -n 160 -c 2 ...`). This will also be useful for a GPU-accelerated version of the stdstar fit (WIP).

The performance is about the same on Cori Haswell:

```
# run cmd using multiprocessing for parallelism
> time srun -n 1 -c 32 --cpu-bind=cores $CMD
...
real	1m3.543s
user	0m0.061s
sys	0m0.043s

# run cmd using MPI for parallelism
> time srun -n 16 -c 2 --cpu-bind=cores $CMD --mpi
...
real	0m57.977s
user	0m0.070s
sys	0m0.099s
```

Full test setup:

```
cd $SCRATCH
git clone git@github.com:dmargala/desispec.git
cd desispec
git checkout mpi-stdstars

salloc -C haswell -N 1 -q interactive -t 30 --exclusive 

source /global/common/software/desi/desi_environment.sh 21.2
module unload desispec
export PATH=$(pwd)/bin:$PATH

INDIR=/global/cfs/cdirs/desi/spectro/redux/cascades
OUTDIR=$SCRATCH
CMD="desi_fit_stdstars --delta-color 0.1 \
    --frames \
        $INDIR/exposures/20201214/00067781/frame-b4-00067781.fits \
        $INDIR/exposures/20201214/00067782/frame-b4-00067782.fits \
        $INDIR/exposures/20201214/00067783/frame-b4-00067783.fits \
        $INDIR/exposures/20201214/00067784/frame-b4-00067784.fits \
        $INDIR/exposures/20201214/00067781/frame-r4-00067781.fits \
        $INDIR/exposures/20201214/00067782/frame-r4-00067782.fits \
        $INDIR/exposures/20201214/00067783/frame-r4-00067783.fits \
        $INDIR/exposures/20201214/00067784/frame-r4-00067784.fits \
        $INDIR/exposures/20201214/00067781/frame-z4-00067781.fits \
        $INDIR/exposures/20201214/00067782/frame-z4-00067782.fits \
        $INDIR/exposures/20201214/00067783/frame-z4-00067783.fits \
        $INDIR/exposures/20201214/00067784/frame-z4-00067784.fits \
    --skymodels \
        $INDIR/exposures/20201214/00067781/sky-b4-00067781.fits \
        $INDIR/exposures/20201214/00067782/sky-b4-00067782.fits \
        $INDIR/exposures/20201214/00067783/sky-b4-00067783.fits \
        $INDIR/exposures/20201214/00067784/sky-b4-00067784.fits \
        $INDIR/exposures/20201214/00067781/sky-r4-00067781.fits \
        $INDIR/exposures/20201214/00067782/sky-r4-00067782.fits \
        $INDIR/exposures/20201214/00067783/sky-r4-00067783.fits \
        $INDIR/exposures/20201214/00067784/sky-r4-00067784.fits \
        $INDIR/exposures/20201214/00067781/sky-z4-00067781.fits \
        $INDIR/exposures/20201214/00067782/sky-z4-00067782.fits \
        $INDIR/exposures/20201214/00067783/sky-z4-00067783.fits \
        $INDIR/exposures/20201214/00067784/sky-z4-00067784.fits \
    --fiberflats \
        $INDIR/calibnight/20201214/fiberflatnight-b4-20201214.fits \
        $INDIR/calibnight/20201214/fiberflatnight-r4-20201214.fits \
        $INDIR/calibnight/20201214/fiberflatnight-z4-20201214.fits \
    --outfile $OUTDIR/stdstars-4-00067781.fits"
```

### DIFF
--- a/py/desispec/scripts/stdstars.py
+++ b/py/desispec/scripts/stdstars.py
@@ -43,6 +43,7 @@ def parse(options=None):
     parser.add_argument('--template-error', type = float, default = 0.1, required = False, help = 'fractional template error used in chi2 computation (about 0.1 for BOSS b1)')
     parser.add_argument('--maxstdstars', type=int, default=30, \
             help='Maximum number of stdstars to include')
+    parser.add_argument('--mpi', action='store_true', help='Use MPI')
 
     log = get_logger()
     args = None
@@ -140,7 +141,7 @@ def unextinct_gaia_mags(star_mags, unextincted_mags, ebv_sfd):
                  )*gaia_a0
             unextincted_mags['GAIA-'+band] = star_mags['GAIA-'+band] - dmag
 
-def main(args) :
+def main(args, comm=None) :
     """ finds the best models of all standard stars in the frame
     and normlize the model flux. Output is written to a file and will be called for calibration.
     """
@@ -148,7 +149,29 @@ def main(args) :
     log = get_logger()
 
     log.info("mag delta %s = %f (for the pre-selection of stellar models)"%(args.color,args.delta_color))
-    log.info('multiprocess parallelizing with {} processes'.format(args.ncpu))
+
+    if args.mpi or comm is not None:
+        from mpi4py import MPI
+        if comm is None:
+            comm = MPI.COMM_WORLD
+        size = comm.Get_size()
+        rank = comm.Get_rank()
+        if rank == 0:
+            log.info('mpi parallelizing with {} ranks'.format(size))
+    else:
+        comm = None
+        rank = 0
+        size = 1
+    
+    # disable multiprocess by forcing ncpu = 1 when using MPI
+    if comm is not None:
+        ncpu = 1
+        if rank == 0:
+            log.info('disabling multiprocess (forcing ncpu = 1)')
+
+    if ncpu > 1:
+        if rank == 0:
+            log.info('multiprocess parallelizing with {} processes'.format(ncpu))
 
     # READ DATA
     ############################################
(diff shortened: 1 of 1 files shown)


py/desispec/scripts/stdstars.py
```diff
@@ -468,19 +493,21 @@ def main(args) :
         model_filters["GAIA-" + band] = load_gaia_filter(band=band, dr=2)
 
     log.info("computing model mags for %s"%sorted(model_filters.keys()))
```
sbailey, line 495: please move this log.info into the "if rank == 0" block so that the log messages don't incorrectly imply that all ranks are computing model mags.

py/desispec/scripts/stdstars.py
```diff
@@ -637,26 +669,32 @@ def main(args) :
...
-    data['BLUE_SNR'] = snr['b'][fitted_stars]
-    data['RED_SNR']  = snr['r'][fitted_stars]
-    data['NIR_SNR']  = snr['z'][fitted_stars]
-    io.write_stdstar_models(args.outfile,normflux,stdwave,starfibers[fitted_stars],data)
+    # Now write the normalized flux for all best models to a file
+    if rank == 0:
+        fitted_stars = np.where(chi2dof != 0)[0]
+        if fitted_stars.size == 0 :
+            log.error("No star has been fit.")
+            sys.exit(12)
```
sbailey, line 686: Please update this logic so that if it needs to exit with an error, then that information is broadcast to all ranks and they all exit together.  As-is I think we'd end up with rank 0 exiting but the others hanging around.  Since there isn't any barrier after this that might be ok, but it is fragile.  In other code we've had problems with one rank exiting with an error but the other ranks not exiting until slurm kills the job for a timeout.

sbailey (changes requested):
Thanks.  This lays the groundwork for MPI parallelism of standard stars.  I put a comment inline about having all ranks exit if there is a problem instead of just rank 0, which should be addressed before merging.

For consideration: right now all ranks read the same input frame files, and the same template files N>>1 times.  That works ok at low parallelism and when CFS I/O is behaving, but could be problematic in the future.  Could you investigate having only one rank read + broadcast?  If too messy, that could be punted to the future since this still works as-is for both MPI and multiprocessing.

I also verified that the two parallelism modes produce the same output (other than timestamps and header versions).

Note: To get further MPI flexibility we'll need to push the parallelism down into match_templates, since the current parallelism (multiprocessing and MPI) maxes out at the number of standards, which could be as few as 10.  That can be a future PR, and @ziyaointl has expressed interest in looking into that.

dmargala:
Thanks for the suggestions. I just pushed changes to address your inline comments. 

I considered having one rank read + broadcast but quickly ran into serialization issues when broadcasting dictionary of Frame objects. I'm sure they're possible to overcome but it would involve a deeper investigation. I vote for punting on the IO optimization to keep this PR from getting too complicated.

sbailey (approved):

sbailey:
Thanks, looks good.  We can save MPI I/O optimization for a future PR.  It is odd that coveralls failed again, which makes me wonder if you are based on an old fork or if there is something misconfigured for submitting to coveralls from a forked PR instead of a branch PR, but I'm going to ignore that and merge, and address coveralls separately if it comes up again (it is working for other active PRs from recent branches).

PR #1186, opened 2021-03-15
Fix workflow bug that arises when dither is followed by non-dither

akremin:
This PR fixes the bug observed in cascades (the data was fixed, but by hand) where an "undithered" exposure was kept and flux calibrated with a subsequent science tile as opposed to being flux calibrated on its own. This was due to a simple logic change that arose when `LASTSTEP` was implemented. 

This pull request fixes the incorrect logic. It goes further to do other sanity checks on the list of science exposures to ensure they are meant to be run (`row['LASTSTEP'] != skysub`) and that the tiles are the same. If tiles are different, it processes the exposures from the most common tile. This should add robustness to bugs similar to the primary motivation of this fix where a single extra exposure sneaks in with an otherwise correct set.

This was tested using the `desi_run_night` code and the `desi_daily_proc_manager` code. I also tested with several contrived scenarios interactively. The code handled each correctly.

### DIFF
--- a/py/desispec/workflow/procfuncs.py
+++ b/py/desispec/workflow/procfuncs.py
@@ -802,7 +802,7 @@ def checkfor_and_submit_joint_job(ptable, arcs, flats, sciences, arcjob, flatjob
         internal_id, int, if no job is submitted, this is the same as the input, otherwise it is incremented upward from
                           from the input such that it represents the smallest unused ID.
     """
-    if lasttype == 'science':
+    if lasttype == 'science' and len(sciences) > 0:
         log = get_logger()
         skysubonly = np.array([sci['LASTSTEP'] == 'skysub' for sci in sciences])
         if np.all(skysubonly):
--- a/py/desispec/workflow/procfuncs.py
+++ b/py/desispec/workflow/procfuncs.py
@@ -826,6 +826,7 @@ def checkfor_and_submit_joint_job(ptable, arcs, flats, sciences, arcjob, flatjob
             log.warning("Identified more than one tile in a joint fitting request")
             log.info("Expid's: {}".format([row['EXPID'] for row in sciences]))
             log.info("Tileid's: {}".format(tiles))
+            log.info("Returning without joint fitting any of these exposures.")
             # most_common, nmost_common = counts.most_common()[0]
             # if most_common == -99:
             #     most_common, nmost_common = counts.most_common()[1]
--- a/py/desispec/workflow/procfuncs.py
+++ b/py/desispec/workflow/procfuncs.py
@@ -764,7 +764,7 @@ def make_joint_prow(prows, descriptor, internal_id):
     return joint_prow
 
 def checkfor_and_submit_joint_job(ptable, arcs, flats, sciences, arcjob, flatjob, \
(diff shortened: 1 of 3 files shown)


py/desispec/workflow/procfuncs.py
```diff
@@ -803,21 +802,53 @@ def checkfor_and_submit_joint_job(ptable, arcs, flats, sciences, arcjob, flatjob
...
+            log.info("Tileid's: {}".format(tiles))
+            most_common, nmost_common = counts.most_common()[0]
+            if most_common == -99:
+                most_common, nmost_common = counts.most_common()[1]
+            log.warning(f"Given multiple tiles to jointly fit: {counts}. Only processing the most common non-default" +
+                        f" tile: {most_common} with {nmost_common} exposures")
+            sciences = (np.array(sciences,dtype=object)[tiles == most_common]).tolist()
+            log.info("Tiles and exposure id's being submitted for joint fitting:")
+            log.info("Expid's: {}".format([row['EXPID'] for row in sciences]))
+            log.info("Tileid's: {}".format([row['TILEID'] for row in sciences]))
```
sbailey, line 837: Having some but not all LASTSTEP=skysub and having more than one tile in the group are both error conditions that shouldn't occur if everything else is working correctly, right?  Here you make it robust and proceed, but if this case really isn't supposed to occur, perhaps we should print a message and exit so that we can fix the underlying problem that led to it?

sbailey (commented):
Overall looks good, with one inline comment about whether it would be better to just fail early and stop instead of robustly proceeding and perhaps missing an underlying bookkeeping problem that could lead to other processing problems.

py/desispec/workflow/procfuncs.py
akremin: It's a great point and I thought about this when I was developing it. If you feel strongly I can remove it, but below is my thought process:

It's true that if we're sub-selecting from the list then something in the workflow logic went wrong. The goal of this robustness is that if it fails in a similar mechanistic way to the failure that spurred this PR -- but obviously through a different chain of logic issues -- it would still do the right thing in that circumstance. It's effectively an independent solution to the initial bug that is more general and could theoretically apply to a different unknown problem. 

This protects against bookkeeping issues where e.g. an exposure was added to the list accidentally or the joint fitting wasn't triggered and so two tiles worth of exposures were left in the list until the next triggering event. In either of those cases, this will perform a correct action and process at least one of the tiles. It may not process both if two tiles were lumped together, but it goes part-way and at least does one of them. It also throws up a log WARNING that is grep-able and provides a lot of output to work backward through its decision-making.

There could certainly be an unknown-unknown situation in which this generic catch-all would plod forward with running a subselection and that would be the incorrect thing to do. In that case, we'd have to delete those files and clean up by hand. The only difference between that situation and the situation created by not running at all is in the processing time used on the useless outputs and the effort of removing the bad output files. Recall that these post-extraction steps are quite quick from a processing perspective. 

Given that I can't imagine one of the unknown-unknown scenarios in which that would occur due to a breakdown in logic, whereas I can imagine failures that would be caught, I think the net advantage is to plod on.

PR #1189, opened 2021-03-17
Flux calibration: fiber acceptance correction as a function of seeing, focal plane position, source morphology

julienguy:
This PR provides corrections to the flux calibration to account for variations of the fiber acceptance with seeing, focal plane position (because of the change of plate scale), source morphology.

 - add two functions in `desispec.fiberfluxcorr`:
   -  `flat_to_psf_flux_correction`:  Multiplicative factor to apply to the flat-fielded spectroscopic flux of a fiber
    to calibrate the spectrum of a point source, given the current exposure seeing and change of platescale in the focal plane.
   - `psf_to_fiber_flux_correction`: Multiplicative factor to apply to the psf flux of a fiber to obtain the fiber flux, given the current exposure seeing, accounting for the shape/morphology of the source. The fiber flux is the flux one would collect for this object in a fiber of 1.5 arcsec diameter, for a 1 arcsec Gaussian seeing, FWHM (same definition as for the Legacy Surveys).
 - changes to `desispec.fluxcalibration.FluxCalib` object and the I/O functions in `desispec.io.fluxcalibration`: add `astropy.table.Table` attribute `self.fibercorr` to the class, save it in new HDU 'FIBERCORR'. 
 - changes to `desispec.fluxcalibration.compute_flux_calibration`: 
    -  use `flat_to_psf_flux_correction` to correct the flux of standard stars before averaging, apply it to the flux calibration vectors, and save the coefficient in `FluxCalib.fibercorr["FLAT_TO_PSF_FLUX"]` for the record.
    - use `psf_to_fiber_flux_correction` to compute the multiplicative factor to apply to the calibrated flux to get the fiber flux. The ratio **is not applied** to the calibration vector but is saved in `FluxCalib.fibercorr["PSF_TO_FIBER_FLUX"]`.
 - changes to  `desispec.fluxcalibration.apply_flux_calibration`: add a column to the fibermap, ` frame.fibermap["PSF_TO_FIBER_SPECFLUX"]=fluxcalib.fibercorr["PSF_TO_FIBER_FLUX"]`so that the user can compute the fiberflux for each object.
 - add utility functions in `desispec.magnitude` to compute broadband fluxes and magnitudes of potentially noisy spectra.
 - add option `--seeing-fwhm` to `desi_compute_fluxcalibration`

Validation test:
I recomputed the flux calibration of all the cframes of the night 20201216 assuming a seeing of 0.89". I then computed the broadband fluxes in R-band for all the spectra (labeled SPECFLUX_R in the following figures) and the fiber flux SPECFIBERFLUX_R =  SPECFLUX_R * fibermap["PSF_TO_FIBER_SPECFLUX"]. 
The following figures show the ratio SPECFLUX_R/FLUX_R (which is only valid for point sources) and SPECFIBERFLUX_R/FIBERTOTFLUX_R (which should be valid for all sources)  as  a function of various parameters.

ratio-vs-fiberfrac

ratio-vs-fp-radius

ratio-vs-offset

ratio-vs-rshape

### DIFF
--- a/py/desispec/fiberfluxcorr.py
+++ b/py/desispec/fiberfluxcorr.py
@@ -12,10 +12,16 @@ import numpy as np
 from desimodel.fastfiberacceptance import FastFiberAcceptance
 from desimodel.io import load_platescale
 
-def point_source_fiber_flux_correction(fibermap,exposure_seeing_fwhm=1.1) :
+def flat_to_psf_flux_correction(fibermap,exposure_seeing_fwhm=1.1) :
     """
-    Multiplicative factor to apply to the flatfielded spectroscopic flux of a fiber
-    for a point source, given the current exposure seeing
+    Multiplicative factor to apply to the flat-fielded spectroscopic flux of a fiber
+    to calibrate the spectrum of a point source, given the current exposure seeing
+
+    Args:
+      fibermap: fibermap of frame, astropy.table.Table
+      exposure_seeing_fwhm: seeing FWHM in arcsec
+
+    Returns: 1D numpy array with correction factor to apply to fiber fielded fluxes, valid for point sources.
     """
     #- Compute point source flux correction and fiber flux correction
     fa = FastFiberAcceptance()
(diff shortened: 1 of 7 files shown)


sbailey:
Thinking/commenting aloud:

Code looks good, validation tests on the flux ratios look good.

Requires desimodel >= 0.14.1 to get desimodel.fastfiberacceptance (moved from specsim).

Redshifts are a small positive gain compared to a test run from the fiber_crosstalk PR #1138, but they aren't completely unchanged as one would expect from a strictly achromatic multiplicative factor.  I think that is due to other changes that occurred in the meantime (PR #1157 simplify fluxcalib and #1160 interpolating over ISM Na and Ca) and I haven't tracked down exactly where this branched off compared to what is currently in master, so I haven't succeeded in performing the null test that the as-run spectra are achromatically changed with no changes at all to redshifts.  I'm not sure if that is worth pursuing as a cross check, considering that the rest looks good.

Itemizing some data model changes to consider the impact: 
* cframe, spectra, coadd, zbest FIBERMAP: new column PSF_TO_FIBER_SPECFLUX
* cframe, spectra, coadd SCORES: new columns TSNR2_LYA_B, TSNR2_LYA_R, TSNR2_LYA_Z, TSNR2_LYA (coadd-only)
  * I don't actually see in the code where the these come in; is it a side effect of it being automagically discovered in desimodel/trunk/data/tsnr/tsnr-ensemble-lya.fits ?

In the past adding new columns has broken combining data from the previous release + daily (can't trivially stack the tables anymore).  Some workarounds are in place for desi_coadd_spectra, but we keep breaking that code with special cases so let's check carefully and revisit the potential impact of those data model changes on Monday with a fresh collective brain.

julienguy:
Yes the addition of TSNR_LYA comes from the new ensemble file desimodel/trunk/data/tsnr/tsnr-ensemble-lya.fits.
I don't think you need to aim at a large redshift fit test for this PR because it is applying only an achromatic correction to the fluxes, so it should be totally immune for redrock (but generic pipeline tests are still very useful to test the robustness to missing data, bad fibermap entries ...).

michaelJwilson:
Specifically, the automagick is: 
https://github.com/desihub/desispec/blob/f3895a1ba9b5a36ba3778040143f2210daf01a99/py/desispec/tsnr.py#L41

The only relevant addition to the code base is the notebook that creates the ensemble file. 
https://github.com/desihub/desispec/blob/master/doc/nb/Lya-tsnr-signal.ipynb

It's not clear to me whether the pipeline only uses tagged releases of desimodel and how often this is updated.

sbailey:
@michaelJwilson the daily pipeline runs out of desimodel svn trunk which is updated nightly ~1am, so the addition of desimodel/data/tsnr/tsnr-ensemble-lya.fits (and thus SCORES TSNR2_LYA*) was automatically picked up in the middle of the processing of NIGHT 20210318 between exposures 80954 and 80955, which fortunately were different tiles.

The change of the data model for the SCORES table breaks the ability to merge cframes generated before/after into the same spectra file, and thus breaks multi-night redshifts for those tiles without re-processing the earlier nights.  But that change is unrelated to this PR and already in the processed data so this is more of a heads up @akremin, since we may trip across it during sv2 daily processing that may pickup sv1 tiles too.

The pipeline handles the FIBERMAP additional columns from this PR somewhat more gracefully by just propagating forward the subset of columns found in all files, meaning that for SV1 tiles that we had previously observed, new daily observations of them will not get PSF_TO_FIBER_SPECFLUX in the output until we re-process earlier exposures too (e.g. with Denali).

Robustness check: I have tested 3048 frames from 103 exposures on 10 nights and didn't encounter any corner-case crashing problems from this PR, so that looks good.  Merging now.

michaelJwilson:
Ah, apologies.  In my defence, I ok'ed the svn commit with Julien.  It might be straightforward to update coaddition to assume zero TSNR_X if not present and add appropriately.  I don't know cframes -> spectra well enough to know if a similar approach would work there.

sbailey:
1. My commentary is clarifying the situation and the side effects, not distributing blame. We did approve that addition, and we've always asserted the right to break the daily prods without notice, and this falls within that.  ok.  The main thing I'm trying to get ahead of is warning users of consequences before they trip on them and report back to us as a surprise.
2. It would be "straightforward to update coaddition to assume zero TSNR_X if not present" or otherwise "drop TSNR2_X if present in some files but not others", but it's not clear to me that either is what we actually would want, vs. just investing developer effort in moving on with sv2 and Denali so that we have complete and meaningful TSNR2_LYA in all files.

Issue #1177, opened 2021-03-08
TSNR: add a Lya SNR 


julienguy:
Add a Lya SNR , based on some assumed n(z) for Lya QSOs.

michaelJwilson:
Solved by #1190.

PR #1190, opened 2021-03-18
Lyman-alpha signal template for tsnr

michaelJwilson:
Adds a dedicated notebook under docs/nb for deriving a template Lya signal.  The output of which has been committed to desimodel/data by svn.

### DIFF
--- a/doc/nb/Lya-tsnr-signal.ipynb
+++ b/doc/nb/Lya-tsnr-signal.ipynb
@@ -1,0 +1,1785 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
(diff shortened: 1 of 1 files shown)


michaelJwilson:
Also solves #1177

julienguy:
Tested successfully. Thanks. 

I merge this pull request and will then : 
 - calibrate an LYA_EFFTIME
 - save it in the tsnr-exposures table
 - provide a separate routine to compute LYA_EFFTIME from the TSNR2_LYA per target that could be used in MTL to decide on the number of reobservations for LYA quasars.

Issue #1179, opened 2021-03-09
spectra.select does not work with cascades OR cascades fibermap does not have NIGHT key


AlexGKim:
At issue is that the cascades fibermap does not have a NIGHT key.,

```
from desispec.io import read_spectra, write_spectra
filename = "/global/project/projectdirs/desi/spectro/redux/cascades/tiles/80606/20201215/coadd-1-80606-20201215.fits"
pspectra = read_spectra(filename)
print('NIGHT' in pspectra.fibermap.keys())
filename = "/global/project/projectdirs/desi/spectro/redux/daily/tiles/80606/20201215/coadd-1-80606-20201215.fits"
pspectra = read_spectra(filename)
print('NIGHT' in pspectra.fibermap.keys())

False
True
```

sbailey:
Indeed, we dropped "NIGHT" from the coadds because coadds in general could be across multiple nights so having a single "NIGHT" entry for the target doesn't apply.  However, this broke `Spectrum.select()` for coadds, even if selecting on keys that aren't NIGHT:
```
from desispec.io import read_spectra
filename = "/global/cfs/cdirs/desi/spectro/redux/cascades/tiles/80606/20201215/coadd-1-80606-20201215.fits"
pspectra = read_spectra(filename)
pspectra.select(fibers=[520,521])
...
KeyError                                  Traceback (most recent call last)
<ipython-input-1-663839648f44> in <module>
      2 filename = "/global/cfs/cdirs/desi/spectro/redux/cascades/tiles/80606/20201215/coadd-1-80606-20201215.fits"
      3 pspectra = read_spectra(filename)
----> 4 pspectra.select(fibers=[520,521])

/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/0.39.3/lib/python3.8/site-packages/desispec-0.39.3-py3.8.egg/desispec/spectra.py in select(self, nights, bands, targets, fibers, invert)
    275         keep_nights = None
    276         if nights is None:
--> 277             keep_nights = [ True for x in self.fibermap["NIGHT"] ]
    278         else:
    279             keep_nights = [ (x in nights) for x in self.fibermap["NIGHT"] ]

/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/conda/lib/python3.8/site-packages/astropy/table/table.py in __getitem__(self, item)
   1600     def __getitem__(self, item):
   1601         if isinstance(item, str):
-> 1602             return self.columns[item]
[...]

PR #1195, opened 2021-03-21
handle missing NIGHT in coadded fibermap

moustakas:
Should be a non-controversial merge: in a coadded fibermap `NIGHT` is no longer (necessarily) part of the data model, but the  `spectra.Spectra.select` method was still looking for it.


sbailey:
Thanks.  This issue fixes issue #1179 originally reported by @AlexGKim .  I tested it with the failing case in that issue and it now works.  Merging now.

moustakas:
@armengau this looks fine to me.

Issue #1193, opened 2021-03-19
specscore.compute_coadd_scores fails for TSNR2_XYZ


AlexGKim:
This spectra fits file

/global/project/projectdirs/desi/spectro/redux/daily/tiles/80721/20210305/coadd-1-80721-20210305.fits

includes spectra.scores names

TSNR2_ELG_B
TSNR2_BGS_B
TSNR2_QSO_B
TSNR2_LRG_B
TSNR2_ELG_R
TSNR2_BGS_R
TSNR2_QSO_R
TSNR2_LRG_R
TSNR2_ELG_Z
TSNR2_BGS_Z
TSNR2_QSO_Z
TSNR2_LRG_Z
TSNR2_ELG
TSNR2_BGS
TSNR2_QSO
TSNR2_LRG

The latter 4 fail in line 72 of specscore.py, because the band is expected.

sbailey:
Thanks for reporting @AlexGKim.  To facilitate testing and fixing, specifically what command were you running that triggered specscore.py line 72 on this file?

@julienguy @akremin @sbailey we collectively need to be better at testing and announcing data model changes.

AlexGKim:
The error occurred when running `coadd_cameras(spectra)`.

sbailey:
Specifically the error occurs when trying to coadd a coadd:
```
import desispec.io
import desispec.coaddition
spectra = desispec.io.read_spectra('spectra-1-80721-20210305.fits')
coadd0 = desispec.io.read_spectra('coadd-1-80721-20210305.fits')
coadd1 = desispec.coaddition.coadd_cameras(spectra)   # WORKS
coadd2 = desispec.coaddition.coadd_cameras(coadd1)    # FAILS
coadd3 = desispec.coaddition.coadd_cameras(coadd0)    # FAILS
```
Traceback is
```
...
/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/specscore.py in compute_coadd_scores(coadd, specscores, update_coadd)
     70         for colname in specscores.dtype.names:
     71             if colname.startswith('TSNR2_'):
---> 72                 _, targtype, band = colname.split('_')
     73                 scores[colname] = np.zeros(num_targets)
     74                 comments[colname] = f'{targtype} {band} template (S/N)^2'

ValueError: not enough values to unpack (expected 3, got 2)
```

michaelJwilson:
[...]

PR #1197, opened 2021-03-22
TSNR camera coadd fix

michaelJwilson:
Addresses #1193.  Runs through the use case example provided there and passes other vetting on values. 

For future consideration: rename TSNR_QSO to TSNR_QSO_BRZ etc. to mirror nomenclature for flux etc.


sbailey:
Thanks for the fix, looks good.  I'm open to discussing `TSNR2_XYZ` vs. `TSNR2_XYZ_BRZ` but let's merge this bugfix now without changing the datamodel, and independently decide if we want to change the column name (and let's decide that before running Denali, i.e. soon).

PR #1205, opened 2021-03-25
Add support for using GPUs in standard star fit

dmargala:
This PR adds support for using GPUs in the standard star fitting step of the pipeline. 

The main change is to move the median filter operation in `desispec.fluxcalibration.applySmoothingFilter` to the GPU. Here is a profile of the CPU version of the standard star fit that motivates this:

profile

The GPU code is disabled by default and must be explicitly enabled by setting `desispec.fluxcalibration.use_gpu = True`. After importing, `desispec.fluxcalibration.gpu_available` is set to `True` if the `cupy` module is installed and able to detect a GPU.

I also made some changes to `desi_proc_joint_fit` for testing but those are still WIP. I'm open to feedback on how to integrate those changes with the existing pipeline setup.

```
module load python cuda/11.1.1 gcc openmpi
source activate gpu-stdstars-dgx
export DESI_SPECTRO_CALIB=/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk
export DESI_BASIS_TEMPLATES=/global/cfs/cdirs/desi/spectro/templates/basis_templates/v3.2
export DESI_SPECTRO_DATA=/global/cfs/cdirs/desi/spectro/data
export DESI_SPECTRO_REDUX=/global/cfs/cdirs/desi/spectro/redux
export SPECPROD=$USER/$SLURM_JOBID
INDIR=/global/cfs/cdirs/desi/spectro/redux/cascades
OUTDIR=$DESI_SPECTRO_REDUX/$SPECPROD
mkdir -p $OUTDIR

NIGHT=20201214
EXPIDS=67710,67711,67712,67713

# setup output directories and symlinks to input files
ln -fs $INDIR/calibnight $OUTDIR/calibnight
for e in $(echo $EXPIDS | sed "s/,/ /g")
do
    # zero pad expid
    ze=$(printf "%08d" $e)
    # create directory for exposure outputs
    mkdir -p $OUTDIR/exposures/$NIGHT/$ze
    # create symlinks to reference sky files
    for file in $INDIR/exposures/$NIGHT/$ze/sky-*-$ze.fits
    do 
        ln -fs $file $OUTDIR/exposures/$NIGHT/$ze/${file##*/}
    done
    # create symlinks to reference frame files
    for file in $INDIR/exposures/$NIGHT/$ze/frame-*-$ze.fits; 
    do 
        ln -fs $file $OUTDIR/exposures/$NIGHT/$ze/${file##*/}
    done
done

time srun -n 32 -c 2 --cpu-bind=cores ./mps-wrapper.sh ./cycle_gpus.sh desi_proc_joint_fit --traceshift --obstype science --cameras a0123456789 -n $NIGHT -e $EXPIDS --mpi
...
real	0m56.789s
user	0m0.049s
sys	0m0.101s
```

... and from the timing summary: `stdstarfit.duration.max = 32.922` (a good chunk of the runtime is from mps-wrapper, imports, and mpi connect)

### DIFF
--- a/bin/desi_proc_joint_fit
+++ b/bin/desi_proc_joint_fit
@@ -23,7 +23,6 @@ from desispec.sky import subtract_sky
 from desispec.util import runcmd, mpi_count_failures
 import desispec.scripts.extract
 import desispec.scripts.specex
-import desispec.scripts.stdstars
 
 from desitarget.targetmask import desi_mask
 
@@ -401,23 +400,10 @@ if args.obstype in ['SCIENCE']:
     starmodels = os.path.join(
         os.getenv('DESI_BASIS_TEMPLATES'), 'stdstar_templates_v2.2.fits')
 
-    # - Spread ranks across 10 spectrographs
-    mpi_stdstarfit = True
-    if comm is not None and mpi_stdstarfit:
-        group_size = max(1, size // 10)
-        group = rank // group_size
-        num_groups = (size + group_size - 1) // group_size
-        if rank == 0:
-            log.info(f'Splitting {size} ranks into {num_groups} groups of {group_size}')
-        comm_group = comm.Split(color=group)
-    else:
-        comm_group = None
-        group = rank
-        num_groups = size
-
     # - Fit stdstars per spectrograph (not per-camera)
     spectro_nums = sorted(framefiles.keys())
-    for i in range(group, len(spectro_nums), num_groups):
+    ## for sp in spectro_nums[rank::size]:
+    for i in range(rank, len(spectro_nums), size):
         sp = spectro_nums[i]
         # - NOTE: Saving the joint fit file with only the name of the first exposure
         stdfile = findfile('stdstars', args.night, args.expids[0], spectrograph=sp)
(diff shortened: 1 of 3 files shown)


sbailey:
gpu/cupy acceleration of standard stars looks good, but I find the "use_gpu" logic to be a bit circuitous.  The fluxcalibration module has `use_gpu=False` by default even if it has determined that cupy is installed and a GPU is available, but then the stdstars script overrides that default to make `use_gpu=True` the default instead, i.e. whether the GPU is "default" depends upon the code path to get there.  Let's make the module-level default to use the gpu if cupy is installed and a GPU is available, and that any other code would override that (opt-out) if `--ignoregpu`.  (Minor: my brain keeps reading "ignoregpu" as "ig no reg pu"; let's use "--ignore-gpu" instead).

Thinking aloud: it might be useful to make a desispec.gpu convenience module for centralizing the logic for whether cupy+GPU is available and whether it should be used.  This can be done later after we settle what use pattern is most convenient.

Regarding desi_proc_joint_fit: let's spin that off to a PR separate from the GPU addition.  I like the idea of splitting the communicator at the desi_proc_joint_fit level and passing in sub-communicators to stdstar fitting, but the default wrapper by the pipeline would also need to be updated to make use of that, since the default batch script wrapper calls it with non-optimal ranks and runs 3x slower.  Trying to find what the optimal srun options would be, I started to get hammered by every rank reading the same files, and we need to test on MWS tiles with lots of stars to make sure we don't blow memory.  I.e. I think this is a good direction but it needs to be integrated in the pipeline operations as a whole but I don't want that to hold up getting the initial GPU support in.

dmargala:
Thanks for the feedback. I reverted the changes to desi_proc_join_fit and removed the extra layer of logic around `use_gpu`.

Are the coverage checks failing because I submitted this PR from a fork?

sbailey:
Thanks.  I'm not sure if the coverage submission failure is due to this being a fork-PR instead of a branch-PR, but it appears unrelated to these changes.  Merging now.

PR #1209, opened 2021-03-26
Use sparse matrix types in compute_polynomial_times_sky

ziyaointl:
This is similar to #1182

`fitsdiff` shows no difference in output other than the checksums and timestamps.
For 20200227/00052426 on camera z5, this change brings run time down from ~6m20s to ~1m40s.

P.S. I accidentally pushed this commit to master and then reverted it. Apologies! While I will pay extra attention before I push in the future, let me know if another solution (e.g. force-push) is preferred should this happen again.


sbailey:
Looks good, thanks!  Merging now.

PR #1212, opened 2021-03-26
add LASTNIGHT column to tile completeness table

sbailey:
This PR adds a LASTNIGHT column to the tile completeness table (e.g. daily/tiles.csv), containing the last YEARMMDD on which each tile was observed to simplify looking up `tiles/cumulative/TILEID/YEARMMDD/*-thruYEARMMDD.*`

@julienguy


julienguy (approved):
thanks (tested)

Issue #1162, opened 2021-03-01
add input provenance to stdstar files


sbailey:
stdstar files don't include header keywords for what input exposures contributed to the stdstar fit.  That information is contained in the logs, scripts, and the processing tables, but should also be in the stdstar headers themselves.

moustakas:
Would be great to store the derived scale factors in the headers, too.

segasai:
It would also be nice (given that the stdstars files are small) to also have ra,dec or some object identifiers in the tables as well. ATM it's not easy to scrape stdstars files and systematically analyze them. The only columns there currently are ('LOGG','TEFF','FEH','CHI2DOF','REDSHIFT','DATA_GAIA-BP-RP','MODEL_GAIA-BP-RP','BLUE_SNR','RED_SNR','NIR_SNR')

PR #1213, opened 2021-03-27
improve stdstar provenance

sbailey:
This PR improves standard star provenance by including 2 additional HDUs in the stdstars*.fits files:
  * FIBERMAP: fibermap entries for the standard stars used in the fit, not including the columns like DELTA_X/Y, FIBER_X/Y that apply to specific exposures rather than the input targets (see note below).
  * INPUT_FRAMES: a table with columns NIGHT, EXPID, CAMERA indicating which input frames were used for the joint standard star fit.

And example output file is in /global/cfs/cdirs/desi/users/sjbailey/dev/stdstars/stdstars-2-00081899.fits, to be compared with the original daily run in /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20210324/00081899/stdstars-2-00081899.fits .  This was run with
```
cd /global/cfs/cdirs/desi/spectro/redux/daily/
desi_fit_stdstars --delta-color 0.1 \
  --frames exposures/20210324/00081899/frame-?2-00081899.fits exposures/20210324/00081900/frame-?2-00081900.fits \
  --skymodels exposures/20210324/00081899/sky-?2-00081899.fits exposures/20210324/00081900/sky-?2-00081900.fits \
  --fiberflats calibnight/20210324/fiberflatnight-?2-20210324.fits \
  --outfile /global/cfs/cdirs/desi/users/$USER/dev/stdstars/stdstars-2-00081899.fits
```

Notes:
* the "FIBERMAP" HDU is actually the same as the original input FIBERASSIGN columns, though it derives from the fibermap of an input frame minus the exposure-specific columns.  I'm ambivalent about whether this should be called "FIBERMAP" or "FIBERASSIGN", so if someone feels strongly either way feel free to comment.
* this update does not cache the per-exposure information like DELTA_X/Y which could potentially be useful for some debugging studies, but I'm not sure they are sufficiently useful to be worth yet another HDU in this file.  If anyone disagrees, please suggest exactly what table you would want.

@moustakas @segasai does this address what you would want?

@moustakas upon typing that, I realize this doesn't include "store the derived scale factors in the headers" you requested in #1162.  Are you referring to the overall r-band normalization, e.g. logged as 
```
INFO:stdstars.py:674:main: scaling R mag 6.539 to 18.905 using scale 1.130817831075739e-05
```
That seems reasonable to add to the METADATA HDU, but since that is a change to an existing HDU rather than the addition of new HDUs, I wanted to check that I'm adding the right value that you are requesting.

### DIFF
--- a/py/desispec/io/fluxcalibration.py
+++ b/py/desispec/io/fluxcalibration.py
@@ -18,7 +18,8 @@ from desiutil.io import encode_table
 from .util import fitsheader, native_endian, makepath
 from . import iotime
 
-def write_stdstar_models(norm_modelfile,normalizedFlux,wave,fibers,data,header=None):
+def write_stdstar_models(norm_modelfile, normalizedFlux, wave, fibers, data,
+        fibermap, input_exposures, header=None):
     """Writes the normalized flux for the best models.
 
     Args:
@@ -27,6 +28,8 @@ def write_stdstar_models(norm_modelfile,normalizedFlux,wave,fibers,data,header=N
         wave : 1D array of wavelengths[nwave] in Angstroms
         fibers : 1D array of fiberids for these spectra
         data : meta data table about which templates best fit
+        fibermap : fibermaps rows for the input standard stars
+        input_exposures : Table with NIGHT, EXPID, CAMERA of input frames used
     """
     log = get_logger()
     hdr = fitsheader(header)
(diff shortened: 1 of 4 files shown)


sbailey:
I see I broke the standard star tests due to the updated HDUs needed as input for `desispec.io.write_stdstar_models`.  I will fix that later; I'm pretty sure the problem is with the test, not desi_fit_stdstars itself (which I did test).

segasai:
I've looked at the file (not the patch), and it looks good to me

PR #1215, opened 2021-03-29
Add KNL hooks for current workflow

sbailey:
This PR adds hooks for desi_proc to use KNL nodes, with the configuration parameters for different systems kept in py/desispec/data/batch_config.yaml.  Default is to use cori-haswell, but `--system-name cori-knl` can override that.  I tested this with cori-haswell and cori-knl, with placeholders for dirac, perlmutter-cpu, and perlmutter-gpu, but I'm sure those will require additional updates to work for real.

desi_proc on an individual exposure end-to-end takes 17 minutes on 5 cori-haswell nodes, or 53 minutes on 3 cori-knl nodes.    Although that is much longer wallclock, at the current charge factors that is only 8% more expensive on KNL.

I'm opening this for review by @akremin, but noting a few items left to do:
  * test with individual prestdstar / joint stdstar / poststdstar steps
  * add --system-name option to desi_dailyproc and desi_dailyproc_manager
  * check stdstar rank allocation

Note: I also updated to allow 4 stdstar ranks per node (with additional parallelism fanout to multiprocessing) for better matching with the 3 KNL nodes.  The original 2 per node was driven by early CMX data with mostly standards star targets that was blowing memory and taking forever, but I think we throttle that now by taking the best N stars, but I want to check this on a tile with lots of stars.

### DIFF
--- a/bin/desi_proc
+++ b/bin/desi_proc
@@ -129,7 +129,8 @@ if args.batch:
         #    jobdesc = 'science'
     scriptfile = create_desi_proc_batch_script(night=args.night, exp=args.expid, cameras=args.cameras,\
                                                jobdesc=jobdesc, queue=args.queue, runtime=args.runtime,\
-                                               batch_opts=args.batch_opts, timingfile=args.timingfile)
+                                               batch_opts=args.batch_opts, timingfile=args.timingfile,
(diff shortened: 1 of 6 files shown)


sbailey:
superseded by #1220, closing.

PR #1220, opened 2021-03-31
workflow hooks for KNL

sbailey:
This PR replaces #1215, adding `--system-name` support for `desi_run_night` and `desi_tile_redshifts`, and resolving merge conflicts due to PR #1217 (thus the new PR with a newly rebased "knl2" branch).

@akremin please review.

### DIFF
--- a/bin/desi_run_night
+++ b/bin/desi_run_night
@@ -23,6 +23,8 @@ def parse_args():  # options=None):
                         help="The queue to submit jobs to. Default is realtime.")
     parser.add_argument("-r", "--reservation", type=str, required=False, default=None,
                         help="The reservation to submit jobs to. If None, it is not submitted to a reservation.")
+    parser.add_argument("--system-name", type=str, required=False, default=None,
+                        help="Batch system name, e.g. cori-haswell, cori-knl, perlmutter-gpu, ...")
     parser.add_argument("--exp-table-path", type=str, required=False, default=None,
                         help="Directory name where the output exposure table should be saved.")
     parser.add_argument("--proc-table-path", type=str, required=False, default=None,
--- a/bin/desi_tile_redshifts
+++ b/bin/desi_tile_redshifts
@@ -416,14 +416,14 @@ for tileid in tileids:
         batchscript, batcherr = batch_tile_redshifts(
             tileid, exptable[tilerows], args.group, submit=submit,
             queue=args.batch_queue, reservation=args.batch_reservation,
-            dependency=args.batch_dependency, system_name=args.system_name,
+            dependency=args.batch_dependency, system_name=args.system_name
             )
     else:
         for i in range(len(exptable[tilerows])):
             batchscript, batcherr = batch_tile_redshifts(
                 tileid, exptable[tilerows][i:i+1], args.group, submit=submit,
                 queue=args.batch_queue, reservation=args.batch_reservation,
-                dependency=args.batch_dependency, system_name=args.system_name,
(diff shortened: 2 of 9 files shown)


bin/desi_tile_redshifts
```diff
@@ -393,14 +416,14 @@ for tileid in tileids:
         batchscript, batcherr = batch_tile_redshifts(
             tileid, exptable[tilerows], args.group, submit=submit,
             queue=args.batch_queue, reservation=args.batch_reservation,
-            dependency=args.batch_dependency
+            dependency=args.batch_dependency, system_name=args.system_name,
```
akremin, line 419: Trailing comma here. Copy paste relic perhaps?

bin/desi_tile_redshifts
```diff
@@ -393,14 +416,14 @@ for tileid in tileids:
...
             )
     else:
         for i in range(len(exptable[tilerows])):
             batchscript, batcherr = batch_tile_redshifts(
                 tileid, exptable[tilerows][i:i+1], args.group, submit=submit,
                 queue=args.batch_queue, reservation=args.batch_reservation,
-                dependency=args.batch_dependency
+                dependency=args.batch_dependency, system_name=args.system_name,
```
akremin, line 426: Trailing comma here

bin/desi_run_night
```diff
@@ -23,6 +23,8 @@ def parse_args():  # options=None):
                         help="The queue to submit jobs to. Default is realtime.")
     parser.add_argument("-r", "--reservation", type=str, required=False, default=None,
                         help="The reservation to submit jobs to. If None, it is not submitted to a reservation.")
+    parser.add_argument("--system-name", type=str, required=False, default=None,
```
akremin, line 26: Can we add a default here of cori-haswell?

bin/desi_tile_redshifts
```diff
@@ -57,6 +57,8 @@ p.add_argument("--batch-reservation", type=str,
         help="batch reservation name")
 p.add_argument("--batch-dependency", type=str,
         help="job dependencies passed to sbatch --dependency")
+p.add_argument("--system-name", type=str,
```
akremin, line 60: Can we add a default here of cori-haswell?

akremin (commented):
I'll note that I did not do any testing of this. I'll trust that it works properly unless you think it needs to be run through its paces.

bin/desi_tile_redshifts
sbailey: cleaned up comma syntax (what was there was legal, but not necessarily intended)

bin/desi_run_night
akremin: In my review I missed the fact that None is set to an auto-detected default by the new `batch.get_config()` via a call to `batch.default_system()`. That works for my needs so I'll retract this comment

sbailey: Default of `None` will propagate through to `desispec.workflow.batch.default_system()` which defaults of 'cori-haswell' if on cori, 'dirac' if on hpcs/dirac, and 'perlmutter-gpu' if on perlmutter (guessing that NERSC will set $NERSC_HOST=perlmutter...).  We chatted offline and agreed this is ok.

akremin (approved):
This all looks good assuming that the docstring failure is under the hood and not part of this update that had nothing to do with docstrings.

sbailey:
Cocs build locally, unit test problems are some intersphinx thing impacting multiple repos.  merging for now to prepare a tag for 21.3 with or without intersphinx.

PR #1242, opened 2021-04-15
wrap desi_proc in a function

dmargala:
This PR wraps the `desi_proc` script in a function `desispec.scripts.proc.main(...)`. The function has an optional `comm` argument that allows the caller to pass an MPI communicator and skip the communicator setup that's provided by `desispec.workflow.desi_proc_funcs.assign_mpi` by default. This could allow a caller to split a communicator into sub-communicators and make separate calls to "desi_proc" for separate exposures.

The diff on github looks scarier than the actual number of changes made in this PR. I recommend using:

```bash
git checkout wrap-desi-proc
git diff -w refs/heads/master:bin/desi_proc py/desispec/scripts/proc.py
```

Perhaps the only surprising modifications are the ones related to the `import desiutil.timer` line. Without those changes I was seeing:

```
  File "/global/homes/d/dmargala/desiproc-refactor/desispec/bin/desi_proc", line 7, in <module>
    proc.main(args)
  File "/global/homes/d/dmargala/desiproc-refactor/desispec/py/desispec/scripts/proc.py", line 83, in main
    timer = desiutil.timer.Timer(silent=(rank>0))
UnboundLocalError: local variable 'desiutil' referenced before assignment
```
If there's a better way to solve that error, let me know.

I tested this on a single Cori Haswell node at NERSC:

```
salloc -C haswell -N 1 -q interactive -t 120
source /global/cfs/cdirs/desi/software/desi_environment.sh 21.3
module unload desispec
export PATH=$(pwd)/bin:$PATH
export DESI_SPECTRO_CALIB=/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk
export DESI_BASIS_TEMPLATES=/global/cfs/cdirs/desi/spectro/templates/basis_templates/v3.2
export OMP_NUM_THREADS=1

srun -n 20 -c 2 python bin/desi_proc_wrapper_test
```

Contents of `bin/desi_proc_wrapper_test`:

```python
#!/usr/bin/env python

import desispec.scripts.proc as proc

from mpi4py import MPI
comm = MPI.COMM_WORLD

night = 20201218
expids = [68657, 68658, 68659]
for expid in expids:

    cmd = 'desi_proc --traceshift --nostdstarfit --nofluxcalib'
    cmd += f' --night {night}'
    cmd += f' --expid {expid}'
    cmd += f' --cameras r0'

    cmdargs = cmd.split()[1:]
    args = proc.parse(cmdargs)

    proc.main(args, comm)
```


sbailey:
Looks good, thanks.  I dug into the "local variable 'desiutil' referenced before assignment" issue because it seemed so bizarre and I wanted to understand the root cause.  It turns out to be a side effect of importing `desiutil.iers` later within the `main` code instead of with the standard imports at the top.  I still don't understand how an import *after* the code in question could confuse python local variables, but I've at least been able to isolate it into a test case to submit to desiutil.  To avoid other unpleasant side effects in the future I moved the `import desiutil.iers` into the top block of imports, and cosmetically changes the `desiutil.timer` calls back to the originally full path for slightly more clarity on where `Timer` and `compute_stats` are coming from.

I tested standard `desi_proc` usage with processing an exposure end-to-end and it worked as expected.  Will merge as soon as tests finish running.

PR #1244, opened 2021-04-16
wrap desi_proc_joint_fit in a function

dmargala:
This PR wraps the `desi_proc_joint_fit` script in a function `desispec.scripts.proc_joint_fit.main(...)`. This is similar to #1242 . I used the same `import desiutil.iers` "fix" from #1242 .

I also took the liberty of fixing two minor issues that I noticed:
 * removed a duplicate instantiation of `timer`. 
 * added missing parens on `time.asctime` in final log message.


sbailey:

PR #1248, opened 2021-04-18
Add MPI to match_templates

ziyaointl:
This work is a continuation of #1184. In addition to parallelizing over standard stars, this PR adds the capability for `match_templates` to utilize multiple MPI ranks as well. This means if there are enough ranks each standard star now gets more than 1 rank at its disposal.

### Correctness
Both multiprocessing and mpi invocations of this code passed `fitsdiff` manual inspection for 20201214/00067765.

### Performance
There seems to be some modest improvement running on a single haswell node. I did not test how well the code scales on multiple nodes because I didn't fix the "file reads performed by each rank" problem yet.

#### MPI
```
real    0m34.602s
user    0m0.082s
sys     0m0.120s

real    0m35.900s
user    0m0.110s
sys     0m0.125s

real    0m38.513s
user    0m0.104s
sys     0m0.131s

real    0m35.980s
user    0m0.090s
sys     0m0.141s

time srun -n 32 -c 2 --cpu_bind=cores desi_fit_stdstars --mpi --frames /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/frame-b5-00067765.fits /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/frame-r5-00067765.fits /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/frame-z5-00067765.fits --skymodels /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/sky-b5-00067765.fits /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/sky-r5-00067765.fits /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/sky-z5-00067765.fits --fiberflats /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm9/fiberflatnight-b5-20201214.fits /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm9/fiberflatnight-r5-20201214.fits /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm9/fiberflatnight-z5-20201214.fits --starmodels /global/cfs/cdirs/desi/spectro/templates/basis_templates/v3.2/stdstar_templates_v2.2.fits --outfile /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/stdstars-5-00067765.fits.test --delta-color 0.1
```

#### Multiprocessing
```
Multiprocessing
real    0m46.155s
user    0m0.065s
sys     0m0.094s

real    0m45.500s
user    0m0.064s
sys     0m0.089s

real    0m45.957s
user    0m0.071s
sys     0m0.089s

time srun -n 1 -c 32 --cpu_bind=cores desi_fit_stdstars --frames /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/frame-b5-00067765.fits /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/frame-r5-00067765.fits /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/frame-z5-00067765.fits --skymodels /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/sky-b5-00067765.fits /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/sky-r5-00067765.fits /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/sky-z5-00067765.fits --fiberflats /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm9/fiberflatnight-b5-20201214.fits /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm9/fiberflatnight-r5-20201214.fits /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm9/fiberflatnight-z5-20201214.fits --starmodels /global/cfs/cdirs/desi/spectro/templates/basis_templates/v3.2/stdstar_templates_v2.2.fits --outfile /global/cfs/cdirs/desi/users/ziyaoz/skysub/exposures/20201214/00067765/stdstars-5-00067765.fits.test --delta-color 0.1
```

### TODO (Probably goes into future PRs)
1. Deduplicate IO; benchmark on multiple nodes
2. Deduplicate code that does Pool.map
3. Profile performance impact of`allreduce` and change it to `reduce` and `scatter` a subset of the data if too slow.

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -718,10 +718,7 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
         func_args.append( arguments )
 
     if comm is not None and comm.Get_size() > 1: # MPI mode & more than one rank per star
-        delta, remainder = len(func_args) // size, len(func_args) % size
-        start = rank * delta + min(rank, remainder) # Start of the args to process
-        end = (rank + 1) * delta + min(rank + 1, remainder) # End of the args to process
-        results = list(map(_func, func_args[start:end]))
+        results = list(map(_func, func_args[rank::size]))
         # All reduce here because we'll need to divide the work out again
         results = comm.allreduce(results, op=MPI.SUM)
     elif ncpu > 1:
@@ -805,10 +802,7 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
             func_args.append(arguments)
 
     if comm is not None and comm.Get_size() > 1: # MPI mode & more than one rank per star
-        delta, remainder = len(func_args) // size, len(func_args) % size
-        start = rank * delta + min(rank, remainder) # Start of the args to process
-        end = (rank + 1) * delta + min(rank + 1, remainder) # End of the args to process
-        results = list(map(_func2, func_args[start:end]))
+        results = list(map(_func2, func_args[rank::size]))
         results = comm.reduce(results, op=MPI.SUM, root=0)
     elif ncpu > 1:
         log.debug("divide templates by median filters using multiprocessing.Pool of ncpu=%d"%ncpu)
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -582,7 +582,7 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
         logg : 1D[nstd] model surface gravity
         feh : 1D[nstd] model metallicity
         ncpu : number of cpu for multiprocessing
-        comm : MPI communicator; if given, ncpu will be ignored
+        comm : MPI communicator; if given, ncpu will be ignored and only rank 0 will return results that are not None
 
     Returns:
         coef : numpy.array of linear coefficient of standard stars
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -567,7 +567,7 @@ def interpolate_on_parameter_grid(data_wave, data_flux, data_ivar, template_flux
     return final_coefficients,chi2
 
 
-def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, logg, feh, ncpu=1, z_max=0.005, z_res=0.00002, template_error=0):
+def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, logg, feh, ncpu=1, z_max=0.005, z_res=0.00002, template_error=0, comm=None):
     """For each input spectrum, identify which standard star template is the closest
     match, factoring out broadband throughput/calibration differences.
 
@@ -582,6 +582,7 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
         logg : 1D[nstd] model surface gravity
         feh : 1D[nstd] model metallicity
         ncpu : number of cpu for multiprocessing
+        comm : MPI communicator; if given, ncpu will be ignored
 
     Returns:
         coef : numpy.array of linear coefficient of standard stars
@@ -599,11 +600,15 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
     # Each data(3 channels) is compared to every model.
     # flux should be already flat fielded and sky subtracted.
 
-
-
     cameras = list(flux.keys())
     log = get_logger()
     log.debug(time.asctime())
+
+    # Initialize MPI
+    if comm is not None:
+        from mpi4py import MPI
+        size, rank = comm.Get_size(), comm.Get_rank()
+        log.debug("Using MPI. rank: {}, size: {}".format(rank, size))
 
     # fit continuum and save it
     continuum={}
@@ -712,8 +717,14 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
                    "template_id":template_id}
         func_args.append( arguments )
 
-
-    if ncpu > 1:
+    if comm is not None and comm.Get_size() > 1: # MPI mode & more than one rank per star
+        delta, remainder = len(func_args) // size, len(func_args) % size
+        start = rank * delta + min(rank, remainder) # Start of the args to process
+        end = (rank + 1) * delta + min(rank + 1, remainder) # End of the args to process
+        results = list(map(_func, func_args[start:end]))
+        # All reduce here because we'll need to divide the work out again
+        results = comm.allreduce(results, op=MPI.SUM)
+    elif ncpu > 1:
         log.debug("creating multiprocessing pool with %d cpus"%ncpu); sys.stdout.flush()
         pool = multiprocessing.Pool(ncpu)
         log.debug("Running pool.map() for {} items".format(len(func_args))); sys.stdout.flush()
@@ -793,7 +804,13 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
             arguments={"template_id":t,"camera_index":index,"template_flux":template_flux[t][ii]}
             func_args.append(arguments)
 
-    if ncpu > 1:
+    if comm is not None and comm.Get_size() > 1: # MPI mode & more than one rank per star
+        delta, remainder = len(func_args) // size, len(func_args) % size
+        start = rank * delta + min(rank, remainder) # Start of the args to process
+        end = (rank + 1) * delta + min(rank + 1, remainder) # End of the args to process
+        results = list(map(_func2, func_args[start:end]))
+        results = comm.reduce(results, op=MPI.SUM, root=0)
+    elif ncpu > 1:
         log.debug("divide templates by median filters using multiprocessing.Pool of ncpu=%d"%ncpu)
         pool = multiprocessing.Pool(ncpu)
         results  =  pool.map(_func2, func_args)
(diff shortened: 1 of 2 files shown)


sbailey:
Thanks @ziyaointl .  @dmargala could you take a look at this for consistency with how you need stdstars to work for GPU integration, and general MPI organization / optimization?

py/desispec/fluxcalibration.py
```diff
@@ -712,8 +717,14 @@ def match_templates(wave, flux, ivar, resolution_data, stdwave, stdflux, teff, l
                    "template_id":template_id}
         func_args.append( arguments )
 
-
-    if ncpu > 1:
+    if comm is not None and comm.Get_size() > 1: # MPI mode & more than one rank per star
+        delta, remainder = len(func_args) // size, len(func_args) % size
+        start = rank * delta + min(rank, remainder) # Start of the args to process
+        end = (rank + 1) * delta + min(rank + 1, remainder) # End of the args to process
+        results = list(map(_func, func_args[start:end]))
```
dmargala, line 724: This looks like it is dividing `func_args` into contiguous chunks for each mpi rank. Is that done to preserve the ordering of results? If there's a reason not to use the strided chunking pattern here (e.g. `func_args[rank::size]`), I suggest adding that reason in a comment.

py/desispec/scripts/stdstars.py
```diff
@@ -688,23 +699,19 @@ def main(args, comm=None) :
...
+        normflux = head_comm.reduce(normflux, op=MPI.SUM, root=0)
+
+    if rank != 0: # All the work left belongs to rank 0
+        return
+
+    # Check at least one star was fit.
+    fitted_stars = np.where(chi2dof != 0)[0]
+    if fitted_stars.size == 0:
         log.error("No star has been fit.")
         sys.exit(12)
```
dmargala, line 716: I don't think we want to remove this, for the reason described in the comment. Also see here:

https://github.com/desihub/desispec/pull/1184#discussion_r594042298

dmargala:
This looks good to me! @sbailey this does not interfere with any GPU integration on my end. @ziyaointl I added a few comments inline. It might also be helpful to describe MPI behavior in more detail in the `match_templates` docstring. For example, rank 0 returns output while other ranks return `None`.

It will be hard to improve the performance much further without addressing the data and model loading. I added some timing statements and tested this on a DGX node using 32 MPI ranks and 4 GPUs.

```
TIMER-STOP:stdstars.py:182:main: Stopping startup at 2021-04-20T13:23:49.870320; duration 6.87 seconds
TIMER-STOP:stdstars.py:185:main: Stopping imports at 2021-04-20T13:23:56.302276; duration 6.43 seconds
TIMER-STOP:stdstars.py:188:main: Stopping mpi_connect at 2021-04-20T13:24:02.013725; duration 5.71 seconds
TIMER-STOP:stdstars.py:506:main: Stopping load_and_prep_data at 2021-04-20T13:24:08.720065; duration 6.70 seconds
TIMER-STOP:stdstars.py:603:main: Stopping load_and_prep_models at 2021-04-20T13:24:15.925648; duration 7.21 seconds
TIMER-STOP:stdstars.py:735:main: Stopping fit_stdstars at 2021-04-20T13:24:22.640940; duration 6.72 seconds
TIMER-STOP:stdstars.py:774:main: Stopping write_output at 2021-04-20T13:24:22.868987; duration 0.23 seconds

real	0m43.026s
user	0m0.086s
sys	0m0.121s
```

py/desispec/fluxcalibration.py
sbailey: I agree the `func_args[rank::size]` is cleaner to mentally parse, and follows a common MPI striding pattern.  If we do need/want to maintain order (usually desireable all other things being equal), this might be simpler to maintain:
```
my_args = np.array_split(func_args, size)[rank]
results = list(map(_func, my_args))
```
i.e. `np.array_split` takes care of splitting the array into `size` different sub-arrays, while taking care of boundary conditions, off-by-one bugs, etc.

But: if you do go that route, test that it does what you intend; I haven't parsed your code in detail to make sure...

py/desispec/scripts/stdstars.py
ziyaointl: Yes, I think you are right. I added an early return before this check for all the ranks != 0, but I just realized that an early return does not gurantee that MPI won't hang on a statement outside of this function. I'll go ahead and add the broadcast back.

py/desispec/fluxcalibration.py
ziyaointl: I tried the unordered striding pattern and the results did not change, so for brevity I'll use `func_args[rank::size]`. @sbailey Thanks for mentioning `np.array_split`! I was not aware of this function before.

ziyaointl:
@sbailey @dmargala Thanks for reviewing! I addresssed all the comments and reran the correctness test to confirm that the output remains correct. Performance got slightly worse, with runs on a single node typically taking 39-40s and one run taking 1m34.171s. I'm pretty sure the 1m34.171s run was due to I/O from looking at the logs, but I am less sure about the extra couple of seconds for other runs.

py/desispec/scripts/stdstars.py
dmargala: Now I see that #1213 introduced a bug for the MPI stdstar fit after #1184 was merged. It looks like the code introduced in #1213 should have been under the `if rank == 0:` statement. Having ranks != 0 return before the write fixes a bug with the existing code. (Although, if rank 0 throws an exception during that final code block...)

ziyaointl: Yeah that is not ideal. We could consider using https://mpi4py.readthedocs.io/en/stable/mpi4py.run.html, which has an option to automatically call `MPI_Abort()` for any unhandled exceptions. This should also obviate the need to broadcast errors to all ranks.

sbailey: #1213 was my PR so if I broke it I'll take a look at fixing it.

dmargala:
Further performance comparisons confirming negligible impacts to performance on a GPU node. Benchmarks are performed using half of a Cori GPU node (20 physical cores, 4 GPUs).

## Joint Fit (4 Exposures, 10 petals)

This benchmark uses a modified `desi_proc_joint_fit` (see #1254) which divides the global MPI communicator into 10 subcommunicators, one for each of the 10 petals. This PR modifies stdstar fit to further subdivide the communicator when `nstars > comm.size`. However, in this case, `comm.size == 2` so no further subdivision of the MPI communicator will occur. I used the same test setup from #1205. The following excerpts confirm that this PR does not substantially affect performance using a single GPU node.

Current master (with bugfix):

```
time srun -n 20 -c 2 --cpu-bind=cores ./mps-wrapper.sh ./cycle_gpus.sh desi_proc_joint_fit --traceshift --obstype science --cameras a0123456789 -n $NIGHT -e $EXPIDS --mpi
...
    "duration.min": 31.324223518371582,
    "duration.max": 55.54797959327698,
    "duration.mean": 44.32634682655335,
    "duration.median": 43.72269296646118,
    "n": 20
...
real    1m16.158s
user    0m0.053s
sys 0m0.093s
```

This PR:

```
time srun -n 20 -c 2 --cpu-bind=cores ./mps-wrapper.sh ./cycle_gpus.sh desi_proc_joint_fit --traceshift --obstype science --cameras a0123456789 -n $NIGHT -e $EXPIDS --mpi
...
    "duration.min": 33.307000398635864,
    "duration.max": 56.45742225646973,
    "duration.mean": 45.577242290973665,
    "duration.median": 45.02499055862427,
    "n": 20
...
real    1m17.080s
user    0m0.032s
sys 0m0.098s
```

This PR modifies stdstar fit to further subdivide the communicator when `nstars > comm.size` but in this case, comm.size == 2 so no further subdivision of the MPI communicator occurs.

## Single petal, single exposure:

This benchmark runs the stdstar fit for a single petal on a single exposures. In this case, there are 11 std stars, 9 of which pass the blue SNR cut, and 20 MPI ranks. 

In the current master, only the first 9 MPI ranks will do work:

```
time srun -n 20 -c 2 --cpu-bind=cores ./mps-wrapper.sh ./cycle_gpus.sh desi_fit_stdstars --mpi --delta-color 0.1 \
    --frames \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/frame-b8-00067710.fits \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/frame-r8-00067710.fits \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/frame-z8-00067710.fits \
    --skymodels \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/sky-b8-00067710.fits \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/sky-r8-00067710.fits \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/sky-z8-00067710.fits \
    --fiberflats \
        $DESI_SPECTRO_REDUX/denali/calibnight/20201214/fiberflatnight-b8-20201214.fits \
        $DESI_SPECTRO_REDUX/denali/calibnight/20201214/fiberflatnight-r8-20201214.fits \
        $DESI_SPECTRO_REDUX/denali/calibnight/20201214/fiberflatnight-z8-20201214.fits \
    --outfile stdstars-8-00067710.fits
...
INFO:stdstars.py:587:main: rank 0: finding best model for observed star #0
INFO:stdstars.py:587:main: rank 1: finding best model for observed star #1
INFO:stdstars.py:587:main: rank 2: finding best model for observed star #2
INFO:stdstars.py:587:main: rank 3: finding best model for observed star #3
INFO:stdstars.py:587:main: rank 4: finding best model for observed star #4
INFO:stdstars.py:587:main: rank 5: finding best model for observed star #5
INFO:stdstars.py:587:main: rank 6: finding best model for observed star #6
INFO:stdstars.py:587:main: rank 7: finding best model for observed star #7
INFO:stdstars.py:587:main: rank 8: finding best model for observed star #8
...
real    0m41.911s
user    0m0.040s
sys 0m0.030s
```


This PR will use all 20 MPI ranks. The MPI communicator that is initialized at the beginning of the std star fit is subdivided into 9 subcommunicators. The MPI rank to stdstar mapping can be read from the log:

```
time srun -n 20 -c 2 --cpu-bind=cores ./mps-wrapper.sh ./cycle_gpus.sh desi_fit_stdstars --mpi --delta-color 0.1 \
    --frames \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/frame-b8-00067710.fits \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/frame-r8-00067710.fits \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/frame-z8-00067710.fits \
    --skymodels \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/sky-b8-00067710.fits \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/sky-r8-00067710.fits \
        $DESI_SPECTRO_REDUX/denali/exposures/20201214/00067710/sky-z8-00067710.fits \
    --fiberflats \
        $DESI_SPECTRO_REDUX/denali/calibnight/20201214/fiberflatnight-b8-20201214.fits \
        $DESI_SPECTRO_REDUX/denali/calibnight/20201214/fiberflatnight-r8-20201214.fits \
        $DESI_SPECTRO_REDUX/denali/calibnight/20201214/fiberflatnight-z8-20201214.fits \
    --outfile stdstars-8-00067710.fits
INFO:stdstars.py:621:main: rank 0: finding best model for observed star #0
INFO:stdstars.py:621:main: rank 1: finding best model for observed star #1
INFO:stdstars.py:621:main: rank 2: finding best model for observed star #2
INFO:stdstars.py:621:main: rank 3: finding best model for observed star #3
INFO:stdstars.py:621:main: rank 4: finding best model for observed star #4
INFO:stdstars.py:621:main: rank 5: finding best model for observed star #5
INFO:stdstars.py:621:main: rank 6: finding best model for observed star #6
INFO:stdstars.py:621:main: rank 7: finding best model for observed star #7
INFO:stdstars.py:621:main: rank 8: finding best model for observed star #8
INFO:stdstars.py:621:main: rank 9: finding best model for observed star #0
INFO:stdstars.py:621:main: rank 10: finding best model for observed star #1
INFO:stdstars.py:621:main: rank 11: finding best model for observed star #2
INFO:stdstars.py:621:main: rank 12: finding best model for observed star #3
INFO:stdstars.py:621:main: rank 13: finding best model for observed star #4
INFO:stdstars.py:621:main: rank 14: finding best model for observed star #5
INFO:stdstars.py:621:main: rank 15: finding best model for observed star #6
INFO:stdstars.py:621:main: rank 16: finding best model for observed star #7
INFO:stdstars.py:621:main: rank 17: finding best model for observed star #8
INFO:stdstars.py:621:main: rank 18: finding best model for observed star #0
INFO:stdstars.py:621:main: rank 19: finding best model for observed star #1
...
real    0m42.974s
user    0m0.019s
sys 0m0.054s
```

sbailey:
Thanks for the extra GPU timing confirmation, @dmargala .  I also updated the logic to follow the pattern used in PR #1254:
```
if rank == 0:
    stuff
# end of function
```
instead of 
```
if rank != 0:
    return
stuff
# end of function
```

Both are equivalent but the first pattern is how the rest of the desispec code uses MPI, where all ranks make it to the end of the function without exiting early.  The idea is that this pattern is more maintainable because you can add all-rank updates to the code without having to check whether some ranks may have exited earlier.

py/desispec/scripts/stdstars.py
sbailey: Documenting a separate conversation: the reason to not use `MPI_Abort()` is because that causes a hard exit of python for all ranks of all communicators without the ability to catch an exception.  This is problematic for workflow wrappers that might split a large communicator to process multiple exposure in parallel.  If there is a problem with one of those, then the wrapper should be able to catch the exception and move on without one exposure forcing the others to exit too.

PR #1252, opened 2021-04-22
bin/desi_average_flux_calibration: version used for DocDB-6157

araichoor:
This PR provides the desi_average_flux_calibration script version used in DocDB-6157.
The modifications are:
- list of exposures are now provided through the name of file with the exposure paths
- option to correct or no for transparency for each exposure, before fitting
- option to use FWHM_ASEC or RADPROF_FWHM_ASEC for the GFA seeing measurement
- relies on SV1 GFA catalogs to extract, for each exposure: airmass, seeing, transparency, fiber_fracflux
- the seeing-dependence is now corrected for for each exposure, and not fitted.

The dependence on SV1 GFA catalogs may be re-worked to be more general, for future SV3 or MAIN.
Now that the seeing-dependence is corrected for before fitting, the seeing-term fitting may be obselete; or maybe this seeing-dependence correction could turned optional.

### DIFF
--- a/bin/desi_average_flux_calibration
+++ b/bin/desi_average_flux_calibration
@@ -20,6 +20,9 @@ import sys
 import fitsio
 import scipy.interpolate
 from pkg_resources import resource_exists, resource_filename
+from desimodel.fastfiberacceptance import FastFiberAcceptance
+from glob import glob
+import yaml
 import matplotlib.pyplot as plt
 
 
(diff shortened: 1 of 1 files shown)


sbailey:
Looks good, thanks.  While recognizing that this code documents what was used for DESI-6157, I made a small update for the future to use `desimodel.io.load_desiparams()` instead of manually constructing the filepath and yaml loading it.  Currently those options are equivalent, but using `desimodel.io.load_desiparams` gives us flexibility if things get re-organized on the desimodel side in the future.  I'll merge after tests re-pass.

PR #1256, opened 2021-04-27
fix pernight redshift logic with multiple input nights

sbailey:
This PR fixes a logic bug in `desi_tile_redshifts --group pernight --night N1 N2 N3` when called with multiple nights.  The intended behavior was to submit one job per (TILEID, NIGHT) combination, using only the exposures from each night.  Instead, it was submitting the equivalent of a "cumulative" job, but calling it a "pernight" job.

Fortunately we did not hit this problem with denali because we only submitted one night at a time.

@akremin please review this.

### DIFF
--- a/bin/desi_tile_redshifts
+++ b/bin/desi_tile_redshifts
@@ -115,6 +115,12 @@ def batch_tile_redshifts(tileid, exptable, group, spectrographs=None,
         log.error(msg)
         raise ValueError(msg)
 
+    nights = np.unique(np.asarray(exptable['NIGHT']))
+    if (group in ['pernight', 'pernight-v0']) and len(nights)>1:
+        msg = f'group=pernight requires all exptable rows to be same night, not {nights}'
+        log.error(msg)
+        raise ValueError(msg)
+
     spectro_string = ' '.join([str(sp) for sp in spectrographs])
     num_nodes = len(spectrographs)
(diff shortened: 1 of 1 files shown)


bin/desi_tile_redshifts
```diff
@@ -115,6 +115,12 @@ def batch_tile_redshifts(tileid, exptable, group, spectrographs=None,
         log.error(msg)
         raise ValueError(msg)
 
+    nights = np.unique(np.asarray(exptable['NIGHT']))
+    if (group in ['pernight', 'pernight-v0']) and len(nights)>1:
+        msg = f'group=pernight requires all exptable rows to be same night, not {nights}'
+        log.error(msg)
+        raise ValueError(msg)
```
akremin, line 122: Should there be a similar check for tile?

akremin (approved):
The logic all looks good. I had one comment about whether we should also check for a being a single tileid given that you are now checking for a single night inside the redshift script generator. It isn't required, but I think we should check if we're already questioning the data the user is submitting.

Assuming you tested that it runs under the conditions you're tweaking, then I think this is ready to be merged after you respond to the minor comment (one way or another).

sbailey:
Superseded by #1260 (including also adding the multi-tile consistency check).  Closing this PR.

PR #1257, opened 2021-04-29
add an option to stdstars allowing to use arbitrary TARGETIDs as standards

segasai:
I propose a simple option to stdstars script allowing to override isStdStar()  by directly specifying TARGETIDs of standards. 
That should make it easier to do tests expanding standard selection, processing fields without enough standards etc.

### DIFF
--- a/py/desispec/scripts/stdstars.py
+++ b/py/desispec/scripts/stdstars.py
@@ -43,6 +43,9 @@ def parse(options=None):
     parser.add_argument('--template-error', type = float, default = 0.1, required = False, help = 'fractional template error used in chi2 computation (about 0.1 for BOSS b1)')
     parser.add_argument('--maxstdstars', type=int, default=30, \
             help='Maximum number of stdstars to include')
+    parser.add_argument('--std-targetids', type=str, default=None,
+                         nargs='*',
+                         help='List of TARGETIDs of standards overriding the targeting info')
     parser.add_argument('--mpi', action='store_true', help='Use MPI')
     parser.add_argument('--ignore-gpu', action='store_true', help='Ignore GPU, if available')
 
@@ -189,6 +192,10 @@ def main(args, comm=None) :
         if rank == 0:
             log.info('GPU not available')
 
+    std_targetids = None
+    if args.std_targetids is not None:
+        std_targetids = [ int(_) for _ in args.std_targetids]
+            
     # READ DATA
     ############################################
     # First loop through and group by exposure and spectrograph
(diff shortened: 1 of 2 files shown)


py/desispec/scripts/stdstars.py
```diff
@@ -189,6 +192,10 @@ def main(args, comm=None) :
         if rank == 0:
             log.info('GPU not available')
 
+    std_targetids = None
+    if args.std_targetids is not None:
+        std_targetids = [ int(_) for _ in args.std_targetids]
```
sbailey, line 197: Minor: targetids are int, so is there a specific reason to use `type=str` for argparse only to have to convert those strings to integers here?  i.e. I would suggest
```
parser.add_argument('--std-targetids', type=int, ... )
...
if args.std_targetids is not None:
    std_targetids = args.std_targetids...
```
I think that has the added benefit of if someone passes in a non-int, it stops at the level of argparsing rather than later in the code.

py/desispec/scripts/stdstars.py
```diff
@@ -235,7 +242,10 @@ def main(args, comm=None) :
             # Set fibermask flagged spectra to have 0 flux and variance
             frame = get_fiberbitmasked_frame(frame,bitmask='stdstars',ivar_framemask=True)
             frame_fibermap = frame.fibermap
-            frame_starindices = np.where(isStdStar(frame_fibermap))[0]
+            if std_targetids is None:
+                frame_starindices = np.where(isStdStar(frame_fibermap))[0]
+            else:
+                frame_starindices = np.nonzero(np.in1d(frame_fibermap['TARGETID'], std_targetids))[0]
```
sbailey, line 248: Minor: I recently learned that numpy docs say "We recommend using :func:`isin` instead of `in1d` for new code."  They are equivalent for this case.

sbailey (approved):
Good idea, thanks.  I added two minor comments inline for consideration.

At the same time I'll still leave it to you to use this feature for custom processing and testing rather than hardcoding specific stdstar target lists for standards processing (i.e. it doesn't get us around needing to fix up how our extinctions are calculated).  IIUC this list of standards still passes through the same quality cuts that were tripping us up on Gaia standards, so using this e.g. for M31 would also require relaxing the default `--delta-color` option.  But regardless, this is a nice feature to have for testing alternate standards; thanks for adding this.

py/desispec/scripts/stdstars.py
segasai: That was an oversight on my part. I initially was planning to pass targetids as comma separated strings. I'll fix that

py/desispec/scripts/stdstars.py
segasai: will fix

segasai:
Thanks for the review @sbailey , I fixed the highlighted issues. And yes, I'm expecting that the delta-color will probably need to be increased at least for now in say m31 fields.

PR #1259, opened 2021-04-30
enable performing fluxcalibration using whatever fibers are in the model file without additional checks

segasai:
This is a continuation of PR #1257 that enables performing fluxcalibration using whatever 
fibers are in the model files without double checking again whether these are standards according to the
fibermap
It adds  --nostdcheck parameter. 
I didnt' add a test for this, as I wasnt' sure if it's worth extra CPU cycles, but I can if needed. 

With this PR and a previous one I was able to easily reprocess the M31 exposures using a custom standard selection.

### DIFF
--- a/py/desispec/scripts/fluxcalibration.py
+++ b/py/desispec/scripts/fluxcalibration.py
@@ -43,6 +43,8 @@ def parse(options=None):
                         help = 'only consider stars with color greater than this')
     parser.add_argument('--delta-color-cut', type = float, default = 0.2, required=False,
                         help = 'discard model stars with different broad-band color from imaging')
+    parser.add_argument('--nostdcheck', dest='nostdcheck',
+                        help='Do not check the standards against flags in the FIBERMAP; just use objects in the model file', action='store_true')
     parser.add_argument('--outfile', type = str, default = None, required=True,
                         help = 'path of DESI flux calbration fits file')
     parser.add_argument('--qafile', type=str, default=None, required=False,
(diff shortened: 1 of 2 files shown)


sbailey:

py/desispec/scripts/fluxcalibration.py
```diff
@@ -172,26 +175,30 @@ def main(args) :
...
-            log.error("inconsistency with spectrum {}, OBJTYPE={}, {}={} in fibermap".format(
+    if stdcheck:
+        fibermap_std_indices = np.where(isStdStar(fibermap))[0]
+        if np.any(~np.in1d(model_fibers%500, fibermap_std_indices)):
+            target_colnames, target_masks, survey = main_cmx_or_sv(fibermap)
+            colname =  target_colnames[0]
+            for i in model_fibers%500:
+                log.error("inconsistency with spectrum {}, OBJTYPE={}, {}={} in fibermap".format(
                 i, fibermap["OBJTYPE"][i], colname, fibermap[colname][i]))
         sys.exit(12)
```
sbailey, line 193: Lesson learned: always run test case before merging.  This sys.exit(12) should have been indented under the `if np.any(~np.in1d(model_fibers%500, fibermap_std_indices))` check, and caused the pipeline to exit early last night.  Credit to @akremin for debugging that.

segasai: Oops. My apologies for missing that/not testing other than running pytest.

Issue #1241, opened 2021-04-15
Possibly unintended behavior of desi_tile_redshift


akremin:
Using `desi_tile_redshift` for cleanup today I observed a behavior I think is unintended. I wanted to rerun cumulative redshifts for the previous night (20210413) because of a bug fix. 

What I expected was `desi_tile_redshifts` to use the specified night and give a cumulative redshift up to and including that night.

What it did instead was found ever exposure (including exposures on future dates to the one specified) and went to redshifts those. It correctly labeled the redshifts with the latest night and puts them in the correct directory for what it is doing (see in the outputs below that it puts them in the /20210414/ directory), but it doesn't allow me to do what I had intended.

Maybe this is okay because we really don't care about old cumulative redshifts, but I thought I'd raise the behavior just in case.

`desi_tile_redshifts -n 20210413 -t 176 -g "cumulative"`

```
INFO:desi_tile_redshifts:349:<module>: Loading production exposure tables for nights [20210413]
INFO:desi_tile_redshifts:403:<module>: 1 tiles; searching for exposures on prior nights
INFO:desi_tile_redshifts:416:<module>: Tile 176 nights=[20210413 20210414] expids=[84643 84644 84816]
INFO:batch.py:60:default_system: Guessing default batch system cori-haswell
INFO:desi_tile_redshifts:262:batch_tile_redshifts: Wrote /global/cfs/cdirs/desi/spectro/redux/daily/run/scripts/tiles/cumulative/176/20210414/coadd-redshifts-176-thru20210414.slurm
Submitted batch job 41656380
INFO:desi_tile_redshifts:278:batch_tile_redshifts: submitted coadd-redshifts-176-thru20210414.slurm
```

sbailey:
I confirm that is unintended.  It should have made the cumulative redshifts through 20210413 and ignored 0414, as you expected.

PR #1260, opened 2021-05-03
Redshifts by Breakfast

akremin:
### Overview
This PR modifies the daily processing and re-processing pipeline managers to incorporate redshift fitting throughout the night with proper Slurm job dependencies on the flux calibration steps of the spectral reductions. It also improves the simulation capabilities by adding a second class of simulated submissions where scripts are written to disk, but the submittions are only simulated. This is in addition to the previous --dry-run that output the filename and simulated submission but didn't write the scripts to disk.

This PR also solves the issue of cumulative redshifts looking into the future for exposures: https://github.com/desihub/desispec/issues/1241,
and re-implements a bug fix for per-night exposures made by Stephen in PR https://github.com/desihub/desispec/pull/1256.

#### New command line arguments
This also refactors the `desi_tile_redshifts` script, placing the bulk of the code under /scripts/tile_redshifts.py and factoring out the path name definitions into functions for use in the pipeline, along with the the main body of the code.

The daily script (`desi_daily_proc_manager`) and the re-processing script (`desi_run_night`) have been modified to include three new keywords:
` --z-submit-types`, type=str, default='perexp,pernight' or 'cumulative' (see below). The group types of redshifts that should be submitted with each exposure. If not specified, default  is 'perexp,pernight'. If 'false' or 'none' then no redshifts are submitted.

`--dry-run-level`, type=int, default=0. If nonzero, this is a simulated run. If dry_run=1 the scripts will be written or submitted. If dry_run=2, the scripts will not be writter or submitted. Logging will remain the same for testing as though scripts are being submitted. Default is 0 (false).

`--no-redshifts`, type=bool, unset is False. Whether to submit redshifts or not. If set, redshifts are not submitted.

#### New defaults
The default types of redshifts run are:
`['cumulative']` for the daily processing. 
`['perexp','pernight'] `for the re-processing. 

(Cumulatives are tricky when running re-processing over many nights in parallel, so those should be done separately after the bulk processing of spectra).

#### Flag Uses
The redshift fitting component can be turned off by setting a flag `--no-redshifts` in either script. This overrides any user inputs to ` --z-submit-types`.

The `--dry-run` flag can still be used as well. If `--dry-run-level` isn't specified with `--dry-run` set, then it is set to 2 (simulated everything without writing to disk).

### Tests
I tested with numerous combinatorics in the directory `/global/cfs/cdirs/desi/users/kremin/spectro/redux/zbybreakie`. I have vetted the pipeline manager behavior to be correct in all tests, and the redshift scripts also appear to be correct. I further ran without simulation to ensure jobs were submitted with proper dependencies (and they were). I compared the redshift scripts generated vs. those using master and the only difference was the fixing of a typo in an `echo` statement. (The typo was in the line "echo Generating files in [...]").

What I have not tested is every iteration of the desi_tile_redshifts command line script under the new refactor using an input exposure table, etc.. I performed several tests, but not an exhaustive set.


bin/desi_tile_redshifts
```diff
@@ -39,26 +39,27 @@ Not supported yet: multiple tiles on a single night in a single call:
...
+               help="batch queue name")
 p.add_argument("--batch-reservation", type=str,
-        help="batch reservation name")
+               help="batch reservation name")
 p.add_argument("--batch-dependency", type=str,
-        help="job dependencies passed to sbatch --dependency")
+               help="job dependencies passed to sbatch --dependency")
 p.add_argument("--system-name", type=str,
-        help="batch system name, e.g. cori-haswell, cori-knl, perlmutter-gpu")
+               help="batch system name, e.g. cori-haswell, cori-knl, perlmutter-gpu")
```
sbailey, line 62: Minor, but for consistency, the standard for the other scripts is that this parser logic should move into `args = desispec.scripts.tile_redshifts.parse()`, and that there should be a `desispec.scripts.tile_redshifts.main(args)`.  That `main` could be itself a super-simple to call `generate_tile_redshift_scripts(**args.__dict__)`, since that function also gets used by the pipeline without going through argument parsing etc.

sbailey:
One bookkeeping hiccup:
```
desi_tile_redshifts --nosubmit --tile 80606 --night 20201215 --group pernight
```
generated `run/scripts/tiles/80606/pernight/coadd-redshifts-80606-pernight.slurm` instead of `run/scripts/tiles/pernight/80606/20201215/coadd-redshifts-80606-20201215.slurm`.  However, not specifying the tile does work, i.e.
```
desi_tile_redshifts --nosubmit --night 20201215 --group pernight
```
generated `coadd-redshifts-80606-20201215.slurm` along with all the other tiles for that night.

Related:
```
desi_tile_redshifts --nosubmit --tile 80606 --night 20201215 20201216 --group pernight
```
also wrote only `coadd-redshifts-80606-pernight.slurm` instead of `coadd-redshifts-80606-20201215.slurm` and `coadd-redshifts-80606-20201216.slurm`.  It looks like the bugfix from deprecated PR #1256 didn't make it over here intact.

bin/desi_tile_redshifts
akremin: Thanks, I have implemented these changes. I will note for thoroughness that this approach does come at a small speed cost in outputting "`desi_tile_redshifts --help`" because it now imports all the packages when importing `desispec.scripts.tile_redshifts`, but I really don't mind.

akremin:
I wasn't able to reproduce either of your cases. I do see a different issue when running your second command, in that it only produces results for the first night and ignores the second night, but I observe the proper naming conventions in what scripts are generated. (Note I ran using 20201214 instead of 20201215  in the second example so it wouldn't overwrite the script from the first example).

When I run:

`desi_tile_redshifts --nosubmit --tile 80606 --night 20201215 --group pernight`

I get `./run/scripts/tiles/pernight/80606/20201215/coadd-redshifts-80606-20201215.slurm`

When I run:

`desi_tile_redshifts --nosubmit --tile 80606 --night 20201214 20201216 --group pernight`

I (only) get: `./run/scripts/tiles/pernight/80606/20201214/coadd-redshifts-80606-20201214.slurm`

I did move the arg parser into a function call as requested, but I don't know why that would play a role here. Could you please try the commands again?

In the meantime, I will double-check the code associated with PR #1256 and identify why it isn't producing the second night when specifying two nights.

sbailey:
Apologies for the noise.  In my flopping back and forth between branches for testing, I was testing the wrong branch when reporting that bug.  Your branch is fine and works well.  Thanks.  Merging now.

PR #1266, opened 2021-05-06
add support for stuck positioners assigned to sky locations

sbailey:
This PR adds support for stuck positioners assigned to blank sky locations (i.e. if OBJTYPE='SKY' it doesn't discard the fibers just because FIBERSTATUS STUCKPOSITIONER is set).

It additionally corrects a mismatch between the desispec and fiberassign bit definitions that were resulting in RETRACTED positioners being discarded, even if they were assigned to valid targets.  FIBERSTATUS changes on the desitarget side to match fiberassign:
  * bit 3 BADTARGET -> bit 0 UNASSIGNED
  * bit 3 now RETRACTED and not treated as a fatal bit

Note: there is more that we can/should do for FIBERSTATUS bit masking and not treating all bits as fatal, but I'm trying to focus this PR on the minimum necessary to support stuck positioners assigned to sky (first used on 81097 observed 20210430/00086750, albeit requiring a special case here because UNASSIGNED bit was also set).  This entered a fiberassign tag with 2.4.0 on 20210505 and tile 81098.

I tested this with 3 tiles from 20210430 in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/stucksky, including 81097 with stuck positioners on sky and normal tiles 210 and 117.

@akremin the normal pipeline will likely fail on tile 81098 on 20210505 but will hopefully succeed with this branch.

@dstndstn @djschlegel @julienguy @tskisner FYI

### DIFF
--- a/py/desispec/fiberbitmasking.py
+++ b/py/desispec/fiberbitmasking.py
@@ -50,7 +50,7 @@ def get_fiberbitmasked_frame_arrays(frame,bitmask=None,ivar_framemask=True,retur
               a bad FIBERSTATUS
 
     example bitmask list:
-        bitmask = [fmsk.BROKENFIBER,fmsk.BADTARGET,fmsk.BADFIBER,\
+        bitmask = [fmsk.BROKENFIBER,fmsk.UNASSIGNED,fmsk.BADFIBER,\
                     fmsk.BADTRACE,fmsk.MANYBADCOL, fmsk.MANYREJECTED]
         bitmask = get_fiberbitmask_comparison_value(kind='fluxcalib')
         bitmask = 'fluxcalib'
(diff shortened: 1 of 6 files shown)


py/desispec/fiberbitmasking.py
```diff
@@ -129,28 +129,25 @@ def get_fiberbitmask_comparison_value(kind='fluxcalib'):
...
     return get_all_fiberbitmask_val()
 
 def get_all_nonamp_fiberbitmask_val():
-    return (fmsk.STUCKPOSITIONER | fmsk.BROKENFIBER | fmsk.BADTARGET | fmsk.MISSINGPOSITION | fmsk.BADPOSITION | \
+    """Return a mask for all bad FIBERSTATUS bits except BADAMPB/R/Z
+
+    Note: does not include STUCKPOSITIONER or RESTRICTED, which could still
+    be on a valid sky location, or even a target for RESTRICTED.
+    """
+    return (fmsk.UNASSIGNED | fmsk.BROKENFIBER | fmsk.MISSINGPOSITION | fmsk.BADPOSITION | \
```
akremin, line 150: I see fmsk.BADTARGET -> fmsk.UNASSIGNED  (good).   But is there a reason fmsk.STUCKPOSITIONER was removed?

py/desispec/sky.py
```diff
@@ -177,6 +177,13 @@ def compute_uniform_sky(frame, nsig_clipping=4.,max_iterations=100,model_ivar=Fa
     skyfibers = np.where(frame.fibermap['OBJTYPE'] == 'SKY')[0]
     assert np.max(skyfibers) < 500  #- indices, not fiber numbers
 
+    #- Hack: test tile 81097 (observed 20210430/00086750) had set
+    #- FIBERSTATUS bit UNASSIGNED for sky targets on stuck positioners.
+    #- Undo that.
+    if (frame.meta is not None) and ('TILEID' in frame.meta) and (frame.meta['TILEID'] == 81097):
```
akremin, line 183: Do we want to have special code for one test sv2 tile? The answer might well be yes, but we could also just flag the exposure as bad. I'm okay either way on this one

akremin (commented):
Overall this looks good. I made two comments. I will rerun last night's tile and verify. I'll comment again once that test is done.

akremin:
I'm still seeing outputs from qproc that it can't find skyfibers, is that expected?

e.g.:
`WARNING:qsky.py:90:qproc_sky_subtraction: No sky fibers! I am going to declare the faintest half of the fibers as sky fibers
`

akremin (approved):
The test on TILE 81098 (Exp 87359 from May 5th 2021) was successful, so I will approve this PR, but please reply to my two comments/questions for the record before we merge

py/desispec/sky.py
sbailey: I don't want to get in the habit of hardcoding special cases, but at the time I was writing this code this was the only tile that had skies assigned to stuck positioners and so I needed something like this for testing.  It seemed harmless enough that I left it in.

sbailey:
qproc: I forgot to check that for how it determines sky.  Let me try to fix that too before we merge.

sbailey:
qproc "No sky fibers!" : that happens when running on a raw data file without an assemble_fibermap generated fibermap added to a preproc file ahead of time, e.g. nightwatch which doesn't use asemble_fibermap yet...

I did fix fibermap auto-discovery to use the fibermap in a preproc file instead of the 5k fiber fibermap file in the same driectory.  We could still do work for full fibermaps in nightwatch, but that is unrelated to interpreting SKY fibers.  It does work with
```
desi_qproc -i /global/cfs/cdirs/desi/spectro/redux/daily/preproc/20210430/00086750/preproc-r1-00086750.fits --camera r1 --auto
```

PR #1268, opened 2021-05-07
Read seeing

julienguy:
This PR to use the keyword 'SEEING' in the frame fits file header to compute fiber aperture loss corrections for the flux calibration (so far, a default value of 1.1" was considered because there was no SEEING in the headers).

This seeing value from the frame header, or the value given by an optional argument to desi_compute_fluxcalib is saved with the keyword FCSEEING in the header of the fluxcalib fits file, and then copied to the cframe files. It's to keep a record of the seeing value used for flux calibration. This can then be used for the TSNR calculation for instance.

The class FluxCalibration has a meta attribute which is a dictionary with the fits header content (like the meta attribute of the Frame class).


julienguy:
After this PR is merged @michaelJwilson  can look for the FCSEEING keyword in the fits header or the 'meta' attribute of the FluxCalib object or flux calibrated Frame to compute the TSNR corrections (and consider a default value of 1.1 was used when FCSEEING is absent for backward compatibility).

PR #1275, opened 2021-05-24
TSNR: Add tracer poisson term to model ivar

michaelJwilson:
Include tracer poisson term in tsnr model var for bgs and backup programs. 

(yet another branch)

### DIFF
--- a/py/desispec/tsnr.py
+++ b/py/desispec/tsnr.py
@@ -19,8 +19,10 @@ from desiutil.dust import dust_transmission
 from desispec.io import findfile,read_frame,read_fiberflat,read_sky,read_flux_calibration,iotime
 from desispec.io.spectra import Spectra
 from desispec.calibfinder import findcalibfile
+from astropy import constants as const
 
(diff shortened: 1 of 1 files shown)


py/desispec/tsnr.py
```diff
@@ -455,6 +457,71 @@ def fb_rdnoise(fibers, frame, psf):
...
+    else:
+        # No source poisson term otherwise. 
+        nominal = 0.0                     # [nanomaggie]. 
+        fiberfrac = 1.0
+
+        return nominal                    # [e/specbin].
+
+    
+    nominal *= fiberfrac
+    nominal /= fiber_area_arcsec2         # [nMg / ''].
```
julienguy, line 507: I don't think we should divide by the fiber area here because the output should be in `(electrons/specbin)` (it's done differently on the wiki because there it is to be added to the sky flux per arcsec2).

julienguy (changes requested):

michaelJwilson:
oops, trying to follow the wiki too closely.  Thanks for the catch.

PR #1277, opened 2021-05-25
Retrieve etc efftime from json

michaelJwilson:
Retrieve etc efftime from json if ACTTEFF not in frame headers (as true of early main survey data). 

Aside: do not add missing prod exposures if args.expids is not None.

### DIFF
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -4,6 +4,7 @@ Calculate stand alone tsnr tables for a given desispec prod.
 '''
 
 import os,sys
+import json
 import glob
 import itertools
 import argparse
@@ -184,7 +185,7 @@ def update_table(table1,table2,keys) :
     else :
         return table2
 
-def compute_summary_tables(summary_rows,efftime_config,preexisting_tsnr2_expid_table,preexisting_tsnr2_frame_table,specprod_dir) :
+def compute_summary_tables(summary_rows,efftime_config,preexisting_tsnr2_expid_table,preexisting_tsnr2_frame_table,specprod_dir, add_missing=True) :
     """ Compute summary tables.
 
     Args:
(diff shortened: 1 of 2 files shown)


michaelJwilson:
Note: rewrites EFFTIME_ETC=-99 if not present in frame header and json does not exist (not expected).  Previous behaviour was EFFTIME_ETC=0.0 in such a case.

Tested with:
export SPECPROD=daily # denali                                                                                                                                                                                                                  
export DESI_SPECTRO_CALIB=/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/
export DESIMODEL=/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desimodel/master

desi_tsnr_afterburner --outfile /global/cscratch1/sd/mjwilson/tsnr-ensemble-updates/tsnr-summary.fits --expids 88359,88360,89881 --recompute --nproc 12 --cameras b0

Yielding:

EXPID EFFTIME_ETC
int32   float32  
----- -----------
88359   181.04936
88360   185.40895
89881   184.74846

for exposures previously marked as EFFTIME_ETC in daily tsnr_exposures.fits.

dkirkby:
Have you compared the FITS header and json values when both are available?  If they disagree, I would use the json value, not the header value.

michaelJwilson:
I had not, here's a comparison for all exposures where etc 'expinfo' was definded (excludes some early exps. in daily tsnrs) and I found ACTTEFF in the corresponding cframe for petal 0 or 1; 

image

One of the worst offenders:
EXPID   JSON_EFFTIME  ACTTEFF     DIFF
84654  1414.137329      1406.3414   7.795928999999887

Although far from ideal, I think we have bigger problems than ~10s and there seems a hesitancy to avoid the jsons where possible, so I'll make the change immediately to defaulting to jsons.

julienguy:
merging after successful test. thanks.

PR #1283, opened 2021-05-26
Add fiberfracs again.

michaelJwilson:
include etc fiberfracs in tsnr calc., also for gpbdark, gpbbright etc.  Plots to follow after we agree approach.

p.s.  What's the record for DESI PRs in a week?


michaelJwilson:
Main survey exposures after calibrating against EFFTIME_ETC for the same exposures (split by program). 

image
image

We should also include purpose in tsnr-exposures.fits?  As main survey tiles are occasionally used for commissioning?

michaelJwilson:
image
image

dkirkby:
@michaelJwilson Why does fiberfrac=True/False change values on the x axis (EFFTIME_ETC) in your scatter plots?

michaelJwilson:
This is just an artificial shift to separate the points.

dkirkby:
Ok, phew.  The improvement in the bulk of exposures at efftime ~ 1000s / 180s looks a bit better than your histograms.  Perhaps the correlation with skylevel is contributing to the scatter?  Could you remove exposures with spec_flux_r > 5 to check this?

dkirkby:
Do you see normalization shifts between fiberfrac=True/False?

julienguy:
- in tsnr.py, line 778, 'elf' -> 'elg' (in case we miss the ETC info)
- in tsnr.py , line 788 (and after), please change variable name avg_platescale to avg_inverse_platescale or take the inverse in order to get something consistent with the isotropic_platescale. (there is no bug, it's just the variable name is confusing)
- in tsnr.py, line 880 and 900 and 916, I think we should not divide by the isotropic_platescale**2 because this platescale correction has already been applied to the fluxcalib vector.

julienguy:
```
 File "/global/common/software/desi/users/jguy/desispec/py/desispec/tsnr.py", line 771, in calc_tsnr_fiberfracs
    log.warning('Failed to inherit from etc json at {}.  Assuming nominal surveyspeed etc fiberfracs at 1.1 arcsecond fwhm.'.format(etcpath))
NameError: name 'etcpath' is not defined
```

julienguy:
Thanks Mike. I will merge when the tests are done.

julienguy:
We would not to make sure to preserve the normalization of the TSNRs. With the inclusion of the ffrac term, they have changed.
tsnr2

PR #1285, opened 2021-05-28
TSNR ensemble updates etcrefcat

michaelJwilson:
Originally, TSNR derived EFFTIME_SPECs were normalized to the GFA derived EFFTIMES (offline sky mags. and nominal calib vector), recorded @ data/tsnr/sv1-exposures.csv.  No etc information was available for these exposures, specifically fiber fracs for psf, elg and bgs used in the latest tsnr defintions.  Further the GFA efftimes generated by the pipeline later differed as the standard calibration assumed changed, e.g. post mirror wash.  

We seek to preserve this original EFFTIME_SPEC normalization for future exposures, which will have new tsnr algorithms. 

To do so, here we record these original EFFTIME_SPECS for the set exposures for which etc fiber fracs are available.  Hence, the new tsnr values (that relies on these fiberfracs) and derived efftimes may be calibrated off this set, where _both the original normalization and etc info. required by the algorithm is available_ (default fiberfracs would be used otherwise).  I.e. with a regression of EFFTIME_[DARK/BRIGHT] new vs old for this set of exposures, see scripts/calibrate_tsnr.

Note: including LYA, desispec commit a056732 on Mar 19 2021, as recorded in the history to data/tsnr/tsnr-efftime.yaml, allows the generation of these original TSNR values for future exposures.  We then have one standard ruler, by regenerating these where required, rather than continually boot strapping in time.


michaelJwilson:
In the calibration run, we should check there are no instances of:

> Failed to inherit from etc json.  Assuming nominal surveyspeed etc fiberfracs at 1.1 arcsecond fwhm.

in the logs to the afterburner, which would indicate etc ffracs where not available, and an error with this
PR.  Hopefully, the added checks mean we'll avoid this.

julienguy:
Calibration plots with SV1 and with SV3(data with ETC) for the record.

efftime-202102-denali
efftime-202105-daily
tsnr2-BGS-202102-denali
tsnr2-BGS-202105-daily
tsnr2-ELG-202102-denali
tsnr2-ELG-202105-daily
tsnr2-LRG-202102-denali
tsnr2-LRG-202105-daily
tsnr2-QSO-202102-denali
tsnr2-QSO-202105-daily

PR #1291, opened 2021-06-02
DESI average throughput from SV1 and SV3

julienguy:
Use the new measurement of the throughput presented in https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=6157 .
 - define new attributes to the class  desispec.averagefluxcalibration.AverageFluxCalib (median_seeing, median_ffracflux, fac_wave_power, ffracflux_wave,first_night)
 - use this when computing the sky magnitudes
(a new version of the average flux calibration following this new data model is saved in the spectro calibration svn repo, see https://desi.lbl.gov/trac/browser/calib/desi_spectro_calib/trunk/spec/fluxcalib)


julienguy:
merging this PR that affects only the sky magnitude after the improved measurement of the throughput (3% change in calibrated sky flux).

PR #1302, opened 2021-06-11
Spectroscopic QA

julienguy:
Spectroscopic QA code

 * QA parameters given in https://github.com/desihub/desispec/blob/spectro-qa/py/desispec/data/qa/qa-params.yaml
 * Exposure QA:
   * functions in https://github.com/desihub/desispec/blob/spectro-qa/py/desispec/exposure_qa.py
   * script `desi_exposure_qa`, example: `desi_exposure_qa -n 20210606 --prod daily --outdir . --nproc 8 --expids 91382:92191`
   * looks at fiber positioning, SNR, number of std stars, rms of std stars, flux calibration relative to other petals, sky chi2pdf, sky normalisation rms.
   * writes one file per exposure `$DESI_SPECTRO_REDUX/$SPECPROD/exposures/NIGHT/EXPID/exposure-qa-EXPID.fits` with two tables, FIBERQA, PETALQA in 2 hdus,
     and summary scores in the header.
 * Tile QA:
   * functions in https://github.com/desihub/desispec/blob/spectro-qa/py/desispec/tile_qa.py
   * script `desi_tile_qa`, example: `desi_tile_qa --prod daily --outdir /global/cfs/cdirs/desi/users/jguy/exposure_qa --night 20210504:20220101 --nproc 16`
   * look for coadd,zbest in `$DESI_SPECTRO_REDUX/$SPECPROD/tiles/cumulative/TILEID/NIGHT/`, finds list of exposures in spectra file and read the exposure-qa files
     or generate them if missing.
   * writes one file per tile/night `$DESI_SPECTRO_REDUX/$SPECPROD/tiles/cumulative/TILEID/NIGHT/tile-qa-TILEID-night.fits` with two tables, FIBERQA, PETALQA in 2 hdus,
     and summary scores in the header.
   * writes a png image summarizing the tile QA (see example below)
 * `desi_tile_vi` script
   * goes through the list of possible tiles to inspect from the input tiles.csv file
   * display the tile-qa png file
   * prompt to ask for decision on tile
   * toggles zdone to true in tiles table if valid.
   * saves changes, state of visualization, and name of inspector in table

Example

```
desi_tile_qa --prod daily --outdir /global/cfs/cdirs/desi/users/jguy/exposure_qa --night 20210514:20220101 --nproc 16

desi_tile_vi --prod daily --qa /global/cfs/cdirs/desi/users/jguy/exposure_qa --night 20210514:20220101 -o /global/cfs/cdirs/desi/users/jguy/exposure_qa/mytiles.csv -u jguy
desi_tile_vi --prod daily --qa /global/cfs/cdirs/desi/users/jguy/exposure_qa --night 20210514:20220101 -i /global/cfs/cdirs/desi/users/jguy/exposure_qa/mytiles.csv -u jguy

INFO:desi_tile_vi:165:main: Inspecting TILE=1404 NIGHT=20210514 ...
INFO:meta.py:142:findfile: Tile-based files selected; healpix-based files and input will be ignored.
Showing /global/cfs/cdirs/desi/users/jguy/exposure_qa/tiles/cumulative/1404/20210514/tile-qa-1404-20210514.png ...
PROMPT: Is tile 1404 a valid tile? (y/n/save/exit/pass/help)
>>> help
USER ENTERED >>> help
y: yes, this tile is valid.
n: no, it's bad and we should discard one or several exposures or fix the processing.
pass: skip this one, because unsure.
exit: exit without saving anything in this session.
save: save output file /global/cfs/cdirs/desi/users/jguy/exposure_qa/mytiles.csv
PROMPT: Is tile 1404 a valid tile? (y/n/save/exit/pass/help)
```

tile-qa-1404-20210514


py/desispec/data/qa/qa-params.yaml
```diff
@@ -0,0 +1,86 @@
...
+ tsnr2_petal_minfrac : 0.5
+ tsnr2_petal_maxfrac : 2.
+ tsnr2_for_efftime_dark : 'TSNR2_ELG'
+ tsnr2_for_efftime_bright : 'TSNR2_BGS'
+ tsnr2_for_efftime_backup : 'TSNR2_BGS'
+
+ # bad POSITIONING flag
+ bad_positions_mask : "STUCKPOSITIONER|BROKENFIBER|RESTRICTED|MISSINGPOSITION|BADPOSITION|POORPOSITION"
+ # bad QAFIBERSTATUS flag
+ bad_qafstatus_mask : "UNASSIGNED|STUCKPOSITIONER|BROKENFIBER|RESTRICTED|MISSINGPOSITION|BADPOSITION|POORPOSITION|BADFIBER|BADTRACE|BADFLAT|BADARC|MANYBADCOL|MANYREJECTED|BADAMPB|BADAMPR|BADAMPZ|BADPETALPOS|BADPETALSKY|BADPETALSTDSTAR|BADPETALFLUXCAL|BADPETALSNR|BADREADNOISE"
```
sbailey, line 33: For both `bad_positions_mask` and `bad_qafstatus_mask`, I don't think STUCKPOSITIONER and RESTRICTED should count as bad, since they could be validly used for SKY targets (or even science targets for RESTRICTED positioners).  If they couldn't be used for that it will appear through UNASSIGNED, and if they aren't at the expected/requested location it will show up through POORPOSITION, BADPOSITION regardless of whether they are RESTRICTED or STUCK.

bin/desi_tile_vi
```diff
@@ -0,0 +1,218 @@
...
+
+import os,sys
+import argparse
+import glob
+import numpy as np
+import multiprocessing
+from astropy.table import Table
+import fitsio
+import subprocess
+from PIL import Image
```
sbailey, line 20: PIL is not one of our standard dependencies (though it happens to be installed via some other dependency), and it appears that this import is no longer used.  Let's remove it.

py/desispec/maskbits.py
```diff
@@ -68,6 +69,13 @@
     - [BADAMPB,        22, "Issues in the amplifier readouts of camera B make this unusable"]
     - [BADAMPR,        23, "Issues in the amplifier readouts of camera R make this unusable"]
     - [BADAMPZ,        24, "Issues in the amplifier readouts of camera Z make this unusable"]
+#- Following bits are reserved for spectro QA
+    - [BADPETALPOS,    25, "Too many fibers with bad positioning (bits 1,2,3,8,9) in petal"]
```
sbailey, line 73: See earlier comment about which bits should be included in BADPETALPOS; update here as well if we change the config definition (or cite config here instead of replicating which bits are bad)

py/desispec/tile_qa.py
```diff
@@ -0,0 +1,221 @@
...
+import glob
+
+from desiutil.log import get_logger
+
+from desispec.exposure_qa import compute_exposure_qa,get_qa_params
+from desispec.io import read_fibermap,findfile,read_exposure_qa,write_exposure_qa
+from desispec.maskbits import fibermask
+
+
+def mystack(tables) :
```
sbailey, line 23: What's the processing case where this was needed?  Let's make this a `_mystack` hidden function and document what it does that is different from normal vstack.

bin/desi_exposure_qa
```diff
@@ -0,0 +1,185 @@
...
+    parser = argparse.ArgumentParser(
+                description="Calculate exposure QA")
+    parser.add_argument('-o','--outfile', type=str, default=None, required=False,
+                        help = 'Output summary file')
+    parser.add_argument('--recompute', action = 'store_true',
+                        help = 'recompute')
+    parser.add_argument('--prod', type = str, default = None, required=False,
+                        help = 'Path to input reduction, e.g. /global/cfs/cdirs/desi/spectro/redux/blanc/,  or simply prod version, like blanc, but requires env. variable DESI_SPECTRO_REDUX. Default is $DESI_SPECTRO_REDUX/$SPECPROD.')
+    parser.add_argument('--outdir', type = str, default = None, required=False,
+                        help = 'Path to ouput directory, default is the input prod directory')
```
sbailey, line 37: Please clarify in help string whether output goes directly into outdir without any sub-hierarchy, or into outdir/exposures/NIGHT/EXPID/, or outdir/night/expid, or something else.

bin/desi_tile_qa
```diff
@@ -0,0 +1,207 @@
...
+    parser.add_argument('-o','--outfile', type=str, default=None, required=False,
+                        help = 'Output summary file')
+    parser.add_argument('--recompute', action = 'store_true',
+                        help = 'recompute')
+    parser.add_argument('--prod', type = str, default = None, required=False,
+                        help = 'Path to input reduction, e.g. /global/cfs/cdirs/desi/spectro/redux/blanc/,  or simply prod version, like blanc, but requires env. variable DESI_SPECTRO_REDUX. Default is $DESI_SPECTRO_REDUX/$SPECPROD.')
+    parser.add_argument('--exposure-qa-dir', type = str, default = None, required=False,
+                        help = 'Path to exposure qa directory, default is the input prod directory')
+    parser.add_argument('--outdir', type = str, default = None, required=False,
+                        help = 'Path to ouput directory, default is the input prod directory')
```
sbailey, line 40: Please clarify in help string whether output goes directly into outdir without any sub-hierarchy, or into outdir/tiles/cumulative/TILEID/NIGHT/, or something else.

py/desispec/tile_qa.py
```diff
@@ -0,0 +1,221 @@
...
+
+    # EFFTIME
+    tile_petalqa_table["EFFTIME_SPEC"]=np.zeros(npetal)
+    for petal in petals :
+        entries=(tile_fiberqa_table['PETAL_LOC'] == petal)
+        tile_petalqa_table['EFFTIME_SPEC'][petal]=np.median(tile_fiberqa_table["EFFTIME_SPEC"][entries])
+
+    # add meta info
+    tile_fiberqa_table.meta["TILEID"]=tileid
+    tile_fiberqa_table.meta["NIGHT"]=night
```
sbailey, line 200: Since tile QA could span multiple nights, "NIGHT" seems a misnomer.  Consider "LASTNITE" (and see next comment about why not "LASTNIGHT").

py/desispec/tile_qa.py
```diff
@@ -0,0 +1,221 @@
...
+    tile_fiberqa_table.meta["NGOODFIBERS"]=good_fibers.size
+    tile_fiberqa_table.meta["NGOODPETALS"]=good_petals.size
+    tile_fiberqa_table.meta["EFFTIME_SPEC"]=np.mean(tile_petalqa_table['EFFTIME_SPEC'][good_petals])
```
sbailey, line 203: These meta keywords end up as FITS header keywords and thus should be 8-char (alas).  As-is they result in headers like
```
HIERARCH NGOODFIBERS = 3777                                                     
HIERARCH NGOODPETALS = 9                                                        
HIERARCH EFFTIME_SPEC = 1192.576296427857                                       
```
Maybe that is ok, but could be a hassle for some FITS readers so let's avoid if possible.

sbailey (commented):
I put several minor comment inline, and one substantive comment for consideration about exactly which FIBERSTATUS bits should be considered bad.

py/desispec/data/qa/qa-params.yaml
julienguy: ok. good point.

py/desispec/tile_qa.py
julienguy: replaced by vstack after preparing the tables (rm some meta keywords, rm COEFF column for zbest).

py/desispec/exposure_qa.py
```diff
@@ -0,0 +1,377 @@
...
+
+    frame_header = None
+    fibermap_header = None
+
+    # EFFTIME
+    goaltype="dark"
+    if fibermap_header is not None and "GOALTYPE" in fibermap_header :
+        goaltype=fibermap_header["GOALTYPE"]
+    else :
+        log.warning("no GOALTYPE info, assume 'dark'")
```
sbailey, line 128: Just before this if/else you set `fibermap_header = None`, which means that it will always not find GOALTYPE and default to dark.  I think you mean to get the `fibermap_header` from the fibermap file instead?  Note that below there is another check on `fibermap_header is None` in which case it is ready from the cframe file, which also might be a leftover.

In particular the fibermap_header is then read from the cframe and its GOALTYPE is is propagated to the output QA.

py/desispec/exposure_qa.py
```diff
@@ -0,0 +1,377 @@
...
+                continue
+            petalqa_table["NCFRAME"][petal]+=1
+            head=fitsio.read_header(cframe_filename)
+            if frame_header is None :
+                frame_header = head
+            if fibermap_header is None :
+                try :
+                    fibermap_header = fitsio.read_header(cframe_filename,"FIBERMAP")
+                except OSError as e :
+                    log.error(e)
```
sbailey, line 160: This looks like a debugging leftover.  If a cframe doesn't have a FIBERMAP HDU, then something is really wrong with the processing.  If you are going to use the original fibermap file (instead of the cframe FIBERMAP HDU), then it seems you should also use that header.  Otherwise if you want to use the cframe FIBERMAP instead, you should probably use that for everything and require it to exist.

sbailey (changes requested):
Review v2: there is a logic problem with reading fibermap header and how goaltype is derived; see comments inline.

py/desispec/exposure_qa.py
julienguy: Thanks for finding this one. I now use directly fibermap.meta.

py/desispec/exposure_qa.py
julienguy: I now use directly fibermap.meta.

julienguy:
I think I addressed all the comments. Thanks for the careful review.

py/desispec/exposure_qa.py
```diff
@@ -0,0 +1,377 @@
...
+        stdstars_filename = findfile("stdstars",night,expid,spectrograph=spectro,specprod_dir=specprod_dir)
+        if os.path.isfile(stdstars_filename) :
+            t = fitsio.read(stdstars_filename,'METADATA')
+            # SNR cut is same as in stdstars.py, this is redundant,
+            # but for clarity on the selection, I repeat the cuts here
+            # CHI2DOF and color cut are used in flux calibration
+            # generous color cut here.
+            good=(t["CHI2DOF"]<2.)&(t["BLUE_SNR"]>=4.)
+            if "MODEL_G-R" in t.dtype.names :
+                good &= (np.abs(t["MODEL_G-R"]-t["DATA_G-R"])<0.3)
```
sbailey, line 209: Our favorite gotcha: off-footprint Gaia-std only tiles like TILEID 80715 observed 20210115/00072498, have stdstars*.fits METADATA columns DATA_GAIA-BP-RP and MODEL_GAIA-BP-RP instead of DATA_G-R and MODEL_G-R, thus this QA will crash.  See scripts/stdstars.py around line 594.

At minimum we should catch the Gaia special case here, but better if this could be abstracted into a way that we won't keep tripping on it.

julienguy: The code tests for "MODEL_G-R" and will not do this color test for GAIA magnitudes (which are  outside of the main footprint and sometimes affected by large extinction corrections).

sbailey: Got it; not applying cut on Gaia is purposeful, and I was wrong about this crashing.  Looks fine, thanks for confirming.

sbailey (approved):

PR #1309, opened 2021-06-16
Pipeline robustness when reading ETC values from raw data header

akremin:
This PR fixes a corner case where ACTTEFF is defined in the raw data header but is set to an empty (non-numeric) value. This caused the pipeline to crash. Now we test that it can be transformed into a float within a `try` statement.

This was part of a larger discussion that led to the choice of deprecating `ACTTEFF` in favor of `ETCTEFF`, which is also in the raw data header. For future dates on or after 20210614, only `ETCTEFF` will be checked. The old keyword is kept for compatibility with old data (under an assertion that `night<20210614`).

I tested this on nights in Dec 2020, Jan 2021, Feb 2021, April 2021, May 2021, and June 14 2021. No issues on any date.

### DIFF
--- a/py/desispec/workflow/exptable.py
+++ b/py/desispec/workflow/exptable.py
@@ -781,9 +781,9 @@ def summarize_exposure(raw_data_dir, night, exp, obstypes=None, colnames=None, c
                     orig = str(dat_header['ACTTEFF'])
                 except:
                     orig = ''
-                reporting = keyval_change_reporting('ACTTEFF', orig, outdict['ACTTEFF'])
+                reporting = keyval_change_reporting('ACTTEFF', orig, outdict['EFFTIME_ETC'])
                 outdict['HEADERERR'] = np.append(outdict['HEADERERR'], reporting)
-                log.error(f"Couldn't convert 'ACTTEFF' with value {orig} to float. Using the default value.")
+                log.error(f"Couldn't convert ACTTEFF with value {orig} to float. Using the default EFFTIME_ETC.")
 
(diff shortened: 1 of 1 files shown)


sbailey (approved):
Looks fine, though please update try/except blocks to catch the specific exception that you are expecting so that we don't silently "work around" future bugs  (e.g. I think ValueError for the float casts, I'm not sure about the what might go wrong with the str() cast).  OK to self merge when ready, including merging as-is if you need this for tonight and want to post-facto discuss try/except philosophy :)

akremin:
Given the need for this for processing tonight, I'll self merge this PR and make the requested changes tomorrow in a new one. 

It's a very reasonable (and PEP-8 proper) request. That said, I genuinely do want it to pretend like nothing happened if anything goes wrong in trying to use that keyword. If the keyword can't be used I'd rather it move on with its life with its default value rather than crash the pipeline. But I know there are coding best-practices reasons why that's a bad idea. So maybe we do need to have that philosophical chat :)

sbailey:
Fine to be pragmatic for tonight.  In another semi-recent case we had something like the equivalent of a new bug causing `dat_header` to be undefined.  Generic try/except blocks that were intended to catch things like missing or malformed keywords were instead cascading through "robustness" logic instead of alerting us early that we had mis-merged and deleted the line that was supposed to be reading the header in the first place.  Or something like that.  But it reminded me of the general wisdom behind catching specific exceptions unless you really really need it to continue no matter what.

PR #1312, opened 2021-06-17
Afterburner for QSO catalog

echaussidon:
Hi @sbailey, 

I create two afterburners in the bin directory called: `desi_qso_mgii_afterburner` and `desi_qso_qn_afterburner`. I add also the  mgii_fitter in the py directory: `mgii_afterburner.py` (the code is provided by @corentinravoux and was presented [here](https://desi.lbl.gov/trac/attachment/wiki/TargetSelectionWG/TSTeleconMinutes/DESITS210113/QSO_completeness_Ravoux_13_01_2021_v2.pdf). I just speed up a lit a bit the process and change the i/o of the initial version.)

These two codes are designed to be run on the tile directory of each tile for each petal. The outputs give all the information that I need to reconstruct the catalog that I presented in the DESI collaboration meeting in June 2021 ([session  1b](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=6272)).

Example to run the code on the cumulative tile 80605 -- last_night 20210205 -- petal 3:
1.  `desi_qso_mgii_afterburner --target_version SV1 --target_selection restricted --coadd /global/cfs/cdirs/desi/spectro/redux/denali/tiles/cumulative/80605/20210205/coadd-3-80605-thru20210205.fits --zbest /global/cfs/cdirs/desi/spectro/redux/denali/tiles/cumulative/80605/20210205/zbest-3-80605-thru20210205.fits --output qso-mgii-afterburner-3-80605-thru20210205.fits
` --> in this configuration it takes 6s.
2. `desi_qso_qn_afterburner --target_version SV1 --coadd /global/cfs/cdirs/desi/spectro/redux/denali/tiles/cumulative/80605/20210205/coadd-3-80605-thru20210205.fits --zbest /global/cfs/cdirs/desi/spectro/redux/denali/tiles/cumulative/80605/20210205/zbest-3-80605-thru20210205.fits --output qso-qn-afterburner-3-80605-thru20210205.fits` -->  in this configuration it takes 15s.

It will require the last version of Redrock (not this which is actually loaded with the desi modules, I don't know how to select it so I tested it with my master branch updated) even if the code will run without errors.

**Remark:** Since some people may be interested by running the mgii fitter on all targets and not only on QSO targets + SPECTYPE != QSO. Set --target_selection all_targets runs the mgii fitter on all targets in the coadd file. In this configuration it takes (on a log node) 15s.

**PS:** When QuasarNP will be update I will remove _load_desi_coadd_ from `desi_qso_qn_afterburner` but the error is still not fix. (cf issue raises in QuasarNP)

Can you test it under the nominal desi environment ? 

@Cyeche, for the notification.

Edmond


sbailey:
Thanks @echaussidon .  I need a bit more time to test and review in detail, but so far only some fairly minor comments:

* please add an EXTNAME to the output files so that they can be ready by name instead of number (gives future flexibility for adding extensions or combining files)
* please add docstrings for all functions defining inputs and outputs.  e.g. `load_desi_coadd` is good, but others like `collect_redshift_with_new_RR_run` don't define all the inputs, and others like `extract_redshift_info_from_RR` don't have a docstring at all.
* let's make all floating point outputs *except* RA,DEC,redshifts single precision (float32) instead of double precision.  Although the individual files are small, the size difference adds up if stacking tables for all desi targets?
* I tried running your two examples (same inputs, different afterburners) and got a different number of targets in the output file.  What picks the default targets that are output?
* for the QN afterburner it would be helpful to have an option to apply it to *all* targets, including non-QSO targets, so that its outputs could be re-used for passing forward to the merged target list decisions (zmtl currently re-runs QN which is fast, but we may want to have it directly use this afterburner in the future).
* Please convert all print statements to use `log.debug()`, `log.info()`, `log.warning()`, or `log.error()` using `log = desiutil.log.get_logger()`.  That produces uglier output, but provides runtime control of verbosity and standardized messages (e.g. with line numbers).
* Please save the QN line list into header keywords so that we can know from the file itself which index is which line
* Change hardcoded /global/cfs/cdirs/desi/science/lya/qn_models/boss_dr12/qn_train_coadd_indtrain_0_0_boss10.h5 to use $QN_MODEL_FILE instead (or at least check that environment variable first, and then fall back to the hardcode if it isn't defined).  Currently that variable is set if load desitarget master, but soon we'll switch to having desispec itself set it).
* Let's make `c_thresh` and `n_thresh` command line options (with the same defaults 0.5 and 1) as you currently have.

Thanks

echaussidon:
Hi @sbailey. I resume here what I changed:

- Add EXTNAME for the two outputs (MGII and QN+RR)
- Add docstrings for each function
- Set f4 instead f8 when it was necessary
- Set logger instead of print
- Save C_Lines and Z_lines in independent columns
- Add n_thresh and c_thresh in the argpars
- Load $QN_MODEL_FILE by default

Additional points:
> Although the individual files are small, the size difference adds up if stacking tables for all desi targets?
- I don't understand, do you want to combine the two files ? I can do that setting the option clobber of fitsio in the argpars but I just want to be the more independent. 

>I tried running your two examples (same inputs, different afterburners) and got a different number of targets in the output file. What picks the default targets that are output?
- I was not clear, I only save objects which are selected by the afterburner ! So in the example you should find 2 targets for the afterburner and 10 for the QN. It is okay ? or do you want the result for every target ? for exemple in the QN afterburner I don't save QSO targets which are SPECTYPE == QSO & QN == QSO and |z_QN - Z_RR| < 0.05
- This leads to the point to add QN for all targets, it will depend if we stay like that we couldn't use the file that I produce as the output of QN since it will not contain every targets.
- I can change that but the size of the file that we will save will be higher. The basic configuration is the minimal afterburner for the clustering QSO Catalog but I agree to enlarge that to the non qso desi target. 

Edmond

echaussidon:
I add the same flag than mgii_afterburner in the qn_afterburner. Set --target_selection restricted or all_targets. 

- restricted: for QSO targets and QN+RR will be run on SPECTYPE != QSO and on SPECTYPE == QSO & |z_QN - z_RR| < 0.05
- all_targets: for all targets in the coadd file and QN+RR will be run on SPECTYPE != QSO or on SPECTYPE == QSO & |z_QN - z_RR| < 0.05

sbailey:
>> Although the individual files are small, the size difference adds up if stacking tables for all desi targets?

> I don't understand, do you want to combine the two files ? I can do that setting the option clobber of fitsio in the argpars but I just want to be the more independent.

You did what I was asking, converting f8 to f4 when possible in the output files, though let's make `Z_RR` and `Z_QN` also f8 instead of f4.  We don't need anything to combine the files in this PR; I was just thinking about when we have a production with 1000 qso_qn*.fits files someone is going to want to stack all of them into a single file and then f8 vs. f4 can start to make a difference on the side.

`--target_selection` option: we need some "all_targets" option that really means *all* targets so that the output file has the same targets as the input coadd.  When I tried `--target_selection all_targets` I got more targets than with `restricted`, but still far less than the 500 in the input file.  I think that is meaningful for QN even though most galaxy targets will result in very low confidence lines.  For MgII there is only a limited input redshift range where you can measure the MgII line, but even for redshifts outside of that range it would be helpful to have dummy values so that every input target is represented.

Please clarify:
* in the MgII files, is "Z" the original redshift from redrock, or a refit redshift just on the MgII line?
* in the QN files, I see "Z" and "Z_RR"; is one of them the original input redrock redshift and the other the refit one to match the Z_QN?  Please clarify which is which.

Thanks

echaussidon:
> --target_selection option: we need some "all_targets" option that really means all targets so that the output file has the same targets as the input coadd. When I tried --target_selection all_targets I got more targets than with restricted, but still far less than the 500 in the input file. I think that is meaningful for QN even though most galaxy targets will result in very low confidence lines. For MgII there is only a limited input redshift range where you can measure the MgII line, but even for redshifts outside of that range it would be helpful to have dummy values so that every input target is represented.

If you want a file with all outputs for each object without any selection, I will set a new option for this. (Keeping --target_selection to isolate qso_targets). **BUT** the weight of the generated files will be much larger than the current ones! (for the moment 3 objects in the mgii files and 25 in the qn files)

### Clarification:
- In  MgII: Z is the best fit in redrock from the zbest file, I will rename it Z_RR.
- In Qn: Z is the new redshift from the new run of RR, I will rename it Z_NEW. Z_RR is the current best fit in RR before the new run (ie) the value in the zbest file.

echaussidon:
I add:
- Column 'IS_QSO_MGII' and 'IS_QSO_QN_NEW_RR' to help the construction of the catalog
- Flag : `--save_target` with two options: restricted (save only objects with IS_QSO_MGII==True or IS_QSO_QN_NEW_RR==True (in function of each afterburner) and the option all saving all the information for the sample pass to the afterburner. For instance to have the output of 500 objects, you have to run `--target_selection all_targets --save_target all`. The minimal files are obtained with `--target_selection restricted --save_target restricted`.

bin/desi_qso_mgii_afterburner
```diff
@@ -0,0 +1,250 @@
...
+    param_mgii_fitter = {'lambda_width':args.lambda_width, 'max_sigma':args.max_sigma, 'min_sigma':args.min_sigma,
+                         'min_deltachi2':args.min_deltachi2, 'min_signifiance_A':args.min_signifiance_A, 'min_A':args.min_A}
+
+
+    if os.path.isfile(args.coadd) & os.path.isfile(args.zbest):
+        # open best fit file generated by redrock
+        zbest_file = fitsio.FITS(args.zbest)
+        zbest = zbest_file[1]
+        fibermap = zbest_file[2]
+        index_sorted = np.argsort(fibermap['TARGETID'][:500]) # in zbest targetid are sorted, it is not the case in the fibermap
```
sbailey, line 214: In the fibermap I think this works by luck of how they are written, but not for a general order of the rows.  This would be more robust and could handle any ordering of repeated TARGETID, and well as any number of TARGETIDs (e.g. the healpix grouping will not have multiple of 500s and different TARGETIDs could appear a different number of times):
```
index_sorted = np.unique(fibermap['TARGETID'], return_index=True)[1]
```

bin/desi_qso_mgii_afterburner
```diff
@@ -0,0 +1,250 @@
...
+            DESI_TARGET = 'SV1_DESI_TARGET'
+        elif args.target_version == 'SV3':
+            DESI_TARGET = 'SV3_DESI_TARGET'
+        elif args.target_version == 'MAIN':
+            DESI_TARGET = 'DESI_TARGET'
+        else:
+            log.error("**** CHOOSE CORRECT TARGET_VERSION FLAG (SV1 / SV3 / MAIN) ****")
+            sys.exit()
+
+        is_qso_target = fibermap[DESI_TARGET][:500][index_sorted] & 2**2 != 0
```
sbailey, line 226: See above; assuming that fibermap[DESI_TARGET][:500] are unique TARGETIDs that just have to be sorted is currently true, but not guaranteed in the future.  If using the above `np.unique(..., return_index=True), then you just need `fibermap[DESI_TARGET][index_sorted]`.

Please also use `desitarget.targetmask.desi_mask.QSO`, which happens to be 2**2, but is more formally correct.  Or at least add a comment that bit 2 is QSOs for SV1-3 and the DESI main survey.

bin/desi_qso_mgii_afterburner
```diff
@@ -0,0 +1,250 @@
...
+
+        if args.target_version == 'SV1':
+            DESI_TARGET = 'SV1_DESI_TARGET'
+        elif args.target_version == 'SV3':
+            DESI_TARGET = 'SV3_DESI_TARGET'
+        elif args.target_version == 'MAIN':
+            DESI_TARGET = 'DESI_TARGET'
+        else:
+            log.error("**** CHOOSE CORRECT TARGET_VERSION FLAG (SV1 / SV3 / MAIN) ****")
+            sys.exit()
```
sbailey, line 224: To simplify incorporating this into the pipeline, let's use `desitarget.targets.main_cmx_or_sv(fibermap)` to auto-derive which flavor this input is and which column name to use (with the added benefit that also supports SV2 and will support any other flavors of targeting that may appear in the future).

bin/desi_qso_mgii_afterburner
```diff
@@ -0,0 +1,250 @@
...
+            # Run MgII fitter only on QSO targets with SPECTYPE!=QSO objects to save time !
+            sel_to_mgii = is_qso_target & ~sel_RR
+        elif args.target_selection == 'qso_targets':
+            # Run MgII fitter only on QSO targets
+            sel_to_mgii = is_qso_target
+        elif args.target_selection == 'all_targets':
+            # Run MgII fitter only on all targets
+            sel_to_mgii = np.ones(500, dtype='bool')
+
+        log.info(f"Nbr objetcs for mgii: {sel_to_mgii.sum()}")
```
sbailey, line 239: typo (objetcs -> objects)

bin/desi_qso_qn_afterburner
```diff
@@ -0,0 +1,376 @@
...
+         #    err_redshift (numpy array): Array containing the associated error for the redshift
+         #    coeffs (numpy array): array containing the coefficient for the best fit given by RR
+         #                   even we work only with QSO template, it has a "shape" of redshift.size x 10
+         #                    WARNING: they have to be converted into a list (with .tolist()) to be added in the pandas dataframe
+
+        with fitsio.FITS(filename_zbest) as zbest:
+            redshift = zbest[1]['Z'][:]
+            err_redshift = zbest[1]['ZERR'][:]
+            coeffs = np.zeros((redshift.size, 10))
+            coeffs[:, :4] = zbest[1]['COEFF'][:]
```
sbailey, line 139: Please read DESI FITS file HDUs by name not number

bin/desi_qso_qn_afterburner
```diff
@@ -0,0 +1,376 @@
...
+                'filename_output_rerun_RR':args.filename_output_rerun_RR,
+                'filename_zbest_rerun_RR':args.filename_zbest_rerun_RR,
+                'delete_RR_output':args.delete_RR_output}
+
+    if os.path.isfile(args.coadd) & os.path.isfile(args.zbest):
+        # open best fit file generated by redrock
+        zbest_file = fitsio.FITS(args.zbest)
+        zbest = zbest_file[1]
+        fibermap = zbest_file[2]
+        index_sorted = np.argsort(fibermap['TARGETID'][:500]) # in zbest targetid are sorted, it is not the case in the fibermap
```
sbailey, line 346: See comments in desi_qso_mgii_afterburner about using `np.unique(..., return_index=True)` to be more robust for the future, as well as using `desitarget.targets.main_cmx_or_sv` to derive DESI_TARGET.

bin/desi_qso_mgii_afterburner
```diff
@@ -0,0 +1,250 @@
...
+        QSO_sel (pandas dataframe): contains all the information useful to build the QSO cat
+    """
+
+    # just to speed up the process and don't go in mgii_fitter function for nothing..
+    if sel_to_mgii.sum() == 0:
+        QSO_sel = pd.DataFrame()
+    else:
+        (index_selected_with_mgii_fit,
+        fits_result,
+        index_with_mgii_fit) = mgii_fitter(spectra_name, zbest_name, sel_to_mgii,
```
sbailey, line 83: Since the mgii fit is semi-expensive, please check that the `save_targets` input is a valid option before running the fit instead of risking running a long fit only to crash later because `save_targets = 'everything'`

bin/desi_qso_mgii_afterburner
```diff
@@ -0,0 +1,250 @@
...
+
+        QSO_sel['TARGETID'] = zbest['TARGETID'][:][index_to_save]
+        QSO_sel['RA'] = fibermap['TARGET_RA'][:500][index_sorted][index_to_save]
+        QSO_sel['DEC'] = fibermap['TARGET_DEC'][:500][index_sorted][index_to_save]
+        QSO_sel['Z_RR'] = zbest['Z'][:][index_to_save]
+        QSO_sel['ZERR'] = zbest['ZERR'][:][index_to_save]
+        QSO_sel['COEFFS'] = zbest['COEFF'][:][index_to_save].tolist()
+        QSO_sel['Z_RR'] = zbest['Z'][:][index_to_save]
+        QSO_sel['SPECTYPE'] = zbest['SPECTYPE'][:][index_to_save]
+        QSO_sel[DESI_TARGET] = fibermap[DESI_TARGET][:500][index_sorted][index_to_save]
```
sbailey, line 122: Another case of a fragile `[:500]` that won't work in general in the future (at least for for healpix grouping).

bin/desi_qso_mgii_afterburner
```diff
@@ -0,0 +1,250 @@
...
+                                  min_sigma, min_deltachi2, min_signifiance_A, min_A
+        DESI_TARGET (str): name of DESI_TARGET for the wanted version of the target selection
+        save_target (str) : restricted (save only IS_QSO_MGII==true targets) / all (save all the sample)
+    Returns:
+        QSO_sel (pandas dataframe): contains all the information useful to build the QSO cat
+    """
+
+    # just to speed up the process and don't go in mgii_fitter function for nothing..
+    if sel_to_mgii.sum() == 0:
+        QSO_sel = pd.DataFrame()
```
sbailey, line 79: Maintainability question: how committed are you to using pandas DataFrame?  It is a perfectly reasonable data object that happens to be in our dependency stack, but elsewhere we have standardized on astropy Tables and numpy named arrays as our supported table-like objects and prefer to not use DataFrames to minimize the number of slightly different syntaxes to keep in our head.  In this case I think some of the DataFrame/FITS work arounds you have to do for multi-dimensional columns might already be natively handled by astropy Tables .  Is there some specific reason to use DataFrames to get some convenience or performance that isn't provided by astropy Tables or numpy arrays, or was it just a default that you are familiar with?

I don't want to make you undo DataFrames just for the sake of reorganizing code, but if these every become problematic for another reason (like interfacing with fits...) we'd probably want to standardize then.

sbailey (changes requested):
Thanks; we're getting closer.  I added a few comments about clarity of what the mgii vs. qn afterburners mean by the various --target_selection options, but there is also an important thread of comments about assumptions about the input `fibermap` ordering and how `fibermap[:500][index_sorted]` isn't going to work in the future for healpix-based coadds+zbest.

bin/desi_qso_mgii_afterburner
echaussidon: Ok, I supposed that the form will be always the same (ie) by petal and then with only 500 objects in each redrock file which are sorted by redrock by TARGETID in the zbest file. So I will change !

bin/desi_qso_mgii_afterburner
echaussidon: ok

bin/desi_qso_mgii_afterburner
```diff
@@ -0,0 +1,250 @@
...
+    # Param for the MgII fitter see desispec/py/desispec/mgii_afterburner.py for additional informations
+    param_mgii_fitter = {'lambda_width':args.lambda_width, 'max_sigma':args.max_sigma, 'min_sigma':args.min_sigma,
+                         'min_deltachi2':args.min_deltachi2, 'min_signifiance_A':args.min_signifiance_A, 'min_A':args.min_A}
+
+
+    if os.path.isfile(args.coadd) & os.path.isfile(args.zbest):
+        # open best fit file generated by redrock
+        zbest_file = fitsio.FITS(args.zbest)
+        zbest = zbest_file[1]
+        fibermap = zbest_file[2]
```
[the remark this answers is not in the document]
echaussidon: changed

bin/desi_qso_mgii_afterburner
echaussidon: use `desitarget.targetmask.desi_mask.QSO`

bin/desi_qso_mgii_afterburner
echaussidon: ok

bin/desi_qso_qn_afterburner
echaussidon: changed

bin/desi_qso_qn_afterburner
echaussidon: ok

bin/desi_qso_mgii_afterburner
echaussidon: solved with new index_sorted

bin/desi_qso_mgii_afterburner
echaussidon: I check the validity of `--save_target` before running the function `select_targets_with_mgii_fitter`. Add this also in the qn afterburner

bin/desi_qso_mgii_afterburner
echaussidon: I had very bad experiences with astropy tables and fits opening with astropy few years ago. In addition, I was more familiar with pandas dataframe. I used it initially to build a catalog, pandas have great features for matching / unique / concatenation / visualization / ect ...

I agree the work around DataFrame/FITS is not great but I already implemented kind of similar interface so it was ok. 

If there is any problem in the future, I will modify it using `astropy.Table `

sbailey:
Thanks for the updates.  The one remaining issue is what `--target_selection` does for `desi_qso_qn_afterburner` vs. `desi_qso_mgii_afterburner`, so I will try to re-describe:

I think the argparse description for `desi_qso_qn_afterburner --target_selection` is incorrect:
```
--target_selection TARGET_SELECTION
  on each sample the mgII fitter is perform:
  restricted (QSO targets with SPECTYPE==GALAXY + QSO targets with SPECTYPE==QSO and |z_QN - z_RR| < 0.05) /
  all_targets (All targets in the coadd file)
```
i.e. from the code the "restricted" option appears to make a selection on QSO targets without any additional cuts on SPECTYPE or |z_QN-z_RR|.

Also, I find the difference in --target_selection interpretation between the two afterburners to be confusing:

| --target_selection | mgii afterburner | qn afterburner |
| ------------------ | ---------------- | --------------- |
| restricted                | QSO targets with SPECTYPE='GALAXY' | all QSO targets |
| qso_targets            | all QSO targets | --- |
| all                             | all targets | all targets |

Let's change `desi_qso_qn_afterburner --target_selection` "restricted" to "qso_targets" so that both afterburners use the same name for running on all input QSO targets.

Does that make sense?

If this is still confusing, I could just merge this PR and make the updates and you can then correct it if needed (I can't directly edit your fork, so would need to merge first).

Thanks

echaussidon:
I update the argparser (I copy/paste and I forget to modify it !). 

I change restricted --> qso_targets for the qn_afterburner, you are right there is no more selection.

Here the recap:

| --target_selection  | MgII  afterburner  | QN afterburner |
| :--------------- | :---------------| :----------|
| restricted  |   QSO targets with SPECTYPE='GALAXY'|  -  |
| qso_targets | only QSO targets             |   only QSO targets |
| all_targets  | ALL          |    ALL |

Let me know if there is additional stuff to change.

sbailey (approved):

sbailey:
Looks good, thanks for your patience with all the updates.  Merging now.

PR #1313, opened 2021-06-17
Enable I/O of the extra_catalog member.

sybenzvi:
Allow the Spectra I/O module to write and read the `extra_catalog` to and from FITS. The use case is when `extra_catalog` contains data like ZBEST tables from redrock.


sbailey (changes requested):
Looks good.  Please add a unit test to py/desispec/test/test_io.py that tests writing/reading spectra with and without the EXTRA_CATALOG HDU to make sure that someone else doesn't break it in the future.

Not your fault, but we are somewhat organically accreting spectra/coadd features that are beyond the original design, and keep breaking them too, so unit tests will provide some level of insurance for the future.  Thanks.

sybenzvi:
@sbailey, I updated the I/O test using the case with and without the extra_catalog member saved. Instances of Spectra without the extra_catalog HDU shouldn't cause a problem for read_spectra and write_spectra.

sbailey (approved):

sbailey:
Thanks for adding the tests @sybenzvi .  Since the `extra_catalog` check is only done if `extra_catalog is not None`, I added one more check to verify that `extra_catalog is not None` after reading the file back in because otherwise the test would silently incorrectly pass even if the reading was dropping `extra_catalog` (the reader works correctly as intended, this extra level of test just helps ensure that).

I'll merge once tests finish.

Issue #1314, opened 2021-06-17
Updated Daily Pipeline Missing EndOfXYZ Flags


akremin:
In solving a previous issue, we removed the dependency on request*.json files for the daily pipeline in favor of the raw data itself. However, this now ignores "exposures" meant to delineate the end of calibration scripts where there is no raw data file (but there is a `request` file).

A possible solution is to look for `manifest` files in addition to raw data, and run `summarize_exposure` on anything that matches either selection.

In general, the code is somewhat robust to this, as it also runs joint-fitting when a change of `OBSTYPE` occurs in incoming exposures, but it failed yesterday when additional flats were taken for GAIN analysis immediately after the normal flat sequence.

PR #1320, opened 2021-06-21
Look for manifests in nightly processing

akremin:
This solves issue #1314 where we were missing the "`EndOfArcs`", "`EndOfFlats`", etc. flags sent via a `manifest` file because those "exposures" don't contain a raw data file, and we recently updated the pipeline to look for the raw data instead of `request` files.

This PR looks for `manifest` files in addition to the raw data, so it doesn't add a dependency on manifests. If no `manifest` files exist, that is okay.

I tested this on the three most recent nights 20210618, 20210618, and 20210620. It worked for all three of those nights in identifying the manifest files while still identifying all of the data exposures.


sbailey:
Looks good, merge when ready.

Issue #1324, opened 2021-06-28
self.extra_catalog undefined for SpectraLite object in pixgrouping


michaelJwilson:
```
[...]

PR #1325, opened 2021-06-28
fix for self.extra_catalog undefined for SpectraLite object in pixgrouping

michaelJwilson:
One liner fix to #1324.

@sybenzvi wants to add a unit test for the SpectraLite class before merging.

### DIFF
--- a/py/desispec/pixgroup.py
+++ b/py/desispec/pixgroup.py
@@ -263,7 +263,8 @@ class SpectraLite(object):
         #- for compatibility with full Spectra objects
         self.meta = None
         self.extra = None
-
+        self.extra_catalog = None
+        
     def target_ids(self):
         """
         Return list of unique target IDs.
--- a/py/desispec/test/test_pixgroup.py
+++ b/py/desispec/test/test_pixgroup.py
@@ -96,28 +96,7 @@ class TestPixGroup(unittest.TestCase):
             fmap[s]["EXPID"] = s
         cls.fmap1 = encode_table(fmap)
 
-        fmap = empty_fibermap(cls.nspec)
-        fmap = add_columns(fmap,
-                           ['NIGHT', 'EXPID', 'TILEID'],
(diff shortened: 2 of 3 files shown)


sybenzvi:
Fix looks good. I'm going to try to add an I/O test for the `SpecLite` class to catch this issue.

sybenzvi:
An unintended consequence of changes in #1313 broke the I/O of `SpectraLite` with `read_spectra`. Fixed when @michaelJwilson added a null `extra_catalog` to `SpectraLite`. I added a single I/O unit test to the pixgroup test suite to make sure similar issues are caught in the future.

sbailey (approved):
Thanks for the fix @michaelJwilson and thanks for the tests @sybenzvi that would have caught the problem in the first place.

I took the liberty of removing the draft flag which apparently is necessary for submitting a review.  OK to merge as-is, but some curiosity comments / questions that don't have to be blocking factors:

Previously we haven't used `numpy.testing` (nt) but have done this instead:
```
nt.assert_array_almost_equal(A, B)  # this PR
self.assertTrue(np.allclose(A, B))  # previous style
```
the `numpy.testing` version produces more informative output if the test fails (good).  I'm not aware of any downsides to a "bare" exception being raised by `numpy.testing` vs. having the `unittest.TestCase` object (`self`) do the test and handle any failures, but I comment on that difference in case it raises any concerns.

Also, in this case I think the only difference should be due to the float32 cast for I/O, so a slightly more rigorous test would be something like
```
nt.assert_array_equal(spec.flux[band], self.flux[band].astype(np.float32).astype(np.float64))
```

sybenzvi:
@sbailey, FYI, I actually copied the call to `numpy.testing` from the `test_spectra` testing suite, so this is being used elsewhere.

About the float32->float64 cast that attempts to mimic the FITS I/O, it didn't work when I tried it. The arrays are still not quite the same after the manual cast, so I ended up not changing the test.

PR #1328, opened 2021-07-06
Modification for Everest run

echaussidon:
Hi @sbailey, 

To be sure and do not do something wrong with branch / upstream _ect_. with git. I prefer to create a PR for this branch.

I modified the two QSO-afterburners. I tested it and found the same output than previously. And I also checked that It works for healpix scheme. You can run for instance: 
`desi_qso_mgii_afterburner --target_selection all_targets --save_target all --coadd /global/cfs/cdirs/desi/users/sjbailey/dev/coadd/healpix/sv3/dark/272/27239/coadd-sv3-dark-27239.fits --zbest /global/cfs/cdirs/desi/users/sjbailey/dev/coadd/healpix/sv3/dark/272/27239/zbest-sv3-dark-27239.fits --output test.fits`

**PS 1:** This works only with coadd_fibermap branch from redrock

**PS 2:** I don't implement any retro-compatibility since these afterburners were not used for the moment. 

Evereything should be here, 

Edmond


echaussidon:
@Cyeche for the notification

sbailey (approved):

PR #1330, opened 2021-07-07
Coadd fibermap

sbailey:
This is still work in progress, but I'm opening a PR to get it on radar and having a more specific place for others to contribute.

Short version is that this splits the coadd and redrock(zbest) FIBERMAP into COADD_FIBERMAP with per-target information that is meaningful when coadded, vs. EXP_FIBERMAP which has per-exposure information (like FIBER_X, FIBER_Y).

Details from the 2021-07-06 data telecon presentation are in [DESI-6408](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=6408).  I'll add more details here when this is closer to being ready to merge.

@julienguy

PR #1335, opened 2021-07-09
Add sanity check readin the new zbest file

echaussidon:
the zbest HDU from the new zbest file (new run of RR) is not in the same order than previously. The TARGETID order from ZBEST HDU does not match the order from the input targetid. 

I add, a sanity reorder to don't be messed with this later.

Edmond


echaussidon:
It has no impact on the rest and solve the problem for the qn-afterburner. I tested that it gives what we expected. I directly merge it.

sbailey:
Thanks.  Please change this to be a PR to the "fiberassign_coadd" branch which has replaced "pixgroup".  I branched fiberassign_coadd off of pixgroup and then rebased off of master before starting PR #1330 which is where the active work is.

PR #1337, opened 2021-07-14
fix table loading typo

akremin:
Check for comma_replacement instead of a comma when trying to find things to replace with a comma. 

Done as part of table input-output to avoid commas in saved tables but avoid semicolons in working memory since bash doesn't like semicolons.


akremin:
Tested on the night that was a problem (20210503) where a semicolon in the BADAMPS column was crashing because of the typo. That not works.

I also tested on two other nights, 20201214 where no semicolons existed. It ran as before. I also ran on 20210529 where there are semicolons in informational columns and this didn't impact the ability to read that table in or use the table to submit jobs.

PR #1348, opened 2021-07-23
Include TILERA,TILEDEC,MJD in TSNR2/exposures output tables

### LINKED ISSUE #1255: MJD in tsnr file
Feel free to close with WONTFIX.  But looking at the survey margins now, I really want to be able to get things like twilight and the moon phase to correlate with the tsnr sky brightness.  But it's annoying to do this because there's no MJD tabulated, just NIGHT and EXPID.  It would be nice if MJD were tabulated.

OTOH, it only took 10 min to trawl this out of all the headers.

sbailey:
Everest wrapup PR:

This PR fixes #1255 by adding TILERA,TILEDEC,MJD to the production summary exposures (formerly tsnr2) and tiles files.  A side effect was to also fix more SURVEY="unknown" cases (e.g. dither tiles 80262-, now "cmx"), and fix SURVEY for tiles like 80607 (previously "cmx" because FAFLAVOR started with "cmx" even though FA_SURV="sv1"; according to https://desi.lbl.gov/trac/wiki/SurveyOps/TileDesigns those are sv1 tiles). 

Example files are in /global/cfs/cdirs/desi/users/sjbailey/dev/expfiles/ sampling one night per month from December 2020 through June 2021 to get a variety of programs.  The "master" files are current master branch and "new" are this branch.

**Feedback requested**:
For Everest we renamed the tsnr2*.fits to exposures-everest.fits since the file has outgrown the original tsnr-only usage and "exposures" is a more relevant name.  However, I noticed that the EXTNAMEs are still "TSNR2_EXPID" and "TSNR2_FRAME".  I suggest renaming these to "EXPOSURES" and "FRAMES", but I'm open to other suggestions.

@schlafly since you requested #1255, I'll specifically mention you for commentary.  @moustakas is another friendly-expert end-user who may have opinions about HDU names.

As a reminder, EXPOSURES/TSNR2_EXPID are metrics per exposure combined across cameras, while FRAMES/TSNR2_FRAME have per-camera metrics (a "frame" is a single camera from a single expid, but I suspect that name may not be obvious to non-core-pipeline folks).

I also made a cosmetic change of preserving column order from input -> output for the FRAMES/TSNR2_FRAME table; previously astropy was alphabetically sorting the columns when creating the initial table from a list of dictionaries (probably a leftover from pre-py3.6 days when dictionary keys were returned in an undefined order).

Note: this PR is into the "everest" branch, not master yet.

For the record, the command used to generate the test files:
```
desi_tsnr_afterburner --prod everest \
    -o exposures-new.fits \
    --tile-completeness tiles-new.csv \
    --aux /global/cfs/cdirs/desi/survey/observations/SV1/sv1-tiles.fits \
    --gfa-proc-dir /global/cfs/cdirs/desi/survey/GFA/ \
    --compute-skymags --update --nproc 32 --nights 20201214,20210114,20210214,20210314,20210414,20210514,20210614
```


sbailey:
Note: bin/desi_merge_redrock_files was already removed from everest branch in PR #1347 , but is also showing up here as being removed.  I think that is an extraneous side effect of a git rebase.  Focus on the TSNR2 stuff here.

schlafly:
Can I request that there remain a symlink with the original names to the new names?  Afternoon planning looks for the existing tsnr and tiles files.  I'll update AP to use the new names, but I'd appreciate it if there were a transition period where both names worked.  AP just reads in the first extension (I hadn't even realized that there were two extensions to the tsnr file!), so renaming the extensions won't cause any breakage.

Thumbs up for changing the extension names to EXPOSURES and FRAMES.  It's true that I didn't immedately register that as sum(camera) vs. N_exp * 10 petals * 3 cameras, but it's at least true that faced with both I would usually think that EXPOSURES is what I wanted.

sbailey:
Thanks for the comments.  For the daily production we can provide a tsnr-exposures.fits -> exposures-daily.fits link to ease the transition when we start up again in the fall.  For everest we'll just use the exposures-everest.fits name without a tsnr link.

I'll update the HDU names TSNR2_EXPID -> EXPOSURES and TSNR2_FRAME  -> FRAMES before merging.

Issue #1363, opened 2021-08-03
exp_fibermap table in spectra not being copied, stacked, or sliced


sybenzvi:
The `exp_fibermap` table added to the `Spectra` class for everest is not copied and stacked like other tables in `Spectra` objects. It is also not returned when slicing `Spectra`, though admittedly the difference between `fibermap` and `exp_fibermap` make this a little complicated.

sbailey:
Thanks for reporting and submitting PR #1364 to fix it.  I'll reopen this ticket until the PR is actually merged and the fix is in place.

PR #1364, opened 2021-08-03
Sybenzvi/expfibermapfix

sybenzvi:
This should fix issue #1363; slicing and stacking of `Spectra` will now include `exp_fibermap`. Need to update unit tests to confirm.


sybenzvi:
Unit tests committed, should be ready for merging.

sbailey (approved):
Looks good.  Thanks for the fix and for adding tests so that we don't break it again in the future.

PR #1367, opened 2021-08-04
Bad sky fibers

julienguy:
Detect sky fibers with contamination for star or galaxy and mask them.


akremin (approved):
This looks good to me. The only thing I could find is a lot of comparisons and assignments of an array of floats to an int. But the performance hit from that would be negligible, so I don't think it needs to be changed.

PR #1368, opened 2021-08-04
Adjust trace x when computing image model

julienguy:
- This PR adds a fit of a spectral traces offset (along x=cross-dispersion) between the input psf and the current image in the function compute_image_model. This function compute_image_model is called during preprocessing to build an inverse variance image.
- It improves the variance model, because there are variations as large as 1 pixel of the fiber traces location in the CCDs, expecially in z (that have been found to be correlated with humidity in the shack).
- This feature explains why we often have bad sky residual chi2 in z8 in the exposure QA (the default PSF for z8 is off by 1 pixel for some exposures).

This computation takes less than 2 seconds per frame (on one cori haswell process).

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,9 @@ desispec Change Log
 0.45.4 (unreleased)
 -------------------
 
-* No changes yet.
+* Adjust spectral traces when computing CCD variance in preprocessing (PR `#1368`_).
+
+.. _`#1368`: https://github.com/desihub/desispec/pull/1368
 
 0.45.3 (2021-07-29)
 -------------------
--- a/py/desispec/image_model.py
+++ b/py/desispec/image_model.py
@@ -85,13 +85,7 @@ def compute_image_model(image,xytraceset,fiberflat=None,fibermap=None,with_spect
         dx = np.median(dx)
         log.info("measured trace shift dx = {:.3f} pixel".format(dx))
         log.info("dx fit took {:.2f} sec".format(time.time()-t0))
-        print(xytraceset.x_vs_wave_traceset._coeff.shape)
         xytraceset.x_vs_wave_traceset._coeff[:,0] += dx
-
-
-
-
-
 
     # first perform a fast boxcar extraction
     log.info("extract spectra")
--- a/py/desispec/image_model.py
+++ b/py/desispec/image_model.py
@@ -8,6 +8,7 @@ almost not correlated with pixel value.
 import os
 import numpy as np
 import copy
+import time
 
 import numba
 import scipy.interpolate
@@ -18,7 +19,7 @@ from desispec.io.xytraceset import read_xytraceset
 from desispec.qproc.qextract import qproc_boxcar_extraction
 from desispec.qproc.qsky import qproc_sky_subtraction
 from desispec.qproc.qfiberflat import qproc_apply_fiberflat
-
+from desispec.trace_shifts import compute_dx_from_cross_dispersion_profiles
 
 @numba.jit
 def numba_proj(image,x,sigma,flux) :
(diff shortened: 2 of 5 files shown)


julienguy:
This PR (combined with PR#1367) solves the issue with QA sky chi2. As an example, for tile 1093, see the old and new QA plots.
old: https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/1093/20210610/tile-qa-1093-thru20210610.png
new:  https://data.desi.lbl.gov/desi/spectro/redux/jguy-tile-qa/tiles/cumulative/1093/20210610/tile-qa-1093-thru20210610.png

But more importantly, **this changes significantly the variance in spectro SP8 (by 20% for tile 1093)**, and only of 1-2% in the other spectrographs. **This impacts the values of TSNR2 and the values of redrock DELTACHI2**.
This will have to be taken into account when producing the **final SV analysis using the Fudji data release**.

spectro 0 redrock DELTACHI2 ratio new/old= 0.993
spectro 1 redrock DELTACHI2 ratio new/old= 0.983
spectro 2 redrock DELTACHI2 ratio new/old= 0.969
spectro 3 redrock DELTACHI2 ratio new/old= 0.969
spectro 4 redrock DELTACHI2 ratio new/old= 0.975
spectro 5 redrock DELTACHI2 ratio new/old= 0.982
spectro 6 redrock DELTACHI2 ratio new/old= 0.997
spectro 7 redrock DELTACHI2 ratio new/old= 0.991
spectro 8 redrock DELTACHI2 ratio new/old= 0.837
spectro 9 redrock DELTACHI2 ratio new/old= 0.957

spectro 0 TSNR2_ELG ratio new/old= 0.994
spectro 1 TSNR2_ELG ratio new/old= 0.992
spectro 2 TSNR2_ELG ratio new/old= 0.983
spectro 3 TSNR2_ELG ratio new/old= 0.993
spectro 4 TSNR2_ELG ratio new/old= 0.987
spectro 5 TSNR2_ELG ratio new/old= 0.986
spectro 6 TSNR2_ELG ratio new/old= 1.026
spectro 7 TSNR2_ELG ratio new/old= 0.997
spectro 8 TSNR2_ELG ratio new/old= 0.800
spectro 9 TSNR2_ELG ratio new/old= 0.990

TSNR2_ELG

julienguy:
An illustration of the effect, for the preprocessed image preproc-z8-00094685.fits.
Top: signal
Middle: new inverse variance
Bottom: old inverse variance

Screenshot from 2021-08-05 10-11-11


One can see the traces for the inverse variance in the bottom panel are significantly shifted.
They appear irregular because the code switches to the signal value instead of the model value when the discrepancy between the two exceeds a threshold.

akremin:
Q1) Am I reading these changes correctly that this code existed prior to this pull request, but wasn't actually used?

Q2) I see that `fit_x_shift` is set to `True` by default in `compute_image_model()`, which in turn enables it in the preproc code because that parameter isn't set when `compute_image_model()` is called. Do we want to keep this as an implicit low-level choice or explicitly define the option in preproc (or even `desi_proc`)? 

I'll leave Q2 up to you to choose. It is very fast and the correction is clearly important, so I am fine with it being a low-level implicit option.

julienguy:
The code to adjust the trace coordinates exists and is used prior to extracting the spectra (in `desi_compute_trace_shifts`), but it was not used for preprocessing because I was correctly expecting trace coordinates shifts to be only of about 0.1 pixel.

I will add an option `--no-traceshift` to `desi_preproc` to disable this feature.

I will also manage the conflict in  `changes.rst`.

akremin (approved):
Thank you for fixing the readme and addressing my questions. The changes look good. If they pass your tests, I think it is good to merge.

PR #1372, opened 2021-08-12
Add the Poisson noise from the zero exposure dark current (=bias value)

julienguy:
Add the Poisson noise from the zero exposure dark current in the read noise.
A good estimate of this Poisson noise variance is the master bias value (which had the electronic pedestal subtracted), after multiplication by the gain (to convert ADUs to electrons).

This correction increases the read noise std deviation by 10% for the blue cameras. It explains the variation of the measured read noise rms with the CCD row or column.

The average ratio of the noise measured in the active region of the CCDs to its estimate based on the overscan columns and the master bias and darks is now of 0.994,0.985,0.985 in BLUE,RED
and NIR CCDs respectively. 

See https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=6431.

### DIFF
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -623,11 +623,17 @@ def preproc(rawimage, header, primary_header, bias=True, dark=True, pixflat=True
             overscan_col += o
             rdnoise  += r
             if bias is not False :
+                # the master bias noise is already in the raw data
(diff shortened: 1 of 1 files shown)


akremin (approved):
Apologies I didn't notice this request until now. The changes looks good

Issue #1365, opened 2021-08-03
coadding camera is slow


moustakas:
I haven't dug to find the bottleneck(s), but `desispec.coaddition.coadd_cameras` is painfully slow (3-4 seconds for a `Spectra` object with a single galaxy). This issue doesn't affect the pipeline but it does affect downstream analysis with `fastspecfit` and work by at least one GQP WG member.

sybenzvi:
@moustakas, this also affects the transient pipeline, which requires coaddition across the cameras. I'll see if I can profile it and make some headway.

sybenzvi:
Prompted by @sbailey, I got around to running a line profiler. When running `coadd_cameras` on one spectrum, 98% of the execution time is spent on two lines checking the alignment of the camera wavelength grids (see [here](https://github.com/desihub/desispec/blob/a407ce88ca6c30bdc8a36f7202ebba874f836b0c/py/desispec/coaddition.py#L496)) and then repeating the same expensive calculation at the start of coaddition (see [here](https://github.com/desihub/desispec/blob/a407ce88ca6c30bdc8a36f7202ebba874f836b0c/py/desispec/coaddition.py#L533)). When calling `coadd_cameras` on a full petal, these two lines still take up 45% of the execution time.

The line profile for coadding one spectrum is attached.

[coadd_camera_prof.py.001.lprof.txt](https://github.com/desihub/desispec/files/7004067/coadd_camera_prof.py.001.lprof.txt)

sybenzvi:
I added a unit test to check the new calculation. I should have written the test before submitting PR #1377, but I did check offline that the also test passes with the old `coadd_cameras` code, so I'm confident about the changes.

moustakas:
Done in #1377. Thanks @sybenzvi!

PR #1377, opened 2021-08-18
Speed up coadd_cameras

sybenzvi:
Effort to address issue #1365 by reducing and speeding up comparisons of the camera wavelength grids in `coadd_cameras`.

### DIFF
--- a/py/desispec/coaddition.py
+++ b/py/desispec/coaddition.py
@@ -498,8 +498,9 @@ def coadd_cameras(spectra, cosmics_nsig=0., onetile=False) :
         nwind, nspec = len(windices), len(spectra.wave[b])
         log.debug("{}: {} matched indices; {} wavelength values".format(b, nwind, nspec))
         if nwind != nspec:
-            log.error("Cannot directly coadd the camera spectra because wavelengths are not aligned, use --lin-step or --log10-step to resample to a common grid")
-            sys.exit(12)
+            errmsg = 'Cannot directly coadd the camera spectra because wavelengths (band "{}") are not aligned. Use --lin-step or --log10-step to resample to a common grid'.format(b)
+            log.error(errmsg)
+            raise ValueError(errmsg)
         number_of_overlapping_cameras[windices] += 1
         windict[b] = windices
 
--- a/py/desispec/coaddition.py
+++ b/py/desispec/coaddition.py
@@ -490,7 +490,8 @@ def coadd_cameras(spectra, cosmics_nsig=0., onetile=False) :
             wave=np.append(wave,spectra.wave[b][spectra.wave[b]>wave[-1]+tolerance])
     nwave=wave.size
 
-    # check alignment
+    # check alignment, caching band wavelength grid indices as we go
+    windict = {}
     number_of_overlapping_cameras=np.zeros(nwave)
     for b in spectra.bands :
         windices=np.argmin((np.tile(wave,(spectra.wave[b].size,1))-np.tile(spectra.wave[b],(wave.size,1)).T)**2,axis=1)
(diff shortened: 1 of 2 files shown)


sybenzvi:
Achieved some speedup by:
1. caching the search for overlapping wavelength values across cameras;
2. switching from numpy.tile + distance calculations to a tolerance calculation with numpy broadcasting.

Note: fix (2) is leads to a 3x speedup when calling `timeit`, but the line profiler doesn't show any particular advantage.

py/desispec/coaddition.py
```diff
@@ -490,16 +490,18 @@ def coadd_cameras(spectra, cosmics_nsig=0., onetile=False) :
...
+    # check alignment, caching band wavelength grid indices as we go
+    windict = {}
     number_of_overlapping_cameras=np.zeros(nwave)
     for b in spectra.bands :
-        windices=np.argmin((np.tile(wave,(spectra.wave[b].size,1))-np.tile(spectra.wave[b],(wave.size,1)).T)**2,axis=1)
-        dist=np.sqrt(np.max(spectra.wave[b] - wave[windices]))
-        log.debug("camera {} max dist= {}A".format(b,dist))
-        if dist > tolerance :
-            log.error("Cannot directly coadd the camera spectra because wavelength are not aligned, use --lin-step or --log10-step to resample to a common grid")
+        windices = np.argwhere((np.abs(spectra.wave[b][:,None] - wave) < tolerance).any(0)).ravel()
```
sbailey, line 497: This still generates a large temporary 2D array in memory and I suspect that we could still get much much faster with a series of 1D operations, e.g. something like:
```
imin = np.argmin(np.abs(spectra.wave[b][0] - wave))
windices = np.arange(imin, imin+len(spectra.wave[b]), dtype=int)
dwave = spectra.wave[b] - wave[windices]
if np.any(np.abs(dwave) > tol):
    raise ValueError("Input wavelength grids aren't aligned...")
```
But that would need to be profiled to see if it actually is faster (and debugged to make sure it is correct).

py/desispec/coaddition.py
```diff
@@ -490,16 +490,18 @@ def coadd_cameras(spectra, cosmics_nsig=0., onetile=False) :
...
+        nwind, nspec = len(windices), len(spectra.wave[b])
+        log.debug("{}: {} matched indices; {} wavelength values".format(b, nwind, nspec))
+        if nwind != nspec:
+            log.error("Cannot directly coadd the camera spectra because wavelengths are not aligned, use --lin-step or --log10-step to resample to a common grid")
             sys.exit(12)
```
sbailey, line 502: I know this predates this PR, but since we're making changes, it would be better to raise a ValueError (inputs have right type but invalid values) instead of calling `sys.exit(12)`.

sbailey (commented):
Thanks for this speed improvement and for adding tests.  If you are willing to push on this a little bit more, I suspect that we could still get way faster.  See inline comment about a 1D method that doesn't require making a temporary 2D array.  Whether that is worth your time depends upon whether this speedup is already good enough or still irritatingly slow.  Back of the envelope: 2-3x faster on a 3-4 seconds still sounds problematically slow for a coadd operation that smells like it should be well under a second per target.

py/desispec/coaddition.py
sybenzvi: What, the meaning of `sys.exit(12)` isn't obvious to all users ;) ?
JK, I agree, this should be changed. It's an easy fix.

py/desispec/coaddition.py
sybenzvi: True, it's not memory efficient and we may be getting hurt even with a vectorized loop over the 2D structure. I'll mess around to see if there are more gains to be had.

sybenzvi:
> Thanks for this speed improvement and for adding tests. If you are willing to push on this a little bit more, I suspect that we could still get way faster.

@sbailey, your instinct was correct. Moving from a 2D broadcast to the 1D index check reduced the execution time for one spectrum to 40 ms. See the latest commit.

sbailey:
@sybenzvi great, thanks.  Let's call this good enough for now and merge, and if coadding across cameras is still problematically slow we can revisit the next level of profiling and optimizations in future PRs.

sybenzvi:
For the record, with this fix 75% of the execution time of the function is spent instantiating a new `Spectra` and computing new scores. The indexing is now pretty negligible.

PR #1382, opened 2021-08-23
swap fibers 3402 and 3429 if needed

sbailey:
This PR intercepts and corrects for the swapped fibers 3402 and 3429 in DESI data through July 2021.  It checks for the known incorrect FIBER:LOCATION and corrects only if needed.

Example:
```
assemble_fibermap -n 20201215 -e 68040 -o fibermap-00068040.fits
```
Python check:
```
import numpy as np
fa1 = Table.read('/global/cfs/cdirs/desi/spectro/redux/everest/preproc/20201215/00068040/fibermap-00068040.fits')
fa2 = Table.read('fibermap-00068040.fits')
fa1.sort('LOCATION')
fa2.sort('LOCATION')
for col in ['TARGETID', 'LOCATION', 'TARGET_RA', 'TARGET_DEC', 'DESI_TARGET']:
    assert np.all(fa1[col] == fa2[col])

ii1 = (fa1['FIBER'] == 3402) | (fa1['FIBER'] == 3429)
ii2 = (fa2['FIBER'] == 3402) | (fa2['FIBER'] == 3429)
print('Original swapped:\n', fa1['TARGETID', 'LOCATION', 'FIBER'][ii1])
print('New corrected:\n', fa2['TARGETID', 'LOCATION', 'FIBER'][ii2])
```
Prints
```
Original swapped:
      TARGETID      LOCATION FIBER
------------------ -------- -----
616088605995041626     6098  3402
 39627853691620387     6099  3429
New corrected:
      TARGETID      LOCATION FIBER
------------------ -------- -----
616088605995041626     6098  3429
 39627853691620387     6099  3402
```

Issue #1380 .  @akremin and/or @julienguy please double check.

### DIFF
--- a/py/desispec/io/fibermap.py
+++ b/py/desispec/io/fibermap.py
@@ -461,15 +461,17 @@ def assemble_fibermap(night, expid, badamps=None, force=False):
     fa = Table.read(fafile, 'FIBERASSIGN')
     fa.sort('LOCATION')
 
-    #- Dec 2020 - July 2021 had a LOCATION:FIBER swap for fibers 3402 and 3429;
-    #- check and correct if needed; see desispec #1380
+    #- Tiles designed before Fall 2021 had a LOCATION:FIBER swap for fibers
+    #- 3402 and 3429 at locations 6098 and 6099; check and correct if needed.
+    #- see desispec #1380
     #- NOTE: this only swaps them if the incorrect combination is found
-    iloc6098 = np.where(fa['LOCATION'] == 6098)[0][0]
-    iloc6099 = np.where(fa['LOCATION'] == 6099)[0][0]
-    if (fa['FIBER'][iloc6098] == 3402) and (fa['FIBER'][iloc6099] == 3429):
-        log.warning(f'FIBERS 3402 and 3429 are swapped in {fafile}; correcting')
-        fa['FIBER'][iloc6098] = 3429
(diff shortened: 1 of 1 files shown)


akremin (commented):
This looks good. Straight forward swap and the documentation shows that it works. I just have two quick questions to verify we'll never run into a corner case --

Verification one: Do we also want this under a `if night<20210901` check, or are we comfortable with the fact that it is an infinitesimal small chance of two new fibers on a new petal being labeled and placed in this configuration?
Second verification: Could there ever be a circumstance where we don't have a fibermap row with `LOCATION` equal to 6098 or 6099? Lines 467 and 468 will raise index errors if either doesn't exist.

julienguy (commented):
I agree with Anthony testing for the locations and adding a condition night < maxnight would be safer.

sbailey:
Good point about the possibility of not having LOCATION=6098/6099, which I think could occur if petal 6 was not included in the instance.  I added that check.

I had considered the night < maxnight check, but that would prevent the correct swap if we ever re-observed one of the existing tiles, e.g. as a repeatability consistency check.  Given the current focal plane work and changed FP calibs, it is unlikely we'd reobserve an existing tile, but it's perhaps more likely than the 1/(50*49) = 4e-4 chance that a new petal would have its correct mapping be exactly the current wrong mapping (at which point we could then add the maxnight check).

PR #1386, opened 2021-08-26
Emlinefit

araichoor:
This PR adds the `desi_emlinefit_afterburner` script.
This script computes standard emission lines for all 500 spectra of a given set of {redrock,coadd} file.

**Expected calling sequence in the pipeline:**
`desi_emlinefit_afterburner --redrockfn $RRFN --coaddfn $COADDFN --outfn $OUTROOT.fits`

**Functions:**
- `get_emlines()`: called by `main()`, reads redrock/coadd files, calls `emlines_gaussfit()`, outputs results, in the format of a dictionary;
- `emlines_gaussfit()`: called by `get_emlines()`, performs the fit for a single spectrum and a single emission line;
- `get_rf_em_waves()`: called by `emlines_gaussfit()`, defines the vacuum, rest-frame emission wavelengths.
 
**Emission lines:**
Current fitted lines are: OII (doublet, free ratio), Hdelta, Hgamma, Hbeta, OIII (doublet, fixed ratio), Halpha.
For each line, I report the columns {EMNAME}_{QUANT} with QUANT in:
- OBSEMWAVES: observed wavelengths at which the fit is done
- CHI2, NDOF: chi2 and nb of degrees of freedom
- CONT, CONT_IVAR: continuum in 1e-17 * erg/cm2/s/A
- FLUX, FLUX_IVAR: flux in 1e-17 * erg/cm2/s/A
- SIGMA, SIGMA_IVAR: line width in A (observed frame)
- SHARE, SHARE_IVAR: f1/(f0+f1) for OII and OIII doublets
- RFEW, RFEW_IVAR: rest-frame equivalent width
- wave, data, ivar, model: data used for fitting + fitted model, used in case plotting is requested
I also propagate some columns from the REDSHIFTS/FIBERMAP extensions:
- TARGETID, TARGET_RA, TARGET_DEC, OBJTYPE, FIBER, Z, ZWARN, SPECTYPE, DELTACHI2.

**Example files:**
In `/global/cscratch1/sd/raichoor/tmpdir/desispec_emlinefit/`, there are fits and log files:
```
denali/tiles/cumulative/80606/20201219/emlinefit-?-80606-thru20201219.{fits,log}
denali/tiles/cumulative/80613/20210324/emlinefit-?-80606-thru20201219.{fits,log}
everest/tiles/cumulative/1/20210406/emlinefit-?-1-thru20210406.{fits,log}
everest/tiles/cumulative/1000/20210517/emlinefit-?-1000-thru220210517.{fits,log}
everest/tiles/cumulative/20000/20210609/emlinefit-?-20000-thru20210609.{fits,log}
everest/tiles/cumulative/80606/20201219/emlinefit-?-80606-thru20201219.{fits,log}
```
I also provide in the same folder few examples of the (optional) output pdf; it has been useful to me to make diagnoses (and debug!):
```
denali/tiles/cumulative/80606/20201219/emlinefit-0-80606-thru20201219.pdf
denali/tiles/cumulative/80613/20210324/emlinefit-0-80613-thru20210324.pdf
everest/tiles/cumulative/80606/20201219/emlinefit-0-80606-thru20201219.pdf
everest/tiles/cumulative/80613/20210324/emlinefit-0-80613-thru20210324.pdf
```

**Speed:**
On the logging node, processing one file takes 10s-20s; generating the pdf takes much longer (~6mins).

**Choices:**
- I try to output nans for failed/dubious fits
- for simplicity, the code generates outputs for the 500 spectra (though many fits will be nans, or won t make sense)
- emlines_gaussfit() has the flexibility to be run on a subset of TARGETIDs only; same I kept that in case useful for testing; could be removed (that would simplify the code)
- I propagate as input arguments some of the emission line settings (rf_fit_hw, min_rf_fit_hw, rf_cont_w, rv, balmerfit), in case people want to test those; could be removed
- in get_emlines(), I allow pre-everest format (i.e. ZBEST instead of REDSHIFTS extension in redrockfn)
- the infos in  log files are quite minimalist; happy to add more if relevant/useful
- I do not propagate {DESI,BGS,MWS,SCND}_TARGET columns, as that would imply to handle SV1/SV2/SV3/Main; could do if useful
- in args.outpdf, spectra are ordered by increasing redshifts.


sbailey:
Thanks for the explanations @araichoor .

@moustakas any more comments on this?  In particular for data model names (SHARE, RFEW, ...) and units?  Algorithmically fastspecfit and emlinefit can be different, but let's try to have common variable names and units for the same concepts.  Could you explicitly cross check that and many any suggestions?  Thanks.

moustakas:
Apologies for the delay.

Algorithmically, @araichoor has already shown that there is good agreement between his measurements and my independent measurements with `fastspecfit`, so I don't intend to review the code. I'll just add some notes and questions. Apologies in advance for the length of my report!

* First, I agree that we should make the rest-frame wavelengths consistent. I've been using [PyNeb](
https://github.com/Morisset/PyNeb_devel), which uses the latest atomic data, to get these wavelengths and in any case our disagreements are mostly at the third decimal point.

* My second question is: why these lines and not others? For example, the [OII] and [OIII] doublets and H-beta make sense for a code focused on ELGs, but if you're fitting H-alpha why not fit the adjacent [NII] doublet? Presumably it's too weak, but then so are H-gamma and H-delta. In addition, as you know, the Balmer lines are all affected by stellar absorption (becoming fractionally more important for the higher-order lines), so that will affect the measured Balmer fluxes except in galaxies with very large specific star formation rates (equivalent widths). But maybe it doesn't matter given the goals of `emlinefit`?

* Third, regarding the data model, to make it easy to compare our two codes, I've pasted below the reported measurements from `emlinefit` for just the [OII] doublet and for the `[OII] 3726` line from `fastspecfit` (see also the [fastspec data model](https://fastspecfit.readthedocs.io/en/latest/fastspec.html)). Here are my reactions / thoughts / comments but please feel free to ignore any and all of my suggestions!
    * It's odd to me that `OBSEMWAVES` is a byte array rather than a 2-element single-precision array. Also, the observed wavelengths are derivable from the redshift and the rest-frame wavelengths so perhaps they don't need to be stored (although it's clearly convenient!)?
    * `NDOF` should be an integer, not float array.
    * Is your `CHI2` the reduced chi^2 (mine *is* reduced)?
    * How strongly do you feel about the suffix `RFEW` (`emlinefit`) vs `EW` (`fastspecfit`)? In my upbringing as an optical spectroscopist, equivalent widths are *always* rest-frame, hence my simpler suffix. 
    * In my code, the peak amplitude of the line is the critical measurement, not the integrated flux. (The reason is that you can have a broad line totally buried in the noise which is "significant" based on the integrated flux but which is *insignificant* based on the ratio of the line-amplitude to the rms of the surrounding continuum. I've done a lot of visual inspection on these two approaches and have found using the amplitude much more akin to what you and I would do to identify significant lines by eye.)
    * I'm a little unclear about the `SHARE` suffix and whether it should be a reported quantity in the core measurements (as opposed to a quantity-of-convenience that an end-user should construct themselves). Since you're fitting each component of the [OII] doublet independently, in order to construct this quantity, presumably you have to make some choices about the significance of each of the components of the doublet. For example, if the [OII] 3726 line-flux is negative (let's say it's affected by problematic sky-subtraction) but the [OII] 3729 line-flux is positive (and real), do you still use both lines to combine them? In `fastspecfit` I opted to keep each line separate (although in my current branch I've decided to fix the doublet ratio) and let the end-user combine the individual lines however they need or want depending on their application(s).

    **emlinefit data model (one line)**
    ```
           name        dtype  n_bad
    ----------------- ------- -----
       OII_OBSEMWAVES bytes20     0
             OII_FLUX float32   129
        OII_FLUX_IVAR float32   129
            OII_SIGMA float32   129
       OII_SIGMA_IVAR float32   129
             OII_CONT float32    61
        OII_CONT_IVAR float32    61
            OII_SHARE float32   129
       OII_SHARE_IVAR float32   129
             OII_RFEW float32   129
        OII_RFEW_IVAR float32   129
             OII_CHI2 float32   129
             OII_NDOF float32   129
    ```
    **fastspecfit data model (one line)**
    ```
             name           dtype  shape
    ---------------------- ------- -----
              OII_3726_AMP float32
         OII_3726_AMP_IVAR float32
             OII_3726_FLUX float32
        OII_3726_FLUX_IVAR float32
          OII_3726_BOXFLUX float32
           OII_3726_VSHIFT float32
            OII_3726_SIGMA float32
             OII_3726_CONT float32
        OII_3726_CONT_IVAR float32
               OII_3726_EW float32
          OII_3726_EW_IVAR float32
       OII_3726_FLUX_LIMIT float32
         OII_3726_EW_LIMIT float32
             OII_3726_CHI2 float32
             OII_3726_NPIX   int32
    ```
  * It's probably my history with `IDL`, but I'm not at all a fan of using `nan` to represent non-measurements. In my world, `nan` means something went terribly wrong and needs to be investigated. In `fastspecfit` if the `ivar` of a measurement is zero then it means no measurement was possible.
  * It looks like you don't allow a velocity-shift of the line-center with respect to the `redrock` redshift, which `fastspecfit` does. I haven't investigated how much this flexibility is needed, but in Denali I found that the 1-sigma scatter in the Balmer-line redshifts compared to the nominal redrock redshifts were about 30 km/s (less than a pixel over most of our wavelength range); for the forbidden lines the scatter was a bit larger, roughly 60 km/s, although with a tail out to +/- 100 km/s (>3 pixels). It may be worth thinking about this point more (but not for this PR).
* It doesn't look like you use the resolution matrix---why not? I guess this means that your measured line-widths (which are in observed-frame Angstroms) need to be deconvolved for the instrumental resolution before they're converted to km/s.
* Why do you fit all 500 spectra, including the sky spectra? How are the fits to the non-TGT spectra used? 
* You write that it takes 10-20 seconds to fit a single petal (500 spectra), which is shockingly fast---nice! However, I don't see any multiprocessing in the code. Presumably with multiprocessing you could fit 32 spectra simultaneously and get this time down to <1 second. Along the same lines, how is the code MPI-parallelized, if at all?

Finally, a broader question (especially for @sbailey) that is semi-related to this PR but which should not be a blocking factor. We now have two after-burners for emission lines, `emlinefit` for ELGs and the MgII after-burner for QSOs. It has taken some time to evolve, but `fastspecfit` now does both these tasks and much more, delivering quantities that are basically science-ready for all extragalactic objects (well, the QSOs are hard!). 

With more capabilities, of course, there is a computational cost. `fastspecfit` is significantly slower than either of these existing after-burners, although my latest refactor sped things up by about a factor of 10 (https://github.com/desihub/fastspecfit/issues/37) and I think with more streamlining (read: remove as many of the `astropy` wrappers as possible!) the code can be sped up even more. Something to discuss and think about.

moustakas (approved):

araichoor:
thanks a lot for all the relevant comments/questions!

I answer below to each item (sorry for the long message).

But first a general comment relevant for several items: this afterburner is primarily designed to be used for the ELG zreliable criterion , which currently relies on the OII SNR; maybe the OIII SNR will be used later, but probably not much more.
I know it is very tempting to improve/refine things, but then it will just try to reproduce fastspecfit, which is way more precise/developed.
That s why I m more enclined to keep things as simple as possible, with of course knowing that the provided measurements are coarse ones; but those are sufficient for the ELG zreliable criterion.
The other provided measurements may be useful for users looking for simple diagnoses.

Now going per item:
- rest-frame wavelengths. => no problem, I ll switch to your (PyNeb) values;
- why these lines and not others? => as said, that was initially for ELGs; so I started with OII, then added OIII; and the first Balmer lines where suggested then..
- data model:
 - OBSEMWAVES is a byte array => I came up with a string because I ve one entry for the doublets, so two wavelengths stored in one array element; as you mentioned, that is mainly for conveniency; one could always go back to the code, and get it with the redshifts; I can remove that column;
 - NDOF should be an integer, not float array => that s because of the np.nan, I believe; if that s an issue I could try to find a work-around
 - Is your CHI2 the reduced chi^2 (mine is reduced)? => no; I could provide CHI2/NDOF (to me it would sound better to then call it RCHI2, but I guess you re not going to change that in fastspecfit);
- How strongly do you feel about the suffix RFEW (emlinefit) vs EW (fastspecfit)? => sure I can switch to EW, I don t mind;
- peak amplitude vs. integrated flux => in case it s relevant here, I m requesting SIGMA<10 A, so that may prevent such cases, no? (looking at the OII_SIGMA distribution, it peaks at ~2 A, has a std of ~2A, and a tail extending to 10 A)
- SHARE: it is only relevant for the OII doublet, where I let it free; for the OIII doublet, I fix it during the fit (and obviously it is not relevant for the Balmer lines); the distribution is centered on the initial guess (=expected value) of 0.58, with a std of ~0.2; I note that it happens my approach is the same as the one recommended by B. Weiner in [desi-galaxies 703]; for any valid fit, I request that the flux at the observed emission line(s) is above the continuum, so for the OII doublet; I m not sure I make some choice about the significance of each component; for the OII, OIII doublets, I use this simple functional form: https://github.com/desihub/desispec/blob/7e5c17b424dc6e7021f806278e2ac5ade9881f5c/bin/desi_emlinefit_afterburner#L275
- I'm not at all a fan of using nan to represent non-measurements => to make my life simpler, I just set to np.nan all cases where I don t have a "reliable" fit, be it because it is outside the spectral range, there are not enough pixels to fit, the fit hit the boundaries, etc. I agree it would be best to proceed as you suggest, but that may require more coding work, so if that s fine, I ll keep the nans; besides it also simplifies the end-user;
- velocity-shift of the line-center with respect to the redrock redshift => correct, I don t allow that, I just use redrock redshift; for the ELG-purpose I guess it s fine;
- It doesn't look like you use the resolution matrix---why not? I guess this means that your measured line-widths (which are in observed-frame Angstroms) need to be deconvolved for the instrumental resolution before they're converted to km/s => that is correct; I guess the reason is that my main product is the OII SNR, all other products are "for free"; I d more remove those other products than going to refine the analysis...
- Why do you fit all 500 spectra => yes, I fit everything; that is simpler from a file structure point of view; I agree that the non-TGT fits are non-sense; I could force np.nan for those.
- multiprocessing/mpi: my understanding is that the pipeline is launching one task per redrock file, so the parallelization is already done in the pipeline, no?
- fastspecfit now does both these tasks and much more => sure, I agree! the primary initial goal of this emlinefit afterburner is to get the OII SNR in the daily reduction.

araichoor:
for info, I hope to be able to implement those changes today/tomorrow.
also, I plan to factor out the functions in a emlinefit_io.py; so that one can load them through "from desispec.emlinefit_io import .."; and also call get_emlines() for just a set a provided targetids.

sbailey:
Minor: please put the I/O functions in desispec.io.emlinefit to be grouped with the other I/O sub-modules under desispec.io.

araichoor:
oh right, thanks for noticing that; sure, will do.

araichoor:
As said, for conveniency, I also propagate those columns from the redrock file:
`TARGETID, TARGET_RA, TARGET_DEC, OBJTYPE, FIBER, Z, ZWARN, SPECTYPE, DELTACHI2`

However, I notice that - obviously - `FIBER` is not present in the healpix `FIBERMAP` extension.
So I think to remove FIBER from that list, i.e. have the same datamodel for tiles/ and healpix/ catalogs (rather than adding the FIBER column only for the tiles/ catalogs).

araichoor:
With further thinking, I think I ll actually remove the `OBSEMWAVES` column, and instead put that information in the `EMLINEFIT` extension header.
I plan to use the following format, with first listing the fitted lines in `EMNAMES`, and then report for each line the rest-frame wavelength(s) in `RFWAVE00`, `RFWAVE01`, etc, with using the order of appearance in `EMNAMES`:
```
EMNAMES = 'OII,HDELTA,HGAMMA,HBETA,OIII,HALPHA'                                 
RFWAVE00= '3727.092,3729.874'                                                   
RFWAVE01= '4102.892'                                                            
RFWAVE02= '4341.684'                                                            
RFWAVE03= '4862.683'                                                            
RFWAVE04= '4960.295,5008.239'                                                   
RFWAVE05= '6564.613'                                                            
```
That way, along with the `Z` column, one can reconstruct the fitted observed wavelength; and it will offer flexibility, and keep the header key at 8 characters (provided we don t fit more than 99 lines!)

Besides, I also plan to propagate the arguments of `desi_emlinefit_afterburner` relevant to the fit:
```
RRFN    = '/global/cfs/cdirs/desi/spectro/redux/everest/healpix/main/dark/100/&'
CONTINUE  '10000/redrock-main-dark-10000.fits'                                  
COADDFN = '/global/cfs/cdirs/desi/spectro/redux/everest/healpix/main/dark/100/&'
CONTINUE  '10000/coadd-main-dark-10000.fits'                                    
RFHW    =                   40                                                  
MINRFHW =                   20                                                  
RFCONTW =                  200                                                  
RV      =                  3.1                                                  
BALMFIT = 'em      '                                                            
```

araichoor:
I eventually had to time to look at this.

Changes:
-  I did some of the changes suggested by @moustakas (use of fastspecfit rest-frame wavelengths, output reduced chi2, change RFEW to EW, set NDOF as integer);
- I removed OBSEMWAVES, and add that information in the header, along with all fit-relevant infos;
- I added as arguments (rr_keys, fm_keys) the keys to be propagated from redrockfn (defaults: rr_keys="TARGETID,Z,ZWARN,SPECTYPE,DELTACHI2",  fm_keys="TARGET_RA,TARGET_DEC,OBJTYPE");
- I factored out functions in desispec.io.emlines.py

Then I compared the fitted quantities (for OII, OIII, Ha-b-c-d) for the main-healpix everest data, where I can compare with fastspecfit.
With comparing with fastspecfit:
- the EW generally disagree;
- for the ELG targets, emlinefit and fastspecfit are in reasonable agreement for all lines;
- for the BGS_ANY and the LRG:
  - the OII and OIII are in agreement (though with some scatter for the LRGs),
  - but emlinefit fits much lower fluxes for the Balmer lines than fastspecfit; 
- the QSO targets are ~intermediate between the ELG and LRG behaviors.

I think that one reason that the emlinefit Balmer lines are much lower than in fastspecfit could be that no velocity shift is allowed in emlinefit.
And for the ELGs, as redrock mostly picks the redshift from the emission lines, there is ~no velocity shift, so emlinefit is in ~agreement with fastspecfit.

So my feeling would be to:
- remove the EW measurement
- either restrict to OII and OIII measurements (ie remove the Balmer lines fit) or provide measurements (OII, OIII, Ha-b-c-d) for ELG targets only.
I of course welcome any other suggestions!

araichoor:
If that helps, I copy here my diagnosis plots:
- left to right: continuum, flux, flux_snr, ew (two plots for each, scatter plot and ratio histogram)
- top to bottom: OII, Hd, Hg, Hb, OIII, Ha
I put three plots.

For the ELG targets (everest, main, dark): and BGS_ANY (everest, main, bright).
emlinefit-main-dark-everest-fsp-ELG

For the LRG targets (everest, main, dark):
emlinefit-main-dark-everest-fsp-LRG

For the BGS_ANY targets (everest, main, bright):
emlinefit-main-bright-everest-fsp-BGS_ANY

araichoor:
And an example of an LRG target where I find Halpha_SNR of approx. 20, whereas fastspecfit finds approx. 200:
[tmp-39628165072552158.pdf](https://github.com/desihub/desispec/files/7240639/tmp-39628165072552158.pdf)
=>
one can see that the redrock redshift is offset, likely explaining the difference; it has in fastspecfit for the Balmer lines VSHIFT=-300 (extreme case).

Plot made with:

```
>>> from desispec.io.emlinefit import get_emlines
>>> d = get_emlines("/global/cfs/cdirs/desi/spectro/redux/everest/healpix/main/dark/323/32301/redrock-main-dark-32301.fits", "/global/cfs/cdirs/desi/spectro/redux/everest/healpix/main/dark/323/32301/coadd-main-dark-32301.fits", targetids=np.array([39628165072552158]), outpdf="tmp-39628165072552158.pdf")

```

araichoor:
Hopefully last changes here: I've defaulted `desi_emlinefit_afterburner` to only fit the ELG targets, with the introduction of the `--bitnames` argument.
Because:
- the primary goal is to use it for ELGs;
- the current fitter is not always working for non-ELG targets;
- the risk to just try to reproduce fastspecfit;
- my limited available time for development.

A consequence is that now the output file won't row-match the input `redrockfn`, but I think it is safer this way.
As it is also the case for the qso afterburners, I assume it is possible (though I understand it less handy for the pipeline team, sorry).

If a user would like to have the measurement for another class of target, he can run `desi_emlinefit_afterburner` on his side, and adapt the settings if needed.
As the code only looks for a class of targets, it is faster now (I ran the 500k main/dark ELG spectra in 5min with 10 nodes).

In the spirit of simplifying things, I've also removed the option to fit the Balmer lines with an emission plus an absorption Gaussians, as this was not well-tested, and not used for the ELGs.

@akremin : I'm happy to coordinate to implement any changes to allow to plug that as an afterburner.

moustakas:
@araichoor I think you should consider continuing to fit all the target classes, at least for the time being. You've shown that the compute cost is negligible and it would be really helpful to have a comparison dataset for `fastspecfit` for more than just he ELGs (which, as you know, are the faintest and hardest objects we're observing!)

araichoor:
I see, thanks for the feedback.
My worries is that we would then provide measurements, which we know some of them are not reliable / well-tested.

Another point if we go this way: with comparing  BGS/LRG measurements with fastspecfit, I notice that the default SIGMA (observed Gaussian width) settings I adopted, based on the ELGs, are no more valid for those different tracers; I set SIGMA < 10 A, whereas it can be larger for BGS/LRGs; I can see if I can allow larger SIGMA, but with not affecting the ELG results..
I ll put that on top on my to-do-list.

bin/desi_emlinefit_afterburner
```diff
@@ -0,0 +1,188 @@
...
+    )
+    parser.add_argument(
+        "--bitnames",
+        help="comma-separated list of target bits to fit from the CMX_TARGET, SV{1,2,3}_DESI_TARGET, or DESI_TARGET mask; if 'ALL', fits all fibers (default=ELG)",
+        type=str,
+        default="ELG",
+        required=False
+    )
+    parser.add_argument(
+        "--outfn",
```
sbailey, line 47: Minor, but for consistency with other desispec programs, let's call the I/O options `--redrock`, `--coadd`, and `--output` (like the desi_qso_*_afterburner scripts)

py/desispec/io/emlinefit.py
```diff
@@ -0,0 +1,567 @@
...
+import fitsio
+import numpy as np
+from scipy.optimize import curve_fit
+from desitarget.geomask import match_to
+from desiutil.dust import ext_odonnell, ext_fitzpatrick
+from desiutil.dust import ebv as dust_ebv
+from desiutil.log import get_logger
+from matplotlib.backends.backend_pdf import PdfPages
+import matplotlib.pyplot as plt
+from matplotlib import gridspec
```
sbailey, line 21: move these matplotlib imports into the plotting section so that it will only try to import matplotlib if really needed to make a plot (removes a possible fragility from batch processing)

py/desispec/io/emlinefit.py
```diff
@@ -0,0 +1,567 @@
...
+                log=log,
+            )
+            for key in emkeys:
+                mydict[emname][key][i] = tmpmydict[key]
+            mydict[emname]["wave"][i] = w
+            mydict[emname]["fitdata"][i] = d
+            mydict[emname]["fitivar"][i] = iv
+            mydict[emname]["model"][i] = m
+    # AR plot?
+    if outpdf is not None:
```
sbailey, line 527: Usually we want to separate calculation from plotting, e.g. so that if you want to update the plots you can call it with some data to get a new plot without having to redo intermediate calculations.  Could you move out this plotting into a separate function that would take the outputs of `emlines_gaussfit` and make a plot, or does it rely upon some internal variable that isn't returned?

py/desispec/io/emlinefit.py
```diff
@@ -0,0 +1,567 @@
...
+    min_share=1e-1,
+    max_share=1,
+    log=get_logger(),
+):
+    """
+    Fits the [OII] doublet line profile with 2 related Gaussians.
+
+    Args:
+        emname: "OII" or "OIII" or "HALPHA", "HBETA", "HGAMMA", "HDELTA" (string)
+        targetid: targetid (int)
```
sbailey, line 141: targetid appears to be used only for passing through into the output, and thus feels like it shouldn't be a required input.  i.e. from a core algorithm standpoint, fitting gaussians to data shouldn't require a targetid as input.

py/desispec/io/emlinefit.py
```diff
@@ -0,0 +1,567 @@
...
+    if emname == "HBETA":
+        rf_em_waves = np.array([4862.683])
+    if emname == "HGAMMA":
+        rf_em_waves = np.array([4341.684])
+    if emname == "HDELTA":
+        rf_em_waves = np.array([4102.892])
+    return rf_em_waves
+
+
+def emlines_gaussfit(
```
sbailey, line 115: This function is the core algorithm that doesn't do I/O (good), but should be in a separate non-I/O module like `desispec.emlinefit`, leaving `desispec.io.emlinefit` for functions that are very specific to I/O itself.  `get_rf_em_waves` would also go in that module.

py/desispec/io/emlinefit.py
```diff
@@ -0,0 +1,567 @@
...
+from desiutil.dust import ext_odonnell, ext_fitzpatrick
+from desiutil.dust import ebv as dust_ebv
+from desiutil.log import get_logger
+from matplotlib.backends.backend_pdf import PdfPages
+import matplotlib.pyplot as plt
+from matplotlib import gridspec
+
+allowed_emnames = ["OII", "HDELTA", "HGAMMA", "HBETA", "OIII", "HALPHA"]
+
+def get_targetids(redrockfn, bitnames, log=get_logger()):
```
sbailey, line 25: This function opens and reads the redrock file twice to perform a selection of targetids.  I think it would be cleaner to make this a non-I/O function that takes a fibermap and bitnames and returns targetids (or maybe the indices of the fibermap or a boolean array of matches).

Some of this logic looks like it would be cleaner to use `desitarget.targets.main_cmx_or_sv` to figure out what the relevant target column is and what mask to import.  Please check if that would work.

sbailey (changes requested):
Apologies for delayed review.  This is nice and fast and meets the core needs, but I've put some comments inline related to code structure and future maintainability.

The core algorithm is isolated in the `emlines_gaussfit` function (good), but other pieces seem to mix I/O, algorithms, and plotting.  Could you try to separate these more so that the outer wrapper script looks more like:
```python
stuff = desispec.io.emlinefit.read_stuff(coadd_filename, redrock_filename)
results = desispec.emlinefit.emlines_gaussfit(stuff)
desispec.io.emlinefit.write_emlines(outfile, results)
if args.outpdf:
    desispec.io.emlinefit.plot_emlines(args.outpdf, results)
```
(don't actually use the name "stuff", that's a placeholder for whatever would be read and returned)

Maybe that's what `get_emlines` is supposed to be?  But in that case `get_targetids` should be working on stuff that `get_emlines` is already reading instead of working off of filenames directly.  Please try to restructure this into:
  * desispec.io.emlinefit - I/O and convenience reformatting, but no calculations
  * desispec.emlinefit - the core algorithms but no I/O, in a form that would be useful to someone calling on a simulated spectrum or from a Jupyter notebook.
  * bin/desi_emlinefit_afterburner - puts those two together

If that doesn't make sense let me know and we can chat in person to clarify.  Thanks.

araichoor:
Thanks a lot @sbailey for detailed+pedagogical review!

Except the comment on `get_targetids()`, I think I've addressed all of those.
Apologies for the "big commit" (https://github.com/desihub/desispec/pull/1386/commits/fe63705038dc85bf89430fb140a855c5c0cc2854), but I didn't find an easy to split it in smaller commits that wouldn't leave the code in a broken state.

I've checked on very cases that the new version produces the same answer as the old version, but I'll test more tomorrow; though I'm already pushing/sharing the code, so that you have time to re-review when that suits you best.

I now reply to each comment.

bin/desi_emlinefit_afterburner
```diff
@@ -0,0 +1,188 @@
...
+    fd["EMLINEFIT"].write_key("RRFN", args.redrockfn)
+    fd["EMLINEFIT"].write_key("COADDFN", args.coaddfn)
+    fd["EMLINEFIT"].write_key("RFHW", args.rf_fit_hw)
+    fd["EMLINEFIT"].write_key("MINRFHW", args.min_rf_fit_hw)
+    fd["EMLINEFIT"].write_key("RFCONTW", args.rf_cont_w)
+    fd["EMLINEFIT"].write_key("RV", args.rv)
+    fd["EMLINEFIT"].write_key("EMNAMES", args.emnames)
+    for i_emname, emname in enumerate(args.emnames.split(",")):
+        fd["EMLINEFIT"].write_key("RFWAVE{:02d}".format(i_emname), ",".join(get_rf_em_waves(emname).astype(str)))
+    fd.close()
```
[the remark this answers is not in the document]
araichoor: done.

py/desispec/io/emlinefit.py
```diff
@@ -0,0 +1,567 @@
...
+        Returns np.nan in mydict (and NDOF=-99) if not enough pixels to fit or if fit fails.
+        Default settings designed for ELGs (e.g. max_sigma); need to re-assess if run on other targets.
+        For "OII", let the doublet line ratio free during the fit.
+        For "OIII", fits the 4960 and 5007 lines with a fixed line ratio.
+        For the Balmer lines, SHARE is not fitted and set to np.nan.
+    """
+    # AR allowed arguments
+    if emname not in allowed_emnames:
+        log.error("{} not in {}; exiting".format(emname, allowed_emnames))
+        sys.exit(1)
```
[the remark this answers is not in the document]
araichoor: done.

py/desispec/io/emlinefit.py
```diff
@@ -0,0 +1,567 @@
...
+    p0_sigma=2.5,
+    p0_flux=10,
+    p0_share=0.58,
+    min_sigma=1e-5,
+    max_sigma=10.,
+    min_flux=1e-5,
+    max_flux=1e9,
+    min_share=1e-1,
+    max_share=1,
+    log=get_logger(),
```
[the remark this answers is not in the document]
araichoor: done; thanks for the explanation!

py/desispec/io/emlinefit.py
```diff
@@ -0,0 +1,567 @@
...
+
+    Notes:
+        We add TARGETID and Z to rr_keys if TARGETID not present in rr_keys nor in fm_keys.
+        If keys in rr_keys or fm_keys are not present in the redrockfn, those will be ignored.
+    """
+    # AR sanity checks
+    for fn in [redrockfn, coaddfn]:
+        if not os.path.isfile(fn):
+            log.error("no {} file; exiting".format(fn))
+            sys.exit(1)
```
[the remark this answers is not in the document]
araichoor: done.
I've replaced the calls to `sys.exit(1)`; I hope I've picked the relevant raising case each time.

bin/desi_emlinefit_afterburner
araichoor: done.

py/desispec/io/emlinefit.py
araichoor: done.

py/desispec/io/emlinefit.py
araichoor: done, I've moved that to `desispec.io.emlinefit.plot_emlines()`.

py/desispec/io/emlinefit.py
araichoor: done (nice catch).

py/desispec/io/emlinefit.py
araichoor: done.

py/desispec/io/emlinefit.py
araichoor: Here, I've kept the function as is (for the time being).

Maybe that doesn't really matter as the `redrock*fits` files are small, but by my experience, I think it's faster to proceed this way:
- first only read the file structure with `fits.open()`, but not the data, to inspect the column content (that step is super-quick)
- then read only the relevant columns.
than just reading the whole data.
Same for `desitarget.targets.main_cmx_or_sv`: that function assumes that the whole data have been read.

Note that in the new code version, I also use that two-steps approach, e.g.:
https://github.com/desihub/desispec/blob/febddeed3ab1885b56b956c5ada69dd2526b6b8c/py/desispec/io/emlinefit.py#L148-L164

Please let me know if there is a better way to proceed.

py/desispec/io/emlinefit.py
araichoor: mmhh.. apparently the related commit failed to pass the checks; I've no idea why.

araichoor:
And before I forget: besides the code-structure changes, one key question is if we do want to provide those fits for all spectra or not.
My worry doing so is that I'm not super-confident for the fits for non-ELG tracers, and I don't have time to play to see if changing the default fitting parameters could help.
But maybe that is still better than nothing...

py/desispec/io/emlinefit.py
sbailey: the checks  are complaining about test coverage (new code without tests that actually try that code).  With the new structure separating I/O from algorithms from plotting, that will be a bit easier to add tests.  We'll deal with that a bit later.

araichoor:
Again, thanks a lot for the reply (+testing the speed of different calls!).
All that you write makes sense, and should be easily doable, so I'll go for that.

I support the idea of keeping `targetids` as an argument, as it can be useful to be able to call the line fitter only for few desired targets.
I think that, in the case where both bitnames and targetids are provided, I'll make a selection on both of those (i.e. take the provided targetids which pass the bitname selection); makes sense?

Lastly, as long as you provide me useful tips in using fitsio: if I read the redrock/fibermap catalogs with all the columns, I still need to then remove all columns except rr_keys/fm_keys, as those will be propagated in the final output catalog.
My question: once I've read a catalog with fitsio, how do I properly remove a given column from that catalog?

sbailey:
> I support the idea of keeping targetids as an argument, as it can be useful to be able to call the line fitter only for few desired targets.
I think that, in the case where both bitnames and targetids are provided, I'll make a selection on both of those (i.e. take the provided targetids which pass the bitname selection); makes sense?

yes, sounds good.

> Lastly, as long as you provide me useful tips in using fitsio: if I read the redrock/fibermap catalogs with all the columns, I still need to then remove all columns except rr_keys/fm_keys, as those will be propagated in the final output catalog.
> My question: once I've read a catalog with fitsio, how do I properly remove a given column from that catalog?

I had forgotten about the issue of what gets propagated to the output catalog.  You can filter down to just the columns you want with e.g.
```python
blat = fibermap['TARGETID', 'DESI_TARGET']
```
`numpy.lib.recfunctions` also has a set of functions for adding and removing columns from a numpy structured array.  Another option is to cast to an astropy Table and then add/remove columns (which I think is more memory efficient but less CPU efficient, but in this case it probably doesn't matter).

araichoor:
I should have addressed your last batch of comments:
- default to fitting all 500 fibers in `bin/desi_emlinefit_afterburner`;
- use `desitarget.targets.main_cmx_or_sv` in `get_targetids()`
- nest the call to `get_targetids()` inside `read_emlines_inputs()`;
- keep the targetids=None as argument of `read_emlines_inputs()` (though that is not propagated as an argument of `bin/desi_emlinefit_afterburner`, as I don't feel it's reasonable to have this at the command line level);
- cleaner use of fitsio to read files.

Plus few code syntax cleanings.

sbailey:
Thanks for the updates; the new structure looks great.

As compensation for your patience, I have added the final set of changes to add unit tests:
  * Moved most of `bin/desi_emlinefit_afterburner` into `py/desispec/scripts/emlinefit.py` so that the functions become available to call from unit tests
  * Added `py/desispec/test/test_emlinefits.py`
    * at GitHub, this tests the algorithms in `desispec.emlinefit` with 95% coverage
    * at NERSC, it additionally calls the full script on everest inputs to also test `desispec.scripts.emlinefit` (75%) and `desispec.io.emlinefit` (61%).
    * making the pdf and filtering by targetid are the pieces not tested

These aren't sophisticated tests for correctness, but are basic "does it run without crashing" tests that will help us in the future to catch any problems with changes to numpy, astropy, fitsio, etc. before running this in the full pipeline.

Speaking of the full pipeline -- I think this is ready to merge now, but I'll leave it open for you to look at what I did with the tests and complain/question if needed.  After cori is back let's add the final hooks to include this in the daily pipeline desi_tile_redshifts (which can be a separate PR).

My structural requests have caused your original single-file-script to become multiple files:
  * bin/desi_emlinefit_afterburner
  * py/desispec/scripts/emlinefit.py
  * py/desispec/emlinefit.py
  * py/desispec/io/emlinefit.py

While that is admittedly somewhat of a pain, it is motivated by:
  * separating algorithms from I/O from plotting
  * which makes it easier to reuse the algorithms from Jupyter notebooks or other future MPI-parallelized wrappers
  * and makes it easier to test (e.g. algorithms at GitHub without requiring any input data and full tests at NERSC)

Final notes:
  * a new github sphinx failure has appeared, though full docs build without warnings or errors at NERSC.  I think this is related to pinning the version of sphinx that is already fixed in a previous PR that came after this branch, so I'll ignore it here.
  * coverage is still complaining, but that's because the full script tests requires files at NERSC that aren't available to the coverage test at GitHub.  At NERSC I checked coverage with: `pytest --cov=desispec --cov-report=html py/desispec/test/test_emlinefit.py` and then copied the "htmlcov" dir to $CFS/desi/users/sjbailey so that I could view it via data.desi.lbl.gov .

araichoor:
Thanks a lot for the support, I've learned quite some stuff during this exercise.

Just one remark about the test:
I see this test/comment:
https://github.com/desihub/desispec/blob/707223b3c7b00c93a2966d441b4bd569acce57cb/py/desispec/test/test_emlinefit.py#L48

However, the code will return a NaN if it fails, e.g. if there are not enough valid pixels or if the flux is negative at the fitted location.
As the flux is a random variable centered on zero, it can happen it is negative at the [OII] location.
I checked with copy-pasting the code lines, looks like the two random fluxes are positive there, so it should be ok:
tmp-0
tmp-1

Maybe we could add a positive offset to the fluxes, to be sure / explicit about that, e.g.:
```
 fluxes = 10 + rand.normal(size=(nspec, nwave))
```

sbailey:
Interesting.  Then an algorithm choice question: I was expecting that if the data were negative at the location of a line, you'd just get a negative FLUX rather than forcing it to NaN.  To me that is different from a "failed" fit from not having enough good pixels at that wavelength (or having the line completely redshifted out of the wavelength coverage).  Is there a specific reason to *not* propagate a negative flux fit?  To me this is conceptually equivalent to not clipping photometric fluxes to 0 when making an imaging catalog -- let the data fluctuate, and sometimes the fit will be negative even though you know the object photons are strictly positive, but the negative tail can be useful in assessing the noise/statistical properties of the dataset.

Also, can you point to the piece of code that sets negative fits to NaN (or otherwise doesn't overwrite the default NaN because the fit was negative)?  From a quick glance that isn't obvious to me where that happens.

sbailey (approved):

araichoor:
I acknowledge my approach may be a bit "brutal", but the idea was end-user oriented: if there is a not-NaN measurement, then it can be trusted (at least for the ELGs :)); no need to make several cuts to decide if the fit is trustable or not; that of course has been driven by my work on ELGs.

In that spirit, I've put several requirements; 
I described those at the beginning of this PR (Aug, 26! https://github.com/desihub/desispec/pull/1386#issuecomment-906828766); for what we discussed, I think the relevant ones are:

- the flux at the line position is above the estimated continuum (so what I wrote in the previous message is not exactly correct; but the two random fluxes in the test still pass that requirement, I think):
https://github.com/desihub/desispec/blob/707223b3c7b00c93a2966d441b4bd569acce57cb/py/desispec/emlinefit.py#L212-L213

- I set some boundaries for the flux fitting (min_flux=1e-5, max_flux=1e9), and if the fitted flux is close to the boundaries (1% margin), I decide it is a failure:
https://github.com/desihub/desispec/blob/707223b3c7b00c93a2966d441b4bd569acce57cb/py/desispec/emlinefit.py#L225-L234

Around that place in the code you may find other informations on that topic; happy to give more precision / discuss about that.

PR #1387, opened 2021-08-26
allow svn override of fiberassign files

### LINKED ISSUE #1296: allow svn override of fiberassign files
Currently desispec uses the fiberassign files in the raw data stream on the principle that that reflects what was actually used for observations.  However, @dstndstn et al recently discovered that secondary-only targets have incorrect values for some columns that are *not* used by observation but are propagated downstream for analysis, e.g. BRICKID and FLUX_R (fiberassign issue with details forthcoming).

Rather than patching every observed fiberassign file in $DESI_SPECTRO_DATA, it will be more straightforward to patch the files in svn (on disk at $DESI_ROOT/target/fiberassign/tiles/trunk/), and allow those to override the raw data versions when running assemble_fibermap.  (I have previously opposed this, but we've now found a problem that affects every tile so this seems the only viable option).

At the same time, `assemble_fibermap` should explicitly compare every column that is used by ICS/platemaker and raise and exception if the svn version now disagrees (in which case the svn version should be re-fixed).

Since the nightly productions will be processing raw data before the tiles are checked into svn, a consequence is that `assemble_fibermap` should proceed with the raw-data version as-is if the svn version doesn't exist.

sbailey:
This PR adds the hooks to allow fiberassign files in svn to override those in the raw data stream, e.g. to correct the secondary target bookkeeping bug (desihub/fiberassign#375) or the duplicate negative TARGETID bug (desihub/fiberassign#385).  At the same time, it checks columns used by operations to make sure that they didn't accidentally change: TARGETID, TARGET_RA, TARGET_DEC, PMRA, PMDEC, REF_EPOCH, LAMBDA_REF, LOCATION, PETAL_LOC, DEVICE_LOC .

`assemble_fibermap --no-svn-override ...` can be used to prevent using the svn copy if needed (but in that case it is better to fix whatever we don't like about the svn copy).  I did not implement the opposite of `--force-svn-override` to use the svn copy even if one of the operations columns listed about had a discrepancy.

Once special case: tile 80713 on 20210110 had PMRA=PMDEC=NaN for some targets that got "corrected" to 0.0 in svn.  ICS/Platemaker choked on those targets and they weren't positioned, and then ICS/Platemkaer were updated to handle NaNs differently so the tile isn't repeatably reobservable even if svn went back to using NaN.  I'd prefer svn to be reverted to the original version with NaNs so that this special case isn't required, but the current check at least allows us to still process that tile.

@akremin please review.  Thanks.

fixes #1296

## Examples

Override with svn copy
```
assemble_fibermap -n 20201215 -e 68040 -o fibermap-00068040.fits
...
INFO:fibermap.py:451:assemble_fibermap: Overriding raw fiberassign file /global/cfs/cdirs/desi/spectro/data/20201215/00068040/fiberassign-080610.fits with svn /global/cfs/cdirs/desi/target/fiberassign/tiles/trunk/080/fiberassign-080610.fits.gz
...
```
though in this case the svn and raw data fiberassign files are the same other than gzip or not.  I'm not aware of any overrides in svn yet that are purposefully more correct.

Special case override with svn copy despite PMRA/PMDEC NaN vs 0.0 discrepancy (tile 80713):
```
assemble_fibermap -n 20210110 -e 71721 -o fibermap-00071721.fits
...
INFO:fibermap.py:451:assemble_fibermap: Overriding raw fiberassign file /global/cfs/cdirs/desi/spectro/data/20210110/00071721/fiberassign-080713.fits.gz with svn /global/cfs/cdirs/desi/target/fiberassign/tiles/trunk/080/fiberassign-080713.fits.gz
...
WARNING:fibermap.py:531:assemble_fibermap: Ignoring PMRA mismatch NaN -> 0.0 on tile 80713 night 20210110
WARNING:fibermap.py:531:assemble_fibermap: Ignoring PMDEC mismatch NaN -> 0.0 on tile 80713 night 20210110
...
```

CMX fiberassign file where svn copy is inconsistent with raw data for unknown reason; by default raise exception so that we don't silently use the wrong file:
```
assemble_fibermap -n 20200212 -e 48310 -o fibermap-00048310.fits
...
CRITICAL:fibermap.py:536:assemble_fibermap: incompatible raw/svn fiberassign files for columns ['TARGETID', 'TARGET_RA', 'TARGET_DEC', 'PMRA', 'PMDEC', 'REF_EPOCH']
  File "/global/common/software/desi/users/sjbailey/desispec/bin/assemble_fibermap", line 44, in <module>
    fibermap = assemble_fibermap(args.night, args.expid, badamps=args.badamps,
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/io/fibermap.py", line 537, in assemble_fibermap
    raise ValueError(msg)
ValueError: incompatible raw/svn fiberassign files for columns ['TARGETID', 'TARGET_RA', 'TARGET_DEC', 'PMRA', 'PMDEC', 'REF_EPOCH']
```

But if you want to use raw data copy and ignore svn (what is currently done, prior to this PR):
```
assemble_fibermap -n 20200212 -e 48310 -o fibermap-00048310.fits --no-svn-override
...
INFO:assemble_fibermap:50:<module>: Wrote fibermap-00048310.fits
```

### DIFF
--- a/bin/assemble_fibermap
+++ b/bin/assemble_fibermap
@@ -22,6 +22,8 @@ parser.add_argument("--overwrite", action="store_true",
         help="overwrite pre-existing output file")
 parser.add_argument("--force", action="store_true",
         help="make fibermap even if missing input guide or coordinates files")
+parser.add_argument("--no-svn-override", action="store_true",
+        help="Do not allow fiberassign SVN to override raw data")
 
 args = parser.parse_args()
 
@@ -39,7 +41,8 @@ if os.path.exists(args.outfile):
         log.critical(f'{args.outfile} already exists; remove or use --overwrite')
         sys.exit(1)
 
-fibermap = assemble_fibermap(args.night, args.expid, badamps=args.badamps, force=args.force)
+fibermap = assemble_fibermap(args.night, args.expid, badamps=args.badamps,
+        force=args.force, allow_svn_override=(not args.no_svn_override))
 
 tmpfile = args.outfile+'.tmp'
 fibermap.write(tmpfile, overwrite=args.overwrite, format='fits')
--- a/py/desispec/io/fibermap.py
+++ b/py/desispec/io/fibermap.py
@@ -345,7 +345,7 @@ def find_fiberassign_file(night, expid, tileid=None, nightdir=None):
 
     Returns first fiberassign file found on or before `expid` on `night`.
 
-    Raises FileNotFoundError if no fibermap is found
+    Raises FileNotFoundError if no fiberassign file is found
     """
     log = get_logger()
     if nightdir is None:
@@ -375,6 +375,41 @@ def find_fiberassign_file(night, expid, tileid=None, nightdir=None):
 
     return fafile
 
+def compare_fiberassign(fa1, fa2, compare_all=False):
+    """
+    Check whether two fiberassign tables agree for cols used by ICS/platemaker
+
+    Args:
+        fa1, fa2: fiberassign astropy Tables or numpy structured arrays
+
+    Options:
+        compare_all: if True, compare all columns, not just those used by ops
+
+    Returns list of columns with mismatches; empty list if all agree
+
+    Note: if both are NaN, it is considered a match
+    """
+    if compare_all:
+        compare_columns = fa1.dtype.names
+    else:
+        compare_columns = (
+                'TARGETID', 'TARGET_RA', 'TARGET_DEC',
+                'PMRA', 'PMDEC', 'REF_EPOCH', 'LAMBDA_REF', 'LOCATION',
+                'PETAL_LOC', 'DEVICE_LOC')
+
+        badcol = list()
+        for col in compare_columns:
+
+            #- check for mismatches, but allow for NaN
+            match = (fa1[col] == fa2[col])
+            if fa1[col].dtype.kind == 'f':
+                match |= (np.isnan(fa1[col]) & np.isnan(fa2[col]))
+
+            if np.any(~match):
+                badcol.append(col)
+
+        return badcol
+
 def assemble_fibermap(night, expid, badamps=None, force=False):
     """
     Create a fibermap for a given night and expid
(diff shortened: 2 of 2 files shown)


sbailey:
@akremin or anyone else else want to check me on this?  I'd like to merge this soon for the next round of testing.

akremin (approved):
How often are the early CMX data problematic? I don't see any changes to the code that calls `assemble_fibermap`, so I believe this means we can no longer process those problematic CMX tiles via `desi_proc` without doing a separate call to `assemble_fibermap` with the flag `--no-svn-override`.  If it is a few tiles (or only tiles we don't need to worry about anymore), then I am perfectly okay with your choice.

I tested this on two additional exposures, including one from 20211214 where the fiberassign files were originally wrong and replaced. It performed properly in that case too (as expected since the replacements were given the proper name and old files place under *.orig).

Tests:
Verified all the commands given in the PR description. Also did the following, which worked as expected:
`assemble_fibermap -n 20201214 -e 67744 -o fibermap-00067744.fits`
`assemble_fibermap -n 20210306 -e 79555 -o fibermap-00079555.fits`

sbailey:
> How often are the early CMX data problematic?

There are 6 CMX dithering tiles (63070, 72, 74, 75, 76, 78) that appear 81 times in Jan/Feb 2020 raw data that don't match SVN, though many of those cases are tests and failed acquisitions that never resulted in actual matching spectra data.  You are correct that these can't be processed with desi_proc as-is without additional hand-work, but I think they are also very low priority and unlikely to ever be re-reduced.  And in general future mis-matches should be addressed by fixing the SVN copy, not by adding more special cases to desi_proc / assemble_fibermap.

I could add a `--no-svn-override` option to desi_proc, but I'm inclined to *not* "pollute" the command line options for a case that we don't expect to regularly use, and for which we still can work around by running `assemble_fibermap --no-svn-override` by hand if the problem comes up and we need to process some data before fixing svn.  @akremin what do you think?

akremin:
Thank you for the explanation. I agree with your choice. The odds of working with these tiles are low enough that it isn't worth adding (yet-)another option to `desi_proc`.  My goal was to get more information documented here for posterity and to verify that it was a conscious decision, so I'm satisfied.

You are welcome to merge.

PR #1391, opened 2021-08-30
desi_run_night --tiles option

sbailey:
This PR adds a `desi_run_night --tiles ...` option to filter what tiles would be submitted for that night.  This is primarily for testing where we'd like to re-run some test-tiles without rerunning all the tiles on a night, while still benefitting from the job dependency infrastructure to jointly fit standard stars across exposures, run redshifts, etc.

Examples, starting from a blank production with just the everest/exposure_tables/202006/*.* files:

Process exposures for just two tiles on a night:
```
desi_run_night -n 20210601 --dry-run-level 1 --tiles 80978,1399
```
--> generates jobs for exposures 90656, 90657, and 90660 for those tiles

Also include arcs and flats (tileid -99 in the exposure tables):
```
desi_run_night -n 20210601 --dry-run-level 1 --tiles 80978,1399,-99
```
--> same science jobs as above, but also includes arc and flat jobs

Bad and non-existent tiles:
```
desi_run_night -n 20210601 --dry-run-level 1 --tiles 80978,1399,20098,123
```
--> jobs for 20098 not generated because its exposure 90655 is flagged as bad in the exposures table; i.e. this option is not a way to override good/bad
--> skips tile 123 which wasn't observed on that night

This purposefully skips over any tiles that weren't actually observe on that night, with the idea that we may want to do something like looping over nights with a list of tiles without needing to worry about exactly which tiles were observed on which nights, e.g. (pseudo-bash)
```
for night in (list of nights); do
    desi_run_night -n $night --tiles (list of VI tiles) ...
done
```

@akremin please take a look, thanks.

### DIFF
--- a/py/desispec/workflow/procfuncs.py
+++ b/py/desispec/workflow/procfuncs.py
@@ -284,8 +284,8 @@ def create_batch_script(prow, queue='realtime', dry_run=0, joint=False, system_n
         prow, Table.Row or dict. Must include keyword accessible definitions for processing_table columns found in
                                  desispect.workflow.proctable.get_processing_table_column_defs()
         queue, str. The name of the NERSC Slurm queue to submit to. Default is the realtime queue.
-        dry_run, int. If nonzero, this is a simulated run. If dry_run=1 the scripts will be written or submitted. If
-                      dry_run=2, the scripts will not be writter or submitted. Logging will remain the same
+        dry_run, int. If nonzero, this is a simulated run. If dry_run=1 the scripts will be written but not submitted.
+                      If dry_run=2, the scripts will not be written nor submitted. Logging will remain the same
                       for testing as though scripts are being submitted. Default is 0 (false).
         joint, bool. Whether this is a joint fitting job (the job involves multiple exposures) and therefore needs to be
                      run with desi_proc_joint_fit. Default is False.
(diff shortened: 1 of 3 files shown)


py/desispec/workflow/procfuncs.py
```diff
@@ -304,7 +304,8 @@ def create_batch_script(prow, queue='realtime', dry_run=0, joint=False, system_n
     if prow['JOBDESC'] in ['perexp','pernight','pernight-v0','cumulative']:
         if dry_run > 1:
             scriptpathname = get_tile_redshift_script_pathname(tileid=prow['TILEID'],group=prow['JOBDESC'],
-                                                               night=prow['NIGHT'], expid=prow['EXPID'])
+                                                               night=prow['NIGHT'], expid=prow['EXPID'][0])
```
sbailey, line 307: @akremin previously this step was crashing with dry_run=2 since prow['EXPID'] is an array of EXPIDs not just a single EXPID (get_tile_redshift_script_pathname wants EXPID=int, not an array).
This hack isn't quite right either since JOBDESC='perexp' would submit several jobs instead of just a job for the first EXPID in the array.

I patched what I needed to move on with testing the actual PR, but I'm highlighting this for you in case there is a more-correct simple workaround?

akremin: Great catch. I'm not sure how my initial vetting missed this, I must not have tested all 3 dry_runs * 4 redshifts = 12 combinations. Once you're happy with the PR, I'll piggy back on it and add a complete fix for this so that it works properly.

akremin (approved):
The code looks good. 

One subtlety to highlight -- this selection of tiles will also reject calibrations exposures. Is the goal of this to only run specified tiles without the added complexity of arcs+psf+flats+nightlyflat, or was that not the intention?

sbailey:
Thanks for the quick review.

> One subtlety to highlight -- this selection of tiles will also reject calibrations exposures. Is the goal of this to only run specified tiles without the added complexity of arcs+psf+flats+nightlyflat, or was that not the intention?

The immediate use case is to start with a copyprod production that has symlinks to the nightly calibs and re-run only science tiles, but I think if you include -99 in the list of tiles to process it will also do the arcs and flats.  I verified that with `--dry-run-level 1` (generating scripts but not submitting them).

I'll merge this now as "directionally correct progress" and we can pick up with additional updates as needed (including the fix for --dry-run-level 2 EXPIDs, at low-ish priority).

Issue #1370, opened 2021-08-10
desi_zcatalog backwards compatibility


alxogm:
Hi , I should have catch this before, but we are just noticing that with the recent updates to  `desi_zcatalog`   we (Lya WG) can not longer use it in our existing quickquasars mocks, or new recent runs, since we have not updated our simulated spectra files. So I'm wondering if we should try to restore the compatibility in the desi_zcatalog, or we create our own script somewhere else? 
A quick fix for now for us is to use an older tag of the code for existing mocks, and consider update the data model for new mocks,  but I wanted to check if you discard to maintain the backwards compatibility. 

Thanks!

PR #1392, opened 2021-08-31
desi_zcatalog back compatibility 

alxogm:
As  discussed in [issue #1370 ](https://github.com/desihub/desispec/issues/1370)  I'm adding an option to read the older format of zbest files so that desi_zcatalog can be used in existing "zbest" files, specially useful for the quickquasars mocks.

### DIFF
--- a/bin/desi_zcatalog
+++ b/bin/desi_zcatalog
@@ -139,6 +139,9 @@ for ifile, rrfile in enumerate(redrockfiles):
                 expfibermap = fx['EXP_FIBERMAP'].read()
                 tsnr2 = fx['TSNR2'].read()
                 assert np.all(redshifts['TARGETID'] == tsnr2['TARGETID'])
+            else:
+                expfibermap = None
+                tsnr2 = None
 
         else:
             redshifts = fx['REDSHIFTS'].read()
@@ -166,21 +169,29 @@ for ifile, rrfile in enumerate(redrockfiles):
     else:
         fmcols = list(fibermap.dtype.names)
         fmcols.remove('TARGETID')
-        tsnr2cols = list(tsnr2.dtype.names)
-        tsnr2cols.remove('TARGETID')
-        data.append(hstack(
+        if tsnr2 is not None:
+            tsnr2cols = list(tsnr2.dtype.names)
+            tsnr2cols.remove('TARGETID')
+            data.append(hstack(
             [Table(redshifts), Table(fibermap[fmcols]), Table(tsnr2[tsnr2cols])]
             ))
+        else:
+            data.append(hstack(
+            [Table(redshifts), Table(fibermap[fmcols])]
+            ))
 
-    if "expfibermap" in locals():
+    if expfibermap is not None:
         exp_fibermaps.append(expfibermap)
+
 
 log.info('Stacking zcat')
 zcat = np.array(vstack(data))
-
 if exp_fibermaps:
     log.info('Stacking exposure fibermaps')
     expfm = np.hstack(exp_fibermaps)
+else:
+    expfm = None
+
 
 #- untested with new formats, so commenting out for now
 # if args.match:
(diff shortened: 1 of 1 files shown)


alxogm:
Hi @sbailey, I think this PR is ready for review, I have tested in the mock /project/projectdirs/desi/mocks/lya_forest/develop/saclay/qq_desi/v4.7/v4.7.01/desi-0.0/spectra-16, with the instruction: 

`desi_zcatalog -i /project/projectdirs/desi/mocks/lya_forest/develop/saclay/qq_desi/v4.7/v4.7.01/desi-0.0/spectra-16 -o zcat_test --minimal --prefix zbest `

The previous catalog for that mock is /`project/projectdirs/desi/mocks/lya_forest/develop/saclay/qq_desi/v4.7/v4.7.01/desi-0.0/zcat.fits` Please let me know of any fix I should make.

bin/desi_zcatalog
```diff
@@ -154,12 +172,15 @@ for ifile, rrfile in enumerate(redrockfiles):
             [Table(redshifts), Table(fibermap[fmcols]), Table(tsnr2[tsnr2cols])]
             ))
 
-    exp_fibermaps.append(expfibermap)
+    if "expfibermap" in locals():
```
sbailey, line 175: See comment above about explicitly setting `expfibermap=None` if that extension doesn't exist and then checking `if expfibermap is not None` here instead of leaving the variable undefined and checking in locals here.  Checking `if "expfibermap" in locals()` is similar, but slightly less robust in case someone ever added a `expfibermap` or `tsnr2` variable earlier in the code, not realizing that defining the variable (even as None or False) would suddenly break this.

sbailey (commented):
Thanks @alxogm for adding this backwards compatibility and specific example for testing.  I tried it and it works as expected.  However, if you drop `--minimal`, it breaks on an undefined `tsnr2` variable, even though in principle it could have merged in the fibermap columns even though it doesn't have tsnr2 columns to merge.  Please fix that.

I also added some style comments about explicitly defining variables as None and checking that, vs. leaving them as undefined and checking for them in locals().  That's a minor style difference; I think my suggestion is more maintainable/robust in the future, but if there is a specific reason to use undefined+locals() instead, that's fine (and let me know the reason so I can learn...)

One other note: your original zcat.fits has RA/DEC as double float64, while the new catalog has them as single-precision float32.  The original input files have them as float32 so I think this new code is more correct (keeping the output type the same as the input type), though for real-world operations we need the extra precision of double to actually get fibers on target.  When you are updating quickquasars, consider updating to use double RA/DEC even if that isn't really needed for mocks.

alxogm:
Thanks @sbailey, I think I've fixed it according to your comments.

> Thanks @alxogm for adding this backwards compatibility and specific example for testing. I tried it and it works as expected. However, if you drop `--minimal`, it breaks on an undefined `tsnr2` variable, even though in principle it could have merged in the fibermap columns even though it doesn't have tsnr2 columns to merge. Please fix that.

> I also added some style comments about explicitly defining variables as None and checking that, vs. leaving them as undefined and checking for them in locals(). That's a minor style difference; I think my suggestion is more maintainable/robust in the future, but if there is a specific reason to use undefined+locals() instead, that's fine (and let me know the reason so I can learn...)
>
No, I think setting these variables as None is better as you explained.  Thanks. 

> One other note: your original zcat.fits has RA/DEC as double float64, while the new catalog has them as single-precision float32. The original input files have them as float32 so I think this new code is more correct (keeping the output type the same as the input type), though for real-world operations we need the extra precision of double to actually get fibers on target. When you are updating quickquasars, consider updating to use double RA/DEC even if that isn't really needed for mocks.
Ok, I'll check that this week. 

Another unrelated question, for mocks we have the truth DLA and BAL information, as well as truth redshifts in the truth files, so I'm wondering if we could include some lines that whenever these truth files are present we merge them to get the DLA and BAL catalogs, I currently have this in a separate script, but it could be useful to get them at the same stage as we create de mock redshift catalog. What do you think?

sbailey:
Thanks @alxogm ; merging now.

> Another unrelated question, for mocks we have the truth DLA and BAL information, as well as truth redshifts in the truth files, so I'm wondering if we could include some lines that whenever these truth files are present we merge them to get the DLA and BAL catalogs, I currently have this in a separate script, but it could be useful to get them at the same stage as we create de mock redshift catalog. What do you think?

The original `desi_zcatalog` included a `--match` option to merge in arbitrary other catalogs based on TARGETID, but that code is currently commented out.  If that concept would meet your needs, could you re-enable and test it and submit a new PR?  If that wouldn't meet your needs, please open a ticket with more details.  I think I would prefer merging of external catalogs to be "opt-in", i.e. specified from the command-line rather than auto-recognizing specific cases to merge in.

alxogm:
ok thanks!

PR #1400, opened 2021-09-02
Stdstars in fluxcalib

julienguy:
PR to save the information about the final set of std stars used for the calibration of a frame in the fluxcalib file.
The fibermap info for the standard stars participating to the fit is saved in HDU STDSTAR_FIBERMAP.
This information is used in the exposure_qa to verify the calibration of stars if it exists (i.e. the qa is backward compatible).

### DIFF
--- a/py/desispec/exposure_qa.py
+++ b/py/desispec/exposure_qa.py
@@ -14,7 +14,7 @@ from pkg_resources import resource_filename
 
 from desiutil.log import get_logger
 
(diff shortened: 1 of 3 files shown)


sbailey (approved):
For the record: this is a change in the datamodel compared to Everest, but it is only an addition and the code is backwards compatible for reading files without STDSTAR_FIBERMAP, so this looks ok.

PR #1404, opened 2021-09-04
Option to specify path of temporary merged fits file

marcelo-alvarez:
This set of changes allows the user to optionally specify the location of the temporary merged PSF output file written in desispec.specex.merge_psf through the environment variable `DESI_LOCALBUFF`. In particular, setting `DESI_LOCALBUFF=/dev/shm` gives a speed up of the total PSF fit time from ~700 seconds to ~580 seconds when running 
```
DESI_SPECTRO_REDUX=/global/cfs/cdirs/... 
source /global/cfs/cdirs/desi/software/desi_environment.sh 21.7e
srun -N 10 -n 320 desi_proc --traceshift --cameras a0123456789 -n 20210710 -e 00098135
```
on cori. 

A similar speed up is observed when writing and reading from `$SCRATCH`. The nature of the speedup is not clear, although it is possibly related to some known performance issues when using `astropy.io.fits` to write binary tables to a fits file.  

Results are unchanged for this pull request, i.e. this is 'non-algorithmic'. Additionally, the behaviour of the code is identical to the master branch if `DESI_LOCALBUFF` is not defined.

### DIFF
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -60,7 +60,7 @@ def main(args, comm=None):
 
     imgfile = args.input_image
     outfile = args.output_psf
-        
+
     nproc = 1
     rank = 0
     if comm is not None:
@@ -187,7 +187,10 @@ def main(args, comm=None):
 
         log.debug("proc {} calling {}".format(rank, " ".join(com)))
 
+        t0 = time.time()
         retval = run_specex(com)
+        dt = time.time() - t0
+        log.info(f'PSF fit for bundle {b} on camera {camera} took {dt:.1f} sec')
 
         if retval != 0:
             comstr = " ".join(com)
@@ -218,7 +221,10 @@ def main(args, comm=None):
             sys.stdout.flush()
             time.sleep(5.)
 
+            t0 = time.time()
             merge_psf(inputs,outfits)
+            dt = time.time() - t0
+            log.info(f'PSF merging for camera {camera} took {dt:.1f} sec')
 
             log.info('done merging')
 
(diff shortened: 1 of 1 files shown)


moustakas:
> A similar speed up is observed when writing and reading from $SCRATCH. The nature of the speedup is not clear, although it is possibly related to some known performance issues when using astropy.io.fits to write binary tables to a fits file.

In some recent profiling tests with `fastspecfit` I found `astropy.io.fits` to be about 5-times slower than `fitsio` writing a very simple multi-extension HDU (like 1.5 seconds vs 0.3 seconds).

araichoor:
for what is worth, I noticed in Dec. 2020 a similar super-slow performance of writing to $SCRATCH with astropy.io.fits.
I copy here the answer of @sbailey:
"There is a bug in the Lustre filesystem used by $SCRATCH that gets triggered by astropy.io.fits, resulting in very slow writes for table objects.  If you use fitsio instead it will be much faster.  The vendor has acknowledged the performance bug, but doesn't have a workaround."
Suggested options: switch to fitsio or use CFS instead of SCRATCH.

marcelo-alvarez:
Thanks @moustakas and @araichoor for the feedback. 

I discovered the botleneck when writing to SCRATCH and found adding these two lines fixed it without needing to stop everything and change over to fitsio. My initial test writing to CFS indicated a similar increase in performance, but looking back through the logs I am no longer so confident there is any IO issue on CFS, consistent with the recommendation @sbailey gave to use CFS when encountering the astropy.io.fits-triggered performance bug on SCRATCH. 

Changing to fitsio for the writing of the merged PSF would be a more robust long-term solution for addressing SCRATCH-specific specex performance issues. However, since performance of specex when writing to SCRATCH is not a high priority for ongoing development, I am closing this PR.

PR #1407, opened 2021-09-07
Tiles table updates

julienguy:
Minor change to desi_tsnr_afterburner and tilecompletness. 
Only update in the tiles table file the information about tiles affected by a selection of nights or exposures.


sbailey:
Looks good to me, but let's wait for @akremin to verify with the specific cases he was testing.

PR #1410, opened 2021-09-10
Tile VI

julienguy:
Minor updates on Tile QA and VI
 - Petal with only 2 calibration stars are still considered valid if the difference of calib between the two stars is < 0.05*sqrt(2).
 - Option to only selection one survey in VI
 - QA can run only both old and new redrock file name convention (zbest-* and redrock-*)


sbailey:
I think `max_rms_of_rflux_ratio_of_stdstars_if_few_stars` is not a sufficiently descriptive variable name and you really should spell out "maximum" and "root_mean_square" and "flux_standard_stars" and "if_too_few".  But I won't quibble.  :)

looks good, merging now.

PR #1411, opened 2021-09-13
Averaging of PSFs with different wavelength ranges

marcelo-alvarez:
Attempt to fix bugs preventing PSF and trace coefficients from being correctly averaged in `desispec.specex.mean_psf` whenever the wavelength range differs between the input PSF fits:
1. `legval` and `legfit` were not being imported from `numpy.polynomial.legendre`
2. traces were being averaged redundantly for every entry in the table of PSF coefficients 
3. traces were being re-fit using the wrong number of legendre polynomial coefficients

These changes may require further validation; there is no before and after comparison possible for the use case that has changed since it was crashing for these cases (i.e. whenever the wavelength range differs between the input PSF fits).

### DIFF
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -60,7 +60,7 @@ def main(args, comm=None):
 
     imgfile = args.input_image
     outfile = args.output_psf
-        
+
     nproc = 1
     rank = 0
     if comm is not None:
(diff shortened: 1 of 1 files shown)


julienguy (approved):
Thanks for finding those bugs.
Tested on June data.

PR #1414, opened 2021-09-19
Only consider valid sky fibers

julienguy:
Compare the number of valid measurements per wavelength to the number of valid sky fibers (i.e. fibers that have some non null ivar values after the call to get_fiberbitmasked_frame_arrays) instead of comparing to the total number of sky fibers.
This feature caused sky subtraction to set ivar=0 to full frames. It happened in the first few nights after restart because many sky fibers have large positioning errors (>100 um).

### DIFF
--- a/py/desispec/sky.py
+++ b/py/desispec/sky.py
@@ -195,9 +195,18 @@ def compute_uniform_sky(frame, nsig_clipping=4.,max_iterations=100,model_ivar=Fa
         frame.fibermap['FIBERSTATUS'][skyfibers] &= ~1
 
     nwave=frame.nwave
-    nfibers=len(skyfibers)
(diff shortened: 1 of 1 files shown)


akremin (approved):
Should there be a check to see `if np.sum(bad)==skyfibers` (or equivalently `nfibers==0` after it's redefined)?

I'm going to approve and merge this because it's in the right direction and I'd like it in for the pipeline tonight. But we may want to consider throwing an error if the number of sky fibers is 0.

akremin:
I don't see how the Test coverage failures could be due to this simple code change. The error messages point toward issues in coveralls reporting, rather than coveralls actually reporting an issue in the desispec code.

PR #1418, opened 2021-09-22
Job scheduler for specex

marcelo-alvarez:
Adds new functionality to schedule specex jobs dynamically in `py/desispec/scripts/proc.py`, by assigning each ccd image ('camera') to a group of 20 MPI processes ('group') for PSF fitting (one process for each bundle on the image) from a queue, such that when a group finishes PSF fitting for a given camera it is assigned the last undone camera on the list as currently created in `proc.py` . The process continues until all cameras have had their PSFs fit. 

Because an additional single 'scheduler' process is necessary, the number of processes available for PSF fitting is `nranks-1`. For example, if `proc.py` is run for 30 cameras ([b0,...,b9,r0,...,r9,z0,...z9]) and called with a total of 201 MPI processes available in MPI_COMM_WORLD, then 10 groups of 20 MPI processes will be created, and each group will each first be assigned one of [z0,...,z9] (the last 10 cameras in the list created in `proc.py`). Whichever group finishes first will then be assigned r9 (the last undone camera on the list), and so on.  

On cori, running (with desi environment 21.7e, the current version of specex, and `DESI_SPECTRO_CALIB=/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk`)
```
srun -N 10 -n 301 -c 2 desi_proc --traceshift --cameras a0123456789 
  -n 20210710 -e 00098135 --mpi
```
with this change results in a reduction of about 15% of the total time spent for PSF fitting with respect to running 
```
srun -N 10 -n 300 -c 2 desi_proc --traceshift --cameras a0123456789 
  -n 20210710 -e 00098135 --mpi
```
without this change (~390 vs. ~460 sec), with bitwise identical results for all resulting output PSF fit files. 

Ready to be reviewed, tested in a production environment, and/or merged.

### DIFF
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -235,16 +235,31 @@ def main(args, comm=None):
         # all processes throw
         raise RuntimeError("merging of per-bundle files failed")
 
-    return
-
 def run(comm,cmds,cameras,log):
+    """
+    Run PSF fits with specex on a set of ccd images 
 
-    import desispec.scripts.specex
-    from desispec.workflow.schedule import schedule
+    Args:
+        comm:    MPI communicator 
+        cmds:    dictionary of command arguments to main method of this module
+        cameras: entries in cmds to be run in parallel, one entry per ccd image to be fit
+        log:     logging.Logger object 
+    """
+    from desispec.workflow.schedule import Schedule
 
     group_size = 20
-    cameras.reverse() # expects b cameras first; reverse to do them last since they take least time
+    # reverse to do b cameras last since they take least time
+    cameras = sorted(cameras, reverse=True)
+    cameras.reverse() 
     def fitbundles(comm,job):
+        """
+        Run PSF fit with specex on all bundles for a single ccd image 
+
+        Args:
+            comm: MPI communicator 
+            job:  job index corresponding to position in list of cmds entries 
+        """
+
         rank = comm.Get_rank()
         camera = cameras[job]
         if camera in cmds:
@@ -256,7 +271,7 @@ def run(comm,cmds,cameras,log):
                 log.info(f'MPI ranks {rank}-{rank+group_size-1}'+
                          'fitting PSF for {camera} at {timestamp}')
             try:
-                desispec.scripts.specex.main(cmdargs, comm=comm)
+                main(cmdargs, comm=comm)
             except Exception as e:
                  if rank == 0:
                      log.error(f'FAILED: MPI group ranks {rank}-{rank+group_size-1}'+
@@ -269,7 +284,7 @@ def run(comm,cmds,cameras,log):
 
         return
 
-    sc = schedule(fitbundles,comm=comm,Njobs=len(cameras),group_size=group_size)
+    sc = Schedule(fitbundles,comm=comm,Njobs=len(cameras),group_size=group_size)
     sc.run()
 
     return
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -18,6 +18,13 @@ from astropy.io import fits
 
 from desiutil.log import get_logger
 
+class timer:
+    def __init__(self, **kwargs):
+        self.tinit = time.time()
+    def report(self,message):
+        dt = time.time() - self.tinit
+        print('TIMING REPORT: ',message.ljust(40), dt,' seconds',flush=True);
+
 def parse(options=None):
     parser = argparse.ArgumentParser(description="Estimate the PSF for "
         "one frame with specex")
@@ -50,10 +57,10 @@ def parse(options=None):
         args = parser.parse_args(options)
     return args
 
-
 def main(args, comm=None):
 
     log = get_logger()
+    total_timer=timer()
 
     #- only import when running, to avoid requiring specex install for import
     from specex.specex import run_specex
@@ -120,7 +127,6 @@ def main(args, comm=None):
 
     # Now we assign bundles to processes
 
-
     mynbundle = int(nbundle / nproc)
     leftover = nbundle % nproc
     if rank < leftover:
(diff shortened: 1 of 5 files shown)


py/desispec/workflow/schedule.py
```diff
@@ -0,0 +1,127 @@
...
+        return
+
+    def _work(self):
+
+        # listen for job assignments from the scheduler
+        while True:
+            self.comm.Recv(self.job_buff,source=0) # receive assignment from rank=0 scheduler
+            job = self.job_buff[0]                 # unpack job index
+            if job < 0: return                     # job < 0 means no more jobs to do
+            self._workfunc(self.groupcomm,job)     # call work function for job
```
sbailey, line 114: I think this should be wrapped in a generic try/except block so that if the function raises an exception, it doesn't end up blocking the scheduler until the job times out (since that worker will never get to the final barrier).  There might be other solutions, but please test the scheduler with the specific case of some jobs raising exceptions and others succeeding, and making sure that doesn't lead to a hang.

sbailey (changes requested):
Thanks for this update.  I like that the scheduler is implemented as a generic toolkit that we could reuse, without it being tied to just the case of specex.

Inline I added a number of minor style suggestions and requests for docstrings, and one substantive issue to explore to make sure that individual jobs raising exceptions don't cause the scheduler to hang until the batch job times out (that's a bug that we have introduced and fixed and re-introduced several times already...).

Also, please track down where `desi_proc --batch` is generating the `srun -n ...` call for arcs and update it so that it includes the extra rank needed for the scheduler.  As is, this PR works when a human knows how many ranks to give it, but the pipeline will call desi_proc with the wrong number of ranks to make good use of the scheduler.

marcelo-alvarez:
> Thanks for this update. I like that the scheduler is implemented as a generic toolkit that we could reuse, without it being tied to just the case of specex.
> 
> Inline I added a number of minor style suggestions and requests for docstrings, and one substantive issue to explore to make sure that individual jobs raising exceptions don't cause the scheduler to hang until the batch job times out (that's a bug that we have introduced and fixed and re-introduced several times already...).
> 
> Also, please track down where `desi_proc --batch` is generating the `srun -n ...` call for arcs and update it so that it includes the extra rank needed for the scheduler. As is, this PR works when a human knows how many ranks to give it, but the pipeline will call desi_proc with the wrong number of ranks to make good use of the scheduler.

Thanks for the thorough review. 

Minor style suggestions and requests for docstrings have been resolved, and  wrapping the call to workfunc in a generic try/except block has been verified to prevent individual jobs raising exceptions from causing the scheduler to hang until the batch job times out. 

Additionally the place where desi_proc --batch is generating the srun -n ... call for arcs has been tracked down and updated so that it includes the extra rank needed for the scheduler. As is, this PR should now result in the pipeline calling desi_proc with the correct number of ranks to make good use of the scheduler.

Please re-review and either merge or comment further.

py/desispec/scripts/specex.py
```diff
@@ -237,6 +237,60 @@ def main(args, comm=None):
...
     return
 
+def run(comm,cmds,cameras):
+    """
+    Run PSF fits with specex on a set of ccd images 
+
+    Args:
+        comm:    MPI communicator 
+        cmds:    dictionary of command arguments to main method of this module
+        cameras: entries in cmds to be run in parallel, one entry per ccd image to be fit
```
sbailey, line 247: Documentation clarification requests:

`cmds` is a dictionary keyed by a camera string (e.g. 'b0', 'r1', ...) with values being the `desi_compute_psf ...` string that one would run on the command line?  i.e. not the cmdargs for main(cmdargs) themselves, nor the list that is passed to parse(...), right?

`cameras` is a list of keys in the `cmds` dict that should be processed?  It looks like you support the ability to pass `cameras` entries that aren't a key to `cmds`.  Is that an intended use-case, or should that be logged as an error (while still continuing with the others and not crashing)?

The `comm` here would normally be a "big" communicator that would be split into smaller sub-communicators to process individual cameras (with the limiting case of comm having 20 ranks and not getting split).  Are there any constraints on comm having an even multiple of 20 ranks?  Would it ever make sense to support a smaller group_size as an input argument for different packing?

py/desispec/workflow/schedule.py
```diff
@@ -0,0 +1,189 @@
...
+        Args:
+            workfunc: function to do each MPI job defined using 
+                      def workfunc(comm,job):
+                          where comm is an MPI communicator 
+                          and job is the index of the job 
+            
+        Keyword Args:
+            comm:       MPI communicator (default=None)
+            njobs:      number of jobs (default=2)
+            group_size: number of MPI processes per job (default=1)
```
sbailey, line 29: Normally we could define this with
```
def __init__(self, workfunc, comm=None, njobs=2, group_size=1):
    ....
```
so that python introspection (e.g. with pydoc) can automatically see what keywords are available and what their defaults are.  Unless there is a specific need to use generic `**kwargs`, let's use this more specific form.

py/desispec/workflow/schedule.py
```diff
@@ -0,0 +1,189 @@
...
+        # listen for job assignments from the scheduler
+        while True:
+            self.comm.Recv(self.job_buff,source=0) # receive assignment from rank=0 scheduler
+            job = self.job_buff[0]                 # unpack job index
+            if job < 0: return                     # job < 0 means no more jobs to do
+            try:
+                self._workfunc(self.groupcomm,job) # call work function for job
+            except Exception as e:
+                self.log.error(f'FAILED: call to workfunc')
+                self.log.error(e)
```
sbailey, line 174: Commentary: I see that you also have a try/except within the spexec workfunc, but it is good for this general class to not trust the user to manage their own exceptions and be robust to that, even if it is somewhat redundant for this case.

sbailey (approved):
Thanks for the updates.  I put some additional documentation clarification requests inline to benefit our future selves reading this code.

PR #1421, opened 2021-09-24
add desi_update_tiles_specstatus

### LINKED ISSUE #1420: tiles -> tiles-specstatus update script
We need a script that takes the tile processing status (tsnr, efftime, ...) in daily/tiles.csv and propagates it into the surveyops/ops/tiles-specstatus.ecsv file for followup with human QA inspection, afternoon planning, etc. while not stepping on the QA related columns USER, QA, OVERRIDE, and ZDONE.  I'm working on this script, but documenting a few choices and questions here:

Any *new* TILEID rows in `tiles` should be added to `specstatus`, with USER='none', QA='none', OVERRIDE=0, and ZDONE='false'.

Any existing TILEID with ZDONE='false' and tiles['LASTNIGHT'] > specstatus['LASTNIGHT'] should also be updated for the non-qa columns.

But what about cases where ZDONE='true' but for whatever reason we got more data anyway such that tiles['LASTNIGHT'] > specstatus['LASTNIGHT']?  Should we freeze the `specstatus` row to reflect the status at the time the QA was done, even if more data have arrived (and thus have `tiles` and `specstatus` get out of sync for those TILEID), or should we go ahead and update the non-qa columns like EFFTIME_SPEC to reflect the current processing state even if QA was already finished and the tile was released?

I'm inclined to *not* check+update every row every time just in case the TSNR or EFFTIME calculation changed a little and impacted every tile, and leave that for a special case update.  The script could still support it, but I don't think that should be default so that under normal use any updates to tiles-specstatus.ecsv reflect changes due to new data and nothing else.

@julienguy, opinions?

sbailey:
This PR adds `bin/desi_update_tiles_specstatus` to update svn surveyops/ops/tiles-specstatus.ecsv with with processing information from spectro/redux/daily/tiles.csv, without clobbering QA columns.

To decide (but doesn't apply to any current data): if specstatus['ZDONE'] == 'true' and (tiles['LASTNIGHT'] > specstatus['LASTNIGHT']), should we update the non-QA columns or not?  i.e. if we did QA and flagged a tile as done, but for some reason got more observations anyway, what should the default update do?  Freeze the EFFSTATUS_SPEC etc. at the state it was when we declared the tile done, or update it to the current status of data processing?

This script is useable as-is, but if I have time before final merging, some other items to do:
  * add pre-check whether svn update needs to be run (an option to automatically run it)
  * maybe: add option to svn commit afterward (though I'm not sure that is a good idea to do automatically)
  * add unit tests to check corner cases and confirm intended behavior

fixes #1420

### DIFF
--- a/bin/desi_tile_vi
+++ b/bin/desi_tile_vi
@@ -143,7 +143,7 @@ def main():
         tiles_table["OVERRIDE"]=np.zeros(len(tiles_table),dtype=int)
 
     log.info("Found {} tiles in {}".format(len(tiles_table),args.infile))
-    selection=(tiles_table["ZDONE"]=="false")&(tiles_table["EFFTIME_SPEC"]>=(tiles_table["GOALTIME"]*tiles_table["MINTFRAC"]))&(tiles_table["QA"]!="good")&(tiles_table["QA"]!="bad")
+    selection=(tiles_table["ZDONE"]=="false")&(tiles_table["EFFTIME_SPEC"]>=(tiles_table["GOALTIME"]*tiles_table["MINTFRAC"]))&((tiles_table["QA"]!="good")&(tiles_table["QA"]!="bad") | (tiles_table["QA"]=="bad")&(tiles_table["LASTNIGHT"]>tiles_table["QANIGHT"]))
     if args.new :
         selection &= (tiles_table["QA"]=="none")
     if args.survey is not None :
@@ -219,6 +219,7 @@ def main():
                 tiles_table["ZDONE"][index]="true"
                 tiles_table["QA"][index]="good"
                 tiles_table["USER"][index]=args.user
+                tiles_table["QANIGHT"][index]=tiles_table["LASTNIGHT"][index]
                 if software_answer is not None :
                     if software_answer is False :
                         tiles_table["OVERRIDE"][index]=1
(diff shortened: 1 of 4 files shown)


julienguy:
" if specstatus['ZDONE'] == 'true' and (tiles['LASTNIGHT'] > specstatus['LASTNIGHT']), should we update the non-QA columns or not?  " 
yes, (including LASTNIGHT and all timing measures), and we may add a column QA_NIGHT to save the value of LASTNIGHT when the QA was done.

julienguy (approved):
Good for me. Thanks.

julienguy:
Added some changes: 
 - dropping all QA related columns (ZDONE,QA,USER,OVERRIDE) to the file `tiles.csv` written by `desi_tsnr_afterburner` or `desi_tiles_completeness`.
 - add QANIGHT column, initialized in `desi_update_tiles_specstatus` , and tested and filled by `desi_tile_vi` in file `ops/tiles-specstatus.ecsv`
 
(The actual files on disk have not been modified yet, waiting for pipeline update from last night 20210926)

julienguy:
The files on disk have been modified. (tiles.csv in daily and tiles-specstatus.ecsv in surveyops)

PR #1427, opened 2021-10-01
parallelize fiberflatnight; cleanup tempfiles

### LINKED ISSUE #1397: fiberflatnight should remove tmpfiles
The process of generating the fiberflatnight files leaves behind calibnight/YEARMMDD/tmp/fiberflatnight-camera-*-lamp-?.fits files, which in everest added up ~290 GB of leftover tmp files, vs. ~75 GB of actual calibnight files.

Clean those up on-the-fly after successfully building the fiberflatnight.

sbailey:
This PR refactors the fiberflatnight part of `desispec.scripts.proc_joint_fit` to:
  * use MPI parallelism
    * previous code was getting 160 ranks but only using rank 0, taking 10 minutes
    * now it takes < 1 min
  * cleanup the per-lampbox temporary files if all final fiberflatnight files succeeded (fixes #1397 for Fuji)

Example outputs are in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/fiberflat, generated with
```
export DESI_SPECTRO_REDUX=/global/cfs/cdirs/desi/users/$USER/spectro/redux
export SPECPROD=fiberflat
mkdir -p $DESI_SPECTRO_REDUX/$SPECPROD
cd $DESI_SPECTRO_REDUX/$SPECPROD

copyprod --night 20210929 /global/cfs/cdirs/desi/spectro/redux/daily .
rm calibnight/20210929/fiberflatnight*.fits

time srun -N 5 -n 160 -c 2 desi_proc_joint_fit --mpi --traceshift --obstype flat \
    --cameras a0123456789 -n 20210929 \
    -e 102203,102204,102205,102208,102209,102210,102213,102214,102215,102218,102219,102220
    
ls calibnight/20210929/fiberflatnight-*.fits | wc -l
fitsdiff --ignore-keywords DATASUM,CHECKSUM calibnight/20210929/fiberflatnight-b0-20210929.fits /global/cfs/cdirs/desi/spectro/redux/daily/calibnight/20210929/fiberflatnight-b0-20210929.fits
```
The outputs are bitwise-identical with SPECPROD=daily for the data, and differ in the headers only due to timestamps and DEPNAM/VER for desispec.


py/desispec/scripts/proc_joint_fit.py
```diff
@@ -230,84 +230,97 @@ def main(args=None, comm=None):
...
+            for lampbox in range(4):
+                ofile = os.path.join(tmpdir, f"fiberflatnight-camera-{camera}-lamp-{lampbox}.fits")
+                average_flats_for_camera[camera].append(ofile)
+                camera_lampboxes.append( (camera, lampbox, ofile) )
+                arm = camera[0].lower()
+                flats_for_arm[arm].append(ofile)
+
+        #- rank 0 checks if we have enough inputs
+        enough_inputs = True
+        if rank == 0:
```
akremin, line 266: Is there benefit to all ranks performing lines 239-262 here (generating `camera_lampboxes`, `flats_for_arm`, `inflats_for_camera`) as opposed to broadcasting? Maybe the broadcasting is no more efficient than the minor amount of redundant computation, but I just wanted to point it out.

py/desispec/scripts/proc_joint_fit.py
```diff
@@ -230,84 +230,97 @@ def main(args=None, comm=None):
...
+
+        if comm is not None:
+            comm.barrier()
+
+        log.info("Auto-calibration across lamps and spectro  per camera arm (b,r,z)")
+        for camera_arm in ["b", "r", "z"][rank::size]:
+            log.info(f"Rank {rank} autocalibrating across spectro for camera arm {camera_arm}")
+            cmd = f"desi_autocalib_fiberflat --night {args.night} --arm {camera_arm} -i "
+            for flat in flats_for_arm[camera_arm]:
+                cmd += f" {flat} "
```
akremin, line 316: Could remove the trailing space here and in line 314, though it shouldn't matter

akremin (approved):
This looks good and the bitwise identical results show that it works. I include an implementation question and a trivial comment, but I am also fine with merging as-is given that the current version works.

py/desispec/scripts/proc_joint_fit.py
sbailey: Good question.  When MPI ranks are idle at a broadcast barrier waiting for another rank catch up and send them information, they are pegging their individual core doing repeated checks for whether the data have arrived yet.  So in cases where every rank needs the same calculable info, it isn't faster to have one rank calculate + broadcast and it becomes a matter of style for what you consider more readable/maintainable (which is not a strong feature of the MPI parallelism model in the first place...).  I tend to prefer the style above to avoid extra `if rank == 0...`, `if comm is not None` boilerplate.

However, places where I do calculate+broadcast:
  * if the calculation involves I/O (so that only one rank reads the file instead of N>>1 ranks)
  * if the calculation involves significant memory that could cause ranks to compete for the CPU cache

py/desispec/scripts/proc_joint_fit.py
sbailey: It shouldn't matter for the final command, and it was semi-unintentional (from me, I was copying earlier code from Daniel who may have gotten it from you :) ), but it provides the epsilon of safety that if another line of code appends another option it still works even if they forget to include a leading space.  I'll use that to justify leaving it as is :)

FWIW, I had also considered
```
cmd += ' '.join(flats_for_arm[camera_arm])
```
but that seems more clever than obvious.

PR #1430, opened 2021-10-03
tile QA defaults / usability

sbailey:
This PR adds some convenience features to simplify the daily tile VI procedure:
  * updated desi_tile_vi defaults so that you can just run "desi_tile_vi" from a surveyops/ops svn directory without having to specify extra arguments:
    * `--infile` defaults to tiles-specstatus.ecsv instead of tiles.csv (which was both wrong and now crashes anyway since QA columns have been removed from tiles.csv)
    * `--user` defaults to unix username
  * desi_update_tiles_specstatus
    * checks if input tiles-specstatus.ecsv is up-to-date or needs an svn update first; use --force to override (at your own peril... svn merge conflicts are likely if you aren't up-to-date first)
    * --specstatus now defaults to "tiles-specstatus.ecsv"
    * new --dry-run option to calculate what would be updated without saving any files

Standard usage would now be:
```
export SPECPROD=daily
cd (directory where you have surveyops/ops checkout)
svn update .
desi_update_tiles_specstatus --dry-run
desi_update_tiles_specstatus
desi_tile_vi
svn commit -m "tile QA for YEARMMDD"
```

The `desi_update_tiles_specstatus --dry-run` is optional, but a useful cross check before changing anything, and both desi_update_tiles_spectstatus commands will complain if you forgot to svn update first.

@schlafly @araichoor @julienguy

### DIFF
--- a/bin/desi_tile_vi
+++ b/bin/desi_tile_vi
@@ -12,6 +12,7 @@ This script is used to do a visual validation of the tiles based on the automate
 import os,sys
 import argparse
 import glob
+from getpass import getuser
 import numpy as np
 import multiprocessing
 from astropy.table import Table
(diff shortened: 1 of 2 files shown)


schlafly:
I promised I'd suggest some additional usability improvements:
- support for people with long names (already addressed)
- "only new exposure" mode in desi_tile_vi.  I think we want people doing daily maintenance of newly observed exposures to only see the new exposures, and not old "unsure" observations; if someone's said they're unsure, we want an expert to look at it, not for these likely problematic tiles to be mixed in with a bunch of likely good tiles until a non-expert just validates them and whatever issue was present gets lost.
- I think desi_tile_vi should print a summary at the end that would be appropriate for copy/pasting to #survey-ops.  e.g.

| type | valid | unsure | invalid |
| - | - | - | - |
|new dark||||
|new bright||||
|new other||||
|old dark||||
|old bright||||
|old other|||||
|very old dark||||
|very old bright||||
|very old other||||

Here's I'm imagining separating the new ones (defined as "formerly had 'none' as QA"), the old ones (not none, and lastnight within 2 weeks?), and the very old ones (not none, and lastnight older than 2 weeks?).

julienguy:
`desi_tile_vi` already has the option `--new` which I think does exactly what you want. 
I'm adding the option in the command line on the wiki page.

schlafly:
Thanks, that looks exactly right, sorry I missed it.

schlafly:
I added my features:
- save every time
- remove save / exit options (can just Ctrl-D or Ctrl-C to exit, since it saves every time)
- pass is renamed "unsure"
- print a summary table at end
I was able to use this for my VI run earlier today, but I can't say that I've tested it more than that.

sbailey:
Live use for real QA is good enough testing.  Merging now.  Thanks for adding the additional features.

Issue #1432, opened 2021-10-05
sky subtraction runtime variability


dmargala:
The sky subtraction step of the pipeline has a widely varying runtime, likely due to an iterative algorithm and varying data quality.

The plot below shows the median runtime of the sky subtraction step from the everest pipeline run grouped by observation (orange). For comparison, the extraction step is also shown (blue). There appears to be a jump in the median runtimes for sky subtraction around the full moon. The runtimes (y-axis) are in seconds.

![image](https://user-images.githubusercontent.com/1648834/136075440-d6f3b7a1-a83e-44ab-ae88-898a378e1bfc.png)

The figure below shows runtimes for each exposure's sky subtraction step from the everest pipeline run split by FAPRGRM. The runtimes (y-axis) are in seconds.

![image](https://user-images.githubusercontent.com/1648834/136075383-ffbcfdc4-51f0-40f4-b4e7-d2b68af00eb0.png)

The runtimes are from the `duration.max` field of the `skysub` step in the everest run timing files `/global/cfs/cdirs/desi/spectro/redux/everest/run/scripts/night/*/prestdstar-*timing-*.json` and joined with each corresponding exposure's `FAPRGRM` entry in `/global/cfs/cdirs/desi/spectro/redux/everest/exposures-everest.csv`.

PR #1436, opened 2021-10-08
Optimize desi_compute_sky

dmargala:
This PR includes performance optimizations to desi_compute_sky to address #1432 .

I've tested that the outputs using this branch are bitwise-identical according to fitsdiff (except for timestamps/checksums in headers).

```
cd desispec
git checkout optimize-skysub

salloc -C haswell -N 1 -q interactive -t 60

# current desispec
source /global/cfs/cdirs/desi/software/desi_environment.sh master
time desi_compute_sky -i /global/cfs/cdirs/desi/spectro/redux/everest/exposures/20201222/00069407/frame-r0-00069407.fits --fiberflat /global/cfs/cdirs/desi/spectro/redux/everest/calibnight/20201222/fiberflatnight-r0-20201222.fits --o sky-r0-00069407.fits --adjust-wavelength --adjust-lsf
...
real	2m48.997s
user	2m42.319s
sys	0m1.246s

# using this branch
module unload desispec
export PATH=$(pwd)/bin:$PATH
time desi_compute_sky -i /global/cfs/cdirs/desi/spectro/redux/everest/exposures/20201222/00069407/frame-r0-00069407.fits --fiberflat /global/cfs/cdirs/desi/spectro/redux/everest/calibnight/20201222/fiberflatnight-r0-20201222.fits --o optimized-sky-r0-00069407.fits --adjust-wavelength --adjust-lsf
...
real	0m13.179s
user	0m7.202s
sys	0m0.931s


fitsdiff sky-r0-00069407.fits optimized-sky-r0-00069407.fits
```


sbailey:
Thanks @dmargala .  A 13x speedup with bitwise identical results is great.  The loop -> array operation compacting makes the algorithm harder to understand (for me at least), so before we commit to this, would @julienguy like to take a look at whether the long bright time runtimes indicate some other convergence problem in the model?  Or otherwise we might want to leave the original loop-based code behind, commented out, with a comment that the array based logic is bitwise identical but much faster.

At minimum we could do the cheap win of fixing the `log.debug` if @julienguy wants more time with the original code.

julienguy:
Merged changes in master, and adapted to code to the PCA sky correction method (which was implemented after this PR was opened). I obtain the same sky model as with the master version for the test exposure/camera r0-00069407 mentioned above and for the exposure/camera z7-00111597 from yesterday IF I leave a (minor) bug in the computation of the chi2 used to select valid sky lines (first noted by Daniel in a comment in the code). I have now fixed this bug so there is not any more a perfect agreement with the master version.

julienguy:
Merging after running another test in prod (meaning using desi_proc in a batch job).

PR #1443, opened 2021-10-13
Tile qa plot v2

araichoor:
This PR updates the tile qa png file, used to human-validate recent observations, displayed by `desi_vi_tile`.

New features, based on lessons learned so far:
- add a focal plane plot with the fiber positioning precision
- add a Z vs. FIBER scatter plot (including all fibers)
- list individual exposures from the considered tile (with their EFFTIME_SPEC)
- change colormap
- bright zhist: reducing zrange to (0, 1.5), as the Z vs. FIBER should allow to catch unsually high zspecs
- few extra display choices (hopefully improving readibility).

I hope that the png file is not too busy now..
For listing the exposures, I have introduced a dependence on the `tsnr-exposures.fits` file; however, I do not know what is the correct way to obtain the filename, based on specprod (e.g. tsnr-exposures.fits for the daily, but exposures-everest.fits for everest); I came up with an ad hoc approach, which may need to be changed: https://github.com/desihub/desispec/blob/4dd9eddd6a93dc56b52466854a23559826f7ca92/bin/desi_tile_qa#L107-L108

Another thing to do is to report  (in _efftime/goaltime_) the EFFTIME_SPEC which is used throughout the daily process, because here it is recomputed and can differ from what is in other places.

Ideally, I would also like to extend replace the QSO n(z) curve by two curves:
- one for first-time observed QSOs (with the reference, which should be very close to the current reference curve, as pre-shutdown mostly observed PASS=0 tiles);
- one for all observed QSOs on that tile (which is the current curve).
I remind that I only consider for this plot DELTACHI2>50 zspecs.
@ashleyjross : if I recompute the reference to do the above, what files should I use?
the ones previously used don t exist anymore ... https://github.com/desihub/desispec/blob/c0ce23382b1da64f3a22e820807917bce3b96e02/bin/desi_tile_qa_reference#L58-L65
I feel that using some everest files may be better, so that we are sure they are frozen.
I ll give a closer look tomorrow at that script to see what are the needs.

And of course, I would be happy to implement - or try to, at least - any further suggestion.
Lastly, I could also try to clean a bit the `py/desispec/tile_qa_plot.py` code, if useful.

Example for the dark 2061-thru20211010 (current version here: https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/2061/20211010/tile-qa-2061-thru20211010.png)
tile-qa-2061-thru20211010

Example the bright 25553-thru20211010 (current version here: https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/25553/20211010/tile-qa-25553-thru20211010.png)
tile-qa-25553-thru20211010

The cutouts are not present because of apparently some current viewer cutout issue, independent of this PR.

### DIFF
--- a/py/desispec/tile_qa_plot.py
+++ b/py/desispec/tile_qa_plot.py
@@ -893,8 +893,12 @@ def make_tile_qa_plot(
         ax.legend(ncol=2)
         ax.set_xlabel("Z")
         ax.set_ylabel("Per tile fractional count")
-        ax.set_xlim(0, 5)
-        ax.set_ylim(0, 0.2)
+        if hdr["FAPRGRM"].lower() == "bright":
+            ax.set_xlim(0, 1.5)
+            ax.set_ylim(0, 0.4)
+        else:
+            ax.set_xlim(0, 5)
+            ax.set_ylim(0, 0.2)
         ax.grid(True)
         # AR n(z) : ratio
         ratio_nz = n_valid / nref_valid


sbailey:
This looks good; let's merge it and begin using it now even if there are more things you'd like to do.  @araichoor please followup with @ashleyjross for your questions, or otherwise open another ticket to track those so that merging this doesn't cause those additional questions/features to get lost.

araichoor:
argh could be, sorry; I didn t test that part apparently yesterday night; I can look at this ~now.

PR #1445, opened 2021-10-13
Tile qa plot v2bis

araichoor:
This PR fixes a tsnrexpfits file from https://github.com/desihub/desispec/pull/1443, with a bugfix and a safer way to find the filename, looking for:
- tsnr-exposures.fits (daily)
- or exposures-{}.fits (everest)
- or tsnr-{prod}.fits (cascades, denali).

**Unrelated to the current changes:**
note that with testing for denali with some SV1 tile, I encounter the following error, where apparently the --args.outdir is not taken into account: 
```
desi_tile_qa --outdir ./ --nights 20201221 --prod denali --tileids 80622
...
  File "/global/homes/r/raichoor/software_dev/desispec_tile_qa_plot_v2/py/desispec/tile_qa.py", line 72, in compute_tile_qa
    write_exposure_qa(filename, exposure_fiberqa_table , exposure_petalqa_table)
...
PermissionError: [Errno 13] Permission denied: '/global/cfs/cdirs/desi/spectro/redux/denali/exposures/20201221/00069267/exposure-qa-00069267.fits.tmp'
```
Just mentioning in case it is some corner case in the code.

### DIFF
--- a/bin/desi_tile_qa
+++ b/bin/desi_tile_qa
@@ -104,8 +104,27 @@ def main():
     else:
         tileids = None
 
-    # AR FIXME path to tsnr-exposures.fits or e.g. exposures-everest.fits
-    tsnrexpfits = glob.glob(os.path.join(args.prod, "*exposures*fits"))[0]
+    # AR path to:
+    # - tsnr-exposures.fits (daily)
+    # - or exposures-{}.fits (everest)
+    # - or tsnr-{prod}.fits (cascades, denali)
+    tsnrexpfits = os.path.join(args.prod, "tsnr-exposures.fits")
+    if os.path.isfile(tsnrexpfits):
+        log.info("tsnrexpfits = {}".format(tsnrexpfits))
+    else:
+        log.info("no {} ".format(tsnrexpfits))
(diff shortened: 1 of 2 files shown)


julienguy:
One suggestion: different color for the fibers with FIBERSTATUS!=0 in the redshift vs fiber display.

sbailey:
Thanks.  The PermissionError on the old tile is because if tile_qa is missing one of its per-exposure QA inputs (e.g. from before when we were running exposure QA), it tries to automatically recreate that input, but it will fail if you aren't the desi user.  It could be updated to have a fallback directory, but I think the current situation is ok since normally tile_qa is run as the desi user as part of the pipeline anyway.  i.e. we can save that for another update later if really needed.

I'll let @julienguy and @araichoor coordinate for any other features like changing fiber colors, but let's merge this and update NERSC desispec by the end of the day so that the bugfix will be included in tonight's run.

akremin (approved):
Code looks good. Fixes the non-functioning desi_tile_qa in master, which makes this high priority. The coloring of fibers is also useful and your test worked.

Merging now as this is necessary for running the daily processing tonight

PR #1449, opened 2021-10-15
Tile qa plot v3

araichoor:
This PR improves the following on PR https://github.com/desihub/desispec/pull/1445: it does not rely anymore on the `tsnr-exposures.fits` file.
The exposures list from the coadd is obtained from the spectra-*fits file, and the per-exposure EFFTIME_SPEC is computed from the cframe*fits files, in a similar way as it is done in `desi_tsnr_afterburner`.
I checked against the 1133 main exposures observed so far that the EFFTIME_SPEC is consistent with what s in tsnr-exposures.fits

The reason for proceeding this way are:
- the `tsnr-exposures.fits` table is generated *after* the tile-qa*png files, so `tile_qa_plot.py` cannot rely on that table; this is why no expids are listed in the tile-qa plots since PR https://github.com/desihub/desispec/pull/1445;
- in `desi_tsnr_afterburner`, the process to compress the per-fiber, per-camera EFFTIME_SPEC to a single value for an exposure is scattered throughout the code and cannot easily be extracted into a function, that `tile_qa_plot.py` would call.

I also now display both the EFFTIME_SPEC from the offline pipeline (which will be in tsnr-exposures.fits, etc) and the one re-computed during the QA, dubbed _qa_efftime_, which is different because it considers a different set of criteria to compress the per-fiber, per-camera EFFTIME_SPEC into a single value.

Here is an example:
tile-qa-7439-thru20211008

### DIFF
--- a/bin/desi_tile_qa
+++ b/bin/desi_tile_qa
@@ -151,7 +151,7 @@ def main():
                         entry[k]=r['value']
                     summary_rows.append(entry)
                     continue
-            func_args.append({'night':night,'tileid':tileid,'specprod_dir':args.prod,'exposure_qa_dir':args.exposure_qa_dir,'tsnrexpfits':tsnrexpfits,'outfile':filename})
+            func_args.append({'night':night,'tileid':tileid,'specprod_dir':args.prod,'exposure_qa_dir':args.exposure_qa_dir,'outfile':filename})
 
         if args.nproc == 1 :
             for func_arg in func_args :
--- a/bin/desi_tile_qa
+++ b/bin/desi_tile_qa
@@ -52,7 +52,7 @@ def parse(options=None):
         args = parser.parse_args(options)
     return args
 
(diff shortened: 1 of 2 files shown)


julienguy:
I tried
```
desi_tile_qa --prod daily -n 20211010 --outdir . --nproc 16
and
desi_tile_qa --prod daily -n 20211010 --outdir . --nproc 16 --exposure-qa-dir . --recompute
```
and it produces QAs like https://data.desi.lbl.gov/desi/users/jguy/tile-qa4/tiles/cumulative/8115/20211010/tile-qa-8115-thru20211010.png with efftime=0.

julienguy:
yes, the idea of the optional --outdir ... arg was to write the QA in another directory for debugging, but the inputs, including the spectra files, should always be found in the prod directory.

julienguy:
Thanks! Merging.

PR #1454, opened 2021-10-20
Consistent string I/O and encoding in fibermap columns

### LINKED ISSUE #1451: desispec test failure in updated environment
When running 
```
python setup.py test 
```
in an updated conda environment (including [Astropy v4.3](https://docs.astropy.org/en/stable/whatsnew/4.3.html) see [pkg_list.txt](https://github.com/desihub/desispec/files/7356389/pkg_list.txt) for the full package list) the following failures occur: 

```
ERROR: test_QA (desispec.test.test_ql.TestQL)
...
ValueError: Camera None already in /global/homes/m/malvarez/ql_test_io/20150105/00000314/desi-00000314.fits.fz
FAIL: test_frame_rw (desispec.test.test_io.TestIO)
Test reading and writing Frame objects.
...
AssertionError: masked is not true : Fibermap column BRICKNAME mismatch
see
/global/common/software/desi/users/malvarez/testing-desiconda/test_21.10.14_v1/desispec.std[out,err]
```

marcelo-alvarez:
The default behaviour when creating an empty fibermap table with `desispec.io.fibermap.empty_fibermap` is that table entries of type string are initialized to be empty strings of length zero in columns that are instances of the `astropy.table.Column` class. When a fits file with such empy string table entries is read with the `astropy.io.Table.read` method (in Astropy v4.3.x), however, any column containing empty string table entries is an instance of the `astropy.table.MaskedColumn`, instead, and the formerly empty string table entries are masked with a fill value of `'N/A'[:n]` where n corresponds to the length of the string type.

This modification checks for masked columns with masked string entries with fill values of 'N..' in fibermap objects read in from fits files in the `desispec.io.frame.read_frame` method, replaces the masked string entries with the expected empty strings instead, and changes the affected column objects back to instances of `astropy.table.Column`, from `astropy.table.MaskedColumn`, ensuring that the frames read from a file with `desispec.io.frame.read_frame` is the same as the one that was written to that file with `desispec.io.frame.write_frame`. This modification also fixes #1451.

@sbailey please review and verify that this is an appropriate modification to eliminate the unit test failure from #1451 and merge if so.


moustakas:
In the context of this PR, let me state the obvious that `astropy` keeps changing this behavior (i.e., the interplay between `astropy.table.Table` and `astropy.io.fits`), so we may be chasing this again in the future. Given how much slower `astropy.io.fits` is relative to `fitsio`, is it worth thinking about switching to `fitsio` for I/O and better longer-term stability?

Issue #1419, opened 2021-09-22
Comparison of everest redshift catalog files


weaverba137:
# Executive Summary

* There are *expected* changes from SV1 to SV2 to SV3 to main.  *However*:
* `zpix-sv1-dark.fits` differs significantly from all other `zpix-*.fits` files.
* `ztile-sv1-bright-*.fits` differs significantly from `ztile-sv1-dark-*.fits`,
  which in turn differs significantly from all other `ztile-sv1-*.fits` files.
* `ztile-sv1-(bright|dark)-*.fits` files also differ significantly
  from all other `ztile-*.fits` files.
* More research is needed to resolve all of the anomalies in `ztile-sv1-(bright|dark)-*.fits` files.

# zpix

## main

These files are effectively identical in structure:

* `zpix-main-bright.fits`
* `zpix-main-dark.fits`

## sv1

These files are effectively identical in structure:

* `zpix-sv1-backup.fits`
* `zpix-sv1-bright.fits`
* `zpix-sv1-other.fits`

`zpix-sv1-backup.fits` has additional columns compared to `zpix-main-bright.fits`:

* `SV1_DESI_TARGET`
* `SV1_BGS_TARGET`
* `SV1_MWS_TARGET`
* `SV1_SCND_TARGET`

And the `SCND_TARGET` column is missing.

`zpix-sv1-dark.fits` has additional columns compared to `zpix-sv1-backup.fits`:

* `NUMTARGET`
* `BLOBDIST`
* `FIBERFLUX_IVAR_G`
* `FIBERFLUX_IVAR_R`
* `FIBERFLUX_IVAR_Z`
* `HPXPIXEL`

In addition the `RELEASE` column has a different data type (`J` versus `I` in
other files).

## sv2

These files are effectively identical in structure:

* `zpix-sv2-backup.fits`
* `zpix-sv2-bright.fits`
* `zpix-sv2-dark.fits`

`zpix-sv2-backup.fits` has additional columns compared to `zpix-main-bright.fits`:

* `SV2_DESI_TARGET`
* `SV2_BGS_TARGET`
* `SV2_MWS_TARGET`
* `SV2_SCND_TARGET`

And the `SCND_TARGET` column is missing.

## sv3

These files are effectively identical in structure:

* `zpix-sv3-backup.fits`
* `zpix-sv3-bright.fits`
* `zpix-sv3-dark.fits`

`zpix-sv3-backup.fits` has additional columns compared to `zpix-main-bright.fits`:

[...]

PR #1459, opened 2021-10-25
Consistent assemble_fibermap outputs

weaverba137:
This PR is an upstream fix for #1419.  The assumption is that if `desispec.io.fibermap.assemble_fibermap()` produces consistent output files for all surveys (but especially `sv1`), then this consistency will also be reflected in the downstream redshift catalogs.  This assumption should be tested of course.

I am presently testing fibermap file outputs for all existing surveys for comparison with fibermap files produced in `everest`.

*Do not merge yet, more testing is needed.*


weaverba137:
@sbailey, I've completed the round of tests effectively selecting one exposure from every night.  Summary:

* After testing on a large selection of nights, `assemble_fibermap()` produces consistent outputs for each survey type, including `cmx` now.
* See `desispec.test.analyze_fibermap` for the full test logic.

Remaining items:

* Decide what to do about null header cards.
* Review the logic for finding a fiberassign file in an earlier exposure when not available.
* When using `mergedep()` I arbitrarily picked `conflict='dst'`.  Confirm this is OK.
* Should we set units on these quantities?
* In general, carefully review all the steps in `assemble_fibermap()`.

sbailey:
Thanks for these updates.  I independently checked:
* The only header *keywords* differences are either
  * expected / correctly changed (e.g. dropping SIMPLE and adding CHECKSUM and DATASUM); or
  * harmless (e.g. SP1REDP 7.57799999999999e-08 vs. 7.578e-08)
* fitsverify no longer identifies any format *errors*, even if it still does have *warnings* for null values in the input headers (which we agreed to punt to future work, so ok here)
* the columns have the same *values* for every row

However, I found a problem with the *order* of the columns.  On summer 2021 everest-era data this PR matches the same order as before, but on more recent data this PR produces a column order that is different from both current code *and* from the order of summer 2021 data.  e.g.
```
assemble_fibermap --overwrite -n 20210620 -e 95405 -o fibermap-20210620-00095405-new.fits
```
produces the same columns in the same order as
```
/global/cfs/cdirs/desi/spectro/redux/everest/preproc/20210620/00095405/fibermap-00095405.fits
```

but on more recent data:
```
assemble_fibermap -n 20211101 -e 107000 -o fibermap-20211101-00107000-master.fits
```
produces columns in an order that is different from both current master (e.g. `/global/cfs/cdirs/desi/spectro/redux/daily/preproc/20211101/00107000/fibermap-00107000.fits`), and from the order of the columns in older data.  i.e. for new data it is *not* standardizing the column order; it's just generating some other order.

The underlying cause is that fiberassign changed its column order during the summer changes (e.g. compare 20210620/00095405/fiberassign-020903.fits.gz vs. 20211101/00107000/fiberassign-021057.fits.gz), and it appears that this PR is correcting a specific set of older known column order problems, without it being a more generic "I don't care what order you gave me the columns, make it this".

I think we should standardize on the current data column order in the hopes that the remainder of DESI data will naturally follow that (if we stop changing fiberassign output formats...), and that we should implement this in a generic way so that any arbitrary input order is shuffled to the correct output order, e.g. with something like:
```
# deal with any missing or extra columns
...
# then reorder to a standard fixed order no matter what the input order was
fibermap = fibermap[ ['list', 'of', 'correct', 'standard', 'columns'] ]
```

@weaverba137 thoughts?

weaverba137:
@sbailey, thank you and I agree that we should be working on a standard set of columns.  However, I have to ask, if the fiberassign column order can change, what is to stop it from changing again?  In other words, if the fiberassign column order *does* change again, how do we propagate that to fibermap?

sbailey:
> @sbailey, thank you and I agree that we should be working on a standard set of columns. However, I have to ask, if the fiberassign column order can change, what is to stop it from changing again? In other words, if the fiberassign column order does change again, how do we propagate that to fibermap?

For starters, we should keep our eye on any future fiberassign changes to make sure that the column order doesn't change.  I suspect that the last change was just an unintended side-effect of code reorganization, not some fundamental need to change the column order.

At the same time, I think we should make assemble_fibermap robust to *any* input column ordering, past or future, by standardizing it to match the current fiberassign -> fibermap column ordering and keep that output fixed regardless of input column order.  See code snippet above for reordering columns regardless of what the input order is.

Note: if they add/subtract columns again (not just reorder them), that is a separate issue that would need more care in how assemble_fibermap handles it.  But I really hope that won't happen, and it's a pretty high bar to make a change like that now, so it shouldn't happen without really good reason.

weaverba137:
Thanx @sbailey, that sounds fine.  Just a couple more comments:

1. There will of course, not be *one* column ordering, but one ordering for each survey: cmx, sv1, sv2, sv3, main, simply because they have different columns.
2. `fibermap.py` already defines `target_columns`, `fiberassign_columns`, etc. at the module level.  I assume as part of this we would need to reorder and update those definitions as well.  For example, `empty_fibermap()` should return an object with the same columns in the same order as `assemble_fibermap()`.  Do we also need to update `fibermap_new2old()`?

sbailey:
> Thanx @sbailey, that sounds fine. Just a couple more comments:
> 
> 1. There will of course, not be _one_ column ordering, but one ordering for each survey: cmx, sv1, sv2, sv3, main, simply because they have different columns.

Agreed.

> 2. `fibermap.py` already defines `target_columns`, `fiberassign_columns`, etc. at the module level.  I assume as part of this we would need to reorder and update those definitions as well.  For example, `empty_fibermap()` should return an object with the same columns in the same order as `assemble_fibermap()`.

Yes, that would be a good place to standardize the definition of the column ordering so that `empty_fibermap()` also returns them in the correct order.

>   Do we also need to update `fibermap_new2old()`?

We can deprecate that function; I don't think any code is using it any more (it was to ease the transition period after a few updates on column names years ago to match DRn targeting data column names).

weaverba137:
@sbailey, all column ordering tests passed on the sample of fibermap files I created.  If you want to look at the actual test fibermap files, see `${DESI_SPECTRO_REDUX}/bweaver/preproc`.  Please go ahead with final review.

sbailey:
This looks good; I've spot checked a few more cases and don't see any problems.  Let's wait for Monday to merge to not accidentally break something over the weekend.

PR #1464, opened 2021-10-29
Metric for the variation of overscan per row

julienguy:
In preproc.py ,
 - Compute the range of variation of the median overscan per row along the CCD columns
 - Save this value in the header keyword OSTEPX (with X the amplifier identifier).
 - If OSTEPX<1 ADU, use the standard average value in the overscan region
 - if 1<OSTEPX<15 ADU, use the median overscan per row 
 - if OSTEPX>15 ADU (or readnoise>10 elec), set BADREADNOISE CCD mask bit, which is subsequently propagated to the FIBERMAX and effectively discards all the fiber traces of this amplifier.

The motivation for this PR is in described https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=6516

### DIFF
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -9,6 +9,7 @@ import scipy.interpolate
 from pkg_resources import resource_exists, resource_filename
 
 from scipy import signal
+from scipy.ndimage.filters import median_filter
 
 from desispec.image import Image
 from desispec import cosmics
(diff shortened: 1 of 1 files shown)


sbailey:
Looks good, thanks.  I confirmed that amps that don't trigger row-by-row are unchanged, and with Julien adjusted thresholds to make this a little more conservative before triggering row-by-row overscan subtraction.  Will merge once that last test finishes running.

julienguy:
After the PR was merged, lowered the threshold from 15 to 5 ADU for the overscan step (OSTEP) to set the BADREADNOISE bit.
This change is motivated by the tile QA inspection of data from 2021/10/10 to 2021/10/29 where a few tiles with OSTEP>=8 ADU where found to have bad redshifts resulting from the CCD bias variations.

PR #1465, opened 2021-10-29
Add nightly bias and bad column flagging into desi_run_night

akremin:
### Overview
This is the final PR in the effort to add the bias modeling and bad column identification into the pipeline. This pull request incorporates darks into the reprocessing code, `desi_run_night`. 

Most of the underlying functions were updated yesterday. This primarily edits the coordination script `desi_run_night` so that it keeps and submits dark jobs. While doing so I also implemented a small feature that moves the calibrations to the start of the exposures so that cals are always processed first, in order of: 1) the first 300s dark, 2) arcs+flats, 3) science exposures.

I also fixed a bug meant to catch darks of the wrong `EXPTIME`.

### Reprocessing Test
I've tested this on 20211020 with multiple darks and it runs the first dark and skips the subsequent darks as expected. I tested it on one "typical" night, 20211027, and it did the proper thing there as well. Finally I ran a test night of jobs, stopping halfway through the sciences. The dependencies in the queue were correctly set and the jobs ran properly. I verified that the biases were used and the badcolumn files were also used. This was night 20211026 in the test prod.

commands:
`desi_run_night --dry-run-level=1 -n 20211020 &>test_reprocdarks_20211020.log &`
`desi_run_night -n 20211026`
`desi_run_night --dry-run-level=1 -n 20211027`

The prod used for testing:
`/global/cfs/cdirs/desi/users/kremin/spectro/redux/reprocdarks`

### daily processing (re-) Test
In addition, since I changed two thing in the daily processing script, I reran in dry-run mode:
`desi_daily_proc_manager --override-night=20211020 --dry-run-level=1 --ignore-cori-node`

and it once again did the right thing. The prod used for that test was:
`/global/cfs/cdirs/desi/users/kremin/spectro/redux/usedarks`

Note: I had already tested that night in the PR yesterday. Those results are in `*.orig.csv` versions of the output tables, while the standard names are the new ones produced for this test. I chose that night because it had two darks, which tests slightly more of the new coding logic.

### DIFF
--- a/py/desispec/scripts/daily_processing.py
+++ b/py/desispec/scripts/daily_processing.py
@@ -326,6 +326,13 @@ def daily_processing_manager(specprod=None, exp_table_path=None, proc_table_path
             prow = create_and_submit(prow, dry_run=dry_run_level, queue=queue,
                                      strictly_successful=False, check_for_outputs=check_for_outputs,
                                      resubmit_partial_complete=resubmit_partial_complete)
+
+            ## If processed a dark, assign that to the dark job
+            if curtype == 'dark':
+                prow['CALIBRATOR'] = 1
+                darkjob = prow.copy()
+
+            ## Add the processing row to the processing table
             ptable.add_row(prow)
 
             ## Note: Assumption here on number of flats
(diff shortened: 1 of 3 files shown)


sbailey:
Looks good, thanks.  Merge when ready.  I will trust your tests and the upcoming run to catch any remaining issues if any.

PR #1468, opened 2021-11-02
Improve desi_run_night handling of prior processing table in the pipeline

akremin:
This fixes a bug in which joint-fit jobs were not given strictly_successful=True, which led to them being run even with failed inputs. This is what we want for the daily processing but not for reprocessing where we have already dealt with bad exposures and everything being processed should be successful. This was actually introduced in the late spring by another of my previous pull requests. The original version of the script had the proper "True" value set.

The remainder of the PR deals with existing processing tables. If a processing table is found and told to append to the table using `--append-to-proc-table`, the script now skips over exposures in the exposure table corresponding to exposures already present in the exposure table. 

Since this loads previous exposures, it checks to see if the previous jobs were successful. If there are failed jobs or jobs still in the queue, this will print an error message and exit. If outside manipulations have resolved the failed jobs in the processing table, then I have added a new flag `--ignore-proc-table-failures` that can be used to move forward (at the user's own risk).

The code also check if all exposures for the requested obstypes are accounted for in the processing table and exits with a message informing the user of that.

Finally, I include a new `--dry-run-level=3`, which can be used to run `desi_run_night` or `desi_daily_proc_manager` and have it print out all the messaging but not save any files to disk. Previously even the highest level of `dry-run-level` (which was 2) still wrote out the exposure and processing tables.

### DIFF
--- a/bin/desi_run_night
+++ b/bin/desi_run_night
@@ -53,6 +53,10 @@ def parse_args():  # options=None):
     parser.add_argument("--append-to-proc-table", action="store_true",
                         help="Give this flag if you want to submit jobs even if proc table exists."+
                         " Note this will skip existing exposures present in proc table.")
+    parser.add_argument("--ignore-proc-table-failures", action="store_true",
+                        help="Give this flag if you want to submit jobs even if "+
+                        " there are incomplete jobs in the existing proc table."+
+                        " Only to be used after vetting proc table failures.")
     parser.add_argument("--dont-check-job-outputs", action="store_true",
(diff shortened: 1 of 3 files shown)


akremin:
This was tested on several nights in daily using `--dry-run-level=3`, including nights where all exposures were present in the processing table and nights where some of the jobs in the processing table had failed. I also tested on nights with no data for both `desi_run_night` and `desi_daily_proc_manager` to verify that the new `--dry-run-level` worked as expected.

Issue #1470, opened 2021-11-02
desi_compute_sky singular matrix


sbailey:
With the latest daily rerun, we're getting a hundreds of thousands of "Singular matrix" warnings from desi_compute_sky, when previously they were rare (0-2 per frame).

Example:
```
desi_compute_sky -i /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20211016/00104546/frame-r8-00104546.fits --fiberflat /global/cfs/cdirs/desi/spectro/redux/daily/calibnight/20211016/fiberflatnight-r8-20211016.fits -o testsky.fits --adjust-wavelength --adjust-lsf --pca-corr /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm2/skycorr-pca-sm2-r8.fits
```

In some cases this was fatal (e.g. r4 and r8 didn't get sky modesl for that exposure), while for others like 20211016/00104561 it generated >100k warning messages but still wrote output, vs. the previous daily run which had 0 Singular matrix warnings.  (previous = when that daily was first run, not the one that we ran over the weekend and then deleted).

Note that the original inputs are available under daily/attic/rerunarchive-20211029/exposures... to test if the problem is with some upstream issue like masking, or something about the latest sky model code.

julienguy:
This is due to the fact an amplifier is masked such that the inverse variance = 0  for many sky lines causing the fit of the line center and width to fail.
This is not by itself a serious issue (except we can improve the selection of valid fibers upstream and provide a more useful warning message) but it exposes a more serious problem: the FIBERSTATUS bit fibermask.BADREADNOISE is not set for the fiber traces on the CCD quadrant with ccdmask.BADREADNOISE. We have to change this.

sbailey:
Fixed in PR #1472; closing ticket.

PR #1472, opened 2021-11-03
Set fibermap.FIBERSTATUS BADREADNOISE

julienguy:
In this PR the ccdmask.BADREADNOISE is propagated to the fibermap FIBERSTATUS during preprocessing
(it was only set in QA so far). This PR addresses issue #1470.

### DIFF
--- a/py/desispec/io/raw.py
+++ b/py/desispec/io/raw.py
@@ -242,9 +242,11 @@ def read_raw(filename, camera, fibermapfile=None, **kwargs):
 
         for amp in amp_ids :
             kk  = desispec.preproc.parse_sec_keyword(header['CCDSEC'+amp])
-            if np.all(img.mask[kk] & maskbits.ccdmask.BADREADNOISE > 0) :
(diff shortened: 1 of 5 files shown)


julienguy:
After few additional changes (set fibermask badamp bit along with BADREADNOISE bit), and do not set ivar=0 in extraction for mask=ccdmask.BADREADNOISE), 
this PR has been tested successfully with a pipeline run of exposure 20211016/00104546 , only for r8, down to sky subtraction.
The PR is ready for review.

sbailey:
I spot checked exposures 20211016/00104546 and 00104548 which previously had missing sky outputs and hundreds of thousands of Singular Matrix warnings; these are fixed by this PR.

Small update after chatting with Julien: I changed the threshold for flagging a fiber if half the amp is BADREADNOISE, not requiring every pixel of the amp to be flagged (currently they are all flagged, but I could imaging some future where dead columns aren't BADREADNOISE masked and would want that to accidentally turn off this fiber flagging).

OK to merge after tests finish.

PR #1473, opened 2021-11-03
don't use PSFs with bad amps

### LINKED ISSUE #1471: bad PSF fit/merging on bad r8A 20211026/00106162
Extractions of multiple flats on 20211026 failed:
```
srun -n 20 -c 2 desi_extract_spectra -w 5760.0,7620.0,0.8 -i /global/cfs/cdirs/desi/spectro/redux/daily/preproc/20211026/00106177/preproc-r8-00106177.fits -p /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20211026/00106177/psf-r8-00106177.fits -o testframe.fits --psferr 0.1 --mpi
...
ERROR:extract.py:684:main_mpi: extract:  FAILED bundle 0, spectrum range 0:25
ERROR:extract.py:687:main_mpi: Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/scripts/extract.py", line 674, in main_mpi
    mark_extraction = _extract_and_save(img, psf, bspecmin[b], bnspec[b], specmin,
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/scripts/extract.py", line 744, in _extract_and_save
    results = ex2d(img.pix, img.ivar*(img.mask==0), psf, bspecmin,
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/specter/master/py/specter/extract/ex2d.py", line 190, in ex2d
    ex2d_patch(subimg, subivar, psf,
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/specter/master/py/specter/extract/ex2d.py", line 346, in ex2d_patch
    A = psf.projection_matrix(specrange, wavelengths, xyrange, use_cache=use_cache)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/specter/master/py/specter/psf/psf.py", line 667, in projection_matrix
    A = np.zeros( (nspec*nflux, ny*nx) )
ValueError: negative dimensions are not allowed
```

Negative dimensions are of course bad, but I haven't traced back to whether the input PSF is bad, or whether something about the preproc masking is confusing extractions.

Additionally it looks like the failed extractions aren't handled cleanly by desi_proc resulting in the jobs timing out rather than exiting with errors.

sbailey:
This PR fixes #1471 by updating desi_proc to not use PSFs that have a masked bad amp, e.g. recent r8A.

It still generates the PSF file for post-facto debugging, but if the input CCD has a flagged amp it renames the psf to (origname).badreadnoise and skips the bad fiber interpolation step.  Because this doesn't result in a "fit-psf*.fits" file, it won't get picked up by the merging step.

Implementation detail: although currently we flag entire amps with the ccdmask.BADREADNOISE bit, I implemented this as a somewhat more generic >12.5% pixels (half of one amp) have BADREADNOISE set, just in case some future update changes how BADREADNOISE is set (I realize that readnoise physically comes from individual amps, but sometimes it "turns on" or off during readout and we also have effective readnoise coming from the bias etc...)

Also coming along for the ride was a fix to `proc_joint_fit.main` which previously was doing an over-enthusiastic string replacement that would break any $SPECPROD containing "psf".

I tested this in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/badpsf with night 20211026 which ended up rejecting all r8 PSFs because all 5 input arcs had r8 OSTEPA > 5.  Jobs are in the queue to test 20211025 (which should reject one arc but accept the other four and proceed with a nightly PSF) and 20211024 (which should accept all five arcs).

### DIFF
--- a/py/desispec/scripts/proc.py
+++ b/py/desispec/scripts/proc.py
@@ -45,6 +45,7 @@ from desispec.util import runcmd
 import desispec.scripts.extract
 import desispec.scripts.specex
 import desispec.scripts.nightly_bias
+from desispec.maskbits import ccdmask
 
 from desitarget.targetmask import desi_mask
 
@@ -472,6 +473,23 @@ def main(args=None, comm=None):
         # loop on all cameras and interpolate bad fibers
         for camera in args.cameras[rank::size]:
             t0 = time.time()
+
+            psfname = findfile('psf', args.night, args.expid, camera)
+            inpsf = replace_prefix(psfname,"psf","fit-psf")
+
+            #- check if a noisy amp might have corrupted this PSF;
+            #- if so, rename to *.badreadnoise
+            #- currently the data is flagged per amp, but do more generic
+            #- test for 5% of pixels
+            log.info(f'Rank {rank} checking for noisy input CCD amps')
+            preprocfile = findfile('preproc', args.night, args.expid, camera)
+            mask = fitsio.read(preprocfile, 'MASK')
+            noisyfrac = np.sum((mask & ccdmask.BADREADNOISE) != 0) / mask.size
+            if noisyfrac > 0.05:
+                log.error(f"{100*noisyfrac:.0f}% of {camera} input pixels have bad readnoise; don't use this PSF")
+                os.rename(inpsf, inpsf+'.badreadnoise')
+                continue
+
             log.info(f'Rank {rank} interpolating {camera} PSF over bad fibers')
 
             # fibers to ignore for the PSF fit
@@ -482,9 +500,7 @@ def main(args=None, comm=None):
                 for fiber in fibers_to_ignore[1:] :
                     fibers_to_ignore_str+=",{}".format(fiber)
 
-                tmpname = findfile('psf', args.night, args.expid, camera)
-                inpsf = replace_prefix(tmpname,"psf","fit-psf")
-                outpsf = replace_prefix(tmpname,"psf","fit-psf-fixed-listed")
+                outpsf = replace_prefix(psfname,"psf","fit-psf-fixed-listed")
                 if os.path.isfile(inpsf) and not os.path.isfile(outpsf):
                     cmd = 'desi_interpolate_fiber_psf'
                     cmd += ' --infile {}'.format(inpsf)
--- a/py/desispec/scripts/proc.py
+++ b/py/desispec/scripts/proc.py
@@ -477,15 +477,15 @@ def main(args=None, comm=None):
             psfname = findfile('psf', args.night, args.expid, camera)
             inpsf = replace_prefix(psfname,"psf","fit-psf")
 
-            #- check if a noisy amp might have corrupted this PSF;
+            #- Check if a noisy amp might have corrupted this PSF;
             #- if so, rename to *.badreadnoise
-            #- currently the data is flagged per amp, but do more generic
-            #- test for 5% of pixels
+            #- Currently the data is flagged per amp (25% of pixels), but do
+            #- more generic test for 12.5% of pixels (half of one amp)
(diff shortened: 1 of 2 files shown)


sbailey:
Tests finished on 20211024, 1025, and 1026 in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/badpsf with results as expected for r8:
* 20211024: no arcs rejected, psfnight-r8-20211024.fits generated
* 20211025: 1/5 arcs rejected, psfnight-r8-20211025.fits generated with remaining 4
* 20211026: 5/5 arcs rejected, psfnight-r8-20211026.fits *not* generated (but other cameras have psfnight)

Minor change after chatting with Julien: changed threshold to rejecting a CCD if half of one amp (0.5 * 0.25 = 12.5%) of pixels are flagged BADREADNOISE, instead of 5%.  Makes no change for current code where we flag entire amps as bad, but provides some future proofing if that changes.

Ready for review.

akremin:
Also noting this throws out an entire camera due to one bad amplifier. Previously @julienguy had been advocating for just throwing out the problematic fibers, or not throwing them out at all for the cals since the counts are substantially higher than these jumps?

The logic used in assemble fibermap for a badamp is here, if useful:
https://github.com/desihub/desispec/blob/e0068dd8603444ab21c994d29ab15b8d35cbf8a2/py/desispec/io/fibermap.py#L778

akremin:
I'm confused how night 20211025 produced a psfnight given that the code currently requires more than 4 psfs for a given camera to run: https://github.com/desihub/desispec/blob/e0068dd8603444ab21c994d29ab15b8d35cbf8a2/py/desispec/scripts/proc_joint_fit.py#L185

I think this can be changed in the code to run for e.g. psfs >3. The comment is more in surprise to the result of the test

sbailey:
psf vs. fit-psf vs. findfile('fitpsf', ...) -- this is a bit of a mess in the code pre-dating this PR.  Standardizing everything via findfile would be better, but I'm a little hesitant to do that just for fitpsf without also diving into "fit-psf-before-listed-fix" and "fit-psf-fixed-listed" and cleaning up writing files by one name and then renaming in certain cases and copying in others...  Thinking... e.g. in particular I don't want to end up in a situation where one line is using `findfile('fitpsf', ...)` while another line is assuming that returns a `fit-psf` prefix for the purposes of some other replacement to `fit-psf-fixed-listed`...

Regarding `if len(psfs) > 4: ...` -- that code in proc_joint_fit is counting how many PSFs could exist based upon expids without actually checking if they do exists, leaving it to merge_psf to decide what to do with missing files.  I agree with having some minimum, but will come chat about whether proc_joint_fit should be enforcing that or whether merge_psf should be enforcing that...

sbailey:
I updated desi_proc_joint_fit to be the gatekeeper of how many PSFs it wants before passing to merge PSF.  Output now is:
```
...
TIMER-START:proc_joint_fit.py:177:main: Starting psfnight at 2021-11-03T16:36:30.477441
WARNING:proc_joint_fit.py:189:main: Missing /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/badpsf/exposures/20211025/00105975/fit-psf-r8-00105975.fits
INFO:proc_joint_fit.py:191:main: Number of PSF for night=20211025 camera=r8 = 4/5
INFO:proc_joint_fit.py:194:main: Computing r8 psfnight ...
INFO:specex.py:434:mean_psf: Will compute the average of 4 PSFs
...
```

I switched proc_joint_fit to use `findfile('fitpsf', ...)` but postponed any further cleanup of the chain of flavors of psf fit files in desi_proc itself.

PR #1478, opened 2021-11-09
desi_archive_tilenight

sbailey:
This PR adds the infrastructure for archiving tiles/cumulative/TILEID/LASTNIGHT -> tiles/archive/TILEID/ARCHIVEDATE after desi_tile_vi has approved tiles as done.

For tiles that have already been released:
/global/cfs/cdirs/desi/users/sjbailey/svn/surveyops/ops/tiles-specstatus.ecsv is updated to include a new ARCHIVEDATE column, based upon surveyops/trunk/mtl/mtl-done-tiles.ecsv TIMESTAMP.

`desi_freeze_tilenight` is updated and renamed to `desi_archive_tilenight` so that it moves tiles/cumulative/TILEID/LASTNIGHT -> tiles/archive/TILEID/ARCHIVEDATE, removes write permission, and leave a link behind.  The new frozen tiles/archive/TILEID/ARCHIVEDATE is the location that future MTL code will use.

`desi_update_tiles_specstatus` is also updated to set ARCHIVEDATE=0 for any new tiles added to tiles-specstatus.ecsv.

Example production before archiving:  /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/tilearchive1:
```
tiles/cumulative/11881/20211016
tiles/cumulative/22252/20211016
tiles/cumulative/2361/20211016
tiles/cumulative/24133/20211015
tiles/cumulative/25502/20211016
tiles/cumulative/3135/20211015
```
Tiles 22252 and 3135 have not yet been QA released (ZDONE='true'); the others have ARCHIVENIGHT=20211105 in /global/cfs/cdirs/desi/users/sjbailey/svn/surveyops/ops/tiles-specstatus.ecsv .

Example production after running `desi_freeze_tilenight`: /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/tilearchive2
```
tiles/cumulative/11881/20211016
tiles/cumulative/22252/20211016
tiles/cumulative/2361/20211016
tiles/cumulative/24133/20211015
tiles/cumulative/25502/20211016
tiles/cumulative/3135/20211015
tiles/archive/11881/20211105
tiles/archive/2361/20211105
tiles/archive/24133/20211105
tiles/archive/25502/20211105
```

Note that the unreleased tiles 22252 and 3135 have not been archived, the archived directories have write permission removed, and that there are relative symlinks from the original tiles/cumulative locations to the tiles/archive locations.

@akremin can you double check all of this?  Once confirmed and merged, I think we can
1. commit tiles-specstatus.ecsv
2. run desi_archive_tilenight to archive all currently released tiles in the daily prod

### DIFF
--- a/bin/desi_archive_tilenight
+++ b/bin/desi_archive_tilenight
@@ -48,7 +48,10 @@ def archivetile(tiledir, archivedir, dryrun=False):
         os.symlink(src, dst)
 
         #- Remove write access
-        freezedir(archivedir, dryrun=dryrun)
+        err = freezedir(archivedir, dryrun=dryrun)
+        if err != 0:
+            log.error(f'problem removing write access from {archivedir}')
+
 
 def freezedir(path, dryrun=False):
     """
@@ -115,9 +118,10 @@ else:
 
 if args.specstatus is None:
     args.specstatus = os.path.expandvars('$DESI_ROOT/survey/ops/surveyops/trunk/ops/tiles-specstatus.ecsv')
-    if not os.path.exists(args.specstatus):
-        log.critical(f'Missing {args.specstatus}')
-        sys.exit(1)
+
+if not os.path.exists(args.specstatus):
+    log.critical(f'Missing {args.specstatus}')
+    sys.exit(1)
 
 log.info(f'Reading tiles from {args.specstatus}')
 tiles = Table.read(args.specstatus)
@@ -167,5 +171,7 @@ for tileid, lastnight, archivedate in tiles['TILEID','LASTNIGHT','ARCHIVEDATE']:
                 tmpfile = findfile(prefix, expnight, expid, 'b0',
                                    specprod_dir=reduxdir)
                 path = os.path.dirname(tmpfile)
-                freezedir(path, args.dry_run)
+                err = freezedir(path, args.dry_run)
+                if err != 0:
+                    log.error(f'problem removing write access from {path}')
 
--- a/bin/desi_archive_tilenight
+++ b/bin/desi_archive_tilenight
@@ -101,7 +101,7 @@ parser.add_argument('--specstatus', type=str,
 parser.add_argument('--dry-run', action='store_true',
         help='print what directories to archive, without archiving them')
 parser.add_argument('--only-tiles', action='store_true',
-        help='archive only tiles/cumulative dirs; not exposures, preproc')
+        help="archive only tiles dirs; don't freeze exposures, preproc")
 
 args = parser.parse_args()
 log = get_logger()
--- a/bin/desi_archive_tilenight
+++ b/bin/desi_archive_tilenight
@@ -1,20 +1,54 @@
 #!/usr/bin/env python
 
 """
-Utility script to remove write access (i.e. freeze)
-tiles/cumulative/TILEID/NIGHT directories for which spectro operations have
-signed off as done and ready to release overlapping tiles.  If directory
-already has write permission removed, it is skipped.
+Utility script to remove archive tiles for the merged target list (MTL).
 """
 
 import os, sys, glob
 import argparse
 import subprocess
+import shutil
 
 from astropy.table import Table
 
 from desiutil.log import get_logger
 from desispec.io import specprod_root, findfile
+
+def archivetile(tiledir, archivedir, dryrun=False):
+    """
+    Archive tiledir to archivedir, leaving symlink behind
+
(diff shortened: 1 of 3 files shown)


bin/desi_archive_tilenight
```diff
@@ -77,21 +113,27 @@ elif args.prod.count("/") == 0:
...
-        log.info(f'Filtering to just LASTNIGHT={args.night}')
-        keep &= tiles['LASTNIGHT'] == args.night
-
-    tiles = tiles[keep]
-else:
-    tiles = Table(names=('TILEID', 'LASTNIGHT'), dtype=(int, int))
-    tiles.add_row( (args.tileid, args.night) )
+if args.specstatus is None:
+    args.specstatus = os.path.expandvars('$DESI_ROOT/survey/ops/surveyops/trunk/ops/tiles-specstatus.ecsv')
+    if not os.path.exists(args.specstatus):
```
akremin, line 118: I don't think this file existence check should be under the outer "`if __ is None`". As it stands it wouldn't check if a user provided file exists. 
In either case `Table.read` will throw an error anyway, but I think that defeats the purpose of this snippet.

bin/desi_tile_vi
```diff
@@ -136,6 +137,8 @@ def main():
 
     args=parse()
 
+    archivedate = datetime.datetime.now().strftime('%Y%m%d')
```
akremin, line 140: Do we need to concern ourselves with timezones or will these always be used solely at NERSC in pacific times?

akremin (approved):
I don't see any blocking factors. As expected this looks clean and correct. I point out some things that could be considered/changed but I will approve the PR as none of the comments impact the core functionality.

I have not set up a mock dataset and run this code. I will trust the examples you performed unless you think independent tests are warranted/useful.

bin/desi_archive_tilenight
```diff
@@ -1,21 +1,55 @@
...
+
+    Args:
+        tiledir: full path to tiles/cumulative/TILEID/LASTNIGHT
+        archivedir: full path to tiles/archive/TILEID/ARCHIVEDATE
+
+    Options:
+        dryrun: if True, print messages but don't move directories.
+    """
+    log = get_logger()
+    if os.path.isdir(archivedir):
```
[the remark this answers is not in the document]
sbailey: Exactly, I don't want it to trip on previously archived tiles when running over N>>1 nights without wanting to have to specify which ones have already been archived or not.  I could see an argument for even more quietly skipping over previously archived tiles, and/or adding an option along the lines of "I don't expect that any tiles from this night are archived yet, so raise an error if you find any...".

bin/desi_archive_tilenight
```diff
@@ -1,21 +1,55 @@
...
+        os.makedirs(outdir, exist_ok=True)
+        shutil.move(tiledir, archivedir)
+
+        #- Create relative link from original tiledir -> new archivedir
+        src = os.path.relpath(archivedir, os.path.dirname(tiledir))
+        dst = tiledir
+        os.symlink(src, dst)
+
+        #- Remove write access
+        freezedir(archivedir, dryrun=dryrun)
```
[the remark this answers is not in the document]
sbailey: Yeah, that return code came from the previous version of desi_freeze_tilenight, but also wasn't used there.  I think I was laying the groundwork for parallelizing and counting failures from various ranks / processes.  Maybe I'll add a log if nonzero or something.

bin/desi_tile_vi
sbailey: Good question, but I don't think we need to worry about timezones.  I think the place where it would matter would be if collaborators in China or Australia mirrored the data and then did tile_vi from there for daily operations, which seems unlikely.  If someone disagrees, I invite them to contribute the timezone robust code.

bin/desi_archive_tilenight
sbailey: Good point, I'll fix that.

sbailey:
post-facto for the record, the script I used to update tiles-specstatus.ecsv is at /global/homes/s/sjbailey/desi/dev/tilearchive/set-archivedate.py

akremin:
@sbailey Noting that your implementation of using the returned error from `freezedir` will raise an error if the path isn't a directory, even if running in `dry_run` mode. 

I saw that potentiality in my review but since you weren't using the `err`, I didn't mention it.

As with the other comments, it isn't a major problem

PR #1481, opened 2021-11-11
use desitarget.skybricks to check stuck sky locations

sbailey:
This PR uses `desitarget.skybricks` to check OBJTYPE=SKY fibers on stuck positioners for whether they are still on valid sky locations, instead of just setting FIBERSTATUS BADPOSITION if they are "off target".  This addresses a problem first noticed by @araichoor where we were discarding many stuck SKY fibers due to a disagreement between fiberassign and platemaker about where they were expected to be (the underlying cause is still under investigation).

Example outputs for 20211106 are in /global/cfs/cdirs/desi/users/sjbailey/dev/stucksky, with a comparison of how many total SKY are retained:

image

Using `desitarget.skybricks` requires `$SKYBRICKS_DIR=/global/cfs/cdirs/desi/target/skybricks/v3/` [52 GB], which is an unfortunately large new input dependency, but it only adds minimal runtime to `assemble_fibermap`.  Off-footprint work for skybricks might make a v4 that is even bigger.

I chose to *not* try to promote a non-SKY stuck fiber back to SKY in case it was not set to OBJTYPE=SKY for a reason other than positioning (poor throughput fiber, neighboring standard star on CCD, ...)

Note: PR #1459 is another open `assemble_fibermap` PR that might have minor conflicts after this is merged (in particular looking up the tile RA,DEC from the header).

Example usage:
```
assemble_fibermap -n 20211106 -e 107673 -o fibermap-00107673.fits
```

@julienguy @akremin

### DIFF
--- a/py/desispec/io/fibermap.py
+++ b/py/desispec/io/fibermap.py
@@ -19,8 +19,10 @@ from astropy.table import Table, Column, join
 from astropy.io import fits
 
 from desitarget.targetmask import desi_mask
+from desitarget.skybricks import Skybricks
 from desiutil.log import get_logger
 from desiutil.depend import add_dependencies
+from desimodel.focalplane import get_tile_radius_deg
 
 from desispec.io.util import fitsheader, write_bintable, makepath, addkeys, parse_badamps
 from desispec.io.meta import rawdata_root, findfile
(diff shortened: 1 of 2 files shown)


araichoor:
Just a side comment about off-footprint, if relevant: note the desitarget PR https://github.com/desihub/desitarget/pull/771, designed to support the Gaia-based, healpix-based, good sky locations for the backup program.
That PR currently introduces `desitarget.skyhealpixs` and another environment variable `$SKYHEALPIXS_DIR`.

sbailey:
GitHub unit tests are failing because it now needs desitarget to be installed, but NERSC disks are hanging again so I'm going to call it a night and address tests in the morning.

sbailey:
Tests now pass except coverage, because assemble_fibermap has complex/big input dependencies (even before this PR) that aren't viable to download to github actions everytime, so it isn't covered here (but it was tested at NERSC; see original post)

julienguy (approved):
It's a good way to solve the issue if we think that the PM sky coordinates of the stuck positioners are correct (which would mean fiberassign/desimeter has the wrong conversion from X,Y to RA,Dec here but this is still not totally clear ...).

PR #1488, opened 2021-11-18
move the definition of good so it's defined

Waelthus:
This fixes #1486 which did not define the variable for good sky fibers in cases where none are bad.


julienguy (approved):
thanks

PR #1489, opened 2021-11-18
Fixes to TSNR calculation and exposure QA for  2 amp readout mode

julienguy:
Fixes to TSNR calculation and exposure QA for  2 amp readout mode.
Tested successfully on exposure 20211116/00109113.


akremin (approved):
I see no show stoppers. If it worked in your test I will merge and test on rerunning the night of Nov 16 2021. We iterate from there if necessary, but I'm not too concerned.

PR #1493, opened 2021-11-24
tile VI add QA status filter

sbailey:
This PR adds a `desi_tile_vi --qastatus ...` option, e.g. to scan only QA=unsure tiles:
```
desi_tile_vi --qastatus unsure
```


akremin (approved):
You could print an informative error if both `--new` is given and `--qastatus!='none'`, but people can likely also figure it out when they see that no tiles are kept in the selection and it exits.
Otherwise it's a simply addition and useful functionality, so I'll approve it.

PR #1495, opened 2021-11-25
fix tsnr_afterburner "unknown" entries

### LINKED ISSUE #1408: fix "unknown" entries in tiles & exposures files
Fix cases of "unknown" entries in tiles & exposures files and back propagate those into daily/tiles.csv and svn surveyops/ops/tiles-specstatus.ecsv so that there aren't unknown entries.

e.g. tiles-everest.csv includes 10 SURVEY=unknown, 193 OBSTATUS=unknown, and 218 GOALTYPE=unknown.  Resort to hardcoding special cases if we have to so that we don't have a data release with a tiles file claiming that we don't know the survey, goal, or status for some tiles... :)

daily/tiles.csv currently has many more than that because various special cases have been identified and fixed since the time that a particular tile was added to the file.

sbailey:
This PR fixes a number of corner cases in desi_tsnr_afterburner that were resulting in SURVEY='unknown' and other oddities:

* All FAFLAVOR/FAPRGRM dith* tiles now have SURVEY=cmx GOALTYPE=other instead of SURVEY=GOALTYPE=unknown (which previously occurred for some but not all dither tiles)
* Tiles where all exposures are rejected as bad and have EFFTIME_SPEC=0.0 are listed as OBSSTATUS=other instead of OBSSTATUS=unknown.
  * @schlafly I could make this OBSSTATUS=obsstart if that would make life easier to afternoon planning, but currently they only get OBSSTATUS=obsstart if EFFTIME_SPEC>0.0
* Adding bad/missing exposures is now opt-in (`--add-badexp`) with the intention that we would use this for daily processing so that afternoon planning can know about exposures that we have rejected, but we would *not* use it for production runs so that we don't list exposures/tiles that aren't actually included in the production.
* Related: exposures that are listed as LASTSTEP=all but are missing from the production are *not* added, since this reflects an ambiguous processing state that should be fixed by either redoing the processing or flagging them as bad.  Previously these were entering as EFFTIME_SPEC=0.0, which could trigger additional observations even if the underlying problem was a job timeout that just needed to be cleaned up.
* Fixed a bug that resulted in GOALTYPE=other/unknown when the first exposure of a tile was rejected even if there were other good exposures.

Example outputs for everest are in /global/cfs/cdirs/desi/users/sjbailey/dev/tiles ; there are no more "unknown" entries.

Fixes #1408 for Fuji.

### DIFF
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -56,11 +56,11 @@ def parse(options=None):
                         help = 'Path to input reduction, e.g. /global/cfs/cdirs/desi/spectro/redux/blanc/,  ' +
                                'or simply prod version, like blanc, but requires env. variable DESI_SPECTRO_REDUX. ' +
                                'Default is $DESI_SPECTRO_REDUX/$SPECPROD.')
-    parser.add_argument('--cameras', type = str, default = None, required=False,
+    parser.add_argument('-c', '--cameras', type = str, default = None, required=False,
                         help = 'Cameras to reduce (comma separated)')
-    parser.add_argument('--expids', type = str, default = None, required=False,
+    parser.add_argument('-e', '--expids', type = str, default = None, required=False,
                         help = 'Comma separated list of exp ids to process')
-    parser.add_argument('--nights', type = str, default = None, required=False,
+    parser.add_argument('-n', '--nights', type = str, default = None, required=False,
                         help = 'Comma, or colon separated list of nights to process. ex: 20210501,20210502 or 20210501:20210531')
     parser.add_argument('--recompute', action='store_true',
                         help = 'Recompute TSNR values')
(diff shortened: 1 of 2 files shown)


schlafly:
Sounds good to me.  The only status AP cares about in the tiles-specstatus file in obsend; it declares any tile with status = obsend to not need more observations, even if somehow its own tracking of EFFTIME says that the exposure should be observed again.  By contrast, status = obsstart doesn't do anything, even if AP thinks it's done; this needs to instead be communicated through changes to the EFFTIMEs of contributing exposures, and we need to watch out for it.

So I think you should pick whatever status you want.  But I think you could just do obsstart for all of your "other" cases, and you would have consistency with AP.  To AP, a tile is obsstart if there's ever been a spectroscopic exposure on that configuration, and I doubt the spectroscopic pipeline sees any EXPID for which there was no spectroscopic data?

I agree that exposures known by the spectroscopic pipeline to be bad should get added to the daily product, and exposures that fail for some reason should not get added; the described logic sounds correct and complete.

julienguy:
I will merge this PR as is. I looked at the code and it's fine with me. I tested the code runs. What I have not done is rerun the whole exposure list.

akremin:
Sorry this fell off my radar. I agree the code looks good (though I also didn't test it). The feature of not including potential processing failures is a desirable addition.

Issue #1500, opened 2021-11-30
daily cals dependencies afterany -> afterok


sbailey:
Previously I had advocated that daily processing should fall back to default cals (psf, fiberflat) if the nightly cals failed so the pipeline would proceed and we would have *something* in the morning for the science exposures.  In practice, that is never what we want and proceeding just creates more cleanup to do.  I hearby Officially Change My Mind:

change the nightly processing dependency chain to match the reprocessing dependency chain to use "afterok" instead of "afterany", so that if a step fails the other steps don't run without specific action.

akremin:
Resolved with PR #1502

PR #1502, opened 2021-11-30
afterok for all jobs in daily processing

akremin:
This addresses issue #1500 by requiring all dependencies to be "afterok". The consequence of this is that a failure in a calibration exposure (or joint calibration job) will cause all science exposures to be cancelled. This is still the best approach, however, because we want to either flag, rerun, or fix the issue in the cals so that we can use them for the sciences.

The change is straight forward and now mirrors the re-processing script that has used this logic successfully in the past.


sbailey:
Agreed and looks good, thanks.

PR #1503, opened 2021-11-30
Night_qa page

araichoor:
This PR adds scripts making a Night QA html page, which lists some tasks to do to validate as much as possible data from a given night.

I hope that is enough to start with, or easy to improve, as I do not plan to work significantly more on this this week.
I open the PR, so that someone else could take over, if need be, for this week.

The idea is that this should be run by the daily pipeline, after the tile-qa*png have been created; then one inspector would check the various items displayed in the page.

A new folder will be created: `$DESI_ROOT/spectro/daily/nightqa`.
Then, for each night, a `$DESI_ROOT/spectro/daily/nightqa/NIGHT` folder will be created, with containing the following files: `nightqa-NIGHT.html`, `{dark,badcol,ctedet,sframesky,tileqa}.mp4`
Of course that can be modified.

Example of a calling sequence:
```
desi_night_qa -p daily -n 20211126 -o $DESI_ROOT/users/raichoor/nightqa_dev/20211126 --css $DESI_ROOT/users/raichoor/nightqa_dev/nightqa.css
```

Examples of webpages:
https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/20211125/nightqa-20211125.html (no dark, bad columns here)
https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/20211126/nightqa-20211126.html (sframe have been recently rerun)
https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/20211127/nightqa-20211127.html
https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/20211128/nightqa-20211128.html

Remarks:
- it requires `ffmpeg` (can be loaded through `module load ffmpeg` at NERSC);
- it requires a `nightqa.css` file, which could live in `$DESI_ROOT/spectro/daily/nightqa/`;
- instructions can surely be elaborated;
- calibnight: if files are symlink to files from a previous night, I don't know if that is good or not; I've highlighted that case.
- dark: it may be good to review the code, as I'm not familiar with those data, and I don't know if what I have from my development example makes sense (I see quite some features in those dark images);
- CTE detector: I've just put a placeholder (`nightqa.create_ctedet_mp4()`), it should be easy to plug in nightqa.py Stephen's small script
- sframesky:
  - I've restricted to Main exposures (see `nightqa.get_survey_night_expids()`);
  - the movie takes a while to generate (several minutes), as one has to read all the sframe images; could easily be significantly faster if one uses 32 processors from a node; let me know if I should implement that;
- tileqa: same, I restrict here to the Main tiles (and display all them, irrespective of EFFTIME_SPEC).
- another tab that could be added: throughput vs. time.

Credits to @jguy for the choice of the displayed items.


araichoor:
Another remark: for info, for the sframesky, I've kept the lines of code I had written to identify flagged fibers, so that one could either discard those in the image, or plot them in another color:

https://github.com/desihub/desispec/blob/4dea607c1fe687fb457c798ecfefbdee158002f8/py/desispec/night_qa.py#L350

For instance, if one would want to display those in blue, the following lines could be added just before this `ax.set_ylim()` line:
https://github.com/desihub/desispec/blob/4dea607c1fe687fb457c798ecfefbdee158002f8/py/desispec/night_qa.py#L371
```
                    ii = np.where(mydict[camera]["isflag"])[0]
                    for i in ii:
                        ax.plot([0, mydict[camera]["flux"].shape[1]], [i, i], color="b", lw=1, zorder=2)
```

Displaying those in blue is how I noticed that lots of stuck sky fibers where discarded at some point, can be useful.

bin/desi_night_qa
```diff
@@ -0,0 +1,109 @@
...
+        args.css = os.path.join(args.prod, "nightqa", "nightqa.css")
+    log.info("prod    = {}".format(args.prod))
+    log.info("night   = {}".format(args.night))
+    log.info("outdir  = {}".format(args.outdir))
+    log.info("css     = {}".format(args.css))
+
+    # AR existing output folder?
+    if not os.path.isdir(args.outdir):
+        log.info("creating {}".format(args.outdir))
+        os.mkdir(args.outdir)
```
sbailey, line 72: Let's use `os.makedirs(args.outdir, exist_ok=True)` so that it will make the higher level directory too if needed.

bin/desi_night_qa
```diff
@@ -0,0 +1,109 @@
...
+    parser.add_argument("-p", "--prod", type = str, default = None, required = False,
+                        help = "Path to input reduction, e.g. /global/cfs/cdirs/desi/spectro/redux/blanc/,  or simply prod version, like blanc, but requires env. variable DESI_SPECTRO_REDUX. Default is $DESI_SPECTRO_REDUX/$SPECPROD.")
+    parser.add_argument("-n","--night", type = int, default = None, required = True,
+                        help = "night to process. ex: 20211128")
+    parser.add_argument("-o", "--outdir", type = str, default = None, required = False,
+                        help = "Path to ouput folder, default is the input prod directory. Files written in {prod}/nightqa/{night}; several files will be created there")
+    parser.add_argument("--recompute", action = "store_true",
+                        help = "recompute (i.e. overwrite args.outfile if already existing")
+    parser.add_argument("--css", type = str, default = None, required = False,
+                        help = "html formatting css file; default to {prod}/nightqa/nightqa.css")
```
sbailey, line 40: Let's put the CSS file in `py/desispec/data/qa/nightqa.css` so that desispec remains self-contained for what is needed to run this.  Use `pkg_resources.resource_filename('desispec', 'data/qa/nightqa.css')` to find it and then copy that into the output directory.  If you want to leave an option to specify a different CSS ok.

I think we *do* want to copy it into the output directory every time so that if we make an update to the CSS it gets automatically propagated and the old CSS doesn't block the new one from getting installed into the output dir.

py/desispec/night_qa.py
```diff
@@ -0,0 +1,740 @@
...
+        "sframesky" : os.path.join(outdir, "sframesky-{}.mp4".format(night)),
+        "tileqa" : os.path.join(outdir, "tileqa-{}.mp4".format(night)),
+    }
+
+
+
+def get_survey_night_expids(
+    night,
+    survey,
+    datadir = os.getenv("DESI_SPECTRO_DATA")):
```
sbailey, line 49: It is generally considered bad python practice to have the parameter defaults call a function in the definition of the function itself, because of possible side effects of running code at the time of evaluating/parsing/loading the module.  Better would be:
```
def get_survey_night_expids(..., datadir=None):
    ...
    if datadir is None:
        datadir = os.getenv("DESI_SPECTRO_DATA") # or desispec.io.rawdata_root()
    ...
```

e.g. if some test code redefined `DESI_SPECTRO_DATA` and then called `get_survey_night_expids`, I think the current code would still use the value of `$DESI_SPECTRO_DATA` at the time the module was loaded instead of the value at the time the function was called.

py/desispec/night_qa.py
```diff
@@ -0,0 +1,740 @@
...
+        duration (optional, defaults to 15): video duration in seconds (float)
+
+    Notes:
+        Requires ffmpeg to be installed.
+        At NERSC, run in the bash command line: "module load ffmpeg".
+        The movie uses fns in the provided order.
+    """
+    # AR is ffmpeg installed
+    if os.system("which ffmpeg") != 0:
+        log.error("ffmpeg needs to be installed to run create_mp4(); exiting")
```
sbailey, line 127: Please add a message like 'at NERSC, this is available with "module load ffmpeg" '.  Please also do this check earlier in the code so that it doesn't run for several minutes and then exit due to a missing module).

bin/desi_night_qa
```diff
@@ -0,0 +1,109 @@
...
+    create_sframesky_mp4(outfns["sframesky"], args.night, args.prod, expids)
+
+    # AR tileqa
+    create_tileqa_mp4(outfns["tileqa"], args.night, args.prod, expids, tileids)
+
+    # AR create index.html
+    write_nightqa_html(outfns, args.prod, args.night, args.css)
+
+if __name__ == "__main__":
+    main()
```
sbailey, line 109: Thanks for splitting the functions into a separate library, independent of the wrapper script, and also with one function per QA item.  That is good practice.

sbailey (changes requested):
Thanks for your quick work on this.  I'm sure we'll have lots of requests for the *contents* of the QA as we begin to use this, but let's get something in place and then make iterative refinements and not let future ideas block current value.  

That being said, I put a few code structure comments inline; the most important is about having nightqa.css contained in desispec itself (and then copied into final location) instead of requiring an external file to pre-exist before running.

Please also fill out the "TBD" docstrings which will make it easier for others to contribute in the future.  Thanks.

bin/desi_night_qa
araichoor: done

bin/desi_night_qa
araichoor: done

py/desispec/night_qa.py
araichoor: done - thanks for the explanation!

py/desispec/night_qa.py
araichoor: done

bin/desi_night_qa
araichoor: thanks!

araichoor:
I've implemented @sbailey suggestions, thanks.
I've also added two features:
- now the nightlog is displayed with the nightlog tab, thanks to @weaverba137 (https://github.com/desihub/desispec/pull/1503/commits/86c6297d939dd0fb4df347db7a78935300073ea4)
- a args.html_only option, in case one wants to only regenerate the .html page, without having to recompute the .mp4 files (the sframesky one taking a while).

julienguy:
It would be useful to multiply the dark by (ivar>0)*(mask==0) so that one can check that bad columns are properly masked.

julienguy:
for the bad columns, it might be more useful to simply display the number of bad columns as a function of the spectrograph number for each camera arm.

julienguy:
same for the sframe movie. a version with flux*(ivar>0)*(mask==0) would be useful.

araichoor:
I ve changed the badcolumns to a simple plot, e.g. for 20211126:
badcol-20211126

sbailey (approved):

araichoor:
thanks Stephen.
for info I m working on it; on my to-do-list:
- generate pdf instead of mp4 movies, per Julien s request
- remove masked pixels in the dark image
- plug in your CTE small script
- add a per-night Z vs. FIBER plot for sky fibers only.

araichoor:
For info, I ve implemented what I mentioned in the previous message.
I'd be happy to change the display for the CTE plots, if some of the choices I've made are not relevant.

I've launched the current code for all post-shutdown nights, here: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev

From randomly spot-checking some of those:
- weird r8 dark here: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/20211029/nightqa-20211029.html
- z5 ctedet here: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/20211101/nightqa-20211101.html
- z3 ctedet maybe already present here: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/20211104/nightqa-20211104.html

julienguy:
with the new implementation, my browser asks me to download and display in separate windows all of the PDF files from the qa web page. I don't think this is what we want. It's probably dependent on which browser/OS one uses.

araichoor:
thanks for the feedback.
what browser do you use?
I checked with google chrome / safari / firefox, and the pdfs are displayed within the page.

araichoor:
About displaying pdf within firefox: I've quickly looked at that => maybe it s a setting to change?
(see Preferences / General / Applications)
e.g. https://support.mozilla.org/en-US/kb/view-pdf-files-firefox-or-choose-another-viewer#w_firefoxs-built-in-pdf-viewer)

tmp

araichoor:
I'm thinking: would it make sense also generate, in addition to the pdf files, individual png files, for each page of each pdf?

That would be:
- `dark`: one png file with the 3 cameras for each petal (so 10 png files per night);
- `ctedet`: one png file for each {camera}{petal} (so 30 png files per night);
- `sframesky`: one png file for each exposure (so max ~30 png file per night);
- `tileqa`: none, as those already exist in the tiles/cumulative folder.

If we are fine with having many files, that could be handy to:
- for instance share one specific case on slack / github;
- for instance compare a specific {camera}{petal} across different nights.

araichoor:
Unrelated to previous message: I've added in the bad columns plot the 10 previous nights, as a reference, so that one sees if there is significant changes for the considered night:

badcol-20211202

julienguy:
I think the QA code can be merged as is.

sbailey:
Thanks, I'll merge now and update desispec at NERSC.  @araichoor (or others!) if/when you have more updates, please open a new PR.

Issue #1499, opened 2021-11-29
Add 1 sec FLAT for testing CCD CTE QA


sbailey:
To help catch CCD CTE problems (like DESI trac issue 887) sooner than later:

1. Add a 1 sec FLAT to daily cals
2. Have spectro pipeline preprocess it but *not* use it for making fiberflats
3. Add new QA script to use those pre-processed 1-sec flats to look for CTE problems.
  * /global/homes/s/sjbailey/desi/dev/ccd/plot-amp-cte.py may be a starting point (but not along sufficient since it it focused on plotting a specific range of columns, not generally detecting the problem for all CCDs across all columns)

(1) and (2) need to be done in coordination; (3) can come later, in fact easiest after (1) and (2) are done such that there is an example exposure to work with.

akremin:
This is resolved with pull request #1507

PR #1507, opened 2021-11-30
Handle new short QA flats in the pipeline

akremin:
### Summary
This pull request solves issue #1499. It adds new logic in the handling of `OBSTYPE==flat` exposures so that short flats used for monitoring CTE issues in the cameras will not be used to generate `fiberflat`'s. 

Two small changes were made:

- `desi_daily_proc_manager` and `desi_run_night` were updated to process the short flat, but not include it in the joint fit of the `nightlyflat`. This is done by requiring the `EXPTIME > 100s` for a flat to be used for the `nightlyflat` file generation.
- `desi_proc` was updated to load the exposure time of the flat and only produce `fiberflat`'s if `EXPTIME > 10s`. 

The specific hardcoded times are somewhat arbitrary, but the idea is to only use long flats for the ubercal and to not bother creating an individual fiberflat for very short exposures.

### Testing
Note: There are no short flats in past data, so I can only test that there are no typos and that the long flats are still treated properly.

I tested both the daily script and reprocessing script. Both handled the flats correctly and submitted the joint flat script with all 12 long flats. I let two (long) flat jobs process to completion and they produced `fiberflat`'s.

Test 1: `desi_daily_proc_manager --override-night=20211129 --proc-obstypes=flat`
Test 2: `desi_run_night -n 20211129 --dry-run-level=3`

Scripts and logs: `/global/cfs/cdirs/desi/spectro/redux/kremin/run/scripts/night/20211129/`
Outputs: `/global/cfs/cdirs/desi/spectro/redux/kremin/exposures/20211129/0011141[12]/`

### DIFF
--- a/py/desispec/scripts/daily_processing.py
+++ b/py/desispec/scripts/daily_processing.py
@@ -335,7 +335,8 @@ def daily_processing_manager(specprod=None, exp_table_path=None, proc_table_path
             ptable.add_row(prow)
 
             ## Note: Assumption here on number of flats
-            if curtype == 'flat' and flatjob is None and int(erow['SEQTOT']) < 5:
+            if curtype == 'flat' and flatjob is None \
+                    and int(erow['SEQTOT']) < 5 and float(erow['EXPTIME']) > 100.:
                 flats.append(prow)
             elif curtype == 'arc' and arcjob is None:
                 arcs.append(prow)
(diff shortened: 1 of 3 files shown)


sbailey:
Looks good, thanks.  Too bad it took some MPI boilerplate futzing to implement what I initially thought would be 2 lines :(  .  We should get a 1-second flat this afternoon to test this.  Feel free to merge this first and launch, or dryrun test in branch first, whatever is most convenient.

akremin:
I ran the branch on the calibrations for 20211130, which now include the 1s flat. The dry run printouts show it doing the right thing -- submitting the 1s flat but not including it in the `nightlyflat`. I then submitted the 1s flat and the outputs match the intended -- extractions but no `fiberflat`'s.

I'm taking the above comment as approval to merge and will do so.

PR #1509, opened 2021-12-02
fix proctable entries with 0-length arrays

sbailey:
This PR fixes a parsing bug when loading processing tables that have elements with 0-length arrays.  @akremin please review carefully for any potential side effects.  This PR fixes the particular case I had (causing desi_resubmit_queue_failures to crash), but I'm not sure about what else it might touch.

Processing table /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/f2k/attic/processing_tables/processing_table_f2k-20210702.csv starts with:
```
EXPID,OBSTYPE,TILEID,NIGHT,BADAMPS,LASTSTEP,EXPFLAG,PROCCAMWORD,CALIBRATOR,INTID,OBSDESC,JOBDESC,LATEST_QID,SUBMIT_DATE,STATUS,SCRIPTNAME,INT_DEP_IDS,LATEST_DEP_QID,ALL_QIDS
96939|,dark,-99,20210702,,all,|,a0123456789,1,210702000,,dark,51159333,1638164840,PENDING,,|,|,51159333|
```
Note that some entries like INT_DEP_IDS and LATEST_DEP_QID are "|", i.e. a zero-length array.  desispec.workflow.tableio.load_table is incorrectly parsing these into a 1-element array with a blank string instead of a zero-length array which I think is the intended result.  e.g.

```python
from desispec.workflow.tableio import load_table
p = load_table('/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/f2k/attic/processing_tables/processing_table_f2k-20210702.csv')
print(p['INTID', 'INT_DEP_IDS', 'LATEST_DEP_QID', 'ALL_QIDS'][0:2])
```

Current master:
```
  INTID   INT_DEP_IDS LATEST_DEP_QID  ALL_QIDS 
--------- ----------- -------------- ----------
210702000        ['']           [''] [51159333]
210702001 [210702000]     [51159333] [51159334]
```

This PR:
```
  INTID   INT_DEP_IDS LATEST_DEP_QID  ALL_QIDS 
--------- ----------- -------------- ----------
210702000          []             [] [51159333]
210702001 [210702000]     [51159333] [51159334]
```


akremin (approved):
This is the algorithmically correct thing to do, and should allow for either str or int type values to be appended. 

I tested this using desi_edit_exposure_table to verify that it can load an exposure table with no array values and append a string to write out. I then ran it again to verify it works with non-empty arrays and that the outputs are not corrupted.

Finally, I tested by running the desi_daily_proc_manager on one night in dry-run. It was able to write the exposure table and processing table without issue. I then killed the process several times and restarted. Each time it was able to load the tables and continue onward.

What I did not test is the appending of new integer values. Given that this fix was made to allow that, I am trusting your testing of that behavior.

PR #1521, opened 2021-12-10
KNL arcs also 45min

sbailey:
This PR makes a special case for KNL arc jobs to *not* apply the normal 3x extra time factor, since the default Haswell jobs are already 45 minutes (to allow for fluctuations in realtime processing), even though a normal Haswell arc job is <10 min and a normal KNL arc job is <20 minutes.

Histogram of Haswell arc runtimes (from Nov 2021 arc jobs that finished), and a current KNL test production:

image

Motivation: the default 45min x 3 = 2h15m makes the jobs harder to schedule into the regular queue, and 45 minutes is still ~3x longer than typical.

@akremin


akremin (approved):
Straight forward fix that improves the reprocessing runs since the arcs can now be scheduled in smaller time blocks. It strictly impacts knl, so it will have no impact on the realtime queue or daily processing.

PR #1522, opened 2021-12-10
Nightqa v2

araichoor:
This PR handles in the night qa scripts the case where no (main) science exposures are taken.
The same set of files will be created, but some will be empty (sframesky, tileqa, skyzfiber, petalnz); the page is still useful to monitor the rest.

An example here: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/20211209/nightqa-20211209.html

To inform the inspector, I also now report at the top of the page the number of exposures/tiles, so that it is clearly stated if there is no science exposures.


julienguy:
It looks good. I am merging.

Issue #1515, opened 2021-12-08
OMP_NUM_THREADS impact on extractions


sbailey:
In PR #1437 we started setting OMP_NUM_THREADS in the desi_proc batch script to be the maximum number of hyperthreads available per MPI rank based upon the level of parallelism we were using, because @marcelo-alvarez observed that this significantly improved specex performance (C++ code using OpenMP parallelism, wrapped with MPI via python).

However, we may have inadvertently slowed down spectral extractions compared to the previous default OMP_NUM_THREADS=1 case.  TODO: Profile this for extractions and adjust if needed.

Below is a test script (also at /global/homes/s/sjbailey/desi/dev/parallel/linalg-speed.py) that does a 2k x 2k Cholesky decomposition and prints the min and max time across all MPI ranks.  It has the curious property that increasing OMP_NUM_THREADS *decreases* the performance, even when mpi_size * OMP_NUM_THREADS <= number_of_physical_cores.

That's when using `--cpu-bind=cores`.  Without that option, the *fastest* rank gets faster with larger OMP_NUM_THREADS as long as you are staying within physical cores, but the slowest rank is slower than the `--cpu-bind=cores` case such that the total runtime across all ranks is worse, i.e. not a win.

@dmargala thoughts?

@marcelo-alvarez heads up for pipeline testing.

Example runs on a KNL interactive batch node:
```
<cori parallel> OMP_NUM_THREADS=4 srun -n 64 -c 4 --cpu-bind=cores python linalg-speed.py
3.021 - 3.106 seconds
<cori parallel> OMP_NUM_THREADS=2 srun -n 64 -c 2 --cpu-bind=cores python linalg-speed.py
2.556 - 2.643 seconds
<cori parallel> OMP_NUM_THREADS=1 srun -n 64 -c 1 --cpu-bind=cores python linalg-speed.py
1.857 - 1.943 seconds
<cori parallel> OMP_NUM_THREADS=1 srun -n 16 -c 1 --cpu-bind=cores python linalg-speed.py
1.300 - 1.311 seconds
<cori parallel> OMP_NUM_THREADS=2 srun -n 16 -c 2 --cpu-bind=cores python linalg-speed.py
1.931 - 1.942 seconds
<cori parallel> OMP_NUM_THREADS=4 srun -n 16 -c 4 --cpu-bind=cores python linalg-speed.py
2.126 - 2.155 seconds
```

The script:
```python
#!/usr/bin/env python
  
"""
[...]

PR #1523, opened 2021-12-10
restore pipeline operations on KNL

sbailey:
This PR has a potpourri  of updates that I made while trying to restore the ability of the pipeline to run efficiently on KNL.  Most of them turned out to be unrelated, but I'm opening this PR to re-establish a working baseline for further updates.  Changes include:

* dashboard outputs to $DESI_SPECTRO_REDUX/$SPECPROD/dashboard by default instead of $HOME
* more user friendly time logging in desi_proc output:
  * instead of dumping full json into the log, log only the duration.max per step (details are still in the *.json files)
  * include the total duration as the last line in the log
* parallelize over cameras while merging the PSFs in desi_proc_joint_fit
  * The previous code was asking for 200 cores but apparently only using 1; now it asks for 30 ranks and uses all of them.
  * There has been some churn of how many ranks this step asks for, so I highlight this for attention to make sure I'm not misunderstanding the reason for earlier changes
* add `srun --cpu-bind=cores` option for steps that are pure MPI (extraction, redrock), and `--cpu-bind=none` for steps that use multiprocessing (fluxcalib, spectra regrouping).  See desihub/desiutil#180 for context
* when generating the desi_proc batch script, derive from the requested command line whether `--nightlybias` applies instead of passing a separate parameter through the function calls that redundantly tracks that info
* More specex logging, e.g. log the command that will be run, not just log the commands that fail
* Keep OMP_NUM_THREADS=(available threads per rank) for PSF fits (where we have profiled that helps), but go back to OMP_NUM_THREADS=1 for other steps.
  * This was motivated by #1515, though in the end my concern about OMP_NUM_THEADS>1 may have been a red herring.  Regardless, I've kept here was emperically works in test runs, and then we can separately test whether re-increasing OMP_NUM_THREADS helps extractions or not.

I've tested this branch for dark+nightlybias, arc, flat, science, and redshifts in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/f2k (still running, and includes some NERSC-outage inflicted failures, but basically working).

@akremin 

For the record: my biggest problem with testing turned out to be missing KMP_AFFINITY=false which is needed for using MPI+numpy on KNL.  Other pieces that may have helped are being more explicit about --cpu-bind and adding the PSF merging parallelism.

### DIFF
--- a/py/desispec/desi_proc_dashboard.py
+++ b/py/desispec/desi_proc_dashboard.py
@@ -161,7 +161,8 @@ def main(args):
 
     if args.output_dir is None:
         if 'DESI_DASHBOARD' not in os.environ.keys():
-            os.environ['DESI_DASHBOARD']=os.environ["HOME"]
+            os.environ['DESI_DASHBOARD']=os.path.join(
+                    os.environ["DESI_SPECTRO_REDUX"], os.environ["SPECPROD"], "dashboard")
         args.output_dir = os.environ["DESI_DASHBOARD"]
     else:
         os.environ['DESI_DASHBOARD'] = args.output_dir
--- a/py/desispec/workflow/desi_proc_funcs.py
+++ b/py/desispec/workflow/desi_proc_funcs.py
@@ -310,7 +310,7 @@ def determine_resources(ncameras, jobdesc, queue, nexps=1, forced_runtime=None, 
     elif jobdesc in ('ZERO'):
         ncores, runtime = 2, 5
     elif jobdesc == 'PSFNIGHT':
-        ncores, runtime = 20 * nspectro, 5 #ncameras, 5
+        ncores, runtime = 20 * nspectro, 10 #ncameras, 5
     elif jobdesc == 'NIGHTLYFLAT':
         ncores, runtime = 20 * nspectro, 20 #ncameras, 5
     elif jobdesc in ('STDSTARFIT'):
--- a/py/desispec/workflow/desi_proc_funcs.py
+++ b/py/desispec/workflow/desi_proc_funcs.py
@@ -406,7 +406,7 @@ def get_desi_proc_batch_file_pathname(night, exp, jobdesc, cameras, reduxdir=Non
     return os.path.join(path, name)
 
 
-def create_desi_proc_batch_script(night, exp, cameras, jobdesc, queue, nightlybias=False, runtime=None, batch_opts=None,\
+def create_desi_proc_batch_script(night, exp, cameras, jobdesc, queue, runtime=None, batch_opts=None,\
                                   timingfile=None, batchdir=None, jobname=None, cmdline=None, system_name=None):
     """
     Generate a SLURM batch script to be submitted to the slurm scheduler to run desi_proc.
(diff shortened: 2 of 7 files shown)


py/desispec/desi_proc_dashboard.py
```diff
@@ -161,7 +161,8 @@ def main(args):
 
     if args.output_dir is None:
         if 'DESI_DASHBOARD' not in os.environ.keys():
-            os.environ['DESI_DASHBOARD']=os.environ["HOME"]
+            os.environ['DESI_DASHBOARD']=os.path.join(
```
akremin, line 164: This makes sense in that the normal calls used to monitor any given prod use this path structure and it allows us to run the script without having to specify the path.

The old setting was chosen so that anyone could create a custom webpage for any prod with a default that wouldn't crash due to lack of write access or overwrite the existing dashboard page (if run as desi user with the proper permissions). E.g. making a page for specific nights or object types while letting a cron job continue updating the full dashboard of the prod.

This change simplifies the standard use case by the pipeline and makes our lives easier the majority of the time, while making it slightly harder/more dangerous when run by a non-power user.

Thoughts?

(To be clear: I don't feel strongly in either direction. I'm just giving the context.)

py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -560,7 +570,7 @@ def create_desi_proc_batch_script(night, exp, cameras, jobdesc, queue, nightlybi
         else:
             if jobdesc.lower() in ['science','prestdstar']:
                 fx.write('\n# Do steps through skysub at full MPI parallelism\n')
-                srun = f'srun -N {nodes} -n {ncores} -c {threads_per_core} {cmd} --nofluxcalib'
+                srun = f'srun -N {nodes} -n {ncores} -c {threads_per_core} --cpu-bind=cores {cmd} --nofluxcalib'
```
akremin, line 573: Line 567 above is another `srun` where you haven't added `--cpu-bind`. It pertains to arcs and flats. Was that intentional?

akremin (commented):
A useful set of potpourri. I left two comments, one giving full context for the dashboard change to verifying our mutual thoughts and another checking if all the `srun`'s are accounted for.

Note: I didn't do any trial runs of the code or verify that the appropriate `cpu-bind` is used in each circumstance.

py/desispec/workflow/desi_proc_funcs.py
sbailey: Full disclosure: I wasn't sure what the right answer was in that case and not specifying it was working so I left it alone.  We can revisit that one later.

py/desispec/desi_proc_dashboard.py
sbailey: Discussed with Anthony.  Having a non-desi user make a dashboard and forget to specify an alternate output is "harmless" because it can't overwrite anything.  We'll need to be a bit more careful when running custom dashboards as the desi user, but the worse case scenario is we overwrite the dashboard and have to run it again with default parameters, which is fairly fast.  I.e. we'll leave this update in place.

PR #1529, opened 2021-12-13
more robust when fiberassign in earlier expid

### LINKED ISSUE #1272: pipeline using wrong fiberassign file if correct one wasn't transferred
@akremin a new failure mode that will require some hand cleanup, and could use some robustification:

1. 20210518/00089147 for tile 20581 initially had a fiberassign file with the wrong permissions such that data transfer was able to transfer the rest of the exposure but not the fiberassign file itself.
2. the pipeline launched for that exposure and `assemble_fibermap` followed the "walk back in time looking for the most recent fibermap" logic that was necessary in early CMX data
3. the file it found was 20210518/00089056/fiberassign-080607.fits.gz, i.e. a fiberassign file from a test exposure for the completely wrong tile
4. but the pipeline ran with it anyway...

Action items:
1. DONE: fix permissions and transfer the correct fiberassign file
2. cleanup and completely rerun 20210518/00089147 for tile 20581 (fortunately the only exposure for that tile that night)
3. check if the "desi-\*.fits.fz file exists but fiberassign\*.fits.gz does not" case ever legitimately exists anymore.  That was a feature of early CMX data where the acquisition happened separately from the spectrograph exposure, but now we may be able to rely upon desi+fiberassign always existing together or otherwise exit.
4. add a cross check that whatever fiberassign file is found, it matches the TILEID of the spectrograph exposure that we are trying to process.

sbailey:
This PR fixes #1272 where the "look in previous exposures for the fiberassign file" could end up grabbing the wrong fiberassign file.  I originally was just going to turn off this look-back behavior since it isn't needed in normal ops, but there may be special test cases of setting up on a tile and then doing some manual exposures that could result in spectrograph exposures without a matching fiberassign file in the same EXPID, so the updated logic is:

* if the raw data has a TILEID keyword, insist that the fiberassign file matches that
* if looking to previous exposures without knowing the TILEID, check the data TARGTRA,TARGTDEC vs. fiberassign TILERA,TILEDEC for whatever fiberassign file is found and only accept it if they match

It also corrects a case where the coordinates file is missing fiber location information, by correctly flagging the fibers as FIBERSTATUS MISSINGPOSITION.

Example case of previous code finding and using an incorrect fiberassign file (now generates a RuntimeError because it can't find a good fiberassign file): 
```
assemble_fibermap --overwrite -o blat.fits -n 20210114 -e 72405
```

Example case of correctly walking back to find a fiberassign file, albeit with missing info in the coordinates file resulting in FIBERSTATUS flags:
```
assemble_fibermap --overwrite -o blat.fits -n 20210220 -e 77103
```

This PR also introduces a utility function `desispec.io.util.checkgzip(filename)` that will look for a filename with or without the .gz extension; this is handy for cases like fiberassign where sometimes it is gzipped or not.  We may also use this in the future with the pipeline if infrequently used files get post-facto gzipped and we want the code to easily find the right file either way.

### DIFF
--- a/py/desispec/io/fibermap.py
+++ b/py/desispec/io/fibermap.py
@@ -24,7 +24,8 @@ from desiutil.log import get_logger
 from desiutil.depend import add_dependencies, mergedep
 from desimodel.focalplane import get_tile_radius_deg
 from desimodel.io import load_focalplane
-from desispec.io.util import fitsheader, write_bintable, makepath, addkeys, parse_badamps
+from desispec.io.util import (fitsheader, write_bintable, makepath, addkeys,
+    parse_badamps, checkgzip)
 from desispec.io.meta import rawdata_root, findfile
 from . import iotime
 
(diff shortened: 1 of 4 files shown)


sbailey:
Note: the test coverage is better when run at NERSC, due to tests that require large/complex inputs and thus are not included in the github actions tests, but I have run unit tests of this branch directly at NERSC to include the assemble_fibermap tests.

PR #1532, opened 2021-12-14
Approximate fix to fibermap if FVC but no FP coordinates from correction move.

julienguy:
For the special case of exposure 80502 from 20210314, estimate DX_1, DY_1 using the FVC coordinates of both the blind and correction move along with the values of DX_0, DY_0.  Use this to estimate FIBER_X, FIBER_Y, DELTA_X, DELTA_Y.
It is an approximation because we cannot redo PM here. So the DELTA_X,DELTA_Y are a bit pessimistic. But most fibers of exposure 80502 are still within 30 microns.


julienguy:
Addresses issue #1527.

sbailey:
Wow, this turned out to be more complicated than I thought, but I appreciate the work to save half the data from that program.  There is some substantial structure between the derived fiber locations on the two exposures with petal 6 sticking out as being different from the others, but it smells like this may be as good as we'll get given the input data, so I'm fine with merging this as-is.

image

Issue #1531, opened 2021-12-14
desi_run_night --survey option


sbailey:
Add a `desi_run_night --survey` option so that we can split fuji (cmx,sv1,sv2,sv3) from guadalupe (main) during the transition period.  I'll submit a PR soon that does this, but I'm filing this ticket to document the nights of overlap between sv and main:
```
20210512     sv1 sv3 main
20210513     sv1 sv3 main
20210517         sv3 main
20210518         sv3 main
20210521         sv3 main
20210529 cmx     sv3 main
20210610         sv3 main
```
interleaved between those are nights with only SURVEY=main.

After that but still within the timeframe of fuji/guadalupe are some nights with SURVEY=special:
```
20210615 main special
20210627 main special
20210705 main special
20210706 main special
20210708 main special
20210709 main special
```
@araichoor @schlafly @djschlegel can you please identify those tiles and whether they should go with fuji (to become public with SVDA) or with guadalupe (processed at the same time as fuji but held in reserve until the DA1 release).  The default would be to keep them with guadalupe/DA1 to minimize the number of nights that we have to process in both.

schlafly:
FWIW, the survey = special tiles with 20210615 <= NIGHT <= 20210709 are bright tiles, TILEID 80978 - 80981.  According to
https://desi.lbl.gov/trac/wiki/SurveyOps/TileDesigns
these correspond to some "test" tiles for horrible weather (978, 979) and to ToO test tiles (980, 981).  The horrible weather tiles have some value for setting speed threshholds but IMO need not delay anything.  Transient group would need to weigh in on 980, 981.

sbailey:
Also: my bookkeeping was based upon the current pipeline exposure tables, falling back to FA_SURV when the SURVEY column was not present in those tables.  This brings up cases like tile 81100, which in the fiberassign file has FA_SURV='main' and SURVEY='special', which I think means it uses main survey targets, but was a special test tile that shouldn't be considered a main survey tile (but got lumped into my main survey accounting for the moment).

I hope we can define 20210610 as the last night of SV, and any tiles observed after that (main or special or even cmx dithers) will be part of guadalupe/DA1.

sbailey:
for the record, from @djschlegel on #survey-ops slack:
> Those first two nights of May 12+13 are before the start of the main survey.  There were no main survey tiles those nights.

[...]

PR #1533, opened 2021-12-15
desi_run_night --surveys option

sbailey:
This PR addresses #1531 by adding a `desi_run_night --surveys` option to filter by survey when processing a night.

However, it requires "SURVEY" to be in the exposures table, but exposure tables from the sv/main overlap time 20210517 - 20210610 don't yet have that column, so I might be getting ahead of myself.  OTOH, that column is only required if the `--surveys` option is requested, so this doesn't break current functionality.

In the input fiberassign files, by understanding in that "FA_SURV" refers to what kind of targets were used, but "SURVEY" refers to what output survey the tile should belong to, e.g. to enable test tiles with main survey targets by using FA_SURV=main but SURVEY=special.

@akremin heads up for coordination on your cleaned up exposure tables (and review of this in general).

### DIFF
--- a/bin/desi_run_night
+++ b/bin/desi_run_night
@@ -25,6 +25,9 @@ def parse_args():  # options=None):
                              "for testing as though scripts are being submitted. Default is 0 (false).")
     parser.add_argument("--tiles", type=str, required=False,
                         help="Comma separated list of TILEIDs to include; use -99 to include arcs/flats")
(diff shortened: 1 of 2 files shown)


akremin (approved):
This looks great. You properly handle the case where the table column doesn't exist, so I see no reason why we can't merge this now.

PR #1535, opened 2021-12-16
Nightqa v3

araichoor:
This PR adds few modifications to the nightqa page creation.
The principal motivation is to move away from restricting to Main tiles only.
The chosen approach now considers all OBSTYPE="SCIENCE" exposures.

For the sframesky, tileqa, skyzfiber plots, we thus now include any SCIENCE observations.
The reason is because the type of survey should not prevent to inspect the exposures quality on those plots.

For the petalnz plot, we proceed a bit differently, as we stack the n(z) per tracer, per night.
We here:
- restrict to sv1, sv2, sv3, main (which conveniently have the same bits for the 5 plotted tracers);
- we make a serie of plots for each survey of the night;
- we still generate one file only.

Along the way, several changes - hopefully improvements:
- get_surveys_night_expids(): not using NTSSURVY anymore, we now rely on the fiberassign file which should be in the same folder as the desi-EXPID.fits.fz file
- write_nightqa_html(): look for possible nightlog filenames
- desi_night_qa: add a --steps argument to allow to execute only some steps
- protect against cases where no DARK exposure is found
- replace sys.exit(1) by some raise Errors

I've run it on what is currently processed in $DESI_ROOT/spectro/redux/f3:
https://data.desi.lbl.gov/desi/users/raichoor/nightqa_v3_dev/f3
(though note that those files may change if I do further developments/tests for this PR)

From a quick random spot checking of those: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_v3_dev/f3/20210511/dark-20210511.pdf => r1 has very funny/dubious masked pixels distribution.

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -111,7 +111,6 @@ def get_surveys_night_expids(
             tileids.append(hdr["TILEID"])
             surveys.append(survey)
     expids, tileids, surveys = np.array(expids), np.array(tileids), np.array(surveys)
-    msg = "for NIGHT={} found"
     per_surv = []
     for survey in np.unique(surveys):
         sel = surveys == survey
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -1292,22 +1292,26 @@ def write_nightqa_html(outfns, night, prod, css, surveys=None, nexp=None, ntile=
         )
         html.write("<div class='content'>\n")
         html.write("\t<br>\n")
-        for text_split in text.split("\n"):
-            html.write("\t<p>{}</p>\n".format(text_split))
-        html.write("\t<tr>\n")
-        if os.path.splitext(outfns[case])[-1] == ".png":
-            outpng = path_full2web(outfns[case])
-            html.write(
-                "\t<a href='{}'><img SRC='{}' width={} height=auto></a>\n".format(
-                    outpng, outpng, width,
+        print(outfns[case], os.path.isfile(outfns[case]))
+        if os.path.isfile(outfns[case]):
+            for text_split in text.split("\n"):
+                html.write("\t<p>{}</p>\n".format(text_split))
+            html.write("\t<tr>\n")
+            if os.path.splitext(outfns[case])[-1] == ".png":
+                outpng = path_full2web(outfns[case])
+                html.write(
+                    "\t<a href='{}'><img SRC='{}' width={} height=auto></a>\n".format(
+                        outpng, outpng, width,
+                    )
                 )
-            )
-        elif os.path.splitext(outfns[case])[-1] == ".pdf":
-            outpdf = path_full2web(outfns[case])
-            html.write("\t<iframe src='{}' width={} height=100%></iframe>\n".format(outpdf, width))
+            elif os.path.splitext(outfns[case])[-1] == ".pdf":
+                outpdf = path_full2web(outfns[case])
+                html.write("\t<iframe src='{}' width={} height=100%></iframe>\n".format(outpdf, width))
+            else:
+                log.error("Unexpected extension for {}; exiting".format(outfns[case]))
+                sys.exit(1)
         else:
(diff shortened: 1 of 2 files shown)


py/desispec/night_qa.py
```diff
@@ -85,19 +88,40 @@ def get_survey_night_expids(
...
-            len(expids), np.unique(tileids).size, survey, night,
+            survey = "unknown"
+            # AR look for the fiberassign file
+            # AR - used wildcard, because early files (pre-SV1?) were not gzipped
+            # AR - first check SURVEY keyword (should work for SV3 and later)
+            # AR - if not present, take FA_SURV
+            fafns = glob(os.path.join(os.path.dirname(fns[i]), "fiberassign-??????.fits*"))
+            if len(fafns) > 0:
+                fahdr = fits.getheader(fafns[0], 0)
+                if "SURVEY" in [cards[0] for cards in fahdr.cards]:
```
sbailey, line 103: FYI, somewhat simpler is
```
if "SURVEY" in fahdar:
```
or
```
if "SURVEY" in fahdr.keys():
```

py/desispec/night_qa.py
```diff
@@ -85,19 +88,40 @@ def get_survey_night_expids(
...
+                else:
+                    survey = fahdr["FA_SURV"]
+            if survey == "unknown":
+                log.warning("SURVEY could not be identified for {}; setting to 'unknown'".format(fns[i]))
+            # AR append
+            expids.append(hdr["EXPID"])
+            tileids.append(hdr["TILEID"])
+            surveys.append(survey)
+    expids, tileids, surveys = np.array(expids), np.array(tileids), np.array(surveys)
+    msg = "for NIGHT={} found"
```
sbailey, line 114: It looks like `msg` is an unused leftover?

py/desispec/night_qa.py
```diff
@@ -796,26 +825,27 @@ def create_petalnz_pdf(outpdf, night, prod, survey="main", dchi2_threshold=25):
...
-                            "Z", "ZWARN", "SPECTYPE", "DELTACHI2",
-                            "Z_QN", "Z_QN_CONF", "IS_QSO_QN",
-                        ],
-                    )
-                )
+                d = Table.read(fn, hdu="ZMTL")
+                # AR rename *DESI_TARGET and *BGS_TARGET to DESI_TARGET and BGS_TARGET
+                keys, _, _ = main_cmx_or_sv(d)
+                d.rename_column(keys[0], "DESI_TARGET")
+                d.rename_column(keys[1], "BGS_TARGET")
```
sbailey, line 832: nice trick to simplify downstream code.

sbailey (approved):
Looks good, thanks.  I think you found a cosmic ray shower in r1.  I put two minor comments in line, but ok to merge whenever you are ready.

py/desispec/night_qa.py
araichoor: thanks a lot, I kind of new my coding was bad here, but didn't know those simpler (and nicer!) approach.

py/desispec/night_qa.py
araichoor: correct, nice catch.

py/desispec/night_qa.py
araichoor: thanks.
I know there is some fragility here: it only works because Adam had the nice idea to reserve the same bit values for LRG/ELG/QSO in the *desi_mask and BGS_BRIGHT/BGS_FAINT in the *bgs_mask for sv1/sv2/sv3/main.
trying to display other tracers is likely to break that; but I don't foresee that for that plot, so I've picked this shortcut approach.

araichoor:
Good you waited to merge: I've noticed that I forgot to push my last commit last night: I've just done it (https://github.com/desihub/desispec/pull/1535/commits/0c3bad26012337027b166886f125959492970fc6).
It's a simple "if enough tiles" case in the `create_petalnz_pdf()` function.
And I've also made the two small changes you noticed (thanks for the careful code reading!).

So I think I'm good on my side.

Issue #1538, opened 2021-12-17
Update fiber offset values after turbulence corrections


djschlegel:
The turbulence-correction code is applied to the FVC images starting with expid 114320 on 2021-12-16.  Starting with that exposure, the following columns in the coordinates files no longer correspond to our best estimates of the fiber positioning: FIBER_OFFSET, FIBER_RA/DEC, FIBER_DX/DY.

The following corrections could be applied to apply the turbulence correction to these columns:
FIBER_X = FIBER_X - TURB_X_1
FIBER_Y = FIBER_Y - TURB_Y_1
FIBER_DX = FIBER_DX + TURB_X_1
FIBER_DY = FIBER_DY + TURB_Y_1
FIBER_OFFSET = OFFSET_1 = sqrt(FIBER_DX**2 + FIBER_DY**2)  <-- After first correcting FIBER_DX,FIBER_DY

We should also correct FIBER_RA,FIBER_DEC.  The below isn't exact, but I would propose it may be close enough (and certainly better than making no correction):
FIBER_RA = TARGET_RA - FIBER_DX*1000/(70*3600)/cos(TARGET_DEC)
FIBER_DEC = TARGET_DEC + FIBER_DY*1000/(70*3600)

The above corrections should be made before rejecting fibers as off-target in the spectro reductions.  As an example, exposure 114359 on TILEID 21560 would have 346 fibers rejected according to FIBER_OFFSET (black histogram below), but far fewer rejected after these corrections (red histogram below).
<img width="586" alt="Screen Shot 2021-12-17 at 7 15 20 AM" src="https://user-images.githubusercontent.com/7918018/146574547-28f58797-4d12-4c63-9b14-9dad92e8f68f.png">

PR #1539, opened 2021-12-17
updates for turbulence corrections on exp 114320 onwards

sbailey:
This PR is an implementation of turbulence corrections from issue #1538 .  However, just as I was finishing this update I discovered that the `FPA_X/Y_n`, `DX/DY_n` etc. columns that we have been using do have the turbulence corrections already applied and maybe we should just keep using those.

I also implemented the `FIBER_RA/DEC` recalculation suggested in #1538.  I also considered applying a `TURB_X/Y_n` correction to `FIBER_RA/DEC` instead of completely replacing it, but it looks like the TURB corrections are larger than the FIBER_DX/DY corrections, so I think the original suggestion is better.

Example files are in /global/cfs/cdirs/desi/users/sjbailey/dev/turbulence:
  * expid 114319, before turbulence corrections were turned on
    * fibermap-114319-old.fits - current master
    * fibermap-114319-new.fits - this branch
  * expid 114320, after turbulence corrections were turned on
    * fibermap-114320-old.fits - current master
    * fibermap-114320-new.fits - this branch

@djschlegel 

I've flagged this as a draft PR; don't merge yet until we settle on what we really want to do with these columns

### DIFF
--- a/py/desispec/io/fibermap.py
+++ b/py/desispec/io/fibermap.py
@@ -664,39 +664,22 @@ def assemble_fibermap(night, expid, badamps=None, badfibers_filename=None,
         log.info('%d/%d fibers in coordinates file', len(pm), len(fa))
 
         #- Exposures 114320 onwards (part way through night 20211216)
-        #- have FVC turbulence correction info, but aren't applied to
-        #- FIBER_* columns.  Fix that.
+        #- have FVC turbulence corrections applied to FPA_X/Y_n and DX_n/DY_n
+        #- columns, but this isn't applied to FIBER_RA/DEC yet, so make
+        #- an approximate correction to TARGET_RA/DEC instead.
         #- NOTE: in the future there may also be an end date after which
         #- they are applied, but we don't know if/when that is yet.
         i = numiter-1
         if (expid >= 114320 and
-            f'TURB_X_{i}' in pm.colnames and
-            f'TURB_Y_{i}' in pm.colnames
+            f'TURB_X_{i}' in pm.colnames and f'TURB_Y_{i}' in pm.colnames and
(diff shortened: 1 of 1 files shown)


sbailey:
@djschlegel I reverted to using the FPA_X/Y and DX/DY values that already had the turbulence correction applied, and so now this just updates FIBER_RA/DEC for exposures 114320 onwards (and when we have an end-exposure, we can add that to the filter).  Please check the fibermap files above to make sure that they have the correct FIBER_RA/DEC, FIBER_X/Y, and DELTA_X/Y  (aka FIBER_DX/DY).

djschlegel:
@sbailey I can confirm that the first set of files don't change any values (as intended), and that I can reproduce and agree with the change of values in the second set (as intended).

sbailey:
Thanks for the check.  I'm converting back to a draft PR state to remind me that I want to add the check that turbulence corrections actually were turned on for a given exposure before applying those turbulence corrections to FIBER_RA/FIBER_DEC.  i.e. the existence of the TURB_X/Y columns doesn't necessarily mean that the fibers were corrected for that.

sbailey:
I added an end date of 20211222 expid 115150 based on email with Steve Kent.  In the end this PR makes FIBER_RA/DEC approximations based upon turbulence corrections for 20211216/114320 until (but not including) 20211222/115150 but doesn't otherwise muck with the columns because we were already using the columns that have turbulence corrections when they are applied.  1216-1221 is the date range when turbulence corrections were applied to the X/Y values but not propagated to FIBER_RA/DEC yet, thus this special case handling.

sbailey:
Tests are failing due to fitsio/numpy installation incompatibility fixed in test config in PR #1566, so ignoring those here.  Merging now.

Issue #1541, opened 2021-12-19
preproc OSTEP cut rejecting 20201214 b0 flats


sbailey:
The desispec.preproc.preproc OSTEP>5 -> BADREADNOISE cut is rejecting multiple b0 flat exposures on 20201214, which we had previously accepted and used in everest.  I haven't gotten far enough in the processing to evaluate how much this is impacting other nights or science exposures in Dec 2020.

This cut was originally motivated by noisy r8A in October 2021; re-evaluate whether it might be rejecting too much earlier data, especially high S/N flats where a 5 ADU step may not matter.

Example:
```
desi_preproc -o preproc-b0-00067628.fits --cameras b0 -n 20201214 -e 67628
...
INFO:raw.py:250:read_raw: Setting BADREADNOISE bit for fibers of amp A
INFO:raw.py:250:read_raw: Setting BADREADNOISE bit for fibers of amp B
INFO:raw.py:250:read_raw: Setting BADREADNOISE bit for fibers of amp C
INFO:raw.py:250:read_raw: Setting BADREADNOISE bit for fibers of amp D
...
```
Then from python:
```
import numpy as np
import fitsio
mask = fitsio.read('preproc-b0-00067628.fits', 'MASK')
np.all(mask>0)  # True (!)
```

Other impacted flats on 20201214 are 67629, 67634, 67638, 67639, and 67645.

@julienguy

sbailey:
The OSTEP variations were much larger in all of the B CCDs prior to late January 2021, e.g. from b0 and b1 (diff colors = diff amps):

![image](https://user-images.githubusercontent.com/218471/146694699-a78f1182-232d-4fab-9bf9-3daecef55e24.png)
![image](https://user-images.githubusercontent.com/218471/146694727-3ecb419b-a938-4d91-aedc-74d612f4e51d.png)

The change corresponds to 20210128 when the FEE configuration was updated to include 400 prescan reads to minimize the "non-linear dark" effect.  Zooming in on a single night of the early data, the different OSTEP bands correspond to zeros, arcs, flats, and science exposures (longer exptimes = larger OSTEP).

It looks like for this time period we'll need to either loosen the OSTEP cut before discarding the amp, or have a larger edge buffer for where we measure the OSTEP.

[...]

PR #1542, opened 2021-12-19
fix desi_proc MPI bug causing hang if extractions missing inputs

sbailey:
This PR fixes a desi_proc MPI logic bug that resulted in a job hang if one of the cameras was missing a necessary input for extractions.  This was motivated by 20201214/00067628 where the b0 preproc gets completely masked (see #1541), causing the traceshifts to fail, causing one of the extraction groups to be missing the PSF input, causing them to never make it to a barrier where all the other ranks were waiting.

This command hangs on current master, and with this branch it processes 29 good cameras to completion and then exits without waiting for the job to timeout:
```
srun -N 7 -n 200 -c 2 desi_proc --mpi --traceshift --cameras a0123456789 -n 20201214 -e 67628
```

Other changes:
  * all preproc log messages include the camera to assist understanding when 30 cameras are being processed in parallel
  * trace_shifts fail faster with an informative log message in the case that the input image is entirely masked

### DIFF
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -489,6 +489,13 @@ def preproc(rawimage, header, primary_header, bias=True, dark=True, pixflat=True
     #- Check if this file uses amp names 1,2,3,4 (old) or A,B,C,D (new)
     amp_ids = get_amp_ids(header)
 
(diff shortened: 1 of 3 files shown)


sbailey:
I've tested this on multiple nights, so will merge now.  Post facto questions / comments are still welcome.

PR #1543, opened 2021-12-20
OSTEP: use difference between adjacent rows to measure the step

julienguy:
In this PR we modify the estimator of the overscan step derived from an array of overscan values per CCD row.

After a median filter of 7 rows to reduce the noise, we measure the amplitude (max-min) of the difference between consecutive values instead of the max-min value of whole array. This removes the sensitivity to slow variation of the overscan while still detecting sharp changes, steps or horizontal bands.

On 20201214/00067628 B0 the steps originally of about 6 electrons are reduced to 1 to 1.2 electrons depending on the amplifier.

On the example exposures presented in https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=6516,

20211026/00106136 R8 amp A: 14.0 instead of 14.3 electrons
20211026/00106135 R8 amp A: 11.0 instead of 12.7 electrons
20211027/00106260 R8 amp A: 5.3 instead o6 6.3 electrons

and similar values for the other (good) amplifiers:

20211026/00106135 R8 amp B: 0.7 instead of 0.6
20211026/00106135 R8 amp C: 0.7 instead of 0.8
20211026/00106135 R8 amp D: 0.6 instead of 0.6

This addresses issue #1541

### DIFF
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -81,20 +81,24 @@ def _clipped_std_bias(nsigma):
     stdbias = np.sqrt(1 - 2*a*np.exp(-a**2/2.) / (np.sqrt(2*np.pi) * erf(a/np.sqrt(2))))
     return stdbias
 
-def compute_overscan_step(overscan_col) :
+def compute_overscan_step(overscan_col, median_size=7, edge_margin=50) :
     """
     Compute the overscan step score 'OSTEP' from an array of
     overscan values averaged per CCD row
 
     Args:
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks for this update, and for specifically checking both good amp and bad amp cases.  I promoted the hardcoded median filter size and the number of edge pixels into options (with the same defaults as before) and confirmed that it still produces identical results to your original branch.  Will merge after tests re-pass.

PR #1546, opened 2021-12-20
Nightly bias for all

julienguy:
This PR implements (few lines changes) the proposal that we also use the nightly bias for red and nir cameras because this considerably reduces the number of 'BADCOLUMN'. Indeed most residuals in CCD columns are not due to dark current but charges generated at readout on specific pixels in the serial clock row. So we can remove them by using a recent bias.

Example from 2021/12/17 and 2021/12/18. Below are figures of the number of bad columns detected with `desi_inspect_dark` in the 300s dark exposure preprocessed with the default or nightly bias for the red and nir cameras.
badcolumns-20211217
badcolumns-20211218

The changes to the code simply consist in considering a nightly bias for all cameras, not just b, and to give an slight advantage in favor of the nightly bias over the default when comparing the residual levels after preprocessing (because they are very similar for the red and nir camera when the CCD are stabilized).


sbailey:
Looks good; thanks for doing this check.

Issue #1548, opened 2021-12-22
NightlyQA Crash in create_petalnz_pdf


akremin:
There was a failure in the QA plotting code when run on the night 20211220 in the daily specprod. 

Command run:
`desi_night_qa -p daily -n 20211220 -o /global/cfs/cdirs/desi/spectro/redux/daily/nightqa/20211220`

```
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/bin/desi_night_qa", line 155, in <module>
    main()
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/bin/desi_night_qa", line 144, in main
    create_petalnz_pdf(outfns["petalnz"], args.night, args.prod, unq_tileids, unq_surveys, dchi2_threshold=25)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/night_qa.py", line 909, in create_petalnz_pdf
    "dark" : np.unique(ds["dark"]["TILEID"][ds["dark"]["SURVEY"] == survey]).size,
TypeError: list indices must be integers or slices, not str
```

akremin:
Resolved with #1549

PR #1549, opened 2021-12-22
nightqa petalnz: handle faprgrm with no tiles

araichoor:
This PR addresses a failure in `night_qa.create_petalnz_pdf()`, when there are no tiles only for one of the two analyzed BRIGHT and DARK programs (https://github.com/desihub/desispec/issues/1548).

@akremin: sorry for having ill-coded that part, missing that corner case.


akremin:
Thank you for the very quick fix! I tested this branch on the daily prod for the night where the issue was discovered (20211220) and it worked. Merging

Issue #1552, opened 2021-12-23
create_petalnz_pdf crash on 20211222


akremin:
New crash in Nightly QA in `create_petalnz_pdf`:

```
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/bin/desi_night_qa", line 155, in <module>
    main()
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/bin/desi_night_qa", line 144, in main
    create_petalnz_pdf(outfns["petalnz"], args.night, args.prod, unq_tileids, unq_surveys, dchi2_threshold=25)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/night_qa.py", line 907, in create_petalnz_pdf
    ntiles_surv = {
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/night_qa.py", line 909, in <dictcomp>
    ds[faprgrm]["TILEID"][ds["bright"]["SURVEY"] == survey]
  File "astropy/table/_column_mixins.pyx", line 74, in astropy.table._column_mixins._ColumnGetitemShim.__getitem__
  File "astropy/table/_column_mixins.pyx", line 57, in astropy.table._column_mixins.base_getitem
  File "astropy/table/_column_mixins.pyx", line 69, in astropy.table._column_mixins.column_getitem
IndexError: boolean index did not match indexed array along dimension 0; dimension is 19936 but corresponding boolean dimension is 9041
```

Command run:
`desi_night_qa -p daily -n 20211222 -o /global/cfs/cdirs/desi/spectro/redux/daily/nightqa/20211222`

akremin:
Fixed with PR #1554

PR #1554, opened 2021-12-24
petalnz: fix bug

araichoor:
This PR corrects a bug (leftover from development test) introduced in PR https://github.com/desihub/desispec/pull/1549, reported in https://github.com/desihub/desispec/issues/1552.

Apologies for that...

(the test @akremin and I did on 20211220 for PR https://github.com/desihub/desispec/pull/1549 did work, because that night only has bright tiles).


akremin:
Thank you for another quick fix. It worked for the night in question, and the fix is clearly correct. Merging.

Issue #1556, opened 2021-12-26
Interpolation of fiber traces and PSF parameters for broken/masked fibers


julienguy:
Bug in the interpolation of fiber traces and PSF parameters for broken/masked fibers causing the bad fiber indexing reported in issue #1555.
Example to work on:
```
desi_interpolate_fiber_psf --infile /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20211223/00115320/fit-psf-r4-00115320.fits --outfile psf.fits --fibers 36,51,132,156,250,251,252,253,254,255,256,257,258,259,260,261,262,263,264,265,266,267,268,269,270,271
plot_fiber_traces --fibers 250:275 -i psf.fits
```

julienguy:
Solved with PR #1557

PR #1557, opened 2021-12-27
Improvements to desi_interpolate_fiber_psf

julienguy:
Select preferentially fibers in the same bundle as reference to fix bad fibers.
Fix a bug in the interpolated x-coordinate of the bad fiber trace when the reference fiber is in another bundle.

### DIFF
--- a/bin/desi_interpolate_fiber_psf
+++ b/bin/desi_interpolate_fiber_psf
@@ -5,6 +5,7 @@ import numpy as np
 
 import astropy.io.fits as pyfits
 
+from desiutil.log import get_logger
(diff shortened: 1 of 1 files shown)


julienguy:
This fixes issue #1556.

julienguy:
I verified it fixes r4 issue. I also ran it on many cameras with broken fibers and visually checked the code is doing the right thing. For the arc exposures of 20211223 in r4 camera, I computed a nightly PSF from the first series of arc exposures and verified the code now successfully fit the PSFs of the second series of arcs (while using the new nightly PSF as input).

PR #1564, opened 2022-01-03
Readnoise estimation whensubtracting the overscan per row

julienguy:
Simple PR that changes the estimation of the readnoise when we are subtracting the overscan per CCD row
(in case where OSTEP>2 electrons).

PR #1566, opened 2022-01-04
PR to test GitHub test failures

sbailey:
This PR has a minor doc change to trigger tests to study them in isolation from other algorithmic changes.  Original test failures noticed in #1564, where the default installed fitsio+numpy appear to be incompatible leading to fitsio import errors.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -17,9 +17,9 @@ jobs:
             fail-fast: true
             matrix:
                 os: [ubuntu-latest]
-                python-version: [3.8, 3.9]  # fitsio does not like Python < 3.7, astropy does not like Python < 3.8
+                python-version: [3.8]  # fitsio does not like Python < 3.7, astropy does not like Python < 3.8, might be too soon for 3.9.
                 astropy-version: ['==4.0.1.post1', '<4.1', '<5.0', '==5.0']
-                fitsio-version: ['==1.1.2']  #, '==1.1.4', '==1.1.6']
+                fitsio-version: ['==1.1.5']  #, '==1.1.4', '==1.1.6']
         env:
             DESIUTIL_VERSION: 3.2.3
             DESIMODEL_DATA: branches/test-0.12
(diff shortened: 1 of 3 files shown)


sbailey:
@weaverba137 could you please debug our GitHub actions test environment which is now failing due to a fitsio import error?  This branch+PR is a starting point to demonstrate the failure independent of other algorithmic changes.  A few notes:

* our latest 'test-master' environment at NERSC has fitsio/1.1.6 and numpy/1.21.4 and works
* fitsio has a "1.1.6" release on pip and conda-forge, but doesn't appear to have a tag by that name
* it looks like these tests are using fitsio/1.1.6 and numpy/1.20.3; it isn't clear to me which of our other requirements is triggering the older version of numpy (it started out trying numpy/1.22.0)
* maybe related to numpy versions: our tests currently use astropy/4.0.1.post1, which matches what our daily pipeline environment uses (master).  Fuji testing uses astropy/5.0 (test-master).  Switching to 5.0 for most testing would be fine, but it would be helpful to retain at least one version of the test using 4.0.1.post1 to ensure that we don't break daily ops while merging Fuji PRs.
  * if supporting testing with astropy/4.0.1.post1 is an undue burden, let's chat options.  The primary reason for not upgrading daily ops is just to minimize the number of moving parts; I'm not aware of any fundamental problems with promoting test-master to default master if needed (and yes, one of these day we'll make the master -> main renaming, also only limited by minimizing moving parts).

Thanks

weaverba137:
@sbailey, acknowledged.  Is there any restriction on the version of `fitsio`, or can I tune that to adjust numpy dependencies?

weaverba137:
numba 0.54.1 requires numpy < 1.21, so that explains why an older numpy is installed after numpy 1.22.x is downloaded.

weaverba137:
It's looking like fitsio==1.1.6 might be incompatible with numpy<1.21.

weaverba137:
Yikes! It now looks like the version of numpy that fitsio is compiled against may depend on the order in which numpy & fitsio are installed.

weaverba137:
@sbailey, summary so far:

* Forcing fitsio to be reinstalled after all other packages does seem to work.
* However, if fitsio==1.1.6 is being tested, it falls back on a cached build, even with `pip install --force`.  I think there is an additional pip command that will force a rebuild.

weaverba137:
@sbailey, I think this is ready for review and merge.

sbailey:
Looks good, thanks.

PR #1569, opened 2022-01-05
Assemble fibermap; empty map is fall-through case.

### LINKED ISSUE #1315: Preproc writing wrong fibermap file
This was noticed because OBJTYPE is not being identified properly in NightWatch. That's because the correct fibermap info is not being saved in the nightwatch files. I traced this to desispec.scripts.preproc. Line 161, sets the fibermapfile is `fibermapfile = infile.replace('desi-', 'fibermap-').replace('.fits.fz', '.fits')` but the fibermap file should really be fiberassign-<TILE>.fits.fz. This is sent to io.readraw which sets it as an empty fibermap on line 144: `fibermap = desispec.io.empty_fibermap(5000)`

sybenzvi:
PR to fix #1315. IO code will:
1. Attempt to load a fibermap from a named disk file.
2. If (1) isn't successful, assemble a fibermap using NIGHT and EXPID.
3. If (2) isn't successful, use a blank fibermap.

### DIFF
--- a/py/desispec/io/raw.py
+++ b/py/desispec/io/raw.py
@@ -8,6 +8,7 @@ TODO: move into datamodel after we have verified the format
 import os.path
 import time
 from astropy.io import fits
+from astropy.table import Table
 import numpy as np
 
 from desiutil.depend import add_dependencies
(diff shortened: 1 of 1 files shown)


sbailey:
I discussed this with @sybenzvi, and I think it would be better for nightwatch to proactively call `assemble_fibermap` and then call `read_raw` with that fibermap that it just created, instead of having `read_raw` decide to create the missing fibermap on-the-fly (read_raw already does a surprising amount of non-I/O calculation; let's not add to that).  This is similar to how the offline pipeline works, and allows a single process to make the single fibermap which is then used by 30 cameras, instead of 30 processes all creating fibermaps on-the-fly using the same inputs.

So I'm going to close this without merging and punt it back to nightwatch to call assemble_fibermap (the script) before read_raw.

Issue #1562, opened 2021-12-31
discard nightly bias if BADAMP set


sbailey:
amp r1A was in a bad state on 20201217 from exposures 68341 though 68405 (inclusive), covering the ZEROs, 5min DARK, and ARCs, and then it was fixed before the FLATs began with exposure 68416.

All of the bad exposures already have BADAMPS=r1A set, but biasnight-r1-20201217.fits.gz is generated anyway and is junk, polluting preprocessing of later exposures, e.g. causing the flats to fail.  Fix desi_compute_nightly_bias so that if a BADAMP is set for a given camera, it won't use that ZERO (and in this case there won't be enough good ZEROs and the nightlybias won't be generated for r1, which is better than having a junk one).

Note: badcolumns (from DARK) and psf (from ARC) are correctly skipped due to the bad amp.

akremin:
This should be solved with PR #1570

Issue #1511, opened 2021-12-06
darks and nightly bias when only one camera is bad


sbailey:
Do `desispec.ccdcalib.make_dark_scripts`, `desi_proc --nightlybias ...`, and `desi_run_night` do the right bookkeeping if only one camera is flagged as bad instead of the full exposure?

Example: night 20211203 exposures 111915-111939 (ZERO) and 111941 (DARK) had bad z3 (VCCD=off), but were fine for the other cameras.  Observers then re-took ZEROs (111987-112011) and DARK (112013) with all 30 cameras.  For the moment I have flagged daily/exposure_tables/202112/exposure_table_20211203.csv EXPID 111915-111939 with BADCAMWORD=z3 and flagged the first DARK 111941 as bad in the hopes that pipeline re-runs will correctly pickup the second dark (112013) for determining badcolumns.  If we need to, we could flag all the first ZEROs as bad since we have a complete second set, but there may be cases in the future where we don't have a second set and need to have this bookkeeping right when a subset of cameras are bad for a subset of exposures.

@Waelthus could you check `desispec.ccdcalib.make_dark_scripts` and make updates as needed so that it doesn't incorrectly try to use the bad z3 exposures when making darks?

@Waelthus and/or @RaffaellaCapasso please try running `desi_run_night -n 20211203 --proc-obstypes dark --dry-run` and `desi_proc -n 20211203 -e 112013 --nightlybias` in a test prod to see if it incorrectly tries to use the bad z3 exposures.  We may need to get @akremin help for bookkeeping but it would be helpful if you could start with assessing the situation so that Anthony can focus on Fuji instead of dailyops.

Again: for this particular case we can flag 111915-111939 as bad if needed so it doesn't have to be a blocking factor, but it's a good test case for studying the bookkeeping for the future.

Waelthus:
Ok, I did run the test_prod, and to me it does not look like 111915-111939 (ZERO) and 111941 (DARK) have been used for the z3 reduction. 

The processing table produced from `desi_run_night` only contains 112013 and the logs from `desi_proc` only mention that tile as well regarding badcolumns. I guess given that the full DARK was flagged to ignore, and not just Z3 as BADCAM this is the expected behaviour?
Are the nightly biases supposed to be used for R/Z at all? There are only `calibnight/biasnight-*` outputs/log-entries for B. Since this is the case generally and not just in this test_prod, I'd assume we currently don't use nightly biases for R/Z.

[...]

PR #1570, opened 2022-01-05
Removing bad exp for darks and biases

Waelthus:
This PR is supposed to tackle #1511 and #1562.
I.e. for both the standalone make_dark_scripts/desi_compute_dark_nonlinear as well as the nightly desi_compute_nightly_bias it incorporates checking of BADAMPS/BADCAMWORD/CAMWORD.
For both types of analysis exposures will only be used when processing a given camera when it is not part of BADAMPS/BADCAMWORDS and either not be part of the exposure_table at all or part of CAMWORD and has LASTSTEP='all'.

Exposures are treated in the following fashion:
If no exposure_table entry exists, we assume all is fine with them.
Else, every exposure with LASTSTEP!=all will be ignored for all cams (this was used for nightlybias before, but not for standalone). Cameras that are flagged inside BADAMPS or BADCAMWORD as well as cameras that are not inside CAMWORD will not be used either.

For the nightly: if there are enough exposures, i.e. by default nzeros=25, with all cams working, only those are used. Only when there are less, exposures with problems on some cameras will still be used for the others.
In cases where an exposure does not even reach minzeros anymore for some camera (e.g. r1 on 20201217 where all exp are bad), all other cameras will still be processed and an error is written to the log (instead of raising an Exception). Please tell if this is the behaviour we want or if we should fail completely (maybe for simpler bookkeeping?).

For the standalone: Will always use whatever is available without being flagged. This required a modification of the speclog produced by make_dark_scripts to track which cams are good/bad.

I tested this for the nights of 20211203 (some exposures with BAD z3, but second set was taken, for both nightly and standalone BAD z3 exposures are correctly ignored, and depending on the number of zeros requested for the other spectrograph the first set is still used) and 20201217 (BADAMP r1A  leads to discarding all r1 spectra for nightly_bias [with SPECPROD="f4"] no output file is produced for this, other cams are still processed as normal).


py/desispec/ccdcalib.py
```diff
@@ -423,28 +460,38 @@ def compute_nightly_bias(night, cameras, outdir=None, nzeros=25, minzeros=20,
...
     if rank == 0:
-        expids = _find_zeros(night)
+        expdict = _find_zeros(night,cameras=cameras,minzeros=nzeros)
 
     if comm is not None:
-        expids = comm.bcast(expids, root=0)
+        expdict = comm.bcast(expdict, root=0)
+
+    rawfiles_dict={}
+    skipped_cams=[]
```
akremin, line 471: You could tuck all of lines 471-493 under line 465 (the `_find_zeros()` line) within the `if rank==0`. Then broadcast the shortened `expdict` to all ranks assuming you didn't exit with no valid cameras.

Otherwise you're broadcasting a larger dictionary and having all ranks calculate all cams.

py/desispec/ccdcalib.py
```diff
@@ -423,28 +460,38 @@ def compute_nightly_bias(night, cameras, outdir=None, nzeros=25, minzeros=20,
...
         if rank == 0:
-            log.critical(msg)
-        raise RuntimeError(msg)
+            log.info(f'Using {len(expdict[cam])} ZEROs for nightly bias {night} and cam {cam}')
 
-    if len(expids) > nzeros:
-        nexps = len(expids)
-        n = (nexps - nzeros)//2
-        expids = expids[n:n+nzeros]
+        rawfiles_dict[cam] = [io.findfile('raw', night, e) for e in expids]
```
akremin, line 488: You can move this logic to the next for loop so that each rank only determine the rawfiles for its camera(s)

py/desispec/ccdcalib.py
```diff
@@ -458,6 +505,13 @@ def compute_nightly_bias(night, cameras, outdir=None, nzeros=25, minzeros=20,
 
     nfail = 0
     for camera in cameras[rank::size]:
+        if camera in skipped_cams:
+            log.error(f'execution was skipped for camera {camera} due to lack of usable zeros')
+            nfail+=1
+            continue
+
+        expids=expdict[camera]
+        rawfiles=rawfiles_dict[camera]
```
akremin, line 514: Replace with logic from line 488 and use the `expids` you just took from the `expdict`.

py/desispec/ccdcalib.py
```diff
@@ -476,6 +530,7 @@ def compute_nightly_bias(night, cameras, outdir=None, nzeros=25, minzeros=20,
                 compute_bias_file(rawfiles, testbias, camera,
                                   extraheader=dict(NIGHT=night))
             except Exception as ex:
+                nfail+=1
```
akremin, line 533: Thank you for implementing the failure counting!

py/desispec/ccdcalib.py
```diff
@@ -743,12 +799,50 @@ def make_dark_scripts(outdir, days=None, nights=None, cameras=None,
...
+            if entry['EXPID'] in exptable_select['EXPID']:
+                sel=entry['EXPID']==exptable_select['EXPID']
+                badcamwords.append(exptable_select['BADCAMWORD'][sel][0])
+                badamps.append(exptable_select['BADAMPS'][sel][0])
+                camwords.append(exptable_select['CAMWORD'][sel][0])
+                laststeps.append(exptable_select['LASTSTEP'][sel][0])
+            else:
+                badcamwords.append("")
+                laststeps.append("all")
+                camwords.append("a123456789")
```
akremin, line 832: This is missing petal 0

py/desispec/ccdcalib.py
```diff
@@ -743,12 +799,50 @@ def make_dark_scripts(outdir, days=None, nights=None, cameras=None,
...
+        for night in nightlist:
+            expfiles.append(get_exposure_table_pathname(night))
+        exptables = load_tables(expfiles)
+        exptable_all=table_vstack(exptables)
+        select = ((exptable_all['OBSTYPE']=='zero')|(exptable_all['OBSTYPE']=='dark'))
+        exptable_select=exptable_all[select]
+    
     log.info(f'Scanning {len(nightlist)} night directories')
     speclog = io.util.get_speclog(nightlist)
+    if use_exptable:
```
akremin, line 817: As mentioned in another comment, I'd also add logic to make the `OBSTYPE`'s all one case (probably uppercase since that is what the "speclog" uses).

py/desispec/ccdcalib.py
```diff
@@ -782,6 +876,8 @@ def make_dark_scripts(outdir, days=None, nights=None, cameras=None,
             cmd += f" \\\n    --nskip-zeros {nskip_zeros}"
         if first_expid is not None:
             cmd += f" \\\n    --first-expid {first_expid}"
+        if use_exptable:
```
akremin, line 879: Can safely remove this if you implement my other comment removing the option

akremin, line 879: Can safely remove this if you implement my other comment removing the option

akremin (changes requested):
Thank you Michael! Overall this looks great and I think you made the right decisions throughout. It looks like it will work as-is, but I have some requested changes. The requests are to increase clarity and optimize a little bit. There are a lot of things, but they should all be simple changes. Please let me know if you need clarification on anything.

Since some of my requested changes move lines around, I'll wait to test your code until after you address the requests.

If I could make one final request: to speed up testing could you provide the commands you used in your debugging and/or point us to a directory at NERSC with outputs of your tests?

bin/desi_compute_dark_nonlinear
```diff
@@ -144,6 +157,16 @@ if args.last_expid is not None:
         log.critical('No exposures! exiting...')
         sys.exit(1)
 
+if args.use_exptable:
```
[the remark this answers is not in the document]
Waelthus: done

bin/desi_compute_dark_nonlinear
```diff
@@ -104,11 +108,20 @@ else:
     daynight = 'NIGHT'
     nights = args.nights
 
+#if args.use_exptable:
```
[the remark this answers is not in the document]
Waelthus: done

bin/desi_compute_dark_nonlinear
```diff
@@ -47,6 +47,9 @@ parser.add_argument('--nskip-zeros', type=int, default=30, required=False,
                     help='Skip N ZEROs per day while flushing charge')
 parser.add_argument('--mindarks', type=int, default=5, required=False,
                     help='Minimum number of DARKs per EXPTIME to use that EXPTIME')
+parser.add_argument('--use_exptable', action='store_true', required=False,
```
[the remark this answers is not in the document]
Waelthus: done

bin/desi_compute_dark_nonlinear
```diff
@@ -161,8 +184,8 @@ for exptime in np.unique(speclog['EXPTIME_INT']):
 speclog = speclog[keep]
 
 #- Print some summary stats before continuing
-isZero = speclog['OBSTYPE'] == 'ZERO'
-isDark = speclog['OBSTYPE'] == 'DARK'
+isZero = (speclog['OBSTYPE'] == 'ZERO') | (speclog['OBSTYPE'] == 'zero')
```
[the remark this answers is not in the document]
Waelthus: agreed, I'll double check why I needed to put this in at some point

py/desispec/ccdcalib.py
Waelthus: indeed, fixed now

py/desispec/ccdcalib.py
```diff
@@ -337,11 +338,13 @@ def compute_bias_file(rawfiles, outfile, camera, explistfile=None,
 
     log.info(f"done with {camera}")
 
-def _find_zeros(night):
+def _find_zeros(night,cameras,minzeros=25):
```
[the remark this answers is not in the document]
Waelthus: done

py/desispec/ccdcalib.py
```diff
@@ -376,18 +379,52 @@ def _find_zeros(night):
...
+                for camera in goodcamlist:
+                    if camera in cameras and camera not in badampstring:
+                        expdict[camera].append(expid)
+    else:
+        expdict={f'{cam}':expids for cam in cameras}
+    
+    for camera,expids in expdict.items():
+        log.info(f'Keeping {len(expids)} calibration ZEROs for camera {camera}')
+        #make sure everything is in np arrays again
+        expdict[camera]=np.array(expids)
```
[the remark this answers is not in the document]
Waelthus: indeed

py/desispec/ccdcalib.py
Waelthus: done

py/desispec/ccdcalib.py
Waelthus: done, also saved the skipped cameras by just removing from the expdict and checking that later

py/desispec/ccdcalib.py
```diff
@@ -423,28 +460,38 @@ def compute_nightly_bias(night, cameras, outdir=None, nzeros=25, minzeros=20,
...
+    expdict = None
     if rank == 0:
-        expids = _find_zeros(night)
+        expdict = _find_zeros(night,cameras=cameras,minzeros=nzeros)
 
     if comm is not None:
-        expids = comm.bcast(expids, root=0)
+        expdict = comm.bcast(expdict, root=0)
+
+    rawfiles_dict={}
```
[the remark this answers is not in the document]
Waelthus: done

py/desispec/ccdcalib.py
Waelthus: done, please tell if there's a different preferred way on how to do this

Waelthus:
thanks @akremin. I modified the code according to your suggestions. 


The tests can be found on cori: `/global/cscratch1/sd/mwalther/desi/PRs/desispec/1570/`.
For testing I used SPECPROD=f4, such that exposure tables are found which mark things as bad.



- Subdirectory `testing_darks`:
  - steps: 
     - run `create.py` (which calls `make_dark_scripts`)
     - modify output slurm script to load development env
     - run slurm script
  -  this runs on 20211203 and 20211204 and reports 88 ZEROs and 36 DARKs for b2, but 63 ZEROs and 36 DARKs for z3 on 20211203 (i.e. the latter skips 25 bad ZEROs, both skip the bad DARK), for the other date both use the same number of exp.

 - Subdirectory, `testing_nightly_bias`:
   - tests on 20201217 (to check if the behaviour for missing zeros is as expected when BADAMPS was set and when no replacements where taken)
     - steps: submit nightlybias.sl
      this correctly treats r1 which has no valid exposures and is skipped, all other cameras are reduced, "1/30 failed" is reported.
   - tests on 20211203 (to check if the partials work as expected based on BADCAMWORD)
     - steps: submit nightlybias_alt_64.sl
        set desired number of zeros to a high number, but for z3 only 25 exp will be usable and the logic for partial obs is needed. This uses 48 exp for all but z3 and 25 for z3  (2 of the 25 bad zeros where dropped for all as they were the first 2 of the night).
     - steps: submit nightlybias_alt_25.sl
        same as above but requesting to use 25 only, as that amount is found for all cams, the partials are ignored.
      - In both cases all 30 bias frames are generated.

akremin (approved):
Thank you for addressing the changes and the quick turn around. Everything here looks great. 

I tested the bias code on two additional nights: 
1.  20210503 with BADCAMWORD="r6z8" set for a subset of zeros
2. 20211109 with BADAMPS="b5B;b5D" for all zeros.

The logs indicate that it did the right thing in each case and the output files were also as expected. In case (1) the two cameras were missing for biases (leaving behind biasnighttest's instead). In case (2) the one camera was missing for biases, but that is because the dark was also missing and the submitted cameras didn't include b5. I.e. that is a feature rather than a bug, but worth noting for the future.

I have not done an independent test of the dark model code, but will trust Michael's tests. That is not part of the normal processing, so we can patch any subtleties with that later on in the unlikely event that we need to.

PR #1572, opened 2022-01-05
don't flag a tile as archived if archiving failed

sbailey:
This PR fixes a bug in desi_archive_tilenight, where previously it would set tiles-specstatus.ecsv ARCHIVEDATE for a tile as being archived, even if the actual act of archiving it had failed.  The example that triggered this was TILEID=21743 on LASTNIGHT=20211216 which we had initially flagged as QA=good, but then retroactively decided to discard it.  When I ran `desi_archive_tilenight -n 20211216` it tried to archive 21743, found that the tiles/cumulative/21743/20211216 directory was missing but flagged it as ARCHIVEDATE=20220105 anyway and updated tiles-specstatus.ecsv.

Now the code tracks errors while moving / updating permissions on directories, and it will only set ARCHIVEDATE if the actual operations succeeded, and it also prints a CRITICAL log message at the end if any tiles failed so that the individual errors don't get lost in screenfuls of archiving messages (as I had missed originally).

I tested this by creating a munged tiles-specstatus.ecsv file with an invalid tile to archive, and then running `desi_archive_tilenight --dry-run --specstatus ...` on that.

@akremin please check.

### DIFF
--- a/bin/desi_archive_tilenight
+++ b/bin/desi_archive_tilenight
@@ -26,16 +26,19 @@ def archivetile(tiledir, archivedir, dryrun=False):
 
     Options:
         dryrun: if True, print messages but don't move directories.
+
+    Returns int error code, non-zero if there was a problem
     """
     log = get_logger()
     if os.path.isdir(archivedir):
         log.info(f'{archivedir} already exists')
-        return
+        return 0
 
     if not os.path.isdir(tiledir):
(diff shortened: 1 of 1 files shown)


akremin (approved):
The logic and implementation both look good. I did not independently test this as the setup is quite involved. I'll trust the test outlined in the PR writeup.

PR #1574, opened 2022-01-07
Overscan rows

julienguy:
Subtract the mean of the overscan cols to the overscan rows (instead of using the overscan pixels of reach overscan row).
This allows us to fix the bias offset in r4-B due to contamination of the overscan columns by the leftover charges from the active pixels cause by the charge trap.

One can see the exponential tail in the overscan region in the following raw image:

Screenshot from 2022-01-07 10-00-21

Test this PR at NERSC with `export DESI_SPECTRO_CALIB=/global/cfs/cdirs/desi/users/jguy/teststand/desi_spectro_calib`.
because I  add to `sm1/sm1-r.yaml` for the corresponding config.
```
  # also use background subtraction using overscan rows because of this
  USE_ORSEC: true
```
Profile in vertical band to the right (after fiber 499) of the active region of the r4 CCD for exposure 116770 (LED 1s exposure).
before this modification (step of ~-1elec):
preproc-r4-00116770-daily

after (residual positive step of ~0.25 elec):
preproc-r4-00116770-new


julienguy:
my bad ... this PR contains an other one, I restart it.

PR #1578, opened 2022-01-10
tile/night QA pernight

sbailey:
This PR adds a new `--group` option to `desi_tile_qa` and `desi_night_qa` to support "pernight" tile-based redshifts in addition to the default "cumulative" redshifts.  This required updates to `desispec.io.findfile`, and updating tile/night QA to use that.  An example output is at:

https://data.desi.lbl.gov/desi/users/sjbailey/spectro/redux/humidity/nightqa/20211220/nightqa-20211220.html

Generated with
```
desi_tile_qa -n 20211220 -g pernight
desi_night_qa -n 20211220 -g pernight
```
Note: the test prod didn't include darks or zmtl thus the dark plots and n(z) per petal are missing.

A few other changes:
  * Fix exposure-qa bug that was failing to write a "EFFTIME" header keyword if all fibers were bad; now it writes EFFTIME=0.0.
  * Add another even higher-level try/except wrapper around tile QA plotting
  * Add a two-pass approach to tile-QA plotting.
    * This is primarily motivated by cases where the tile-qa*.fits file was generated but not the tile-qa*.png file.  Currently resubmitting the desi_tile_qa command (or the redshifts batch script) will skip making the plot again because the fits file already exists; now it will skip over the fits file and re-try on making the png file.
    * I realize that `--recompute` would also regenerate the fits and png files, but it is convenient to be able to re-run exactly the same pipeline script/command without having to add additional options.
    * Implementation note: I added a `_wrap_make_tile_qa_plot` function to take care of the try/except/logging in a consistent manner when calling `make_tile_qa_plot` from two different places.  The main benefit of leaving the first call at the time of creating the tile-qa*.fits file is that the plots are also made in parallel if using `--nproc`.  If you think that is too messy, we could move the "does the fits file already exist" into the `func` call and just try every tile every time, or stop making the png in `func` and only make it in the second pass.
  * fixes minor night-qa bug when running on the first night of a prod when there aren't any previous nights to lookup previous badcolumns counts.

@araichoor please review

### DIFF
--- a/bin/desi_night_qa
+++ b/bin/desi_night_qa
@@ -37,6 +37,8 @@ def parse(options=None):
                         help = "Path to input reduction, e.g. /global/cfs/cdirs/desi/spectro/redux/blanc/,  or simply prod version, like blanc, but requires env. variable DESI_SPECTRO_REDUX. Default is $DESI_SPECTRO_REDUX/$SPECPROD.")
     parser.add_argument("-n","--night", type = int, default = None, required = True,
                         help = "night to process. ex: 20211128")
+    parser.add_argument("-g","--group", type = str, default = "cumulative", required = False,
+                        help = 'tile group "cumulative" or "pernight"')
     parser.add_argument("-o", "--outdir", type = str, default = None, required = False,
                         help = "Path to ouput folder, default is the input prod directory. Files written in {prod}/nightqa/{night}; several files will be created there")
     parser.add_argument("--css", type = str, default = None, required = False,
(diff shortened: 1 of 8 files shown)


araichoor:
I just had a look: I have no comment, nothing obvious came to my mind; you've obviously thought much more thoroughly than me on how to organize this change!

Only a cosmetic suggestion: changing `--group` to `--groupname`, as in `meta.findfile()`, could make things a bit nicer, with not have `group` and `groupname` here and there in the code; but that really is a detail!

sbailey:
Thanks for the check.  I agree that the group/groupname mismatch is distasteful, but given that I'm running out of time for launching f5 I think I will leave it as-is for now.  `desi_tile_redshifts` also uses `--group`, so if/when we make them consistent I'm inclined to change `findfile` to use "group" instead of the scripts to use `--groupname`, partially because the "name" part of "groupname" doesn't seem to carry any additional information.

PR #1579, opened 2022-01-10
badcolumns with noisy input data (b5 May 2021)

### LINKED ISSUE #1567: b5 large number of "bad" columns
In desihub/specex#60 @julienguy reports:

> There are 206 badcolumns for b5 in /global/cfs/cdirs/desi/spectro/redux/f4/calibnight/20210509/badcolumns-b5-20210509.csv.
This is the cause of the failures.

Something is wrong if we are triggering that many "bad" columns.  Check all badcolumns files in the f4 prod to look for outliers and trends, and investigate whether algorithmic updates are needed to make the bad column identification more robust to high readnoise etc.  Other cases might be genuinely bad data in the dark (or ZEROs used for the nightly bias) that should be flagged as bad.

sbailey:
This PR fixes #1567 (large number of b5 bad columns) by adding a 6-sigma significance cut on identified bad columns, which dramatically reduces the number of false positive bad columns on noisy b5 during May 2021.  Other changes include adding a SIGMA column to the output file and using the same pixel mask as the extractions (which ignores the BADREADNOISE bit).

Example outputs are in:
```
/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/badcol/calibnight/20210507
/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/badcol/calibnight/20210509
```
for comparison with badcolumns*.csv in /global/cfs/cdirs/desi/spectro/redux/f4/calibnight/20210507 and 20210509.

For non-b5 CCDs, all originally identified bad columns are preserved on both nights.

For b5 20210507 464 -> 0 and 20210509 206 -> 1.  I suspect that that single identified b5 column on 0509 is still a false positive, but that seems acceptable given how noisy the CCD is (~10 e- readnoise) and not wanting to crank up the significance cut more and risk false negatives.

I'd like to fast track a review+merge to include with the f5 test prod.  @julienguy please review if you can.

### DIFF
--- a/bin/desi_inspect_dark
+++ b/bin/desi_inspect_dark
@@ -162,7 +162,7 @@ if not args.nopsf : # add fiber info
         camera=head["camera"]
         t["CAMERA"]=np.repeat(camera,len(t))
 
-        odir=os.path.dirname(args.badfiber_table)
+        odir=os.path.dirname(os.path.abspath(args.badfiber_table))
         if not os.path.isdir(odir):
             os.makedirs(odir)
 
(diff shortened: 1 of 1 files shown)


julienguy:
Tested on several nights and cameras.

Issue #1580, opened 2022-01-10
Overestimated background between fiber blocks caused by bright stars


julienguy:
As predicted, the estimate of the background between fiber blocks is sensitive to bright stars.
The worst configuration occurred in 20220107/00117268 camera b7, for a backup exposure,
with 2 bright stars on fibers 324 and 325 (+3500) , biasing the background estimate in the space
between the two fibers, and leading to negative sky fiber residuals.

![preproc-b7-00117268](https://user-images.githubusercontent.com/5192160/148812046-02cc6872-2eba-4fc8-91d7-d1d7947db18c.png)
(blue is with `desi_preproc` option `--bkgsub-for-science`, orange without)
(see motivation for this option in PR #1551 )

sbailey:
desi_preproc uses the non-traceshifted default PSF model from DESI_SPECTRO_CALIB to identify the locations between the bundles, which makes it maximally sensitive to any shifts causing the inter-bundle boxcar extraction to get closer to the fibers.  Using a traceshifted PSF would be ideal.  However, chicken-and-egg: desi_compute_trace_shifts uses a preproc image as input, but here we'd want a traceshifted PSF to use as input to preproc...  Using the nightly PSF is close to ideal (matching the shifted PSF to ~0.02 pixels).

For this case, the shifted PSF differs from the default PSF by ~1.5 pixels, compared to a ~15.8 pixel separation between fibers 324 and 325 (+3500), so I doubt that using a shifted PSF would by itself be sufficient to avoid the wings from the bright stars, but it could be part of the solution.

sbailey:
fixed by #1581.

PR #1581, opened 2022-01-10
Estimate contribution from scattered light in the regions between CCD blocks

julienguy:
PR to fix issue #1580 .

- convolve preproc image with lorentzian profile along cross-dispersion axis as a proxy for scattered light / psf tails. 
- compute value of convolved image in interblock regions
- flag affected interblocks and replace value by average per amp before computing CCD background

Results:
 - flagged 5 blocks of 20220107/00117268 camera b7
 - no blocks of  20211215/00114225 camera b9 (a dark frame) are masked (this is one of the exposures used for the original algorithm development) 

Central band profile of preprocessed image 20220107/00117268 camera b7 before (blue) and after (orange) this fix.

preproc-b7-00117268-new

zoom of same figure in region with largest effect

preproc-b7-00117268-new2

### DIFF
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -201,8 +201,6 @@ def _savgol_clipped(data, window=15, polyorder=5, niter=0, threshold=3.):
     Returns:
 
     """
-    print("Window: {}".format(window))
-
     ### 1st estimation
     array = data.copy()
     fitted = signal.savgol_filter(array, window, polyorder)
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks for the quick fix.  I added docstrings, logging and keywords for min/max of the background per amp, and verified that the extra calculation doesn't cause memory or runtime problems when run in parallel as part of the pipeline.  Merging now.

PR #1582, opened 2022-01-10
Update preproc file creation to fix FITS validation errors

weaverba137:
This PR addresses fitsverify errors identified in desihub/desidatamodel#87.  In particular preproc files now have headers constructed more carefully.  This should hopefully propagate to other files downstream.  Caveats:

* We are still propagating empty header cards as empty header cards, so there will still be warnings about those.
* The checksum errors in the `MASK` HDU need astropy/v5.0 to be fixed.
* *Comments* on header keywords in the `FIBERMAP` HDU are dropped.

Additional side effect: the `fill_header` keyword in `desispec.io.raw.read_raw()` is now promoted to a 'real' keyword input rather than something read from `**kwargs`.

Hopefully the planned `f5`, pre-fuji test will exercise this PR more fully.


py/desispec/io/raw.py
```diff
@@ -46,105 +69,130 @@ def read_raw(filename, camera, fibermapfile=None, **kwargs):
...
         fibermap = desispec.io.read_fibermap(fibermapfile)
     else:
         log.warning('creating blank fibermap')
         fibermap = desispec.io.empty_fibermap(5000)
 
     #- Add image header keywords inherited from raw data to fibermap too
-    desispec.io.util.addkeys(fibermap.meta, img.meta)
+    # BAW: This is unnecssary, because a fibermap file constructed by
+    # assemble_fibermap() will already have the raw data keywords.
+    # desispec.io.util.addkeys(fibermap.meta, img.meta)
```
sbailey, line 195: I think this is necessary for DARK, ZERO, ARC, FLAT exposures that don't have fibermaps from assemble_fibermap.  `desispec.io.util.addkeys` will not overwrite a pre-existing keys (i.e. it only adds new ones) so I think this is harmless to call.  If that causes incorrect keywords to be added, those should be added to skipkeys or stdkeys in desispec.io.util.addkeys.  Or otherwise we need to catch the case where read_raw used `empty_fibermap` and needs to propagate keywords from the raw data.

More generally, we want the raw data headers propagated to the fibermap HDU, regardless of whether that comes via assemble_fibermap -> fibermap file, or whether that comes via empty_fibermap + adding raw data headers.

sbailey (changes requested):
Looks good, except for one inline request regarding propagating raw data keywords into the fibermap HDU of non-science observations that don't have assemble_fibermap generated fibermaps (e.g. DARK, ZERO, ARC, FLAT).

py/desispec/io/raw.py
weaverba137: OK, that's reasonable, but I want to make sure that there are only two possible channels:  *either* `assemble_fibermap()` *or* `empty_fibermap()`.  That is there is no other possible way to add a fibermap HDU.

sbailey: I believe those are the only two ways that a fibermap gets created.  But we should preserve the ability to create an empty_fibermap that isn't tied to a pre-existing raw data file (e.g. for simulations and testing), i.e. header propagation into empty_fibermap could be an option, but shouldn't be required.  Or otherwise it's fine with me if empty_fibermap doesn't pretend to add raw data headers and io.raw.read_raw adds them when needed.

weaverba137: OK, that's what I've got in the latest version.  Please take a look again.

sbailey (approved):

sbailey:
Looks good, thanks, merging now.

sbailey:
Post-facto commentary:

The fibermap created by assemble_fibermap applies to all 10 spectrographs and inherits the HDU 1 extname=SPEC header keywords, but not the per-camera keywords.

When the fibermap get split per-spectrograph and added to the per-camera preproc files as part of preproc+readraw, we also need to propagate the per-camera header keywords that are in the raw data camera HDUs but not the fibermap file.

This was broken in this PR by limiting
```python
desispec.io.util.addkeys(fibermap.meta, img.meta, skipkeys=('BZERO', 'BSCALE'))
```
to only occur when the fibermap was generated with empty_fibermap, but not when it was read from a fibermap file.  I had forgotten that this was the route by which the per-camera keywords come in.  Without those, downstream code breaks e.g. because the FIBERMAP HDU doesn't have the SPECGRPH keyword anymore.

I fixed this in commit 0c7aa720, tested on pre-f5 vetting, and verified that the output preproc files don't have any errors.

weaverba137:
I think we need an inheritance diagram for headers. Right now it is downright obscure how and why headers are passed along the pipeline.

PR #1584, opened 2022-01-11
nightqa: improved 5min DARK expid selection

araichoor:
This PR improves the identification of the 5min DARK exposure used for the `dark-NIGHT.pdf` file in the `nightqa-NIGHT.html` page, with using a method similar to what is done by the pipeline.
Previous approach was just grabbing the first encountered 5min DARK exposure, which was not correct in case that exposure was identified as problematic during the night, and a second exposure has been taken after (the pipeline will then discard the first exposure).

`nightqa.get_dark_night_expid()` now looks for the `processing_table_PRODSHORT-NIGHT.csv` file and search for the exposure with: OBSTYPE=="dark" and OBJDESC=="ccdcalib".

In case there is no processing table or if no exposure is found, None is returned.
In case two or more exposures are identified, same, None is returned: @akremin is that the correct thing to do? Or shall I just pick one of the found exposures in that case? and which?

Lastly:
- I've assumed that, if the full path to a prod is `prod` (e.g. `/global/cfs/cdirs/desi/spectro/redux/daily`, then `os.path.basename(prod)` is the correct thing to grab the prod name (e.g. `daily`);
- I've taken advantage of this PR to correct the docstring for the prod argument, and add a safety call to `os.path.normpath(args.prod)` in `desi_night_qa` (because a trailing `/` in `args.prod` would make `os.path.basename(prod)` return an empty string).


akremin:
As of now there should never be two entries, so I think returning None is as good as any other choice. It would be an ambiguous state. This leaves it missing, which will prompt humans investigate. I think that's reasonable.

It is possible that in the future we have to run multiple darks/biases per night (as David floated yesterday after a hardware state-change). But that will require new development to display both, rather than selecting one or another.

akremin (approved):
Good catch on the trailing "/" problem. The code looks good. If you've tested it and you're happy then I'll merge this into master. Please comment to let me know that you've made all the changes you wanted to.

araichoor:
I'm happy to merge the current version.
Thanks for the precision for the case where two exposures would be identified, we could deal with that when that happens.

For what is worth, I've just re-checked the code for Jan. 2022 nights, looks fine.

```
>>> from desispec.night_qa import get_dark_night_expid
>>> d = Table.read("/global/cfs/cdirs/desi/spectro/redux/daily/exposures-daily.fits", "EXPOSURES")
>>> nights = np.unique(d["NIGHT"][d["NIGHT"] >= 20220101])
>>> expids = [get_dark_night_expid(night, "/global/cfs/cdirs/desi/spectro/redux/daily") for night in nights]
>>> for night, expid in zip(nights, expids):
...     print(night, expid)
... 
20220101 None
20220102 None
20220103 None
20220104 116731
20220105 None
20220106 117013
20220107 117231
20220108 117339
20220109 117589
```

PR #1585, opened 2022-01-11
tile/night QA pernight v2

araichoor:
This PR solves a small bug due to changes in PR https://github.com/desihub/desispec/pull/1578, which introduced the need to specify the `groupname` argument for the `findfile()` call for the tileqa and tileqapng cases.
For the change in `desi_tile_vi`, I've hard-coded `groupname="cumulative"`, as I think we'll only use the cumulative reduction to do the tile vi.

Also, I'm wondering if there would be other places where that update would need to be done.
With looking for "tileqa" and "findfile" in the github desispec repository, maybe those places:

https://github.com/desihub/desispec/blob/fbfcf33867ad774e9627fc38546577db7105c9e2/py/desispec/zmtl.py#L837
https://github.com/desihub/desispec/blob/55cd93510339f077ae3c9667b4ffa0bc396a66b0/py/desispec/test/test_io.py#L801-L803

and maybe here, I don't know:
https://github.com/desihub/desispec/blob/6ed41bff73e0d4eb1addacd4b168fa60b2073921/py/desispec/scripts/tile_redshifts.py#L436

Unfortunately, I'm not familiar with those codes, so someone else input would be needed.

Example of a call raising the error without this PR:
```
desi_tile_vi --prod daily --new -i tiles-specstatus.ecsv --user raichoor --viewer display
…
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/bin/desi_tile_vi", line 296, in <module>
    main()
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/bin/desi_tile_vi", line 224, in main
    qaplot_filename = findfile("tileqapng",night=night,tile=tileid,specprod_dir=args.qa_dir)
  File "/global/common/software/desi/cori/desiconda/20200801-1.4.0-spec/code/desispec/master/py/desispec/io/meta.py", line 242, in findfile
    raise ValueError("Required input '{0}' is not set for type '{1}'!".format(i,filetype))
ValueError: Required input 'groupname' is not set for type 'tileqapng'!```

### DIFF
--- a/bin/desi_tile_vi
+++ b/bin/desi_tile_vi
@@ -221,14 +221,14 @@ def main():
         log.info("Inspecting TILE={} NIGHT={} ...".format(tileid,night))
 
         # find qa plot ...
-        qaplot_filename = findfile("tileqapng",night=night,tile=tileid,specprod_dir=args.qa_dir)
+        qaplot_filename = findfile("tileqapng",night=night,tile=tileid,specprod_dir=args.qa_dir,groupname="cumulative")
         if not os.path.isfile(qaplot_filename) :
             log.error("Missing {}, did you run desi_tile_qa?".format(qaplot_filename))
             continue
 
         software_answer = None
         # find qa fits ...
-        qafits_filename = findfile("tileqa",night=night,tile=tileid,specprod_dir=args.qa_dir)
+        qafits_filename = findfile("tileqa",night=night,tile=tileid,specprod_dir=args.qa_dir,groupname="cumulative")
         if os.path.isfile(qaplot_filename) :
             qahead = fitsio.read_header(qafits_filename,"FIBERQA")
             if "VALID" in qahead :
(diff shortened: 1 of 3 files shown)


sbailey:
Thanks Anand.  It's good to specify groupname='cumulative' in desi_tile_vi because that is the only case that we intend to visually inspect, though the underlying problem here was that I had missed tileqapng in the list of findfile filetypes that are supposed to default to groupname='cumulative'.  I have corrected that and added tests to confirm it, as well as keeping your original updates.

Other items you noted:
  * zmtl: since groupname='cumulative' by default for zmtl, this is ok (and the `-in` and `-out` options can be used to read/write from anywhere)
  * test_io: this test was purposefully not specifying groupname because it was checking that it defaults to 'cumulative'.  My error was missing tileqapng in the list to test (now corrected).
  * tile_redshifts: this is constructing a bash batch script, where the filename is being put together by bash variables.  That involves some non-ideal replication of findfile logic, but I'm going to leave that as is since it is currently working (it's not completely incorrect, but it is more fragile to maintain).

PR #1587, opened 2022-01-12
Fibermap fix jan22

sybenzvi:
Two updates needed for using non-empty fibermaps in qproc:
1. Pass fibermap to `read_raw` to prevent `preproc` from creating an empty default fibermap.
2. Select fibermap rows corresponding to one camera to prevent `qproc_boxcar_extraction` from running past the end of the input traceset.

These changes address #1315. I tried not to touch qproc to fix the issue but the changes are necessary.


sbailey:
This looks fine to me, but for a cross check could you provide a cut-and-pasteable test case that fails (or does the incorrect thing) with current master and works with this branch?

sybenzvi:
> This looks fine to me, but for a cross check could you provide a cut-and-pasteable test case that fails (or does the incorrect thing) with current master and works with this branch?

Yes, I'll add a unit test that demonstrates that _not_ including the fibermap just reproduces the behavior in the main branch.

sbailey:
I don't want to turn down the offer for adding unit tests, but I was asking for something even more minimal than that: what nightwatch or qproc command should someone run and how can they tell whether the right fibermap is being used, even if done as a one-off test of the PR?

sybenzvi:
OK, in the course of writing instructions on how to test this, I discovered a problem with the branch, so I'm converting the PR to a draft until I fix it.

Meanwhile, here are instructions for a one-off test at NERSC. This default mode of creating an empty fibermap in preproc and using that empty fibermap in qproc is tested by running:

`desi_qproc -i /global/project/projectdirs/desi/spectro/data/20220111/00118039/desi-00118039.fits.fz --camera z1 --auto`


To test using a fibermap, create one and use it as follows:

```
assemble_fibermap -n 20220111 -e 118039 -o fibermap-00118039.fits
desi_qproc -i /global/project/projectdirs/desi/spectro/data/20220111/00118039/desi-00118039.fits.fz --fibermap fibermap-00118039.fits --camera z1 --auto
```

Then check the FIBERMAP HDU in the preproc and qproc output FITS files using `desispec.io.read_fibermap`. Here is a script `check_fibermap.py` to do that:

```
from argparse import ArgumentParser
from desispec.io import read_fibermap

p = ArgumentParser(description='Read and dump fibermap to stdout')
p.add_argument('ifiles', nargs='+', help='Input fits file with a FIBERMAP HDU')

args = p.parse_args()

for ifile in args.ifiles:
    print(f'\n{ifile}')
    try:
        fibermap = read_fibermap(ifile)
        print(fibermap)
    except Exception as e:
        print(e)
```

Please let me know if you want this documented in a README or if this comment is good enough.

sbailey:
This comment is good enough; no need for adding README.  Thanks for specific example.

sybenzvi:
I'm converting this PR back from draft mode. The issue was I noticed differences in the Nightwatch QA and wanted to track down the origin. I confirmed the PR does not affect any of the qframe or qcframe output. Only qsky is different, because if a good fibermap with SKY fibers isn't available, qproc sorts the fibers by flux and uses the 50% faintest ones to compute the background.

PR #1588, opened 2022-01-12
update specstatus --lastnight and --all options

sbailey:
This PR adds desi_update_tiles_specstatus --lastnight and --all options to 
1. filter only to tiles with a given LASTNIGHT
2. update all tiles matching that filter even if that LASTNIGHT isn't newer than spectstatus
  (by default, only new tiles and tiles with newer LASTNIGHT are updated)

e.g. if 20211215 was reprocessed and we want to update tiles-specstatus.ecsv with the new entries:
```
desi_update_tiles_specstatus --lastnight 20211215 --all
```
Note that QA-related columns are never updated; this only applies to things like EFFTIME_SPEC.

Because we are building up a bunch of cases for what is supposed to be updated when, I also refactored `update_specstatus` out of the script and added unit tests for a bunch of cases.


py/desispec/specstatus.py
```diff
@@ -0,0 +1,88 @@
...
+    #- but ok if specstatus has TILEID not in tiles
+    assert np.all(np.isin(tiles['TILEID'], specstatus['TILEID']))
+
+    #- For rows with more recent LASTNIGHT (new data), update non-QA columns.
+    #- Note: there is probably a more efficient way of doing this in bulk,
+    #- but let's favor obvious over clever unless efficiency is needed
+    num_updatedtiles = 0
+    for i, tileid in enumerate(tiles['TILEID']):
+        j = np.where(specstatus['TILEID'] == tileid)[0][0]
+        if update_all or tiles['LASTNIGHT'][i] > specstatus['LASTNIGHT'][j]:
```
sbailey, line 74: FYI, this is the part that changed to add the "update_all" option.  Otherwise the function is the same after copying out of the script and into a module function.

sbailey:
And while I'm updating this script, I also added a `--tileids` option to update just a list of specific TILEIDs (instead of or in addition to `--lastnight` filtering).  I renamed `--tiles` to `--tilesfile` to hopefully minimize confusion (that option is for specifying a non-default tiles file input filename, not for specifying a list of TILEIDs).

sbailey:
And one more update: in `desi_archive_tilenight`, missing preproc and exposure directories are ok if they are EFFTIME_SPEC=0.0 in the exposures file.  For daily that's ok because we want to signal to afternoon planning that the exposure was bad even if we didn't process it.

PR #1589, opened 2022-01-13
Bug fix in fiberflat_vs_humidity

julienguy:
This short PR fixes one bug in the fiberflat_vs_humidity correction when
the best fit humidity is on the edge of the humidity range (of the templates).


sbailey:
Thanks.  I confirm that this fixes the crash on 20220112 exp 118339 with very low humidity.
```
desi_proc -n 20220112 -e 118339 --cameras a12 --batch
```
I'll submit a companion PR that does better error tracking in desi_proc so that it will fail faster instead of causing a job hang.

PR #1590, opened 2022-01-13
better desi_proc error tracking

sbailey:
This PR improves desi_proc error tracking:
  * it keeps a running count of errors, and if any of the substeps fail desi_proc exits with an error code at the end, avoiding jobs and logs claiming "SUCCESS"
  * add missing inputs checks to sframe, and wrap with try/except for good measure (this step doesn't use `runcmd` which does that for most other steps)
  * add inputs list for desi_select_calib_stars runcmd (i.e. if an input is missing, it won't run in the first place, instead of running and crashing due to the missing input)

This is a companion PR to #1589 addressing the fiberflat humidity failures on 20220112.  That PR fixed the underlying algorithm (bug when current night has humidity outside the range of the models); this PR fixes the jobs hangs so that they would have failed faster and more obviously instead of waiting for the job to timeout.

Tested with
```
desi_proc -n 20220112 -e 118339 --cameras a12 --batch
```

Note that this branch does *not* have the algorithm fixes from PR #1589 so it is supposed to fail, just faster than before.

### DIFF
--- a/py/desispec/scripts/proc.py
+++ b/py/desispec/scripts/proc.py
@@ -125,6 +125,7 @@ def main(args=None, comm=None):
 
     log = get_logger()
     start_time = time.time()
+    error_count = 0
 
     start_mpi_connect = time.time()
     if comm is not None:
@@ -309,7 +310,9 @@ def main(args=None, comm=None):
             if args.badamps is not None:
                 cmd += ' --badamps={}'.format(args.badamps)
 
-            runcmd(cmd, inputs=[], outputs=[fibermap, tilepix])
+            err = runcmd(cmd, inputs=[], outputs=[fibermap, tilepix])
+            if err != 0:
+                error_count += 1
 
         fibermap_ok = os.path.exists(fibermap)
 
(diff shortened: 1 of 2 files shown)


py/desispec/scripts/proc.py
```diff
@@ -979,22 +1006,43 @@ def main(args=None, comm=None):
...
+                        subtract_sky(frame, sky, apply_throughput_correction=True)
+                        frame.meta['IN_SKY'] = shorten_filename(skyfile)
+                        frame.meta['FIBERFLT'] = shorten_filename(fiberflatfile)
+                        desispec.io.write_frame(sframefile, frame)
+                    except Exception as err:
+                        import traceback
+                        lines = traceback.format_exception(*sys.exc_info())
+                        log.error(f"Camera {camera} sframe raised an exception:")
+                        print("".join(lines))
+                        log.warning(f'Continuing without {sframefile}')
```
akremin, line 1044: Are we sure we want to continue if the file isn't written out?

akremin (commented):
Everything looks good. I only had one comment on failing early vs. continuing on. You're always thoughtful about those choices, so I just wanted to here what they were in this case.

py/desispec/scripts/proc.py
sbailey: Since the sframe file isn't used as input to any downstream step in desi_proc, it isn't necessarily fatal to those steps and they could proceed.  At the same time, error_count is incremented so that the job at the end will register a failure and downstream *jobs* will be blocked.  Even if we did want to stop hard, we still have to let the process continue so that this failing rank can catch up with others that may have succeeded.  Otherwise we're in the situation that prompted this PR: one rank failed and stopped, while the others continued, and then at some downstream MPI barrier/bcast/gather the job got stuck waiting for the failed rank to catch up.

This is similar to the other steps wrapped by runcmd: if it fails, that is logged but the script proceeds and other downstream steps might fail due to missing inputs, but it still lets the N>>1 MPI ranks continue and get to the end of the script and let that decide the ultimate success/fail of the job.

Arguably this warning is a bit to casual about continuing, but it is proceeded by actual error messages.

akremin: Thank you that makes complete sense.

PR #1594, opened 2022-01-14
limit humidity model range considered

sbailey:
This PR fixes a bug on 20220113 exp 118634 and 118635 where the measured humidity (5.5%) was well outside the fiberflat model humidity range (10-45%) and the chi2 scan wasn't able to sample the real minimum and was picking the wrong model extreme:
Screen Shot 2022-01-14 at 10 58 19 AM

The fix is to just limit the considered humidities to be within 20% of the measured humidity.  We can refine that in the future (as well as expand the model as we get more data under very low and high humidity conditions).

Test commands that previously crash and now work:
```
desi_compute_humidity_corrected_fiberflat --use-sky-fibers -i /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220113/00118634/frame-b1-00118634.fits --fiberflat /global/cfs/cdirs/desi/spectro/redux/daily/calibnight/20220113/fiberflatnight-b1-20220113.fits -o blat.fits

desi_compute_humidity_corrected_fiberflat --use-sky-fibers -i /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220113/00118635/frame-b1-00118635.fits --fiberflat /global/cfs/cdirs/desi/spectro/redux/daily/calibnight/20220113/fiberflatnight-b1-20220113.fits -o foo.fits
```

### DIFF
--- a/py/desispec/fiberflat_vs_humidity.py
+++ b/py/desispec/fiberflat_vs_humidity.py
@@ -125,7 +125,6 @@ def _fit_flat(wavelength,flux,ivar,fibers,mean_fiberflat_vs_humidity,humidity_ar
 
     log.info("best fit humidity = {:.2f}".format(best_humidity))
 
-
     return flat , best_humidity
 
 def compute_humidity_corrected_fiberflat(calib_fiberflat, mean_fiberflat_vs_humidity , humidity_array, current_humidity, frame) :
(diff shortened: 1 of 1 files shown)


akremin (approved):
Simple and effective solution. Changes look good, thank you.

PR #1596, opened 2022-01-16
Nightqa: per-program skyzfiber plot

araichoor:
This PR adds three panels to the `NIGHT-skyzfiber.png` plot, one for each of the Main BACKUP/BRIGHT/DARK program (per suggestion of @julienguy).

I identify the tile `FAFLAVOR` with looking for a `fiberassign-TILEID.fits*` file in the `$DESI_ROOT/spectro/data/NIGHT/*/` folders; if no `fiberassign-TILEID.fits*` file is there, or if `FAFLAVOR` is not in the header, I just don't include that tile in the newly added panels.

I've also changed the y-axis (redshifts) to a log scale, as in the `tile-qa-TILEID-thruNIGHT.png` plots.

Example of a new plot:
skyzfiber-20220111


julienguy:
Tested. It's good, thanks.

Issue #1599, opened 2022-01-18
desi_tile_qa MINTFRAC failure


sbailey:
```
source /global/cfs/cdirs/desi/software/desi_environment.sh 22.1a
module swap desispec/0.50.0
export DESI_SPECTRO_REDUX=/global/cfs/cdirs/desi/spectro/redux
export SPECPROD=f5

desi_tile_qa --group pernight -n 20210508 -t 80872 --outdir .
...
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/0.50.0/bin/desi_tile_qa", line 267, in <module>
    main()
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/0.50.0/bin/desi_tile_qa", line 196, in main
    entry = func(**func_arg)
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/0.50.0/bin/desi_tile_qa", line 62, in func
    fiberqa_table , petalqa_table = compute_tile_qa(night,tileid,specprod_dir,exposure_qa_dir,group=group)
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/0.50.0/lib/python3.9/site-packages/desispec/tile_qa.py", line 219, in compute_tile_qa
    minimal_efftime = qa_params["tile_qa"]["fiber_rel_mintfrac"]*exposure_qa_meta["MINTFRAC"]*exposure_qa_meta["GOALTIME"]
KeyError: 'MINTFRAC'
```

I'm guessing that this error applies to all tiles of that era, before MINTFRAC was added to fiberassign files.  @araichoor please check.

araichoor:
Yes, sure... basically, lots of those tile_qa routines are irrelevant for such tiles.
What do we want to do here? 
I've given a try to a small fix:
- GOALTIME: if missing, set GOALTIME=1000/150/30 for dark,cmx/bright/backup, using the TARG header keyword from the fiberassign-TILEID.fits.gz file; if no fiberassign file or no TARG keyword or no correct TARG value, raise an error;
- MINTFRAC: if missing, set MINTFRAC=0.9.
Those are ~nominal values, see here: https://github.com/desihub/fiberassign/blob/c9ef922dae336860959ae8cf1d2c07b1b8eb99a3/bin/fba_launch#L40-L66

Would that make sense? That allows the `tile-qa-TILEID*fits` file to be generated.
[...]

PR #1603, opened 2022-01-19
Tile qa specialtile

araichoor:
This PR addresses the issue #1599, where the tile_qa routines crash for special/early tiles with some missing header keywords.

The workaround is to set MINTFRAC and GOALTIME to ~nominal values (https://github.com/desihub/fiberassign/blob/c9ef922dae336860959ae8cf1d2c07b1b8eb99a3/bin/fba_launch#L40-L66):
- GOALTIME: if missing, set GOALTIME=1000/150/30 for dark,cmx/bright/backup, using the TARG header keyword from the fiberassign-TILEID.fits.gz file; if no fiberassign file or no TARG keyword or no correct TARG value, raise an error;
- MINTFRAC: if missing, set MINTFRAC=0.9.

Along the way, this PR has two additional "fixes" in `tile_qa_plot.py` for this issue case:
- handle cases when SURVEY, FAPRGRM, or EBVFAC is not present in the header (with setting them to "none", "none", -1, respectively);
- if no exposures are found in the `tiles/cumulative` folders, then look for the `tiles/pernight` folders.

For the issue #1599 case, this PR produces this plot:
tile-qa-80872-20210508

### DIFF
--- a/py/desispec/tile_qa.py
+++ b/py/desispec/tile_qa.py
@@ -80,6 +80,33 @@ def compute_tile_qa(night, tileid, specprod_dir, exposure_qa_dir=None, group='cu
             exposure_fiberqa_table , exposure_petalqa_table = read_exposure_qa(filename)
         if exposure_qa_meta is None :
             exposure_qa_meta = exposure_fiberqa_table.meta
+            # AR case no GOALTIME, MINTFRAC (can happen for early tiles):
+            # - set GOALTIME=1000/150/30 for dark,cmx/bright/backup
+            # - set MINTFRAC=0.9
+            if "GOALTIME" not in exposure_qa_meta:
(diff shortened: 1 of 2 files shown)


araichoor:
Note: in this commit https://github.com/desihub/desispec/pull/1603/commits/1e358b8f2175ae89e0b2e09fad120b1586d20df4, I've put some `raise Error` if some information is not found.
I realize that it may be against the spirit of "avoid crashes for special tiles".
If it would be better to default to some nominal values for those cases, please tell me, I can modify the code for that.

sbailey:
Looks good, thanks.  I think the `raise ...` are ok since it only impacts a single special tile and doesn't block other tiles from having their QA.  If these are problematic we can address it in a future PR.

Issue #1604, opened 2022-01-20
coadded spectra when missing one camera


julienguy:
Quasar has valid blue and red camera spectra in 'spectra' file but not in coadded file.

```
plot_spectra -i $DESI_SPECTRO_REDUX/f5/tiles/pernight/139/20210410/spectra-5-139-20210410.fits -t 39627945043561608 --rebin 4 --redrock $DESI_SPECTRO_REDUX/everest/tiles/pernight/139/20210410/redrock-5-139-20210410.fits

plot_spectra -i $DESI_SPECTRO_REDUX/f5/tiles/pernight/139/20210410/coadd-5-139-20210410.fits -t 39627945043561608
```
It is fully justified to have the NIR part of the spectra flagged as bad because the spectrum is affected by known defects in CCD z5, but the good B and R data should be in the coadd. In this example, we should have been able to fit its redshift. 

![Screenshot from 2022-01-20 10-16-02](https://user-images.githubusercontent.com/5192160/150397736-cf081267-3d88-46fb-95a7-4dd464968515.png)

sbailey:
Fixed in PR #1605.

PR #1605, opened 2022-01-20
Bad column fibermask

julienguy:
- Use fibermask bits BADAMPB, BADAMPR, BADAMPZ to flag fibers affected by bad CCD columns instead of using the fibermask bit BADCOLUMN that does not specify the CCD.
-  This avoid having coaddition masked whole spectra when only one camera is affected. 
- Code tested on tile 273 SP5. One Lya quasar that was fully masked in f5 run is now successfully fitted with redrock.

spec3-bis

This addresses issue #1604 .

### DIFF
--- a/py/desispec/badcolumn.py
+++ b/py/desispec/badcolumn.py
@@ -80,14 +80,14 @@ def compute_badcolumn_specmask(frame,xyset,badcolumns_table,threshold_value=0.00
 
     return mask
 
-def compute_badcolumn_fibermask(frame_mask,camera_arm,threshold_specfrac=0.4) :
+def compute_badcolumn_fibermask(frame_mask,camera_arm,threshold_specfrac=0.6) :
     """
     fills a mask for fibers affected by a bad CCD column.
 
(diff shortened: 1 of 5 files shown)


sbailey:
I confirm that for this tile petal, the only targets with significant changes were previously COADD_FIBERSTATUS=BADCOLUMN and are now COADD_FIBERSTATUS=BADAMPR or BADAMPZ, with 1/3 of the spectrum masked instead of all of the spectrum masked, at least making it possible for redrock to fit a redshift.

However, I think we should consider also setting a redrock ZWARN bit (perhaps LITTLE_COVERAGE?) for these cases, since they are fundamentally different from normal 3-camera data in that there are large ranges of redshifts for which we couldn't correctly fit an ELG.  This is similar to how we propagate fiberstatus POORPOSITION and BADPOSITION into zwarn POORDATA and NODATA instead of requiring end-users to inspect both masks.  That's a change on the redrock side, to be coordinated with this PR.

PR #1608, opened 2022-01-21
desi_run_night enhancements and bug fixes

akremin:
This is a potpourri pull request with several minor improvements to the `desi_run_night` and underlying `submit_night()` function, along with a bug fix for the recently implemented `nightlybias` code.

#### Improvements:
1. Avoid submitting `nightlybias` unless either 'bias' or 'dark' is in `proc_obstypes`.  This is a bug fix for both `desi_run_night` and the nightly `desi_daily_proc_manager`.
2. Update the redshift log so that if a tile is observed more than once on a night with other tiles in between, the nightly and cumulative redshifts are only submitted after the final set of observations. `perexp` redshifts are still submitted as they arise though.
3. I also added a check to verify that the 300s dark exposure used for the `ccdcalib` is actually a "calib dark".
4. I've changed the logic such that new "undithered" dither exposures will be `LASTSTEP='skysub'` instead of `fluxcal`. This doesn't change historic table values, but any new exposures will be given the new `LASTSTEP`.
5. Added option to set the `LASTSTEP`'s that we want to run for the given night. The default is now `['all']` (i.e. only exposures that should be fully processed). This is a change from the previous implementation where everything was submitted for processing.
6. Updated the table handling logic so that the unprocessed exposures table is accurate.

#### Testing
I did targeted testing on each of these changes and verified that they do the expected thing.

### DIFF
--- a/py/desispec/scripts/submit_night.py
+++ b/py/desispec/scripts/submit_night.py
@@ -281,7 +281,11 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
                 tile_exps = etable['EXPID'][etable['TILEID'] == lasttile]
                 unprocd_exps = [exp not in ptable_expids for exp in tile_exps]
                 if np.any(unprocd_exps):
+                    print(f"Identified that tile {lasttile} has future exposures"
+                          + f" for this night. Not submitting full night "
+                          + f"redshift jobs.")
                     if 'perexp' in z_submit_types:
+                        print("Still submitting perexp redshifts")
                         cur_z_submit_types = ['perexp']
                     else:
                         cur_z_submit_types = None
--- a/py/desispec/scripts/submit_night.py
+++ b/py/desispec/scripts/submit_night.py
@@ -232,7 +232,8 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
     ## Loop over new exposures and process them as relevant to that type
     tableng = len(ptable)
 
-    if tableng == 0 and np.sum(isdark) == 0:
+    do_bias = ('bias' in proc_obstypes or 'dark' in proc_obstypes)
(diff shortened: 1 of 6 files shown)


py/desispec/scripts/submit_night.py
```diff
@@ -143,16 +156,17 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
...
         log.info(f'Filtering by tiles={tiles}')
         if etable is not None:
             keep = np.isin(etable['TILEID'], tiles)
             etable = etable[keep]
-        if ptable is not None:
-            keep = np.isin(ptable['TILEID'], tiles)
-            ptable = ptable[keep]
+        #if ptable is not None:
+        #    keep = np.isin(ptable['TILEID'], tiles)
+        #    ptable = ptable[keep]
```
sbailey, line 169: Please confirm that this change (and similar one below) is purposeful and not a leftover from testing.  i.e. we used to filter ptable on tiles, now we don't filter on tiles, and that surprises me, but perhaps the equivalent functionality is happening somewhere else in the chain?

akremin: The survey and tile code was added by you. I don't see why we would filter previously submitted jobs based on input criteria for future jobs. It throws away information and I didn't understand what the benefit was. Other than being a record of what was done the only other thing the processing tables are used for is as a list of things that have already been processed (so that we don't re-process them again every time the code is run).

Do you happen to remember what your motivation was for doing that? Is there something that I am missing?

sbailey: OK, it sounds like this might have been a misunderstanding on my part about what needed to be filtered to form the reference set of what needed to be processed.  I verified that
```
desi_run_night -n 20210408 --tiles 191 --dry-run
```
does what is expected with only processing exposures related to tile 191, even if other tiles previously existed in the processing table.

PR #1609, opened 2022-01-21
zcatalog changes

Ragadeepika-Pucha:
Made the following major changes:
- Added ZWARN_NOT_ZERO to give priority to ZWARN = 0 spectra and then select primary spectra based on sort_column for cases where ZWARN != 0
- Added optional keyword to change the sorting column. default = TSNR2_LRG
- Made other minor changes

### DIFF
--- a/py/desispec/test/test_zcatalog.py
+++ b/py/desispec/test/test_zcatalog.py
@@ -1,0 +1,54 @@
+"""
+Test desispec.zcatalog
+"""
+
+import os
+import unittest
+
+import numpy as np
+from astropy.table import Table
+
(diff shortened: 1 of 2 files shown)


sbailey:
Thanks @Ragadeepika-Pucha .  I added some unit tests in py/desispec/test/test_zcatalog.py and didn't find any problems.  These will now run automatically with future PRs to make sure no one else accidentally breaks your code with a future update.  For general future knowledge, please take a look at how those are structured, and you can try running them yourself with python py/desispec/test/test_zcatalog.py .

PR #1627, opened 2022-01-26
if missing fiberassign-xxx.fits.gz file, try fiberassign-xxx.fits bef…

julienguy:
Minor PR for fuji (branched from fuji branch and to be remerged there).
In tile QA, if missing fiberassign-xxx.fits.gz file, try fiberassign-xxx.fits before raising an error.


sbailey:
I'm going to close this and reopen merging into fuji instead of master so that we don't risk bringing in unintended changes merging master -> fuji (I made same target branch mixup in my PR...)

PR #1628, opened 2022-01-26
catch fiberflat splinefit error if almost all input is masked

sbailey:
This PR fixes a fiberflat failure on 20210320/00081166/frame-z5-00081166.fits :

Apparently spline_fit returns a ValueError if all inputs are masked, but a TypeError if only a very small number are unmasked leading to an insufficient number of degrees of freedom for the fit.  The code was already catching the ValueError (everything masked) case; this PR also catches the TypeError (almost everything masked) case.

Example commands that previously failed and now succeed:
```
# from daily
desi_compute_fiberflat -i /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220123/00119849/frame-z5-00119849.fits -o fiberflat.fits
# from fuji
desi_compute_fiberflat -i /global/cfs/cdirs/desi/spectro/redux/fuji/exposures/20210320/00081166/frame-z5-00081166.fits -o fiberflat.fits
```

### DIFF
--- a/py/desispec/fiberflat.py
+++ b/py/desispec/fiberflat.py
@@ -97,7 +97,7 @@ def compute_fiberflat(frame, nsig_clipping=10., accuracy=5.e-4, minval=0.1, maxv
     wave = frame.wave.copy()  #- this will become part of output too
     ivar = frame.ivar.copy()
     flux = frame.flux.copy()
-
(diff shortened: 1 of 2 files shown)


sbailey:
I opened this to merge into the wrong branch.  Trying again...

PR #1630, opened 2022-01-26
if missing fiberassign-xxx.fits.gz file, try fiberassign-xxx.fits before raising an error

sbailey:
originally from @julienguy , but reopening as a PR to fuji instead of master:

Minor PR for fuji (branched from fuji branch and to be remerged there).
In tile QA, if missing fiberassign-xxx.fits.gz file, try fiberassign-xxx.fits before raising an error.


sbailey:
FYI `desispec.io.util.checkgzip` was intended for cases like this where the input may or may not be gzipped, but this is also fine.  I'll merge as soon as tests are done.

PR #1643, opened 2022-02-01
Ignore version names in comments in desi_calib_config

marcelo-alvarez:
Fixes a bug where a comment with a version name in a reference version prevented a new entry to be made based on it. Nothing else is changed.

@julienguy please review and merge.


julienguy (changes requested):
I think it should be line.find("#")<0 instead of <=0

marcelo-alvarez:
> I think it should be line.find("#")<0 instead of <=0

done

julienguy (approved):

PR #1660, opened 2022-02-05
add utilities for qso afterburners 

echaussidon:
No impact on the current spectroscopic pipeline but usefull for LSS catalog.


echaussidon:
Hi @sbailey, 

I will do all the update that your require and ask you a review in the next PR. 

I do it quickly so that the LSS code can run directly with the update..

Issue #1661, opened 2022-02-07
Four nights in Fuji and Guadalupe without biases or badcolumn files


akremin:
There are four nights in Fuji + Guadalupe that do not have enough "calib zeros". Since the biases and bad column identification are performed in the same script, the bias calculation exits and prevents the bad column identification from occurring either.

The darks are available for three of the four nights, but are not run in any of the nights because of the bias code exiting.

Additionally, we could consider reducing the minimum number of zeros from 20 to 18 to accommodate three of the nights. The fourth night (20210629) we will have to decide if non "calib zeros" could be used. Using non "calib zeros" is a potential concern because zeros from special tests or afternoon zeros run before hardware re-initialization may then be used.

All of these nights have already been processed in Fuji+Guadalupe, but vetting should be done and an assessment made as to whether we should rerun using biases from a smaller set of zeros, having create the bad column identifications, or both.

Nights in Fuji:

- **20210422** - 21 zeros taken but after discarding first 2 there weren’t enough to compute biases (19 is less than 20). There is a "calib dark" but code exits before it gets there.
- **20210331** - 20 zeros taken but after discarding first 2 there weren’t enough. There is a "calib dark" but code exits before it gets there.

Nights in Guadalupe:

- Night **20210519**: 18 zeros were taken, so not enough for biases.  No darks, so we can't produce badcolumns in any event.
- **20210629** has dark sequences and zeros associated with it but no "calib dark" or "calib zeros". We would use a “dark seq” dark thanks to a recent PR, but the bias code only uses "calib zeros" and exits before it gets to bad column identification.

julienguy:
I would lower the min threshold to N=15 zeros (while still using more when available).
the total noise rms is then increased by sqrt(1+1/15) /sqrt(1+1/20)) =  0.8% which is very small.

For 20210629, it would be great to be able to declare zeros as good for calib in the exposures table.

sbailey:
@julienguy thanks.  Could you also confirm that the remaining zeros on those nights are ok?  i.e. not getting a complete set of zeros might indicate some other problem with the data leading to canceling the sequence, i.e. it might be more than just a question of statistics.

> For 20210629, it would be great to be able to declare zeros as good for calib in the exposures table.

Unfortunately I think that requires a code update to allow those zeros, but we could do that (as an option to not change previous runs).

julienguy:
- zeros in 20210331 are ok
- zeros in 20210422 are ok
[...]

PR #1662, opened 2022-02-07
default min number of bias lowered to 15

julienguy:
Trivial PR: default min number of bias lowered to 15 instead of 20.
Related to issue #1661 .

### DIFF
--- a/py/desispec/ccdcalib.py
+++ b/py/desispec/ccdcalib.py
@@ -338,13 +338,17 @@ def compute_bias_file(rawfiles, outfile, camera, explistfile=None,
 
     log.info(f"done with {camera}")
 
-def _find_zeros(night,cameras,nzeros=25):
+def _find_zeros(night, cameras, nzeros=25, nskip=2, anyzeros=False):
     """Find all OBSTYPE=ZERO exposures on a given night
 
     Args:
         night (int): YEARMMDD night to search
         cameras (str): list of cameras to process
+
+    Options:
         nzeros (int): number of zeros desired from valid all-cam observations to not worry about partials
+        nskip (int): number of initial zeros to skip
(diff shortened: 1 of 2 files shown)


sbailey:
Thanks; I'll add the code to optionally use non-calib zeros, so please don't merge yet.

sbailey:
I added an option `desi_compute_nightlybias --anyzeros` to allow it to use any ZEROs taken during the night, not just those taken as part of the calibration ZEROs sequence (e.g. for night 20210629 without calib zeros, but with zeros taken as part of darks later).

I also replaced the hardcoded skip 2 ZEROs at the start of the night with an option `--skipzeros`, default 2.

tested with a 2 node interactive job and:
```
srun -n 15 -c 8 desi_compute_nightly_bias --mpi -n 20210629 --anyzeros
srun -n 15 -c 8 desi_compute_nightly_bias --mpi -n 20210331 --nskip 0
```
outputs in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/f9

Extra checks on this are welcome, but I'll merge+tag when needed to proceed with Fuji processing.

sbailey:
sphinx docs pass at NERSC, and are failing here due to
```
WARNING: failed to reach any of the inventories with the following issues:
[17](https://github.com/desihub/desispec/runs/5100582587?check_suite_focus=true#step:5:17)
intersphinx inventory 'https://docs.scipy.org/doc/scipy/reference/objects.inv' not readable due to ValueError: unknown or unsupported inventory version: ValueError('invalid inventory header: ')
```
let's debug that later and move on with Fuji.

weaverba137:
I think missing scipy docs are a known issue right now.  See *e.g.*, https://github.com/scipy/scipy/issues/15545.  I'd give it another day to stabilize.

Issue #1664, opened 2022-02-08
Fuji: desi_tile_qa crash when some exps missing some cameras


sbailey:
desi_tile_qa crashes when some of the input exposures are missing a camera but others have that camera.  e.g. tile 1 observed on 20210406 has all 30 cameras on exposures 83714 and 83715, but a z7 light staturation event on 83716 and so that frame was excluded from the processing.
```
source /global/homes/d/desi/bin/desi_fuji_env.sh
desi_tile_qa -n 20210406 -t 1 --group cumulative --outdir .
...
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/0.51.4/bin/desi_tile_qa", line 267, in <module>
    main()
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/0.51.4/bin/desi_tile_qa", line 196, in main
    entry = func(**func_arg)
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/0.51.4/bin/desi_tile_qa", line 62, in func
    fiberqa_table , petalqa_table = compute_tile_qa(night,tileid,specprod_dir,exposure_qa_dir,group=group)
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/0.51.4/lib/python3.9/site-packages/desispec/tile_qa.py", line 71, in compute_tile_qa
    exposure_fiberqa_table , exposure_petalqa_table = compute_exposure_qa(night, expid, specprod_dir)
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/0.51.4/lib/python3.9/site-packages/desispec/exposure_qa.py", line 309, in compute_exposure_qa
    scores = fitsio.read(cframe_filename,"SCORES")
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/conda/lib/python3.9/site-packages/fitsio/fitslib.py", line 139, in read
    with FITS(filename, **kwargs) as fits:
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/conda/lib/python3.9/site-packages/fitsio/fitslib.py", line 520, in __init__
    self._FITS = _fitsio_wrap.FITS(filename, self.intmode, create)
OSError: FITSIO status = 104: could not open the named file
failed to find or open the following file: (ffopen)
[...]

PR #1665, opened 2022-02-08
handle missing cframe in tsnr2 summing

araichoor:
This PR is a simple suggested fix for issue https://github.com/desihub/desispec/issues/1664.
I let @julienguy review.

My understanding is that for each fiber of a given petal, we get the TSNR2 with summing the values over the three cameras, i.e.: TSNR2 = TSNR2_ELG_B+TSNR2_ELG_R+TSNR2_ELG_Z (using ELG here)
https://github.com/desihub/desispec/blob/769b71295eed2cbc0de78fb2d5bdf94b215cb2f1/py/desispec/exposure_qa.py#L305-L312
(`entries` are the 500 indexes of the fibers for the considered petal).

This fix will just ignore a camera if the cframe file is missing.
In the issue case (z-camera missing for petal=7), we will have for petal=7: TSNR2 = TSNR2_ELG_B+TSNR2_ELG_R, and this will result in a lower efftime:
```
>>> d = Table.read("/global/cfs/cdirs/desi/users/raichoor/tmpdir/exposures/20210406/00083716/exposure-qa-00083716.fits","FIBERQA")
>>> for p in range(10): print("{}\t{:.0f}s".format(p, d["EFFTIME_SPEC"][d["PETAL_LOC"] == p].mean()))
... 
0	698s
1	790s
2	839s
3	690s
4	855s
5	707s
6	963s
7	148s
8	663s
9	723s
```
Is that what we want?

### DIFF
--- a/py/desispec/exposure_qa.py
+++ b/py/desispec/exposure_qa.py
@@ -306,6 +306,9 @@ def compute_exposure_qa(night, expid, specprod_dir):
         for band in ["B","R","Z"] :
              camera="{}{}".format(band.lower(),spectro)
              cframe_filename=findfile('cframe',night,expid,camera,specprod_dir=specprod_dir)
+             if not os.path.isfile(cframe_filename):
+                 log.warning("missing {} => using {}_{}=0".format(cframe_filename, tsnr2_for_efftime_key, band))
+                 continue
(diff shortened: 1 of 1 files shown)


julienguy:
yes, thanks. I am merging.

PR #1668, opened 2022-02-09
Refactor healpix redshifts for Fuji

sbailey:
This PR refactors the healpix-based redshift workflow.  Previously it required desi_group_spectra to be run ahead of time in a monolithic many-frames to many-healpix remapping before proceeding with the healpix redshift jobs.  This created a bottleneck in the processing and became problematic if we needed to reprocess individual healpix (e.g. due to adding a new frame).

Now the healpix-redshift jobs are similar to the tile-redshift jobs, with each one owning their call to desi_group_spectra.  `desi_healpix_redshifts` is analogous to `desi_tile_redshifts`, generating and submitting batch jobs.  This requires an input exposures table with meaningful SURVEY and FAPRGRM entries, or otherwise it derives these from the input prod exposures_tables/ (which still have a lot of 'unknown' entries to be fixed in the future...)

I tested this in a prod /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/zpix, including testing that cumulative redshifts didn't break.

@moustakas or others, if you are available, could you inspect outputs under zpix/healpix/... as a cross check?


sbailey:
Known issues, so far:
  * I completely broke test_pixgroup, because I completely changed pix grouping...
  * SURVEY and FAPRGRM are missing from headers
  * HPXNSIDE missing from headers

I'll work on those, but other cross checks would still be appreciated...

Issue #1657, opened 2022-02-04
Better handle spatial variation in sky spectrum


schlafly:
I'm playing around with modeling the variation in the sky within a petal, with an eye to solving issues like
https://github.com/desihub/desisurveyops/issues/4
The approach that I'm currently taking is the following:
- Do a PCA analysis of the sky in each camera/petal.  Take top 4 PCA components, for example, for r0.
- Given a particular exposure with r0, fit the sky as mean(sky) + sum_i (a_i * PC[i] * (x - mean(x)) + b_i * PC[i] * (y-mean(y)))

Here are the PCs for r0:
![image](https://user-images.githubusercontent.com/1417841/152451642-66069bf8-89ef-41b3-a22f-f007090af0bd.png)
I use the top 4 PCs based on the spectrum of the singular values; there stops being rapid decrease in significance beyond the 4th PC.  PC1 is mostly continuum and is presumably what the moon looks like.  PC2 & PC3 are both liney things that have different behavior in the telluric absorption area in r?  And PC4 is really just controlling the amplitude of some OI lines and is probably not important for us?

So I can go and take those and fit them, for example, to exposure 20211117/00109231, which is one of the tiles in limbo due to large sky residuals.  Only looking at P0, removing just a mean compared with removing a mean + fitting gradients times the first four PCs looks like this:
![image](https://user-images.githubusercontent.com/1417841/152472543-b14c9d17-3b65-4771-a340-d9fd8db299d6.png)
where sigma has dropped by a factor of 2.  That's not great---I was hoping for a factor of 4 or 5---but it's effectively a factor of 2 in sky brightness which is enough to take these tiles from being very bad to being fairly typical, so it's enough, I hope?

I don't see much remaining spatial structure in the residuals
![image](https://user-images.githubusercontent.com/1417841/152472944-d78c3732-2289-415c-9d38-fc89871040fd.png)
so I don't see a clear approach for further improvement.  Maybe the sky touch up from the throughput corrections would fix up some of the remaining residuals.

I think this is a fairly simple and effective approach.  The current approach is to fit ~2000 numbers from the sky spectra---one number per wavelength bin---per petal and camera.  I'm proposing adding an additional 8 numbers---a slope in x & y for each of 4 PCs.  I wouldn't be surprised if I could get away with only fitting one PC, or with forcing the direction of the 4 gradients to be the same, so that I were only fitting 2 or 5 numbers instead.  But I don't see a strong motivation to be more parsimonious here.

[...]

PR #1669, opened 2022-02-09
Fit gradients in sky brightness with a few PCA eigenspectra

schlafly:
This PR reorganizes the main sky computation routine a bit and adds the ability to fit gradients in the sky.  It is a work in progress, and has some ~placeholder code surrounding the I/O needed to get the sky gradient PCA in place in desi_compute_sky that will need to be replaced.  But I expect the core algorithmic elements to be finished.

The sky gradient fitting is motivated by https://github.com/desihub/desispec/issues/1657 where large sky gradients lead to sky residuals that prevent accurate redshift determination.  The proposed fix is simple.
- Derive a few high SNR PCA eigenspectra that are basically the spectra of moonlight close to the moon from a library of all past DESI exposures.  Presently 2 components, though one would be adequate.
- Save these to DESI_SPECTRO_CALIB.
- When fitting the sky, fit two additional number per PCA eigenspectra giving the slope in the ampltude of these terms in the observed sky spectra.  
- Apply to all target spectra according to their location in the focal plane.

In some ways, this is a very "conservative" PR.  We presently fit ~2000 numbers from the sky spectra for each petal & camera: the amplitude of the sky at each wavelength.  This PR increases that number by ~4.  On the other hand, it does make the sky fitting potentially less robust if all of the skies are in a corner, or if for some reason the skies all fall along a line, so one direction of the gradient isn't probed by the sky spectra.

That said, this works pretty well at least for the single example exposure I have run.  night/expid 20211117/00109231 had large sky residuals, and this PR reduces them by a factor of about two.
image
The left panels are the sframe files, the right panels are the median residual sframe fluxes as a function of X & Y, the top panels are the original reduction, and the bottom panels are the reduction from this PR.

There are clearly some residuals remaining.  I believe these are issues in the fiberflat.  
image
But I'm not sure, and I haven't thought hard about how this interacts with the sky ampltiude touch ups based on the fits to the individual lines.

What are the next steps?
- I would like advice on how best to incorporate this into the pipeline.  I have not tried to introduce a new skygradpca.py file with the standard data structures, nor associated desispec.io I/O routines, nor tried to figure out the pipeline flow into desi_compute_sky to get the right arguments & filenames passed.
- I guess I should run this on some large set of images.  I think I personally would be convinced that it's right after running a large number of skies and comparing new & old skies and residuals---I wouldn't need to run all the way through redshifts---but others should tell me what makes the most sense.  There isn't a meaningful difference in compute time, so this is something like 10 s / camera, 5 min / exposure, so I can knock out a lot of these in a single interactive session.

Other concerns and things to note:
- I have moved the sector offset computation from an extra add-on to the "main" linear least squares fit.  I think this is an improvement in terms of the accuracy of the fits.  I haven't thought very hard about how it interacts with the existing LSF & wavelength solution touch-ups, though; those used to get a "pure sky" spectrum in the second iteration of the fit, while presently they do not in my current implementation.  I just haven't tried to understand the LSF & wavelength solution touch ups to decide how to best work there.
- I don't know how the sectors are specified.  I don't have anything checking for degenerate sectors that will fail to invert.  I considered adding a very weak zero prior to the linear least squares fitter so that things always invert.  But I decided against it just due to not wanting to change things.
- I presently always do the gradient fits.  One could imagine only doing the gradient fits if SKYCHI2PIX were large, for example.
- My implementation looks to be ~10% slower than the old one, taking something like 11 s vs. 10 s for desi_compute_sky.  This is mostly because I didn't want to figure out how to do the A^T C^-1 A computation Julien was doing with the added rows for the PCA components and sectors, so I instead explicitly computed A.  I don't think that desi_compute_sky is a major expense, though, so I haven't tried to do better here.
- My implementation is definitely more memory hungry than the old implementation, since I take all the frame & resolution data structures, repackage them as lists of indices and values for incoration into the sparse matrix A, and then make the sparse matrix A---these end up getting copied three times.  I don't know if memory usage during the sky computation step is especially important.
- I don't try to check for weird distributions of skies on the petal and how that might compromise attempts to compute gradients.
That's all the concerns I can think of.

Is this roughly what we want?


julienguy:
- Eddie tells me on slack his refactor of the algorithm gives identical results. I would be good to document that here.
- It is indeed statistically better to have the sectors and the PCA in the same linear system to solve, but I can imagine situation where it is more fragile to a nasty cosmic ray or bad column. So, for the sector part, it would also be nice to report on a run on a few exposures just to show the stability of the output (this branch compared to the main one).

schlafly:
Here was my reproduction check:
- check out desispec/master
- run
> desi_compute_sky -i $DESI_SPECTRO_REDUX/daily/exposures/20220210/00122046/frame-r0-00122046.fits --fiberflat $DESI_SPECTRO_REDUX/daily/calibnight/20220210/fiberflatnight-r0-20220210.fits  -o skyorig-r0-00122046.fits
- check out my version of code
- run
> desi_compute_sky -i $DESI_SPECTRO_REDUX/daily/exposures/20220210/00122046/frame-r0-00122046.fits --fiberflat $DESI_SPECTRO_REDUX/daily/calibnight/20220210/fiberflatnight-r0-20220210.fits  -o skynew-r0-00122046-nosector-nopca.fits
> skynew = fits.getdata('skynew-r0-00122046-nosector-nopca.fits')
> skyorig = fits.getdata('skyorig-r0-00122046.fits')
> np.all(skynew == skyorig)
- observe that this returns True.

But revisiting this, I see that my recomputation of the sky isn't identical to DESI_SPECTRO_REDUX/daily/exposures/20220210/00122046/sky-r0-00122046.fits.  I presume I just need more arguments to desi_compute_sky to reproduce the daily product invocation---what's the correct invocation that brings in all the features?  I think I need the right pca-corr file?

julienguy:
```
grep compute_sky  $DESI_SPECTRO_REDUX/daily/run/scripts/night/20220210/prestdstar-20220210-00122046*.log  | grep RUNNING | grep r0

INFO:util.py:75:runcmd: RUNNING: desi_compute_sky -i /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220210/00122046/frame-r0-00122046.fits --fiberflat /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220210/00122046/fiberflatexp-r0-00122046.fits -o /global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220210/00122046/sky-r0-00122046.fits --adjust-wavelength --adjust-lsf --pca-corr /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm4/skycorr-pca-sm4-r0.fits --fit-offsets
```

schlafly:
Here's some plots of new and "old" skies and their differences compared to the uncertainties on the frames after flat fielding.
https://data.desi.lbl.gov/desi/users/schlafly/sky/skycomparison-new.pdf
I really should not have picked 20211226 for this, given the calib issues on this night---but maybe that's good for this robustness test.

Sectors are enabled for new processings of that data, but in the real daily product sectors were not fit.  So the "old" skies are my reductions of the frames with the old code.

In all but one case, differences are <0.05 sigma per pixel.  In one case, differences are ~0.1 sigma per pixel everywhere.  In this case, the new processing is better by 0.1% in chi^2 over the all the sky spectra, so I guess that's okay.  It doesn't look like cosmics or anything are causing real issues here.

I also checked that using the full invocations Julien pointed out to me, I exactly reproduce the existing daily skies when not using sectors.

```
>>> skynew = fits.getdata('skynew-r0-00122046-nosector-nopca.fits')
>>> skyorigdaily = fits.getdata('/global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220210/00122046/sky-r0-00122046.fits')
>>> np.all(skynew == skyorigdaily)
True
```

schlafly:
Is this enough for robustness tests or should I do a bunch of nights?

schlafly:
Here are comparison plots of sky residuals for all of the exposures touching tiles listed in https://github.com/desihub/desisurveyops/issues/4.
https://data.desi.lbl.gov/desi/users/schlafly/sky/skygradexps/skycompplots.pdf
And an example case:
image
Left colum are the sky spectra (rows: spectra, columns: wavelength bin), right column has the spatial variation of the median fluxes.  Rows are the full sky spectra, the spectra after removing the daily sky, and the spectra after removing the sky as computed in this PR.  The moonlight induced sky gradients are removed by this PR, though there's still clearly more structure than one would like in the sky residuals, which I'm partially attributing to limitations of the flat field.

julienguy:
This looks great. It seems the only missing pieces are 
 - replace `skygradpcafn = os.path.join(os.environ['SKYGRADPCA_DIR'],f'skygradpca-{camera}.fits')` by a mechanism similar to the pcacorr (save in `$DESI_SPECTRO_CALIB`, use keywords in yaml files, ideally add file `desispec.io.skygradpca` with a function `read_skygradpca` files).
 - add option in `desi_proc` to not use the option (I understand we want it 'on' by default).
 - add to this PR the script that generates the skygradpca files.
 - run one night of data with the main pipeline down to redshifts to validate the results.

julienguy:
Summarizing here a discussion with Eddie. The code is good and it's doing what it's been written for: correct for the moonlight or twilight component using PCA spectra multiplied by DX and DY in the focal plane. This does remove the obvious gradient but leaves systematic residuals at a level of 1-2% of the sky that are problematic for some tiles with a very bright sky background. We think those systematic effects could be corrected with a sky flat correction that we could or not include in the skycorr components. This will be the topic of a separate PR.

julienguy (approved):

Issue #1673, opened 2022-02-10
Incorrect specmask.BADFIBER bit set


julienguy:
For frames like `/global/cfs/cdirs/desi/spectro/redux/guadalupe/exposures/20210707/00097851/cframe-b1-00097851.fits` (and same for `z1`) , the specmask bit BADFIBER is set for fibers 250 to 500. This is a bug because there is only the fibermask bit BADAMPR that is set, and this bit signals an issue in the camera r, not b nor z.

julienguy:
The bug affects 86 exposures, all in guadalupe
```
1 prod=guadalupe night=20210704 expid=97428 with badamps=r1d
2 prod=guadalupe night=20210704 expid=97429 with badamps=r1d
3 prod=guadalupe night=20210704 expid=97430 with badamps=r1d
4 prod=guadalupe night=20210704 expid=97431 with badamps=r1d
5 prod=guadalupe night=20210704 expid=97432 with badamps=r1d
6 prod=guadalupe night=20210704 expid=97433 with badamps=r1d
7 prod=guadalupe night=20210705 expid=97537 with badamps=r1d
8 prod=guadalupe night=20210705 expid=97538 with badamps=r1d
9 prod=guadalupe night=20210705 expid=97539 with badamps=r1d
10 prod=guadalupe night=20210705 expid=97541 with badamps=r1d
11 prod=guadalupe night=20210705 expid=97542 with badamps=r1d
12 prod=guadalupe night=20210705 expid=97543 with badamps=r1d
13 prod=guadalupe night=20210705 expid=97544 with badamps=r1d
14 prod=guadalupe night=20210705 expid=97545 with badamps=r1d
15 prod=guadalupe night=20210705 expid=97546 with badamps=r1d
16 prod=guadalupe night=20210705 expid=97547 with badamps=r1d
17 prod=guadalupe night=20210705 expid=97548 with badamps=r1d
18 prod=guadalupe night=20210705 expid=97549 with badamps=r1d
19 prod=guadalupe night=20210705 expid=97550 with badamps=r1d
20 prod=guadalupe night=20210705 expid=97551 with badamps=r1d
21 prod=guadalupe night=20210705 expid=97552 with badamps=r1d
22 prod=guadalupe night=20210705 expid=97553 with badamps=r1d
23 prod=guadalupe night=20210705 expid=97554 with badamps=r1d
24 prod=guadalupe night=20210705 expid=97555 with badamps=r1d
[...]

PR #1674, opened 2022-02-10
specmask.BADFIBER

julienguy:
This PR fixes a bug/feature in the `fiberbitmask` functions, where the `specmask.BADFIBER` bit was set in the `frame.mask` whenever a `fibermask.BADAMP{B,R,Z}` bit was set in the `fibermap['FIBERSTATUS']`,  irrespectively of the frame camera.

As an example, with the change in this PR, the quasar spectra from SP1 in tile 1299 are recovered.
See https://data.desi.lbl.gov/desi/spectro/redux/jguy/tiles/pernight/1299/20210707/tile-qa-1299-20210707.png
compared to https://data.desi.lbl.gov/desi/spectro/redux/guadalupe/tiles/pernight/1299/20210707/tile-qa-1299-20210707.png

This PR fixes issue #1673

### DIFF
--- a/py/desispec/fiberbitmasking.py
+++ b/py/desispec/fiberbitmasking.py
@@ -79,7 +79,9 @@ def get_fiberbitmasked_frame_arrays(frame,bitmask=None,ivar_framemask=True,retur
         if bitmask.isnumeric():
             bad = np.int32(bitmask)
         else:
-            bad = get_fiberbitmask_comparison_value(kind=bitmask)
+            camera = frame.meta["CAMERA"].lower()
+            band   = camera[0]
+            bad    = get_fiberbitmask_comparison_value(kind=bitmask,band=band)
     else:
         bad = bitmask[0]
         for bit in bitmask[1:]:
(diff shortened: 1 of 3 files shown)


py/desispec/fiberbitmasking.py
```diff
@@ -100,46 +105,48 @@ def get_fiberbitmasked_frame_arrays(frame,bitmask=None,ivar_framemask=True,retur
...
+        return get_fluxcalib_fiberbitmask_val(band)
 
 
-def get_skysub_fiberbitmask_val():
-    return get_all_fiberbitmask_val()
+def get_skysub_fiberbitmask_val(band):
+    return get_all_fiberbitmask_with_amp(band)
 
-def get_flat_fiberbitmask_val():
+def get_flat_fiberbitmask_val(band):
```
akremin, line 139: It's unclear why we need to add a dummy parameter here, but it's okay for uniformity.

py/desispec/fiberbitmasking.py
```diff
@@ -100,46 +105,48 @@ def get_fiberbitmasked_frame_arrays(frame,bitmask=None,ivar_framemask=True,retur
...
-    return get_all_fiberbitmask_val()
+def get_fluxcalib_fiberbitmask_val(band):
+    return get_all_fiberbitmask_with_amp(band)
 
-def get_stdstars_fiberbitmask_val():
-    return get_all_fiberbitmask_val() | fmsk.POORPOSITION
+def get_stdstars_fiberbitmask_val(band):
+    # for standard stars, it's important to have blue camera
+    # in order to fit Balmer lines
+    return get_all_fiberbitmask_with_amp("b") | fmsk.POORPOSITION
```
akremin, line 149: Why would you not mask the actual `BADAMPS` for a camera if they are set? I.e. why are you ignoring `band` here? I can understand why blue is especially important, but those fibers will be masked in the blue camera as appropriate if we replace "b" with band.

We jointly fit exposures for stdstars, so this seems like it will throw away useful r and z information that would improve multi-exposure fits.

py/desispec/fiberflat_vs_humidity.py
```diff
@@ -159,7 +159,9 @@ def compute_humidity_corrected_fiberflat(calib_fiberflat, mean_fiberflat_vs_humi
     if frame is not None :
         log.info("using frame to fit for the best fit current humidity")
         ivar = frame.ivar*(frame.mask==0)
-        badfibermask = get_skysub_fiberbitmask_val()
+        camera = frame.meta["CAMERA"].lower()
```
akremin, line 162: Earlier you check for both frame.meta and that CAMERA is in frame.meta. Should you do that here as well?

py/desispec/fiberbitmasking.py
```diff
@@ -79,7 +79,12 @@ def get_fiberbitmasked_frame_arrays(frame,bitmask=None,ivar_framemask=True,retur
         if bitmask.isnumeric():
             bad = np.int32(bitmask)
         else:
-            bad = get_fiberbitmask_comparison_value(kind=bitmask)
+            band = 'b' # default, which is the most constraining (for stdstar fit)
```
akremin, line 82: Perhaps we could do band = 'brz' that masks if any amp is set. Otherwise you're default will miss actually flagged fibers if the true camera is r or z.
Note this suggestion breaks your later logic of only checking the first character of "band" in `get_all_fiberbitmask_with_amp(band)`

akremin (commented):
You can ignore these comments if you wish, but it naively looks to me like you are ignoring amplifier bits on the actual cameras for which they apply, which I'd like more explanation on.

py/desispec/fiberbitmasking.py
julienguy: I left it because in principle the flat_fiberbitmask should also be a function of the band.
It's just because we have decided to run the fiberflat even if BADAMP is set that this argument is not used for now.

py/desispec/fiberbitmasking.py
julienguy: Indeed this was a mistake.

py/desispec/fiberflat_vs_humidity.py
julienguy: yes

py/desispec/fiberbitmasking.py
julienguy: yes, I am replacing 'b' by 'brz' and I am changing the function get_all_fiberbitmask_with_amp to test all bands

py/desispec/fiberbitmasking.py
akremin: Thank you for the explanation

julienguy:
@akremin the code ready for second review.

I checked it's producing the expected results (good redshifts despite lack of r-camera data for some fibers)

```
cd /global/cfs/cdirs/desi/spectro/redux/jguy/tiles/pernight/1299/20210707
plot_spectra -i coadd-1-1299-20210707.fits --rebin 6 --redrock redrock-1-1299-20210707.fits -t 39628397629932704
```

Screenshot from 2022-02-10 16-20-57

akremin:
The updates addressed my comments. I made a doc string update to clarify that multiple bands are acceptable.

Issue #1649, opened 2022-02-02
Fuji: Inconsistencies between top-level exposures and tiles files


weaverba137:
While working with the `f5` test run, I was comparing these two top-level files:

* `$DESI_SPECTRO_REDUX/f5/exposures-f5.fits`
* `$DESI_SPECTRO_REDUX/f5/tiles-f5.csv`

I found the following inconsistencies:

1. Tiles 80605, 80606, 80608, 80609, 80610 have `SURVEY = 'sv1'` in `tiles-f5.csv`, but `SURVEY = 'cmx'` in `exposures-f5.fits`.
2. Tiles 80865, 80869 have `EFFTIME_GFA = (0, 0)` (respectively) in `tiles-f5.csv`, but `EFFTIME_GFA = (14890.1, 10798.3)` (respectively) in `exposures-f5.fits`.
3. Many tiles have non-zero `GOALTIME` in `tiles-f5.csv`, but zero `GOALTIME` in `exposures-f5.fits`.

sbailey:
Tiles 80605, 80606, 80608, 80609, 80610 are cases of faflavor=cmxelg or cmxlrgqso which despite the name are actually sv1 tiles.  i.e. the tiles file is correct.

I'm not sure about what causes the EFFTIME_GFA and GOALTIME discrepancies, but presumably the non-zero values are more correct (easy to say; finding it in the code is the harder part...)

weaverba137:
Good to know.  Do we know where the discrepancy in SURVEY arises, or will that be as hard as EFFTIME_GFA or GOALTIME?

sbailey:
as hard (somewhere in desi_tsnr_afterburner, which organically outgrew its original intent and resulted in a lot of if/then/else logic to work around our evolving understanding of what metadata we needed in fiberassign header keywords)

araichoor:
About the GFA_EFFTIME, if this helps:
- I notice that those two tiles have several EXPIDs, but _some_ have EFFTIME_GFA=0, because they are not in the GFA file (see below)
- I see those lines in what I think is writing the tiles-f5.fits file, where EFFTIME_ETC and EFFTIME_GFA are set to zero if one of the exposures has a zero value:
https://github.com/desihub/desispec/blob/26c627ca6bb1c075bb4edc8cacd3d87b6c1f8a26/py/desispec/tilecompleteness.py#L109-L110
```
                if k == "EFFTIME_ETC" or k == "EFFTIME_GFA" :
[...]

PR #1675, opened 2022-02-10
fix tsnr_afterburner exposures SURVEY

sbailey:
This PR fixes the tsnr_afterburner exposures vs. tiles SURVEY inconsistency reported in #1649.  Admittedly it does this by adding another special case check in the exposures logic rather than attempting a deeper refactor to reduce technical debt.

Other updates:

Adds a PROGRAM column, which is the output of desispec.io.meta.faflavor2program(FAFLAVOR), standardizing the many sv1 flavors/faprgrm into the standardized dark/bright/backup/other to match current usage and to match how the healpix redshifts are grouped.  For sv3 and main tiles, FAPRGRM==PROGRAM, but for sv1 tiles this should make interpreting the healpix structure easier.

It also fixes an MPI logic bug if size>nights (the same night was proceeded by multiple ranks, leading to race conditions), and renames `--multinode` to `--mpi` for consistency with other desispec scripts.  The MPI route still requires --recompute which I haven't tried to fix here.

Lastly, if --tile-completeness is a fits file, it also writes a .csv version, mirroring the behavior of --outfile for the exposures file which writes both a fits and a csv version.

This PR does not try to fix the GOALTIME inconsistency, nor the EFFTIME_ETC.

Example outputs are in /global/cfs/cdirs/desi/users/sjbailey/dev/fuji/exp.fits and tiles.fits, generated with
```
time desi_tsnr_afterburner -o exp.fits -t tiles.fits --aux /global/cfs/cdirs/desi/survey/observations/SV1/sv1-tiles.fits --gfa-proc-dir /global/cfs/cdirs/desi/survey/GFA/ --nproc 16
```
[45 minutes]
Note that I did *not* include --compute-skymags in that run

### DIFF
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -31,6 +31,7 @@ from   desispec.calibfinder import CalibFinder
 from   desispec.io import read_frame
 from   desispec.io import read_fibermap
 from   desispec.io import findfile
+from   desispec.io.meta import faflavor2program
 from   desispec.io.fluxcalibration import read_flux_calibration
 from   desiutil.log import get_logger
 from   desispec.tsnr import calc_tsnr2,tsnr2_to_efftime
@@ -79,7 +80,7 @@ def parse(options=None):
                         help = 'path to table with sky magnitudes. Expected keys=NIGHT,EXPID,SKY_MAG_G,SKY_MAG_R,SKY_MAG_Z')
     parser.add_argument('--compute-skymags', action='store_true',
                         help = 'recompute sky mags')
-    parser.add_argument('--multinode', action='store_true',
+    parser.add_argument('--mpi', action='store_true',
                         help = 'Paralleize TSNR recomputation across multiple nodes using mpi4py. Requires args.update to be False.')
     parser.add_argument('--add-badexp', action='store_true',
                         help = "Include entries for known bad exposures that weren't processed")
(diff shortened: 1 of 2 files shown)


akremin (approved):
The code looks good.

I compare the two `csv` versions of the example tables. Among the columns in common I see agreement on all tiles entries.

Columns compared for each tile: `['TILEID', 'TILERA', 'TILEDEC', 'SURVEY', 'FAPRGRM', 'FAFLAVOR', 'GOALTIME', 'GOALTYPE', 'MINTFRAC']`

Code used:
```
import numpy as np 
exps = Table.read("/global/cfs/cdirs/desi/users/sjbailey/dev/fuji/exp.csv")
tiles =  Table.read("/global/cfs/cdirs/desi/users/sjbailey/dev/fuji/tiles.csv")
overlap = [col for col in exps.colnames if col in tiles.colnames and 'EFFTIME' not in col and 'EXPTIME' not in col] 
for i,tile in enumerate(tiles['TILEID']): 
     itterexps = exps[exps['TILEID'] == tile] 
     ittertile = tiles[i] 
      
     for col in overlap: 
         eversion = np.unique(itterexps[col].data) 
         tversion = np.atleast_1d(ittertile[col]) 
         if not np.all(eversion == tversion): 
             print(f"{tile}\t{col}\t{eversion}\t{tversion}") 
```

The only things printed are related to `GOALTIME`, as expected since it is acknowledged in the PR summary that that issue is unresolved.

sbailey:
Post review update:
  * added EXTNAME=TILE_COMPLETENESs to tiles.fits table
  * added missing PROGRAM column to tiles table (it was in exposures, but not tiles)

Issue #1676, opened 2022-02-11
Fuji: bit fibermask.BADPOSITION not set for some fibers/exposures with missing positioning data


julienguy:
The FIBERSTATUS bit fibermask.BADPOSITION is not set for some fibers/exposures with missing positioning data.
Example:

```
from astropy.table import Table
fmap=Table.read(f"/global/cfs/cdirs/desi/spectro/redux/fuji/tiles/cumulative/80605/20210205/spectra-7-80605-thru20210205.fits","FIBERMAP")
tid=39627676691992850
print(fmap["FIBERSTATUS"][fmap["TARGETID"]==tid])
print(fmap["FIBER_X"][fmap["TARGETID"]==tid])
```

```
  0
  0
  0
  0
  0
  0
  0
  0
  0
  0
  0
  0
512
512
512
512
512
512
512
512
512
512
512
512

fmap["FIBER_X"][fmap["TARGETID"]==tid]
     0.0
     0.0
     0.0
     0.0
     0.0
     0.0
     0.0
     0.0
     0.0
     0.0
     0.0
     0.0
-188.342
-188.355
-189.839
-189.836
 -189.83
-189.833
-189.839
-189.837
-189.843
 -189.85
-189.839
-189.832


```

sbailey:
Argh.  Astropy 5.0 broke this through their Table "feature" that NaN entries are read as masked values instead of just keeping them as NaN.  Astropy 4.0.x used to just preserve them as NaN and we were checking for bad entries with `np.isnan`, but with astropy 5.0 that returns False instead of True because the NaNs are masked...

excerpts from desispec.io.fibermap.assemble_fibermap:
```
        pm = Table.read(coordfile, 'DATA')  #- PM = PlateMaker
        fibermap['DELTA_X'] = pm[f'DX_{numiter-1}']
        fibermap['DELTA_Y'] = pm[f'DY_{numiter-1}']
        dr = np.sqrt(fibermap['DELTA_X']**2 + fibermap['DELTA_Y']**2) * 1000
        badpos = (dr > bad_offset_um) | np.isnan(dr)
```

With astropy 4.0.x this identified input platemaker NaNs as badpos, but now fails to do that.

julienguy:
I scanned all Fuji spectra files, there are 19 affected tiles are:
```
80605
80606
80607
80608
80609
80610
80611
80612
80613
80614
80615
80616
80617
80618
80619
80620
80621
80622
80623
```

[...]

PR #1680, opened 2022-02-11
Fix assemble_fibermap handling of input NaN

sbailey:
This PR fixes #1676 where a change in astropy 5.0 Table.read handling of NaN inputs resulted in us not masking fibers with unknown positions (NaN in input platemaker coordinates file).  The previous logic only worked in astropy 4.0.x (unsure about other in the 4.x series).

This PR introduces `desispec.io.table.read_table(filename, ext)` which does what `astropy.table.Table.read(filename, ext)` used to do: NaNs are left as NaNs, blank strings are left as blank strings, and neither are converted into MaskedColumns.

Test with:
```
assemble_fibermap -n 20201214 -e 67710 -o fibermap.fits
```
and then compare the output to /global/cfs/cdirs/desi/spectro/redux/fuji/preproc/20201214/00067710/fibermap-00067710.fits

The FIBER_X/Y DELTA_X/Y values that used to be NaN are now set to 0.0 (like they were before) and have fiberstatus bit 9 BADPOSITION set.

Silver lining: we've still been using the old astropy 4.0.x desiconda installation for daily operations, so this bug hasn't been affecting daily ops.  And I don't think platemaker is giving us NaN anymore anyway...

PR #1681, opened 2022-02-12
Fix assemble_fibermap handling of input NaN, take 2

sbailey:
Replacement for PR #1680, this time I think properly branched off of fuji with just the minimal changes needed:

This PR fixes https://github.com/desihub/desispec/issues/1676 where a change in astropy 5.0 Table.read handling of NaN inputs resulted in us not masking fibers with unknown positions (NaN in input platemaker coordinates file). The previous logic only worked in astropy 4.0.x (unsure about other in the 4.x series).

This PR introduces desispec.io.table.read_table(filename, ext) which does what astropy.table.Table.read(filename, ext) used to do: NaNs are left as NaNs, blank strings are left as blank strings, and neither are converted into MaskedColumns.

Test with:

```
assemble_fibermap -n 20201214 -e 67710 -o fibermap.fits
```

and then compare the output to /global/cfs/cdirs/desi/spectro/redux/fuji/preproc/20201214/00067710/fibermap-00067710.fits

The FIBER_X/Y DELTA_X/Y values that used to be NaN are now set to 0.0 (like they were before) and have fiberstatus bit 9 BADPOSITION set.

Silver lining: we've still been using the old astropy 4.0.x desiconda installation for daily operations, so this bug hasn't been affecting daily ops. And I don't think platemaker is giving us NaN anymore anyway...

### DIFF
--- a/py/desispec/io/__init__.py
+++ b/py/desispec/io/__init__.py
@@ -39,6 +39,7 @@ from .exposure_tile_qa import (read_exposure_qa, write_exposure_qa, read_tile_qa
 from .raw import read_raw, write_raw
 from .sky import read_sky, write_sky
 from .skycorr import (read_skycorr, write_skycorr, read_skycorr_pca, write_skycorr_pca)
+from .table import read_table
 from .util import (header2wave, fitsheader, native_endian, makepath,
                    write_bintable, iterfiles, healpix_degrade_fixed,
                    healpix_subdirectory, replace_prefix)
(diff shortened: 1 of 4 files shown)


julienguy (commented):
Elegant solution to this astropy.table feature.
Tested on /global/cfs/cdirs/desi/spectro/data/20201214/00067710/coordinates-00067710.fits with NaN if FIBER_X.
I confirm the NaN are preserved with read_table(...) but not with Table.read(). I also checked the meta entries, and that the valid FIBER_X values were identical.

julienguy:
Indeed  option not recognized with astropy 4.0.1.post1

PR #1682, opened 2022-02-14
only use 120s flats for nightlyflat

### LINKED ISSUE #1671: Both nightly submission scripts do wrong thing with LED gain sequences
On 20210616 an LED Gain sequence was run that took two flat exposures at a number of exposure times from 1s to 200s. This broke the logic used to bundle together the calibration flats for the `nightlyflat`.

The current `desi_daily_proc_manager` and `desi_run_night` logic only checks that the sequence total is less than 5 and that the exposure time is over 100s. These were sufficient to avoid the old "short flat sequence" flats, but do not avoid the above gain sequence. 

The result of this is that the 100s and 200s exposures were bundled together with the 12 calibration flats and then the pipeline didn't know what to do with the 16 submitted heterogeneous flats. Thus no `nightlyflat` was produced for 20210616 in Guadalupe.

The solution is simple: we can check for PROGRAM name and/or require that the exposures be 120s.

sbailey:
This is a PR of a branch from @akremin to correct the fiberflat logic on nights that included gain sequences, e.g. 20210616 (fixes #1671).  I've confirmed via a desi_run_night --dry-run that it selects the correct flats on that night, and as a paranoia check that it also selects the same flats as before on a normal calibration night.

Note: post-fuji we can update how calibration exposures are selected (e.g. via a config file, initially populated using PROGRAM filters in addition to OBSTYPE) to avoid hardcoding that calibration flats have to be 120 sec, but this PR is a simple fix for the problem we currently have (we don't currently have cases where the nightly calib fiberflats weren't 120 sec, but we do have cases where other fiberflats had different exposure times confusing the pipelining logic).

### DIFF
--- a/py/desispec/scripts/daily_processing.py
+++ b/py/desispec/scripts/daily_processing.py
@@ -384,7 +384,8 @@ def daily_processing_manager(specprod=None, exp_table_path=None, proc_table_path
 
             ## Note: Assumption here on number of flats
             if curtype == 'flat' and calibjobs['nightlyflat'] is None \
-                    and int(erow['SEQTOT']) < 5 and float(erow['EXPTIME']) > 100.:
+                    and int(erow['SEQTOT']) < 5 \
+                    and np.abs(float(erow['EXPTIME'])-120.) < 1.:
                 flats.append(prow)
             elif curtype == 'arc' and calibjobs['psfnight'] is None:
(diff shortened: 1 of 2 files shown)


sbailey:
Ignoring coverage "failure" of 0.001% decrease for this PR (and other recent PRs).

PR #1696, opened 2022-02-18
Add 'group' feature to io.spectra.read_tile_spectra.

schlafly:
This PR adds a "group" argument to io.spectra.read_tile_spectra, so that one can do group='cumulative' to get the cumulative tile spectra.  I also needed to special case the file name for 'cumulative' spectra so that it looked for thruYYYYMMDD in the file name, rather than just YYYYMMDD.

I have not tried to be exhaustive about handling all of the different possibilities for groups.  But this will now read the "normal" cumulative groups again.  It's possible that the default should be group='cumulative'?


sbailey:
read_tile_spectra could use a rework to make better use of findfile and more flexibly support all the group options, but this fixes something that was previously broken, so I'll accept it as progress.  I'll also merge into the fuji branch, albeit not via a PR this time (like #1708 + #1714).

Issue #207, opened 2016-06-06
DESI_BASIS_TEMPLATES or mention the test data 


rainwoodman:
The integration test needs DESI_BASIS_TEMPLATES to run. It is not mentioned in desispec documentation. A google search suggests installing https://github.com/desihub/desisim-testdata may help -- but would be nice if this is actually mentioned in the README file or in the error message of desispec's integration test.

sbailey:
Agreed with the basic point.  For the record, desisim-testdata is sufficient for basic testing, but not for actually using desispec for real work, which requires a full copy of the templates, e.g. at NERSC `/project/projectdirs/desi/spectro/templates/basis_templates/v2.2` .  This is mentioned at https://desi.lbl.gov/trac/wiki/Pipeline/Install but the documentation in desisim itself should be more helpful about that.

rainwoodman:
What about adding a section in README with a link to https://desi.lbl.gov/trac/wiki/Pipeline/Install in each piece of the pipeline software?

weaverba137:
The documentation here addresses this problem: https://desi.lbl.gov/trac/wiki/Pipeline/Install#SetNecessaryEnvironmentVariablesandData
If this is insufficient, please open a new issue with specific concerns.

Issue #204, opened 2016-05-31
redmonster crash on zero fluxes


ngbusca:
@timahutchinson 

When running redmonster on the zdc2 bricks, I get a crashe when some of the spectra have zero flux. I've traced the problem to file 

redmonster/python/redmonster/physics/zfinder.py

when a flux is zero for a given spectrum, the fits are skipped (zwarning set to unplugged) but the lists  sn2_data, f_nulls and chi2_null aren't increased in size . In this case, for the following fiber, the code tries to index these lists past their size.

I've tried the following git diff, but then other problems show up further down (seemingly in zpicker2.py)

diff --git a/python/redmonster/physics/zfinder.py b/python/redmonster/physics/zfinder.py
index 314ccf8..aa2c581 100644
--- a/python/redmonster/physics/zfinder.py
+++ b/python/redmonster/physics/zfinder.py
@@ -133,6 +133,9 @@ class ZFinder:
             # zwarning flags and don't bother with doing fit
             if len(n.where(specs[i] != 0.)[0]) == 0:
                 self.zwarning[i] = int(self.zwarning[i]) | flag_val_unplugged
+               self.sn2_data.append(0)
+               self.chi2_null.append(-1)
+               self.f_nulls.append(0)
             else: # Otherwise, go ahead and do fit
                 for ipos in xrange(self.npoly):
                     bvec[ipos+1] = n.sum( poly_pad[ipos] * data_pad[i] *

timahutchinson:
I'll look into this first thing tomorrow morning.

moustakas:
@timahutchinson Has this been fixed?

timahutchinson:
Yes, I believe it was fixed.

PR #1702, opened 2022-02-19
Remove mean work per process from subgroup partitioning

marcelo-alvarez:
This modification ensures that the partitioning encoded in `max_per_proc` limits the size of subgroups returned by `desispec.parallel.dist_discrete_all` by removing the `target` and `over` variables.

If the expected behaviour of `dist_discrete_all` is to return a configuration that minimizes the maximum of work done over all subgroups, as is likely, then this modification should be adopted. 

This issue was discovered with @dmargala during review for redrock [PR #204](https://github.com/desihub/redrock/pull/204) for very similar code in redrock. The changes here are essentially identical to those in redrock [PR #207](https://github.com/desihub/redrock/pull/207).

Please review and merge if appropriate.

### DIFF
--- a/py/desispec/parallel.py
+++ b/py/desispec/parallel.py
@@ -217,8 +217,6 @@ def dist_discrete_all(worksizes, nworkers, power=1.0):
         warnings.warn("Too many workers ({}) for {} work items.  Some workers"
             " idle.".format(nworkers, len(worksizes)), RuntimeWarning)
 
-    target = np.sum(weights) / nworkers
-
     dist = []
 
     off = 0
@@ -227,8 +225,7 @@ def dist_discrete_all(worksizes, nworkers, power=1.0):
         if curweight + weights[cur] > max_per_proc:
             ## dist.append( (off, cur-off) )
             dist.append( list(range(off, cur)) )
-            over = curweight - target
-            curweight = weights[cur] + over
+            curweight = weights[cur]
             off = cur
         else:
             curweight += weights[cur]
--- a/py/desispec/test/test_parallel.py
+++ b/py/desispec/test/test_parallel.py
@@ -74,6 +74,10 @@ class TestParallel(unittest.TestCase):
             assert(w[1] == self.taskcheck)
             off += self.taskcheck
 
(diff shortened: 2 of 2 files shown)


sbailey:
Catching up on old PRs:

@marcelo-alvarez please add to `desispec.test.test_parallel` unit tests an example of the case that this fixes (i.e. a test that would fail on master, but pass with this update).  I think that case is in the discussion of desihub/redrock#204 , albeit with some API differences for dist_discreet_all vs. distribute_work.

Otherwise this looks fine, thanks.

sbailey:
Catching up on old PRs:

@marcelo-alvarez please add to `desispec.test.test_parallel` unit tests an example of the case that this fixes (i.e. a test that would fail on master, but pass with this update).  I think that case is in the discussion of desihub/redrock#204 , albeit with some API differences for dist_discreet_all vs. distribute_work.

Otherwise this looks fine, thanks.

marcelo-alvarez:
> Catching up on old PRs:
> 
> @marcelo-alvarez please add to `desispec.test.test_parallel` unit tests an example of the case that this fixes (i.e. a test that would fail on master, but pass with this update). I think that case is in the discussion of [desihub/redrock#204](https://github.com/desihub/redrock/pull/204) , albeit with some API differences for dist_discreet_all vs. distribute_work.
> 
> Otherwise this looks fine, thanks.

Thanks @sbailey. I will add the `desispec.test.test_parallel` unit tests as instructed. 

The same problem was addressed recently in redrock, with [desihub/redrock#207](https://github.com/desihub/redrock/pull/207), by replacing `distribute_work` with the simpler version developed for  [desihub/redrock#204](https://github.com/desihub/redrock/pull/204). 

I'm investigating whether following the same approach here will result in identical result with no loss of performance. If that is the case, then it would make sense to replace the current `dist_discreet_all` with a simpler version that accomplishes the same task, based on `redrock.utils.distribute_work`.

marcelo-alvarez:
@sbailey I've included the desispec.test.test_parallel unit test, ready for review.

PR #1705, opened 2022-02-22
move qso cat utilities to LSS repo

echaussidon:
@sbailey, I move the code to treat the qso catalog into the LSS repo. This PR remove useless files from desispec repo. 

Edmond


sbailey:
Thanks for the cleanup here after moving to LSS.

PR #1706, opened 2022-02-22
do not implicitly load desispec.io.raw.[read,write]_raw routines

moustakas:
For your consideration:

Background
------------

I am working on documentation and installation instructions for `FastSpecFit` (to be run on Fuji+Guadalupe) and would like to minimize the number of code dependencies. `FastSpecFit` explicitly depends on various `desispec` routines, but *not* on `Specter`. 

Problem
--------

In trying to install `FastSpecFit` in a "clean" (laptop) environment, I tripped because `Specter` is implicitly imported in `desispec.io.__init__.py` via the `desispec.io.raw.[read,write]_raw` routines.

Proposed PR
-------------

Explicitly import `desispec.io.raw.[read,write]_raw` (as needed) thereby allowing users (and `FastSpecFit`) to access the various high-level `desispec` utilities without having to import (or depend on) `Specter`.


sbailey:
I had typed a full response, but apparently forgot to click "comment" or otherwise had a internet/github hiccup and thus lost it, so trying again with an abbreviated version:

* I agree with the motivation of this PR
* yes, let's do this into fuji not master so that it can be included in the final fuji tag
* I'm concerned about the API change of removing desispec.io.read_raw, just in case some other piece of code like nightwatch might be using it
* the specific dependency on desispec.io "needing" specter comes in N>>1 levels deep in desispec.trace_shifts `import specter.psf` but it turns out that is a leftover import that isn't even needed, and thus removing line desispec/py/desispec/trace_shifts.py line 19 fixes the problem (I think)
* if there are other cases that you are tripping on that inadvertently require specter, let's try moving those imports into the functions that use specter so that we only really need specter if we are running code that really needs specter...

moustakas:
Thanks @sbailey. I'm closing this PR in lieu of #1709.

PR #1708, opened 2022-02-22
Let plot_spectra show errors even when --rebin is specified.

schlafly:
Add the ability to plot errors on spectra to plot_spectra even when --rebin is specified.  Currently the --errors argument only works when --rebin is not set; with this PR, --errors now works even for rebinned spectra.  I also changed the behavior of --errors in two ways:
- I also show a simple line plot of the errors offset by 2, in addition to the error bars
- I replaced the individual lines with a shaded region.

Here is a resulting spectrum:
`
plot_spectra -i ~/redux/daily/tiles/cumulative/7893/20220219/coadd-7-7893-thru20220219.fits --zrange 5 6 --redrock ~/redux/daily/tiles/cumulative/7893/20220219/redrock-7-7893-thru20220219.fits --rebin 5 --errors
`
image
and zooming in
image


sbailey:
Looks good.  I am also going to attempt to merge this into the fuji branch to include with the final fuji tag, hopefully not causing chaos when fuji is merged back into master.

PR #1712, opened 2022-02-23
add SPGRPVAL to desi_zcatalog output for custom coadds/redshifts

sbailey:
This PR propagates the input SPGRPVAL header keywords into the desi_zcatalog output ZCATALOG tables.  For the perexp, pernight, and cumulative groups this is redundant with EXPID, NIGHT, and LASTNIGHT columns, but for custom coadds like 1x_depth, 4x_depth, lowspeed it provides a generic way to record which input set contributed to each row.

Example outputs are in /global/cfs/cdirs/desi/users/sjbailey/dev/fuji/zcat:
  * ztile-1x_depth.fits  ztile-4x_depth.fits  ztile-lowspeed.fits : proposed redshift catalogs to include with fuji
  * ztile-sv1-dark-pernight.fits example reprocessing of the existing fuji/zcatalog/ztile-sv1-dark-pernight.fits, differing only by the addition of the SPGRPVAL column.

The one future-proofing trick is that it tries to use np.int32 as the dtype if it can (applies to all current cases of SPGRPVAL), otherwise falling back to np.int64 or letting numpy derive the dtype if not an int.

@rongpu @akremin comments?

Note: we may have other PRs soon for zcatalog to patch missing targeting info, etc. but I'm trying to keep each feature in a separate PR.

### DIFF
--- a/bin/desi_zcatalog
+++ b/bin/desi_zcatalog
@@ -243,6 +243,8 @@ for ifile, rrfile in enumerate(redrockfiles):
 
         data.add_column(np.full(nrows, hdr['SPGRPVAL'], dtype=dtype),
                 index=icol, name='SPGRPVAL')
+    else:
+        log.warning(f'SPGRPVAL keyword missing from {rrfile}')
 
     zcatdata.append(data)
 
--- a/bin/desi_zcatalog
+++ b/bin/desi_zcatalog
@@ -214,8 +214,9 @@ for ifile, rrfile in enumerate(redrockfiles):
     if args.group == 'perexp':
         data.add_column(np.full(nrows, hdr['NIGHT'], dtype=np.int32),
                 index=icol, name='NIGHT')
+        icol += 1
         data.add_column(np.full(nrows, hdr['EXPID'], dtype=np.int32),
-                index=icol+1, name='EXPID')
+                index=icol, name='EXPID')
     elif args.group == 'pernight':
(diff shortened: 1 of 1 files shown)


akremin:
If I'm reading this correctly you're only adding this column if it exists in the header, so your catalogs will then differ in their columns, is that right? Or are we adding this keyword with dummy values for all redshift types?

Otherwise the code changes look good to me.

rongpu:
The SPGRPVAL for the custom coadds are always integers. I checked the SPGRPVAL values in the custom coadd files (ztile-1x_depth.fits ztile-4x_depth.fits ztile-lowspeed.fits), and they look good.

sbailey:
Thanks for the checks, @akremin and @rongpu 

> If I'm reading this correctly you're only adding this column if it exists in the header, so your catalogs will then differ in their columns, is that right? Or are we adding this keyword with dummy values for all redshift types?

All standard pipeline redshift groupings (perexp, pernight, cumulative, healpix) have SPGRP and SPGRPVAL, and @rongpu included those in the custom coadds as well, so all of our current cases should have that column, and it should be int32 in all current cases too.  I didn't want desi_zcatalog to become useless if that keyword didn't exist (in which case it doesn't create the column), but I'll add a log WARNING message for that case because it is unexpected.

The case that would crash is if some input files had the column but others did not, but I think that's a crash-worthy situation because the inputs are inconsistently structured compared to each other, and thus likely not valid to be stacked.

akremin (approved):
The changes look good

PR #1713, opened 2022-02-24
desi_assemble_tilepix

sbailey:
This PR replaces `desi_map_tilepix` with a re-written `desi_assemble_tilepix` that uses the pre-calculated `preproc/NIGHT/EXPID/tilepix*.json` files and the `exposures-$SPECPROD.fits` table to filter out petals that are known to be missing.  It also adds a `--check` option to check that all expected healpix redrock files exist, and that no unexpected healpix directories exist (after some cleanup, those checks pass for both Fuji and Guadalupe).

Example outputs are in $CFS/desi/spectro/redux/fuji/healpix/tilepix.* and guadalupe/healpix/tilepix.*, generated with
```
desi_assemble_tilepix -o tilepix.fits --check
```

Extra: I added a `parse_cameras(..., loglevel)` option to enable silencing INFO-level messages for just that call.

I'm pretty sure that the original `desi_map_tilepix` also had a bug from caching the result for a tile for the first healpix it was on, thus missing the other healpix, but I haven't studied that in detail because I was completely replacing it with a faster version anyway.  The `--check` option replicates the original spirit of maintaining consistency between the mapping what is actually on disk.

@akremin @araichoor comments?


bin/desi_assemble_tilepix
```diff
@@ -0,0 +1,166 @@
...
+#- filter out petals that don't have frames
+for tileid in tileids:
+    ii = frames['TILEID'] == tileid
+    camword = parse_cameras(np.unique(frames['CAMERA'][ii]), loglevel='warning')
+    #- should only be complete petals, not individual cameras left
+    assert 'b' not in camword
+    assert 'r' not in camword
+    assert 'z' not in camword
+    for petal in range(10):
+        if camword_union([camword, f'a{petal}']) != camword:
```
akremin, line 97: Given the above assertions I think you can just check `if str(petal) not in camword`, couldn't you?

bin/desi_assemble_tilepix
```diff
@@ -0,0 +1,166 @@
...
+    log.info(f'{survey} {program} {len(healpix)} healpix')
+    for pix in healpix:
+        fn = desispec.io.findfile('redrock', groupname=pix,
+                survey=survey, faprogram=program)
+        if not os.path.exists(fn):
+            log.error(f'MISSING {fn}')
+            nbad += 1
+
+    #- Check for extra directories
+    for dirname in sorted(glob.glob(f'healpix/{survey}/{program}/*/*')):
```
akremin, line 151: Doesn't glob allow you to specify numeric dir names such that you are less likely to have ValueErrors?

akremin (commented):
Overall this looks great. I have comments below, but nothing major. If you merged without any edits I wouldn't be upset.

Comments:
It seemed odd at first glance not to require `outfile` and not set a default. The code runs everything and then just exits. My guess is that this was to allow for `--check` to be used independent of whether we want the file or not? 

I think post-fuji we want to spin the --check code off into another script along with the rest of our vetting scripts, but for now I think it is okay to leave it.

This code makes two file type assumptions for files with two common formats: 
1.  The user provided exposure file is assumed to be fits 
2. The `tilepix` returned by `findfile()` is assumed to be json.
I think both are reasonable assumptions. For the first you specify in the argparser help information that the exposure file should be fits. I'll leave it as optional if you want to also check it explicitly in the code (otherwise we just let the code crash and they'll figure it out). The second is (currently) enforced by the fact that `findfile()` returns the json filename, so we're okay there too.

bin/desi_assemble_tilepix
sbailey: good idea, that is cleaner.  I switched to that.

bin/desi_assemble_tilepix
sbailey: that reduces the chance of picking up a non-int dir, but it can still trip on pathological cases like "123blat", so I still have to use the try/except protection when casting to an int.  What I'd really want is something regex-like `[0-9]+` to match 1 or more digits but only digits, but the glob syntax doesn't seem to support that.  I went ahead and added the `[0-9]*` to the code for possible clarity, while keeping the try/except too.

sbailey:
Thanks for the review.  I implemented your suggestions, with a few comments on them:

> It seemed odd at first glance not to require outfile and not set a default. The code runs everything and then just exits. My guess is that this was to allow for --check to be used independent of whether we want the file or not?

Yeah, that was the reasoning: you can run --check and only commit to writing an output file after you know that all checks pass.  I added a warning so that if the user doesn't specify either --check or --outfile, then it warns them but still proceeds (possibly useful for code testing the loading of the exposures and individual tilepix files).

> The user provided exposure file is assumed to be fits

I added a check that it is either fits or ecsv before running a bunch of logic and then crashing on the wrong output type.

> The tilepix returned by findfile() is assumed to be json

Yes, which findfile enforces.  It is picking up the preproc/NIGHT/EXPID/tilepix*.json files created as part of assemble_fibermap (new with Fuji, and saves a lot of time later so that we don't have to keep re-deriving what tiles cover what healpix).

akremin:
Thank you for addressing my comments.

For the record I think `ecsv` will *not* work for the exposures file use case here, since you read in two different HDU's (and therefore two tables) from the fits file, which I don't believe `ecsv` supports. `ecsv` also isn't one of the currently used formats, but I'm confident you knew that and you were just future proofing the code.

I don't think it is harmful as committed and merged, and don't think we need to retag if that has already been done. Worst case scenario is that it confuses someone if they try to use `ecsv`.

sbailey:
Your comment was clear "The user provided exposure file is assumed to be fits", but somehow I misread that as referring to assumptions about the --outfile rather than the input --expfile.  I added checks that the user was specifying a valid outfile format before proceeding with a bunch of calculations, but I didn't add any checks on the input format.  i.e. I think what I did was correct, just not addressing the comment.  Indeed you are correct that input expfile ecsv (or csv) wouldn't work because it needs both HDUs from the fits file, causing the code to error quickly if given the wrong format.

PR #1715, opened 2022-02-25
validate redshift catalog metadata and generate merged targeting+photometric catalogs for Fuji+Guadalupe

moustakas:
This PR checks in two support scripts to support a given data release (here, Fuji+Guadalupe).

Validate provenance of each object with a redshift
--------------------------------------------------

`desi_validate_zcatalog_spectra` makes sure that the on-disk location of every coadded spectrum and Redrock redshift catalog can be found using the columns and/or header metadata in the final, merged redshifts catalogs, i.e., all the catalogs found in `$DESI_SPECTRO_REDUX/fuji//global/cfs/cdirs/desi/spectro/redux/fuji/zcatalog/z[pix,tile]*.fits`

Generate merged targeting and Tractor photometric catalogs
-------------------------------------------------------------

`match_zcat_photo` generates merged targeting + Tractor (photometric) catalogs for all unique TARGETIDs observed in a given data release. (Photometric catalogs currently only exist for Fuji while they are being vetted, but Guadalupe catalogs will follow.)

There are two key catalogs (to be included in the data release) and two support catalogs (to assist with vetting, which won't appear in the final data release) which can be found in `/global/cscratch1/sd/ioannis/photocatalog`:

**Catalogs (for release)**

* `fuji-targetphot.fits` [2.8GB, N=2,757,937, extname=`TARGETPHOT`] - Targeting information (from `/global/cfs/cdirs/desi/target/catalogs/dr9`) for every unique TARGETID in any of the [Fuji] redshift catalogs. Key notes:
  * All entries are copied from the original targeting catalogs; nothing is copied from the fiberassign tables, which is important for some of the earliest SV observations which are missing some columns like `FLUX_W1_IVAR`.
  * There is special handling for finding DR9 photometry (where available) for secondary targets, but additional vetting is welcome.
  * The data model includes all possible targeting columns for each object, irrespective of which survey/program the target came from, i.e., `CMX_TARGET`, `[DESI,BGS,MWS]_TARGET`, `SV[1,2,3]_[DESI,MWS,BGS]_TARGET`, `SCND_TARGET`, and `SV[1,2,3]_SCND_TARGET`.
  * There are 930,555 objects (34% of all targets) with no DR9 targeting photometry, which can be identified by flagging rows with all null values (zeros or blank strings), e.g., `BRICKNAME==''`. These targets are overwhelmingly skies or stuck positioners but do include some secondary targets. See the discussion of *missing* photometric targets below. **Question: do we want to remove these rows/objects from this catalog?**
  
* `fuji-tractorphot.fits` [3.7GB, N=1,834,701, extname=`TRACTORPHOT`] - For the subset of objects with targeting photometry in `fuji-targetphot.fits`, this catalog includes the full DR9/Tractor photometry for each object (with a `targetid` column added at the end in order to enable joining/matching to the original redshift catalogs).

**Support catalogs (will not appear in final data release)**

Two additional catalogs are written out to help with testing and validation:

* `fuji-targetphot-zcat.fits` [108MB, N=2,757,937] - This is the merged redshift catalog used to generate `fuji-targetphot.fits`, against which it is row-matched. **Help wanted: does this catalog truly include every unique TARGETID in, here, Fuji?**

* `fuji-targetphot-missing.fits` [90MB, N=930,555] - This catalog includes all the objects without *object* targeting information in DR9. (Note: *sky* targets obviously have targeting information, but those catalogs were not searched; I'm talking just about object/source photometry here.) 

  The data model is very simple and designed to help track down and understand the provenance of every object in this catalog. The columns are: `['TARGETID', 'SURVEY', 'PROGRAM', 'TARGET_RA', 'TARGET_DEC', 'OBJID', 'BRICKID', 'RELEASE', 'SKY', 'GAIADR', 'RA', 'DEC', 'GROUP']` with the following meaning:
    * `'TARGETID', 'SURVEY', 'PROGRAM', 'TARGET_RA', 'TARGET_DEC'` come from the original redshift catalog.
    * `'OBJID', 'BRICKID', 'RELEASE', 'SKY', 'GAIADR'` come from using [decode_targetid](https://github.com/desihub/desitarget/blob/master/py/desitarget/targets.py#L136) to unpack all the **positive TARGETIDs** (`TARGETID>=0`).
    * `'RA', 'DEC', 'GROUP'` come from using [decode_negative_targetid](https://github.com/desihub/desitarget/blob/master/py/desitarget/targets.py#L248) to unpack all the **negative TARGETIDs** (`TARGETID<0`). 
  
  The resulting statistics are as following:
    * N=383,435 objects (41%) have `SKY==1`.
    * Among the 547,120 objects with `SKY==0`, 402,552 (74%) have `TARGETID<0` (i.e., stuck positioners), leaving 144,568 objects with potentially recoverable DR9 photometry.
      * Among these 144,568 objects with `SKY==0 & TARGETID>=0` there are 46 unique `RELEASE` numbers ranging from zero to 9999. However, all of these appear to have valid `OBJID` and `BRICKID` entries, which suggests that (perhaps?) they are secondary targets with good DR9 photometry. **I would like to try to recover the DR9 photometry for as many of these objects as possible but would like to get additional input from @geordie666, @djschlegel, @stephjuneau, and others first.**


geordie666:
Regarding this point:
```
Among these 144,568 objects with SKY==0 & TARGETID>=0 there are 46 unique RELEASE numbers ranging from zero to 9999. However, all of these appear to have valid OBJID and BRICKID entries, which suggests that (perhaps?) they are secondary targets with good DR9 photometry.
```
"Standalone" secondary targets are assigned an `OBJID` and `BRICKID` but those values are not necessarily taken from the Legacy Surveys. See section `3.2.1` of the [desitarget pipeline paper](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=6693) for a full description.

For "standalone" secondary targets for which `desitarget` did not find (or look for in the case of `OVERRIDE`s) an LS DR9 match, you would have to perform an RA/Dec match. There is also no guarantee that these targets will be in the Legacy Surveys, or that an RA/Dec match will recover the correct target. Proposers of secondary programs were not required to use DR9 of the Legacy Surveys to construct their target lists.

If you decide to perform an RA/Dec match back to the Legacy Surveys, I'd recommend either:

- Using the same matching radius as used by `desitarget` to construct the `TARGETID` for "standalone" secondary targets that have a match in LS DR9. This radius is stored as `"SEP"` in the header of the relevant "standalone" secondary files (e.g. `/global/cfs/cdirs/desi/target/catalogs/dr9/1.1.1/targets/main/secondary/dark/targets-dark-secondary.fits`). I'm fairly sure that I always used 1 arcsecond.
- Perform a full RA/Dec match yourself for _all_ "standalone" secondary targets. Then you could use whichever matching radius you'd prefer.

stephjuneau:
> Regarding this point:
> 
> ```
> Among these 144,568 objects with SKY==0 & TARGETID>=0 there are 46 unique RELEASE numbers ranging from zero to 9999. However, all of these appear to have valid OBJID and BRICKID entries, which suggests that (perhaps?) they are secondary targets with good DR9 photometry.
> ```
> 
> "Standalone" secondary targets are assigned an `OBJID` and `BRICKID` but those values are not necessarily taken from the Legacy Surveys. See section `3.2.1` of the [desitarget pipeline paper](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=6693) for a full description.
> 
> For "standalone" secondary targets for which `desitarget` did not find (or look for in the case of `OVERRIDE`s) an LS DR9 match, you would have to perform an RA/Dec match. There is also no guarantee that these targets will be in the Legacy Surveys, or that an RA/Dec match will recover the correct target. Proposers of secondary programs were not required to use DR9 of the Legacy Surveys to construct their target lists.
> 
> If you decide to perform an RA/Dec match back to the Legacy Surveys, I'd recommend either:
> 
> * Using the same matching radius as used by `desitarget` to construct the `TARGETID` for "standalone" secondary targets that have a match in LS DR9. This radius is stored as `"SEP"` in the header of the relevant "standalone" secondary files (e.g. `/global/cfs/cdirs/desi/target/catalogs/dr9/1.1.1/targets/main/secondary/dark/targets-dark-secondary.fits`). I'm fairly sure that I always used 1 arcsecond.
> * Perform a full RA/Dec match yourself for _all_ "standalone" secondary targets. Then you could use whichever matching radius you'd prefer.

@moustakas & @geordie666 : I have checked something very relevant to this by doing the comparisons of RA, Dec and would like to send the results in a coordinated fashion so we don't duplicate more work. Should I just put the results here in this thread? I'll start working on that unless one of you wants to chat to coordinate.

sbailey:
boring technical note: it looks like this branch started off of master, not fuji, and thus includes a bunch of changes from other PRs into master that we *don't* want in the fuji branch.  I've converted the PR into a draft so that we don't accidentally merge it, but we'll need to do some rebasing and maybe a new PR before the final merge, but let's continue the catalog conversation here.

moustakas:
Changes carried over to #1716; closing this PR.

PR #1719, opened 2022-02-28
Fit constant offset to CCD y coordinates.

sybenzvi:
Suggestion from @julienguy: add `deggy=0` to the list of automatic qproc options to use a constant offset for the CCD y-coordinates rather than a polynomial. The new default should fix the B camera DY out-of-range issue reported in several Nightwatch tickets (e.g. desihub/nightwatch#263).
- Before fix: https://nightwatch.desi.lbl.gov/20220227/00124229/qa-camera-00124229.html
- After fix: https://data.desi.lbl.gov/desi/users/sybenzvi/nightwatch/20220227/00124229/qa-camera-00124229.html

### DIFF
--- a/py/desispec/scripts/qproc.py
+++ b/py/desispec/scripts/qproc.py
@@ -232,9 +232,9 @@ def main(args=None):
 
         # using the trace shift script
         if args.auto  :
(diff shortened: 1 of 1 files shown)


julienguy:
I am merging. Someone has to update the desispec version ran by nightwatch at KPNO and NERSC.

PR #1721, opened 2022-03-03
Create Summary Catalogs for Fuji/Guadalupe

Ragadeepika-Pucha:
The `zcatalog.py` now has three extra functions. The `create_summary_catalog()` function will be useful for creating summary redshift catalogs that combines all zpix*/ztile*cumulative files and adds the PRIMARY columns that flag the "best" spectra for a given TARGETID. The code can be run as follows: 

```
import zcatalog.create_summary_catalog as cz

## set location of output files
path = ".\"

## For Fuji zpix
filename = path+"zpix-sv-all.fits"
cz(spec_release = 'fuji', version = 'zpix', output_filename = filename)

## For Fuji ztile 
filename = path+"ztile-sv-all.fits"
cz(spec_release = 'fuji', version = 'ztile', output_filename = filename)

## For Guadalupe zpix 
filename = path+"zpix-main-all.fits"
cz(spec_release = 'guadalupe', version = 'zpix', output_filename = filename)

## For Guadalupe ztile 
filename = path+"ztile-main-all.fits"
cz(spec_release = 'guadalupe', version = 'ztile', output_filename = filename)
```

### DIFF
--- a/py/desispec/zcatalog.py
+++ b/py/desispec/zcatalog.py
@@ -142,8 +142,20 @@ def find_primary_spectra(table, sort_column = 'TSNR2_LRG'):
 ####################################################################################################
 ####################################################################################################
 
+def _get_survey_program_from_filename(filename):
+    """
+    Return SURVEY,PROGRAM parsed from zpix/ztile filename; fragile!
+    """
+    # zpix-SURVEY-PROGRAM.fits or ztile-SURVEY-PROGRAM-SPECGROUP.fits
+    base = os.path.splitext(os.path.basename(filename))[0]
+    arr = base.split('-')
+    survey = arr[1]
+    program = arr[2]
+    return survey, program
+
 def create_summary_catalog(specprod, specgroup = 'zpix', \
-                           all_columns = True, columns_list = None, output_filename = './zcat-all.fits'):
+                           all_columns = True, columns_list = None,
+                           output_filename = './zcat-all.fits'):
     """
     This function combines all the individual redshift catalogs for either 'zpix' or 'ztile' 
     with the desired columns (all columns, or a pre-selected list, or a user-given list).
@@ -192,13 +204,15 @@ def create_summary_catalog(specprod, specgroup = 'zpix', \
     specred_dir = specprod_root(specprod)    
     if (os.path.isdir(specred_dir) == False):
         log.error(f'"{specprod}" directory does not exist')
-        raise OSError(f'"{specprod}" directory does not exist')
+        raise ValueError(f'"{specprod}" directory does not exist')
         
     ## Initial check 2
     ## If specgroup = something else by mistake  
-    if (specgroup != 'zpix')&(specgroup != 'ztile'):
-        log.error(f'"{specgroup}" not recognized')
-        raise NameError(f'"{specgroup}" not recognized')
+    valid_specgroups = ('zpix', 'ztile')
+    if specgroup not in valid_specgroups:
+        errmsg = f'{specgroup=} not recognized, should be one of {valid_specgroups}'
+        log.error(errmsg)
+        raise ValueError(errmsg)
         
     ######################################################################################
     
@@ -208,10 +222,10 @@ def create_summary_catalog(specprod, specgroup = 'zpix', \
     ## Find all the filenames for a given specgroup
     if (specgroup == 'zpix'):
         ## List of all zpix* catalogs: zpix-survey-program.fits
-        zcat = glob(f'{zcat_dir}/zpix*.fits')
+        zcat = glob(f'{zcat_dir}/zpix-*.fits')
     elif (specgroup == 'ztile'):
         ## List of all ztile* catalogs, considering only cumulative catalogs
-        zcat = glob(f'{zcat_dir}/ztile*cumulative.fits')
+        zcat = glob(f'{zcat_dir}/ztile-*cumulative.fits')
                     
     ## Sorting the list of zcatalogs by name
     ## This is to keep it neat, clean, and in order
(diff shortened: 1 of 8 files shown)


py/desispec/zcatalog.py
```diff
@@ -1,15 +1,429 @@
...
+    
+    ## For convenience, sort by SURVEY, PROGRAM, and (HEALPIX or TILEID) 
+    if (version == 'zpix'):
+        tab.sort(['SURVEY', 'PROGRAM', 'HEALPIX'])
+    else:
+        tab.sort(['SURVEY', 'PROGRAM', 'TILEID', 'LASTNIGHT'])
+    
+    ## Add the final table as a HDU
+    hdu0 = fits.PrimaryHDU()
+    hdu1 = fits.BinTableHDU(tab)
```
sbailey, line 192: I think this is where the column comments are getting lost (maybe), and I don't think you need to go via HDUList etc. to write a single table.  Consider:
```
tab.meta['EXTNAME'] = 'ZCATALOG'
tab.write(output_filename, overwrite=True)
```
otherwise try using `fits.convenience.table_to_hdu(tab)` which may preserve the column comments.

py/desispec/zcatalog.py
```diff
@@ -1,15 +1,429 @@
...
+    survey : str
+        The SURVEY of the given redshift catalog (cmx|special|sv1|sv2|sv3)
+        
+    program : str
+        The PROGRAM of the given redshift catalog (dark|bright|backup|other)
+        
+    Returns
+    -------
+    tab : Astropy Table
+        The final table after adding all the required '*_TARGET' columns
```
sbailey, line 222: Please clarify in the docstring whether this modifies the input `tab` in-place, or whether it leaves the input unchanged and returns the new table with additional columns.  If modifying the input table in-place, consider *not* returning the table, which sort of makes it more obvious that it is modifying the table in-place.  Context: That's one of the reasons that `list.sort()` doesn't return the sorted copy -- they want to you realize that it is modifying the list in place so that you don't accidentally think that `sorted_list = orig_list.sort()` gives you a new list without changing the original.

py/desispec/zcatalog.py
```diff
@@ -1,15 +1,429 @@
...
+
+####################################################################################################
+####################################################################################################
+
+
+def fix_tables(survey, program, spec_release = 'fuji', version = 'zpix', \
+               all_columns = True, columns_list = None):
+    """
+    This function adds all the required columns and returns the table with all columns, 
+    or pre-selected columns, or with user-given list of columns.
```
sbailey, line 275: Style/Clarity: The function name and the comment here imply that `fix_tables` is fixing a pre-existing table, but it appears that it is reading and patching a table from disk.  Consider separating the reading of the table from the "fixing" of the table, or update the function name to include "read" in the name.

If this function still does the reading, it would be better for it to take `filename` as an input instead of re-deriving what it should be (potentially ending up with a different answer from what `glob` found above if we change the file naming scheme in the future).

py/desispec/zcatalog.py
```diff
@@ -1,15 +1,429 @@
...
+    ## 'backup' and 'bright' consists of six characters and is the maximum for a PROGRAM string
+    col1 = Column(np.array([survey]*len(tab)), name = 'SURVEY', dtype = '<U7')
+    col2 = Column(np.array([program]*len(tab)), name = 'PROGRAM', dtype = '<U6')
+    tab.add_column(col1, 1)
+    tab.add_column(col2, 2)
+    ## The SURVEY and PROGRAM columns are added as second and third columns,
+    ## immediately after TARGETID
+    
+    ## Add the TARGET columns -- only for SV, not needed for Main
+    if (spec_release == 'fuji'):
```
sbailey, line 354: See above comments about not deriving functionality based upon specprod (spec_release) name so that this code can be used with future prods as well.

sbailey (changes requested):
Thanks for this work.  Please see comments inline for requested changes.  The most important is to use the specprod name only to find files on disk, but not for making any logic decisions so that this code will also work for future releases and not just fuji.

py/desispec/zcatalog.py
```diff
@@ -1,15 +1,429 @@
...
+        ## Add the other target columns that are not present in the redshift catalogs
+        ## Depends on the survey and program
+        ## The function `add_target_columns` adds the required columns based on survey and program
+        tab = add_target_columns(tab, survey = survey, program = program)
+        
+        ## Making a list of all the '*_TARGET* columns 
+        ## This is for rearranging the columns into proper order
+        target_cols = ['CMX_TARGET', 'DESI_TARGET','BGS_TARGET','MWS_TARGET', 'SCND_TARGET',\
```
weaverba137, line 362: These backslashes are completely unnecessary and my trigger style warnings in the future.  In general, a line continuation backslash is never needed in any modern version of Python.

py/desispec/zcatalog.py
```diff
@@ -1,15 +1,429 @@
-## This script is based on this discussion - https://github.com/desihub/desispec/issues/1355
-## Version: January 21, 2022
+"""
```
weaverba137, line 1: I haven't seen the test results yet, but the format of the docstrings in this module don't appear to be particularly standard.  Be prepared for test failures due to unrecognized docstring style.

weaverba137 (changes requested):
The style of this module is somewhat different from other code in this package and in the DESI pipeline code in general.

py/desispec/zcatalog.py
Ragadeepika-Pucha: @weaverba137, can you please specifically point what the issues are with the docstring? Or you can send me some link with standard docstring example?

py/desispec/zcatalog.py
Ragadeepika-Pucha: Both of these suggestions do not preserve the column comments. I changed it to your first suggestion.

Ragadeepika-Pucha:
@sbailey @weaverba137 @stephjuneau I finished working on all the comments and have pushed the changes. 
I have created the following eight files which are present here: `/global/cfs/cdirs/desi/users/raga19/public/zcats_from_script/`

The files are:
`zpix-edr-all.fits` : combines all the zpix-*fits files for fuji and includes all columns
`zpix-main-all.fits` : combines all the zpix-*fits files for guadalupe and includes all columns
`ztile-edr-cumulative-all.fits` : combines all the ztile*cumulative.fits files for fuji and includes all columns
`ztile-main-cumulative-all.fits` : combines all the ztile*cumulative.fits files for guadalupe and includes all columns

`zpix-edr-summary.fits`, `zpix-main-summary.fits`, `ztile-edr-cumulative-summary.fits`, and `ztile-main-cumulative-summary.fits` -- These include only the pre-selected list of columns.

Let me know if you have any questions about these files.

sbailey:
Thanks for the updates.  I pushed some cosmetic changes, but for the most part this looks good.  The remaining issue is sorting out exactly what should go into the output files: main vs. sv vs. everything.  The current code allows
```
create_summary_catalog(specprod='fuji', survey_name='main', ... )
```
And it will create an output file with the stack of everything in fuji, even though fuji doesn't contain any survey=main tiles.  I think this will get even more confusing in future productions which have both sv and main, but the code requires you to pick one or the other, but still ingests everything in the prod whether it is sv or main...

So, following up on the discussion from the Tuesday telecon, a specific suggestion for discussion:
  1. in the summary zcat files, include surveys sv1,sv2,sv3,main but *not* cmx,special.
      * in future prods, this would mean that the summary files could have both sv and main, and that's ok in the spirit of end users who don't care about our survey designations, they just want to know about the best spectrum is for their favorite targets.
      * in practice, M31 tile 80715 is the only CMX target that has made it in, and that is a special beast unto itself (only 3 tiles made it through redshifts, and sky subtraction is terrible anyway...)
      * special tiles might be more ambiguous, and I'm open to arguments that they should be included (or additional arguments in support of keeping them out)
  3. have the code auto-derive from the input files it finds which *DESI_TARGET etc. columns it needs to include instead of using `survey_name` to force it to a specific set of sv or main columns.

Discuss.

Ragadeepika-Pucha:
Hi,

I have modified the code to run for future releases, as long as the column names are consistent. It produces errors for `everest` because of the change in the column names, as well as, the `HEALPIX` column is unavailable. However, after commenting out a few lines that include different column names, it worked well for everest. The summary files for `fuji` and `guadalupe` are fine, and have been updated here: `/global/cfs/cdirs/desi/users/raga19/public/zcats_from_script/`.

To run the code, you have to specify the `specprod`:
```
create_summary_catalog(specprod, specgroup = 'zpix', \
                           all_columns = True, columns_list = None, output_filename = './zcat-all.fits')
```

I have fixed the errors that I found. Let me know if there are any further issues. Thanks!

sbailey:
The CI doctest errors are the same as desihub/redrock#210, due to some new incompatibility between the versions of sphinx and jinja2 used in the tests.  As long as "python setup.py build_sphinx" continues to work from NERSC, we can ignore these test errors here (i.e. use the redrock PR to debug and fix the problem, and we can merge this whether or not the doctest environment is fixed because it is unrelated to the changes in this PR itself).

sbailey:
FYI I made a small update to derive the SURVEY and PROGRAM from the file headers instead of the filenames.  Merging now; thanks for this work.

Issue #1685, opened 2022-02-14
For computing EFFTIME for fiber rejection purposes, use median tile E(B-V) rather than per-fiber E(B-V)


schlafly:
We reject fibers when they have low effective time.  Some contributions to the effective time are things that change from observation to observation, but others do not.  High E(B-V) fibers will never be observed at low E(B-V), so marking those fibers as bad doesn't really help to eventually get those fibers a good observation.  In extreme cases, we end up rejecting entire petals because of high E(B-V) on a substantial fraction of the petal.

Here's an example case where we lost a patch of fibers at higher E(B-V):
https://github.com/desihub/desisurveyops/issues/25

It would be good to change the EFFTIME computation for fiber rejection purposes to use the median tile E(B-V) rather than the individual fiber E(B-V)s.

julienguy:
fixed in pr #1722

PR #1722, opened 2022-03-03
use tile-median EBV for LOWEFFTIME

araichoor:
This PR addresses issue https://github.com/desihub/desispec/issues/1685.
Currently, the per-fiber EFFTIME_SPEC used to turned on the LOWEFFTIME flag is based on a per-fiber EBV value, we switch to a tile-median EBV value.

In detail, the LOWEFFTIME flag is turned on here:
https://github.com/desihub/desispec/blob/4d91a1f1e521a83a0195e2d4674e7e81fc4b587a/py/desispec/tile_qa.py#L277-L278
where EFFTIME_SPEC comes from the per-fiber TSNR2 estimate, where it accounts for the per-fiber EBV value.

With this PR:
- we read the EBVFAC (= 10 ** (2.165 * EBV_MED / 2.5) value from the exposure header (as all exposures are for a given tile, there is just a single value); if I'm correct, that value comes from fiberassign, and is based on the median EBV value of all fibers of the tile;
- we apply the LOWEFFTIME condition on the EFFTIME_SPEC using the tile-median EBV:
https://github.com/desihub/desispec/blob/98a0b91f220660593f552f3f6aa5e378fbe21009/py/desispec/tile_qa.py#L288-L290

This is the resulting tile_qa plot for the https://github.com/desihub/desispec/issues/1685 example:
tile-qa-23224-thru20220212

Remarks:
- as this is quite sensitive, I'd appreciate some expert eyes to double-check my changes are correct!
- I chose to *not* modify the EFFTIME_SPEC values, only the criterion to turn on LOWEFFTIME, is that what we want?
- if for whatever reason EBVFAC is not present in the header (early tiles, special tiles, etc), I use a default value of 1.


julienguy:
I like this idea to just modify the criterion for LOWEFFTIME.
I would suggest using the median EBV from the fibers read in the FIBERMAP instead of EBVFAC from the headers, to make sure the same quantity is used in the numerator and the denominator when rescaling EFFTIME_SPEC. 
It would be useful to verify how EFFTIME_SPEC scales with EBV. There might be some difference with EBVFAC.

araichoor:
thanks for the comments.

_using the median EBV from the fibers read in the FIBERMAP instead of EBVFAC from the headers_
=>
I m fine with implementing that,  but for completeness, just one remark here: 
The EBVFAC stored in the header comes from the fiberassign; with looking at the code, it is the median value for the primary targets on the tile; so it provides a better sampling  of the sky than just the 5000 fibers.
For that example tile: EBVFAC=1.281, median of 5000 ebvfac values: 1.288; using 1.288 changes five fibers to LOWEFFTIME.
maybe those are details which do not matter, but I m mentioning in case.
(And maybe a more extreme way would be just re-derive the median value of EBV over the tile; I guess that is done somewhere in desihub/desisurvey, as that information is provided for the tiles files).

_It would be useful to verify how EFFTIME_SPEC scales with EBV. There might be some difference with EBVFAC._
=>
agreed!
I took EBVFAC, as it was what was used in the seminal definition of EFFTIME (now called GFA_EFFTIME).
if that helps, with digging a bit, I think EBV is included in the tsnr computation here:
https://github.com/desihub/desispec/blob/4d91a1f1e521a83a0195e2d4674e7e81fc4b587a/py/desispec/tsnr.py#L1116
but this at a per-wavelength level; that vector is then manipulated and summed to get the TSNR value; and EFFTIME_SPEC is proporational to TSNR**2.
So, as of now, it is not clear to me, actually (I would need to think more about it).

julienguy:
Hi, I checked the scaling formula of EBVFRAC is a good match for EFFTIME_SPEC vs EBV for the tile you mention in the PR.
Screenshot from 2022-03-03 16-03-38

julienguy:
However despite the fact EBVFRAC is estimated with more pointing, I would rather use the EBV values in the current tile to compute the median in order to make sure that we will always have a good normalization of the EFFTIME correction. This is just in case we change something in the future. I feel nervous to have the numerator and denominator computed in different codes (from different repositories).

araichoor:
Thanks for the ebvfac-dependency check; if easy for you, could you also check with a dark tile? if not, I can check that later today.
And I ll also push a commit to use the median ebvfac value of the 5000 fibers, no problem.

Lastly, I m thinking about this now: do we want backwards-compatibility (related to the MTL decisions)? ie add a switch like "if night < xxx do as before else use this new approach"? I *think* it s fine to not be backwards-compatible, as the "MTL replay" for the LSS analysis only cares for hardware problem; pinging @ashleyjross in case.

ashleyjross:
The zmtl files in the archive just need to never change in order to the MTL replay to be fine. (And it is my understanding they will never change for a given archivedate.)

There might be some funny things to sort out about the differences between the archived zmtl files and the ones with the data release, but I think that is ok and I think we want those to just use whatever we think the best choice should have been. (In the worst case, some hybrid between the data release and archived version of zmtls will be required and even if that becomes so, it does not seem so bad to me.)

So, I'm pretty sure that means *do not* add in any kind of night check. Thanks for checking with me!

ID: ***@***.***>

schlafly:
The other place we care about what E(B-V) to use is in the ETC.  It looks like the ETC knows about EBVFAC but doesn't actually use it, and instead uses median(EBV) for the targets
https://github.com/desihub/desietc/blob/5c6ce47c2e086caba2247692f93091305e7c53f3/desietc/etc.py#L1047
At some level I want to treat that as authoritative since that's what we're actually targeting when we're trying to complete a tile.

But really I kind of wish the ETC were using EBVFAC and that this code would too.  Anyway, these all sound like fine options and I'd be satisfied with no change.

We should think about interaction with the backup program, which ignores E(B-V).
https://github.com/desihub/desietc/blob/5c6ce47c2e086caba2247692f93091305e7c53f3/desietc/etc.py#L1052-L1056
I think we want to "hack" in that all fibers have E(B-V) = 0 for that program.

araichoor:
**Summary:**

Before proceeding, I wanted to perform a check similar to Julien s one, but in a more systematic way.
Unless my approach is biased, there is a departure from EFFTIME_SPEC prop. to 10 ** (-2 * 2.165 * EBV_MED / 2.5) for high EBV.
Based on what follows, I m thinking to use instead of the 2.165 coefficient (which correspond to the r-band):
- dark tiles: 1.4 (justified by data + tsnr)
- bright tiles: 1.7 (justified by data)

**Details: data**
In order to compare all tiles, I came up with the following approach.
For each tile:
- I select fibers with EBV>0 and EFFTIME_SPEC>0
- I define:
 - `NORM_EBV = EBV - median(EBV)`
 - `NORM_EFFTIME = EFFTIME_SPEC / median(EFFTIME_SPEC)`
- I plot the (NORM_EBV, NORM_EFFTIME) values with binning along EBV

This plot shows what I get for tileid=23224, to illustrate how that compares with Julien s plot:
efftime-ebvfac-23224
As Julien showed, for that tile, the EFFTIME_SPEC is in agreement with a 10 ** (-2 * 2.165 * EBV_MED / 2.5) dependency.

And here is the plot for all the main backup (left), bright (middle), and dark (right) tiles:
efftime-ebv-data
(I acknowledge we ll have a different treatment for the backup tiles).

The three lines have values motivated by a small tsnr template analysis, described below; the green line is ~similar to the EBVFAC mentioned so far.
For the dark tiles, there is a clear departure from the green line; the blue line looks a good representation; and is in agreement with the dark tiles using the elg tsnr template.
For the bright tiles, it is less clear, but the green line also looks in disagreement with the data; I m not sure why is that (maybe some other parameters are at play, like the sky brightness); but it looks like the red line is a reasonable representation.

**Details: tsnr**
In parallel, I tried to empirically understand how the tsnr depends on EBV.
For that I ve taken the tsnr templates (see Slide 7 of DESI-DocDB-6579), and only applied dust transmission to those, for a grid of EBV values.
I obtain the following plot, where the three lines are just picked by eye:
efftime-ebv-tsnr
My understanding is that the shape of the template "weights" the wavelengths to account for the extinction; for instance, the elg tsnr template has more flux (="weight") in the red wavelengths, hence is less sensitive to extinction.

I remind that for the backup/bright tiles, the EFFTIME_SPEC comes from the bgs tsnr template, and for the dark tiles, it comes from the elg tsnr template: https://github.com/desihub/desispec/blob/4d91a1f1e521a83a0195e2d4674e7e81fc4b587a/py/desispec/data/qa/qa-params.yaml#L34-L36

For clarity, I copy here the code snippet to do that:
```
ebvs = np.linspace(0, 0.7, 71)
tracers = ["bgs", "lrg", "elg"]
tsnrs = {}
for tracer in tracers:
    d = get_ensemble()
    d = d[tracer]
    tsnrs[tracer] = np.zeros(len(ebvs))
    for i in range(len(ebvs)):
        tsnr = 0
        for band in ["b", "r", "z"]:
            fluxes = d.flux[band].copy()
            fluxes *= dust_transmission(d.wave[band], ebvs[i])
            fluxes = fluxes ** 2
            tsnr += np.sum(fluxes)
        tsnrs[tracer][i] = tsnr
```

schlafly:
I'd speculate that BGS is different since the blue-ish moonlight downweights the blue side of the spectra, making the effective wavelength for E(B-V) purposes redder than you'd otherwise expect.  (comparing your proposed 1.7 coefficient with the tsnr-predicted 2.2 coefficient)

That said, it's unfortunate to introduce these new magic constants.  The alternative would be to compute two EFFTIMEs, one for the SFD E(B-V) and one for the tile-median E(B-V), and use the latter only for the LOWEFFTIME cut.  Then we'd be doing exactly what we want without new approximations, but we'd have to pass more fields along.  I'm happy with your current approach as well---just thinking out loud!

julienguy:
Hi, I would use a simple prescription with a single scaling formula for this LOWEFFTIME threshold, and we could keep the same as for EBVFAC even though it's not a perfect match.

araichoor:
Thanks both for the feedback (and taking the time to read the long message!).
I ll go with Julien s suggestion to keep the EBVFAC dependency, ie 10 ** (-2 * 2.165 * EBV_MED / 2.5).
It s by far the simplest approach, doesn t introduce any new coefficient, and hopefully the inaccuracy is reasonable for the purpose here.

And I ll also add a special hack for the backup program, with keeping the same scheme, but with forcing median(ebv)=0.

araichoor:
A last question about the Main/BACKUP tiles (sorry if I m dense): 

I thought that setting EBV_MED=0, and hence "correcting" the EFFTIME_SPEC by 10 ** (-2 * 2.165 * EBV / 2.5) would do the trick.
However, with testing on TILEID=43005, I get the following:
snapshot
In the rightmost plot, I display the quantity used for the LOWEFFTIME criterion on the y-axis.
The result is that fibers in low-EBV regions don t have a "corrected-enough" EFFTIME_SPEC, and hence end up flagged with LOWEFFTIME.

@schlafly: is that behavior ok with what you had in mind? or did you mean that we should just force that zero fibers are flagged with LOWEFFTIME?

schlafly:
Yeah, this is another case where if we had a version of EFFTIME that didn't use E(B-V) it would be useful.  For the backup program, can we set E(B-V) = 0 for the EFFTIME calculation upstream of this?  So for most programs, EFFTIME = what it usually does, and for BACKUP, EFFTIME = EFFTIME(E(B-V) = 0)?

What happens to the LOWEFFTIME fibers?  I guess this isn't a regression, at least, so we could go ahead and merge as it is for now and not worry about the backup program?

araichoor:
_What happens to the LOWEFFTIME fibers?_
=>
Good question; I always thought those would be put back for re-observations by the MTL, but with looking at the codes, it looks like not.
So, I m wondering if that LOWEFFTIME is actually used anywhere (except for the tile_qa_plot)...

My understanding:
- MTL discards observations with NODATA|BAD_SPECQA|BAD_PETALQA:
https://github.com/desihub/desitarget/blob/1c7edc091ba7b8c628914826abcd5ee9c7a8bf24/py/desitarget/mtl.py#L532
- BAD_SPECQA is a zwarn bit defined in the zmtl catalogs:
https://github.com/desihub/desispec/blob/4d91a1f1e521a83a0195e2d4674e7e81fc4b587a/py/desispec/zmtl.py#L263-L271
which corresponds to bad_qafstatus_mask here:
https://github.com/desihub/desispec/blob/4d91a1f1e521a83a0195e2d4674e7e81fc4b587a/py/desispec/data/qa/qa-params.yaml#L42
and there is no LOWEFFTIME in that.

schlafly:
Thanks.  @geordie666 , do we intend to be flagging LOWEFFTIME out of the MTL?  I thought we did?  Indeed, that was the primary motivation for trying to get rid of the LOWEFFTIME flag on these targets?

Re the backup program... since we aren't using the MTL yet, let's just not special case the backup program targets at all, I guess.  I think the ideal thing to do would be to use EBV = 0 for their original EFFTIME computation, and then again here for their median EBV for tile completion purposes (so that EBV just doesn't play a role).  But I hadn't appreciated that that was going to be hard so let's just do what you've already implemented for all programs.

Sorry!

ashleyjross:
I had thought the same as Eddie, but should have noticed what Anand flagged (thanks Anand!): There does not actually seem to be any LOWEFFTIME cut in the MTL logic.

I can confirm that when I apply the MTL logic to the redshift catalogs I compile for the daily LSS catalogs, I do not see any kind of sharp cutoff in TSNR2_LRG (which I believe has a 1 to 1 relationship with EFFTIME used here). The attached plot shows the low-end of the distribution.
[Chart, histogram  Description automatically generated]
It would be good to get confirmation from @geordie666<https://github.com/geordie666> , though!

(And if this is correct, sorry for wasting people's time during our discussion of the issue!)




.com>

araichoor:
@ashleyjross : your plot didn t came through on the github page.

ashleyjross:
Right, sorry, I forgot github just gives you a description (of sorts) of the plot rather than displaying it when you (or at least I) respond via email:
Screen Shot 2022-03-09 at 5 22 12 PM

araichoor:
Summarizing what was said above + at the Mar. 14 surveyops telecon:
- I ll add LOWEFFTIME in the `bad_qafstatus_mask` in `desispec/data/qa/qa-params.yaml`;
- I ll correct with the 10 ** (-2 * 2.165 * EBV_MED / 2.5), where EBV_MED is the median of the fibers EBV;
- I won t implement any backup-specific behavior.
@schlafly : correct?

araichoor:
I ve went ahead and implemented what I mentioned in the previous message.
Two last comments, for completeness, about this inclusion of LOWEFFTIME in bad_qafstatus_mask and its possible interplay in tile_qa.py.
I just want to check that this is the intended behavior.

(1)
The per-fiber EFFTIME_SPEC value is computed for fibers with bad_qafstatus_mask not set:
https://github.com/desihub/desispec/blob/1fc93bf2a10acdc78a4b9f7b108841311410c8ad/py/desispec/tile_qa.py#L269-L273
But those lines appear *before* with turn on the LOWEFFTIME flag, so nothing will change here; i.e. fibers with LOWEFFTIME will still count in the summing of the EFFTIME_SPEC.
I *think* that it is what we want.

(2)
The *tile* EFFTIME_SPEC is computed from fibers with bad_qafstatus_mask not set:
https://github.com/desihub/desispec/blob/1fc93bf2a10acdc78a4b9f7b108841311410c8ad/py/desispec/tile_qa.py#L316-L319
But, as this line of code to select the fiber uses the *per-exposure* column (`or_qafiberstatus`), I think that nothing will change, because we turn on the LOWEFFTIME flag on the `tile_fiberqa_table['QAFIBERSTATUS']` column.
I m not sure exactly what we want, but I suspect that excluding fibers with LOWEFFTIME from this computation shouldn t change much the *tile* EFFTIME_SPEC, as we take the median.

julienguy:
I think it correct

schlafly:
That sounds reasonable to me.

With this change, will we still kill petals like this one?
https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/21507/20211113/tile-qa-21507-thru20211113.png

schlafly:
That looks correct to me as well.

araichoor:
ok thanks; I ll implement that tomorrow, when cori is back up.

araichoor:
I ve pushed a commit to do the tile-median EBV correction in `exposure_qa.py` for the per-petal TSNR2 (https://github.com/desihub/desispec/pull/1722/commits/ee578849d81149a8b8ab0692c1960d2910388a83).
I had to add more line of codes than initially expected because such correction has to be done _after_ having read the 10 petals, so I had to store some quantities along the way.

For that particular case, it works as expected:
Before:
tmp-before

After:
tmp-after

Though, I d appreciate another pair of eyes on this code change, or suggestion for further testing/validation (@schlafly , @julienguy?).

schlafly:
Do we worry about EFFTIME/GOALTIME = 0 in the "after" plot, or the missing entry in the exposure table?

araichoor:
Thanks for spotting that!
I think we don t care, it s unrelated to the current PR.

I think it s just an "artefact" of me running desi_tile_qa not from the not-official place.
In detail: the list of exposures is grabbed by `desispec.tile_qa_plot.get_expids_efftimes()`, which looks for `spectra*fits` files in the same folder as the input `tile-qa*fits` file.
As I ve not copied/symlinked those `spectra*fits` files to my working folder, no exposure is found (none is reported on top), i.e. an empty `exps` Table is returned.
Which then implies a "total efftime" of zero:
https://github.com/desihub/desispec/blob/1fc93bf2a10acdc78a4b9f7b108841311410c8ad/py/desispec/tile_qa_plot.py#L1610

araichoor:
Confirming, with re-running with copying beforehand the `spectra*fits` files in my working folder, that "artefact" is corrected:
tmp-after2

schlafly:
Looks good to me.  I'm embarrassed that I don't understand the discussion of "_R" in the code.  I gather there are TSNR2_LRG_{BRZ} fields (or something of that ilk?), and that TSNR2_LRG = TSNR2_LRG_B + TSNR2_LRG_R + TSNR2_LRG_Z?  And you only correct the R ones for extinction?

Sorry, I'm probably just not understanding what qa_params['tsnr2_key'], qa_params['tsnr2_band'] are.

araichoor:
Thanks for looking carefully at the changes.

My understanding is that the confusion may come from the fact that:
- we re dealing here with the `BADPETALSNR` flag, which relies on the `petal_tsnr2` values;
- the `petal_tsnr2` values are *not* the ones to compute the efftime (those are `tsnr2_for_efftime_vals`); `petal_tsnr2` are just used for `BADPETALSNR`;
- the `petal_tsnr2` computation relies on `tsnr2_vals`, which reads the `qa_params["tsnr2_key"] = TSNR2_LRG_R` column from the `qa_params["tsnr2_band"] = r` camera; so it s all r-band; that s why using the 2.165 coefficient sounds legitimate to me;
- but if for whatever reason those r-band settings in `qa-params.yaml` are changed, then the 2.165 coefficient won t be relevant anymore; I don t that it will happen, but I thought that would be a useful safety check..

schlafly:
Okay, sounds good.  And, just verifying, this is a petal cut, so we don't need to worry about B or Z cameras being cut due to low SNR?  Thanks!

julienguy:
The petal_tsnr2 is used only to to compute a relative fraction of one petal with respect to the others and set a flag. So it's ok to apply some scaling with E(B-V) here, but then we should rename the variable, because it is not any more a TSNR2 value. In fact, I am not sure we need this at all. Can we simply remove this BADPETALSNR score?

araichoor:
_Can we simply remove this BADPETALSNR score?_
=>
I ve looked at the main survey exposures where a petal is discarded because of BADPETALSNR only.
This happens fairly rarely (6 exposures over ~5k exposures).

I checked the guadalupe reduxs + daily post-20210900 exposures with EFFTIME_SPEC>0.
I searched for fibers in the FIBERQA extension of the `exposure-qa-EXPID.fits` files, with:
- BADPETALSNR turned on,
- BADPETALPOS|BADPETALSKY|BADPETALSTDSTAR|BADPETALFLUXCAL|BADREADNOISE turned off.
I get six exposures from the daily/post-20210900:
```
 NIGHT   EXPID  TILEID  FAFLAVOR       EXPTIME          EFFTIME_SPEC            EBV          NBADPETALSNR
-------- ------ ------ ---------- ------------------ ------------------ -------------------- ------------
20211005 103142   2507   maindark 1266.2879638671875 1097.5108642578125   0.1198229119181633          500
20211023 105797  20967 mainbright   1265.77197265625   69.2481460571289 0.046704355627298355          500
20211113 108645  21507 mainbright  966.3599853515625 285.59417724609375  0.11941109597682953          500
20211115 108885  21266 mainbright  1860.043701171875 142.08111572265625  0.06350825726985931          500
20211130 111580   6255   maindark  981.6066284179688  934.3463134765625   0.1376459002494812          500
20220112 118315  43005 mainbackup  604.9677124023438  7.934487342834473   0.6106953620910645         1000
```
The two dark tiles had fairly good observing conditions, and the high EBV likely explains the petal rejection.
The three bright tiles had crappy observing conditions; for two tiles, it s not clear to me what s the culprit.
The backup tile also had crappy observing conditions, and the high EBV likely explains the petal rejection.

In details:

https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/2507/20211005/tile-qa-2507-thru20211005.png
=> petal 5 with high-ebv

https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/20967/20211023/tile-qa-20967-thru20211023.png
=> not sure which petal is discarded here..
note that this tile got a second exposure: : https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/20967/20211119/tile-qa-20967-thru20211119.png

https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/21507/20211113/tile-qa-21507-thru20211113.png
=> tile that prompt the PR discussion, petal 5 with high-ebv

https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/21266/20211115/tile-qa-21266-thru20211115.png
=> the petal 3 may have higher ebv, but not obviously... and there are other warning turned on; may need a closer look
note that this tile got a second exposure: https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/21266/20211117/tile-qa-21266-thru20211117.png

https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/6255/20211130/tile-qa-6255-thru20211130.png
=> petal 2 has higher ebv

https://data.desi.lbl.gov/desi/spectro/redux/daily/tiles/cumulative/43005/20220112/tile-qa-43005-thru20220112.png
=> backup tile, petals 1&2 may have higher ebv; tons of warnings anyway...

julienguy:
Thank you Anand for carefully checking this. Based on your finding, I think it is safe to drop this BADPETALSNR score.
It's either high EBV which we want to get rid of or terrible tiles/exposures with very low EFFTIME anyways.

So I suggest we simplify the code and remove all of the code written to set BADPETALSNR.

araichoor:
With looking at the code: I notice that there are two header keywords using `petal_tsnr2_frac`:
https://github.com/desihub/desispec/blob/fce0cf0fcf7c51ef0a37a918828e55f728a8545e/py/desispec/exposure_qa.py#L394-L395
From the github search engine, those only appear in desispec in that part of `exposure_qa.py`, so are not used by any other code.

Is it ok to remove those also? (i.e. to change the list of keywords in the header)

If we prefer to keep those keywords, then we d have to keep the `petal_tsnr2_frac` calculation (but that of course doesn t prevent to remove the BADPETALSNR flag).

araichoor:
@julienguy : I apologize, I realize only now that this `petal_tsnr2_frac` is also used to set the `TSNR2FRA` column in the `PETALQA` table:
https://github.com/desihub/desispec/blob/e6f8d0f2dc6ff52f106013e92fde2c4167535b49/py/desispec/exposure_qa.py#L373-L374

So, I m not sure what we d want to do: also remove that `TSNR2FRA` column from the exposure-qa\*fits files (and as a consequence in the tile-qa\*fits files)?

I see that this `TSNR2FRA` column appears in:
- tile_qa.py: propagating in the `PETALQA` extension the `TSNR2FRA` column from the exposure-qa*fits files;
- tile_qa_plot.py: I think here it s just reading the column, and check against tsnr2_petal_minfrac, tsnr2_petal_maxfrac, so doing the same as `BADPETALSNR`.

In a similar spirit, the two qa keywords tsnr2_petal_minfrac, tsnr2_petal_maxfrac:
https://github.com/desihub/desispec/blob/e6f8d0f2dc6ff52f106013e92fde2c4167535b49/py/desispec/data/qa/qa-params.yaml#L32-L33
appear only in exposure_qa.py and tile_qa_plot.py, discussed above.
So, we could also remove those, I think.

julienguy:
Yes. I think it's the logical consequence of all of this. We will see later if we find the need to have this redundant SNR information. I think the QA we have in place, with the visualization of the EFFTIME per fiber per tile, the threshold with the number of good fibers per tile , and is the average tile EFFTIME sufficient.

araichoor:
@julienguy, @schlafly  : for info, I ve pushed a commit which removes:
- the `TSNR2FRA` columns in the PETALQA tables (exposure and tile);
- the `BADPETALSNR` flagging in `QAFIBERSTATUS`;
- the `PMINEXPF` and `PMAXEXPF` header keywords in the exposure-qa fits file;
- the `tsnr2_petal_minfrac` and `tsnr2_petal_maxfrac` in the `qa-params.yaml`.

And summarizing what the current changes do, in addition to the above (as there has been quite some traffic here):
- set bit of LOWEFFTIME per fiber using the median of EBV>0 fibers, with a EBVFAC-like dependence (original motivation of the PR);
- add`LOWEFFTIME` to the mask list in `bad_qafstatus_mask` in `qa-params.yaml` (so that MTL will not consider those fibers as "done").

I ve checked on the TILEID=23224, NIGHT=20220212 that it performs as expected.
I d be happy to perform further suggested sanity checks, if any, please let me know.

araichoor:
Remark: my last commit failed some checks, I m not sure it s related to my changes:
```
Run sphinx-build -W --keep-going -b html doc doc/_build/html
...
ImportError: cannot import name 'environmentfilter' from 'jinja2' (/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/jinja2/__init__.py)
```

julienguy (approved):
I tested successfully several tiles with areas of large E(B-V). I noted one has to set the environment variable `SPECPROD=daily` despite the option `--prod daily` to `desi_tile_qa` to get the information about the list of exposure and the efftime/goaltime in the figure. I fixed that and retested.

julienguy:
ok to remove them.

PR #1734, opened 2022-03-16
Use gpu_specter in desi_proc

marcelo-alvarez:
### Description

Incorporates gpu_specter into desi_proc, based on changes @dmargala made in a [separate branch](https://github.com/desihub/desispec/tree/dmargala-scale-run-wip). 

Testing compared the frame file for camera z0 of exposure 90659, i.e. frame-z0-00090659.fits, obtained with the following commands on Cori and Perlmutter:
```
source /global/common/software/desi/desi_environment.sh 22.2
DESI_SPECTRO_REDUX=/global/cfs/cdirs/desi/users/$USER/spectro/redux

branchdir=$SCRATCH/tmpdir-$NERSC_HOST && mkdir -p $branchdir && rm -rf $branchdir/desispec
git clone --branch gpu_specter https://github.com/desihub/desispec $branchdir/desispec
PYTHONPATH=$branchdir/desispec/py:$PYTHONPATH && PATH=$branchdir/desispec/bin:$PATH

cargs="-n 20210601 -e 00090659 --traceshift --batch --system-name"

preprod () {
  SPECPROD=$1 
  target=$DESI_SPECTRO_REDUX/$SPECPROD/calibnight && mkdir -p $target
  source=$DESI_ROOT/spectro/redux/guadalupe/calibnight/20210601
  ln -s $source $target >& /dev/null
}

if   [ $NERSC_HOST == "cori" ]       ; then 
  module load gpu_specter
  preprod cori_cspec_cpu ; desi_proc $cargs cori-haswell  
  preprod cori_gspec_cpu ; desi_proc $cargs cori-haswell   --gpuspecter  
elif [ $NERSC_HOST == "perlmutter" ] ; then 
  preprod prlm_cspec_cpu ; desi_proc $cargs perlmutter-gpu --queue regular
  preprod prlm_gspec_cpu ; desi_proc $cargs perlmutter-gpu --queue regular --gpuspecter 
  preprod prlm_gspec_gpu ; desi_proc $cargs perlmutter-gpu --queue regular --gpuspecter --gpuextract
fi
```
to the corresponding frame file in Guadalupe. The commands above correspond to running specter on Cori (**cori_cspec_cpu**), gpu_specter on Cori (**cori_gspec_cpu**), specter on Perlmutter CPUs (**prlm_cspec_cpu**), gpu_specter on Perlmutter CPUs (**prlm_gspec_cpu**), and gpu_specter on Perlmutter GPUs (**prlm_gspec_gpu**).

#### 1. Guadalupe vs specter on Cori Haswell CPUs

Specter on Cori produces the same results as for Guadalupe at machine precision (see [guadalupe_vs_cori_cspec_cpu.txt](https://github.com/desihub/desispec/files/8337772/guadalupe_vs_cori_cspec_cpu.txt)). "All-in-one" science processing with desi_proc as above is currently broken on KNL (see #1735) but is expected to give bitwise identical results to Guadalupe. 

#### 2. Guadalupe vs gpu_specter on Cori/Perlmutter CPUs/GPUs
- Percent-level median differences in the flux (normalized with the variance) when ignoring pixels that are masked or have ivar=0 in either file (as in all figures shown here), with some differences for specific pixels occasionally reaching 10-20%, and even higher very near the edges of the CCD. The comparison plots for CPUs and GPUs are indistinguishable, so only one is shown here:
guadalupe_vs_cori_gspec_cpu
This is consistent with what was found by @dmargala in [specter #83](https://github.com/desihub/specter/issues/83)
- Running with gpu_specter results in an additional HDU called "SCORES" being written to the frame file, which is the intended behavior (see [guadalupe_vs_cori_gspec_cpu.txt](https://github.com/desihub/desispec/files/8337785/guadalupe_vs_cori_gspec_cpu.txt) and #1737).

#### 3. gpu_specter on GPUs vs gpu_specter on CPUs
- Very small (<1e-4) differences were found in a very small number of pixels
prlm_gspec_gpu_vs_cori_gspec_cpu
- Non-masked pixels near the edge of the CCD (e.g. pixel 2851 on fiber 0) have zero variance and flux in the CPU calculation, but a non-zero flux in the GPU version (see [prlm_gspec_gpu_vs_cori_gspec_cpu.txt](https://github.com/desihub/desispec/files/8337791/prlm_gspec_gpu_vs_cori_gspec_cpu.txt)).


py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -73,6 +73,9 @@ def get_shared_desi_proc_parser():
     parser.add_argument("--timingfile", type=str, help='save runtime info to this json file; augment if pre-existing')
     parser.add_argument("--no-xtalk", action="store_true", help='diable fiber crosstalk correction')
     parser.add_argument("--system-name", type=str, help='Batch system name (cori-haswell, perlmutter-gpu, ...)')
+    parser.add_argument("--extract-size", type=int, default=22, help="Size to use for GPU extract subcomm")
```
sbailey, line 76: Let's try to find a more descriptive name for this; at first glance I thought that "extract-size" referred to the extraction algorithm itself (e.g. something about the divide and conquer or number of wavelength bins or something) rather than something about the MPI communicator size.

py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -595,6 +605,9 @@ def create_desi_proc_batch_script(night, exp, cameras, jobdesc, queue, runtime=N
             fx.write("export OMP_NUM_THREADS={}\n".format(threads_per_core))
         else:
             fx.write("export OMP_NUM_THREADS=1\n")
+        if system_name == 'perlmutter-gpu':
+            fx.write("export MPICH_GPU_SUPPORT_ENABLED=1\n")
+            fx.write("module load cudatoolkit\n\n")
```
sbailey, line 610: Is there any risk if cudatoolkit was already loaded by the environment?  For everything else we require that the environment is properly configured before launching the job instead of having the job load additional modules.

py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -73,6 +73,9 @@ def get_shared_desi_proc_parser():
     parser.add_argument("--timingfile", type=str, help='save runtime info to this json file; augment if pre-existing')
     parser.add_argument("--no-xtalk", action="store_true", help='diable fiber crosstalk correction')
     parser.add_argument("--system-name", type=str, help='Batch system name (cori-haswell, perlmutter-gpu, ...)')
+    parser.add_argument("--extract-size", type=int, default=22, help="Size to use for GPU extract subcomm")
+    parser.add_argument("--gpuspecter", action="store_true", help="Use GPU specter")
+    parser.add_argument("--gpuextract", action="store_true", help="Use GPU extraction")
```
sbailey, line 78: For the record (no action needed): gpu_specter includes a CPU-based code path which is faster than the original specter, even without using GPUs.  i.e. it could make sense to use `--gpuspecter` without `--gpuextract` on a CPU-only node.

py/desispec/scripts/proc.py
```diff
@@ -722,26 +731,61 @@ def main(args=None, comm=None):
...
                         print('RUNNING: {}'.format(cmds[camera]))
 
-                    desispec.scripts.extract.main_mpi(extract_args, comm=comm_extract)
+                    if args.gpuextract:
+                        desispec.scripts.extract.main_gpu_specter(extract_args, coordinator=coordinator)
+                    elif args.gpuspecter:
+                        desispec.scripts.extract.main_gpu_specter(extract_args, comm=comm_extract)
+                    else:
+                        desispec.scripts.extract.main_mpi(extract_args, comm=comm_extract)
+                    comm_extract.barrier()
```
sbailey, line 787: Since this barrier is inside a loop and inside an if statement, double check the logic to make sure that no ranks could miss the barrier under some special case (e.g. if there are more ranks provided than needed for the number of cameras)

py/desispec/scripts/proc.py
```diff
@@ -694,6 +694,15 @@ def main(args=None, comm=None):
                 cmd += ' -o {}'.format(framefile)
                 cmd += ' --psferr 0.1'
 
+                if args.gpuspecter:
+                    cmd += ' --gpu-specter'
+                    #- default for CPU is nsubbundles=6 but gpu_specter only allows 1, 5, or 25
+                    cmd += ' --nsubbundles 5'
+                    cmd += ' --mpi'
```
sbailey, line 701: I'm a little surprised that the `--mpi` option appears necessary for the gpu_specter branch but not the old-specter branch, even though it is called with MPI parallelism as well later:
```
extract_args = desispec.scripts.extract.parse(cmdargs)

if args.gpuextract:
    desispec.scripts.extract.main_gpu_specter(extract_args, coordinator=coordinator)
elif args.gpuspecter:
    desispec.scripts.extract.main_gpu_specter(extract_args, comm=comm_extract)
else:
    desispec.scripts.extract.main_mpi(extract_args, comm=comm_extract)
```
i.e. what would happen if `main_gpu_specter` was called with a `comm` or `coordinator` but with `args.mpi = False`?

sbailey (commented):
Thanks for these updates to support GPUs within the normal pipeline.  We can address the remaining algorithmic differences in the output separately.  I posted some minor comments inline.  Please also post the performance numbers from your test runs.

py/desispec/scripts/proc.py
dmargala: I recall adding `--mpi` there so that it would show up when the command is logged. I don't think it is necessary.

Here is the logic in `main_gpu_specter`:

https://github.com/desihub/desispec/blob/fce0cf0fcf7c51ef0a37a918828e55f728a8545e/py/desispec/scripts/extract.py#L215-L230

py/desispec/scripts/proc.py
```diff
@@ -722,26 +731,61 @@ def main(args=None, comm=None):
...
+                extract_ranks = list(range(extract_size))
+                #- Create a comm group for the ranks that will perform extraction
+                if rank in extract_ranks:
+                    extract_incl = comm.group.Incl(extract_ranks)
+                    comm_extract = comm.Create_group(extract_incl)
+                    from gpu_specter.mpi import ParallelIOCoordinator
+                    coordinator = ParallelIOCoordinator(comm_extract)
+                else:
+                    run_extraction=False
+            else:
```
dmargala, line 761: There should be another case here for setting up the communicator for the CPU code path of gpu_speceter (when `--gpuspecter` is specified without `--gpuextract`.

py/desispec/scripts/proc.py
marcelo-alvarez: Leaving this in so that it is logged for gpu_specter.

py/desispec/workflow/desi_proc_funcs.py
marcelo-alvarez: Agreed. Will be interesting to try this on Cori, for example.

py/desispec/scripts/proc.py
marcelo-alvarez: Would the additional case be for any other reason than to [set extract_size = 16](https://github.com/desihub/desispec/blob/dmargala-scale-run-wip/py/desispec/scripts/proc.py#L780) instead of 20?

dmargala: `extract_size = 16` should be a decent default for the cpu version of gpu_specter on cori haswell and perlmutter gpu. It does look like it be okay to share the existing cpu-path logic for comm_extract although it needs `extract_size = args.extract_size`.

py/desispec/workflow/desi_proc_funcs.py
dmargala: `--extract-subcomm-size` ?

py/desispec/scripts/stdstars.py
```diff
@@ -47,7 +47,7 @@ def parse(options=None):
                          nargs='*',
                          help='List of TARGETIDs of standards overriding the targeting info')
     parser.add_argument('--mpi', action='store_true', help='Use MPI')
-    parser.add_argument('--ignore-gpu', action='store_true', help='Ignore GPU, if available')
+    parser.add_argument('--use-gpu', action='store_true', help='Use GPU, if available')
```
dmargala, line 50: desi_extract_spectra uses `--gpu`, I don't have a strong preference between `--use-gpu` or `--gpu` but it's probably good to be consistent.

https://github.com/desihub/desispec/blob/0eede4e36b488cbbc560f90f94d4a37ac3f7f543/py/desispec/scripts/extract.py#L69

py/desispec/scripts/proc.py
dmargala: I suspect this barrier will prevent the parallel IO coordinator from doing it's magic

marcelo-alvarez: It worked for 32 ranks and extract_size=22 (so 10 idle cameras), but I agree it's better not to have it generally. I  removed the barrier and moved the check for missing files that it was introduced for in the first place to after the global barrier / serial extraction.

py/desispec/scripts/proc.py
marcelo-alvarez: I think it is best to set extract_size = 20, 16, and 22 for specter, gpu_specter_cpu and gpu_specter_gpu, respectively, when the `--extract-size` flag is not used, and to the value specified otherwise.  

I will wait to get feedback before making this change.

py/desispec/workflow/desi_proc_funcs.py
marcelo-alvarez: > `--extract-subcomm-size` ?

Done.

py/desispec/workflow/desi_proc_funcs.py
marcelo-alvarez: Thanks for catching this. There is no risk, but it is better to be consistent and have the necessary modules loaded by the environment rather than in the Slurm script. I've removed the loading of the module, created a [desiconda issue](https://github.com/desihub/desiconda/issues/51) about this for future installations, and modified `20220119-2.0.1/modulefiles/desiconda/20220119-2.0.1` by hand on Perlmutter to load the cudatoolkit.

py/desispec/scripts/stdstars.py
marcelo-alvarez: Thanks for catching this, have changed `--gpu` to `--use-gpu` for desi_extract_spectra.

marcelo-alvarez:
> Please also post the performance (timing) numbers from your test runs.

Commands to generate tests:
```
# base DESI environment
source /global/common/software/desi/desi_environment.sh 22.2
DESI_SPECTRO_REDUX=/global/cfs/cdirs/desi/users/$USER/spectro/redux

# gpu_specter branch of desispec
branchdir=$SCRATCH/tmpdir-$NERSC_HOST && mkdir -p $branchdir && rm -rf $branchdir/desispec
git clone --branch gpu_specter https://github.com/desihub/desispec/$branchdir/desispec
PYTHONPATH=$branchdir/desispec/py:$PYTHONPATH
PATH=$branchdir/desispec/bin:$PATH

# process exposure 90659
ngt=20210601; exp=00090659
args=" -n $ngt -e $exp --traceshift --batch "

argscori="$args --system-name cori-haswell"
argsprlm="$args --system-name perlmutter-gpu --queue regular"

args_cori_cspec_cpu="$argscori"
args_prlm_cspec_cpu="$argsprlm"
args_cori_gspec_cpu="$argscori --gpuspecter"
args_prlm_gspec_cpu="$argsprlm --gpuspecter"
args_prlm_gspec_gpu="$argsprlm --gpuspecter --gpuextract"

# run desi_proc
if   [ $NERSC_HOST ==  "cori"      ] ; then
    module load gpu_specter
    SPECPROD=cori_gspec_cpu && desi_proc $args_cori_gspec_cpu
    SPECPROD=cori_cspec_cpu && desi_proc $args_cori_cspec_cpu
elif [ $NERSC_HOST == "perlmutter" ] ; then
    SPECPROD=prlm_cspec_cpu && desi_proc $args_prlm_cspec_cpu
    SPECPROD=prlm_gspec_cpu && desi_proc $args_prlm_gspec_cpu
    SPECPROD=prlm_gspec_gpu && desi_proc $args_prlm_gspec_gpu
fi

# command to show extraction time after runs are complete
extractiontime () {
   dir=$DESI_SPECTRO_REDUX/$1/run/scripts/night/$ngt
   echo "Time for $1: "
   echo "  $(grep "log_timer:   extract" $dir/*log | head -n 1 | awk '{print $3}')"
}
```
Results:
```
% runs=$(echo cori_cspec_cpu cori_gspec_cpu \
%            prlm_cspec_cpu prlm_gspec_cpu \
%             prlm_gspec_gpu )
% for run in $runs; do
%   extractiontime $run
% done
Time for cori_cspec_cpu: 395.76
Time for cori_gspec_cpu: 252.88
Time for prlm_cspec_cpu: (on queue)
Time for prlm_gspec_cpu: (on queue)
Time for prlm_gspec_gpu: 240.11
```
Summary: (will update when runs on queue are done) 
```  
           run  N    n   wall  fpnh  speedup

cori_cspec_cpu  5  160   396s    54      1.0
cori_gspec_cpu  5  160   253s    85	 1.6
prlm_cspec_cpu  4  200    ---   ---      ---
prlm_gspec_cpu  4  200    ---   ---      ---
prlm_gspec_gpu  1   32   240s   450      8.7
```

marcelo-alvarez:
> Thanks for these updates to support GPUs within the normal pipeline. We can address the remaining algorithmic differences in the output separately. I posted some minor comments inline.

@sbailey @dmargala I commented inline and made changes where appropriate, including adding using `desi_mps_wrapper` for `--system-name perlmutter-gpu`.  Please see performance numbers above. Assuming those are ok, this should be ready to be merged.

sbailey:
I confirm that I'm able to run on Perlmutter GPUs by specifying `desiproc --system-name perlmutter-gpu ...` and that it still works the same on cori haswell.  Thanks!

PR #1254, opened 2021-04-23
add option to use MPI for parallelism at stdstar fitting in desi proc

dmargala:
This PR adds an option to `desi_proc` and `desi_proc_joint_fit` to use MPI for parallelism at the stdstar fitting step. If provided, the MPI comm will be split by the number of spectrographs to be processed. The sub-comms will then be passed to `desispec.scripts.stdstars.main`. This PR should not alter the default MPI+multiprocessing execution path.

Note the changes to `stdstars.py` are necessary for testing due to a bug introduced in #1213 (also fixed in #1248 but not yet merged). See discussion: https://github.com/desihub/desispec/pull/1248#discussion_r618926077

I tested this using a single Cori Haswell node with data from denali for comparison (night=20210319, expids=81059,81060,81061)

There are some minor differences in the output fits files when compared with denali. Perhaps due to Haswell vs KNL?

```
     Data differs at [6, 5]:
        a> 0.00052696513
         ?            ^^
        b> 0.00052696525
         ?            ^^
     Data differs at [7, 5]:
        a> 0.0005269778
         ?            ^
        b> 0.00052697794
         ?            ^^
     ...
     47195 different pixels found (1.37% different).
```

This ran in about ~235 seconds using a single Haswell node. The corresponding run in denali ran in about ~950 seconds using 5 KNL nodes. 

Test details:

```
salloc -C haswell -N 1 -q interactive -t 120

source /global/cfs/cdirs/desi/software/desi_environment.sh 21.3
module unload desispec

# git clone desispec
# cd desispec
# git checkout branch
export PYTHONPATH=$(pwd)/py:$PYTHONPATH
export PATH=$(pwd)/bin:$PATH
export DESI_SPECTRO_CALIB=/global/cfs/cdirs/desi/spectro/desi_spectro_calib/0.2.5
export SPECPROD=$USER/$SLURM_JOB_ID
OUTDIR=$DESI_SPECTRO_REDUX/$SPECPROD
mkdir -p $OUTDIR

COMPARE_SPECPROD=denali
NIGHT=20210319
EXPIDS=81059,81060,81061
# create symlinks to stdstar inputs for these exposures
for expid in ${EXPIDS//,/ }; do
    ZEXPID=$(printf "%08d" $expid)
    INEXPDIR=$DESI_SPECTRO_REDUX/$COMPARE_SPECPROD/exposures/$NIGHT/$ZEXPID
    OUTEXPDIR=$OUTDIR/exposures/$NIGHT/$ZEXPID
[...]

PR #1735, opened 2022-03-18
Change Slurm script for science exposures with desi_proc

marcelo-alvarez:
Fixes an issue for `desi_proc` with the `--system-name cori-knl` argument in which science exposures are only allocated 3 nodes, instead of 5, causing jobs to run out of memory at the stdstarfit step. E.g.
```
source /global/common/software/desi/desi_environment.sh 22.2
desi_proc -n 20210601 -e 90659 --traceshift --system-name cori-knl --batch 
```
crashes. Behavior in this case for `--system-name cori-haswell` remains unchanged because of the 5-node limit.

@sbailey @akremin please have a look and see if we should merge this.

### DIFF
--- a/py/desispec/workflow/desi_proc_funcs.py
+++ b/py/desispec/workflow/desi_proc_funcs.py
@@ -307,7 +307,7 @@ def determine_resources(ncameras, jobdesc, queue, nexps=1, forced_runtime=None, 
         ncores, runtime = ncores + 1, 45             # + 1 for worflow.schedule scheduler proc
     elif jobdesc in ('FLAT', 'TESTFLAT'):
         ncores, runtime = 20 * nspectro, 25
-    elif jobdesc in ('SKY', 'TWILIGHT', 'SCIENCE','PRESTDSTAR','POSTSTDSTAR'):
+    elif jobdesc in ('SKY', 'TWILIGHT', 'PRESTDSTAR','POSTSTDSTAR'):
         ncores, runtime = 20 * nspectro, 30
     elif jobdesc in ('DARK'):
         ncores, runtime = ncameras, 5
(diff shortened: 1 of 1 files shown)


sbailey:
Context: when the full pipeline runs, it splits prestdstar (3 nodes), stdstar (5 nodes), and poststdstar (3 nodes) into separate jobs, which is why this didn't run out of memory for guadalupe (it was already using 5 nodes for the stdstar step).

I confirmed that this PR does *not* change that 3/5/3 for productions runs (good), via:
```
desi_run_night --dry-run-level 1 --system-name cori-knl -n 20210601 --tiles 1207
```
and inspecting the job scripts generated.

This PR does make the desi_proc "all-in-one" job less efficient because it is now using 5 nodes even for the extractions even though it maxes out at 3 nodes worth of cores (with current specter), and extractions dominate the runtime.  That's not necessarily super-critical because doing `desi_proc --batch` on cori-knl isn't how we typically run things.

At the same time, as we're moving back towards all-in-one jobs for Perlmutter, I don't want to accidentally bake in an inefficiency driven by the memory needs of one step.  Could we throttle the multiprocessing size to reduce the memory footprint instead?  PR #1254 to use MPI instead of multiprocessing for stdstars make make this moot, so I don't want it to spin off into a huge amount of work, but I thought I'd raise the efficiency issue -- i.e. is there a better way to fit within the memory footprint of 3 nodes, instead of expanding to 5 nodes even though the runtime-dominant step can't use a full 5 nodes?

marcelo-alvarez:
> At the same time, as we're moving back towards all-in-one jobs for Perlmutter, I don't want to accidentally bake in an inefficiency driven by the memory needs of one step. Could we throttle the multiprocessing size to reduce the memory footprint instead? PR #1254 to use MPI instead of multiprocessing for stdstars make make this moot, so I don't want it to spin off into a huge amount of work, but I thought I'd raise the efficiency issue -- i.e. is there a better way to fit within the memory footprint of 3 nodes, instead of expanding to 5 nodes even though the runtime-dominant step can't use a full 5 nodes?

@sbailey thanks for checking how this affects the rest of the pipeline. For a bit more context on why submitted this, I wanted to check changes in #1734 produced bitwise identical results to Guadalupe with the 22.2 environment when processing a science exposure, so had to run on KNL, but it crashed, so rather than try to fix it in #1734 I spun it off here. I do get machine-precision agreement with Guadalupe on Cori, so I will proceed with #1734 using modified KNL batch scripts and/or comparing to Cori for testing, and come back to address your question above afterwards.

sbailey:
This PR is superseded by #1820 which fixed the same underlying problem by optimizing stdstar memory usage instead of increasing the nodes allocated.  I tested the specific failing case here (which also required PR #1822 for the desi_proc stdstar args) and it now passes on 3 KNL nodes without requiring the jobs to use 5 nodes.  Closing without merging.

PR #1736, opened 2022-03-18
add cte flat class to dashboard so it reflects expected files

akremin:
This is a small pull request to correct the file counting in the spectro pipeline dashboard for the 1s flats used for CTE identification. These are extracted in the pipeline, but aren't used to create fiberflats, which then show up as red (meaning failed) in the dashboard based on the `OBSTYPE` being 'flat'. 

This update creates a new category of expected files called 'cteflat' that is detected as `OBSTYPE == 'flat' and EXPTIME < 2.0`. Normal flats have much longer integration times, so this should be appropriate, but I could just as easily select on `'cte' in PROGRAM` if people don't like the `EXPTIME` based selection.

An example of the dashboard run with this branch can be found here:
https://data.desi.lbl.gov/desi/users/kremin/PRs/dash_ignore_short_flats/dashboard.html

Screenshot showing that the "cteflat" is now colored properly, while the "normal" flats are also colored appropriately:
 
image


sbailey:
Yay

PR #1742, opened 2022-04-13
Iteration limit

sybenzvi:
This PR will fix an infinite loop in qproc that occurs when it tries reducing the polynomial order of trace fit offsets. The fix is needed because the existing logic at [line 277 of trace_shifts.py](https://github.com/desihub/desispec/blob/98fc62ef166c05557c62027562ec240074c26102/py/desispec/scripts/trace_shifts.py#L277) is incompatible with the initial condition `degyy=0`. The PR has two fixes:

1. Line 277 is changed to `if degxy>0 or degyy>0 and degxy>degxx and degyy>degyx:`
2. An iteration limit is imposed to avoid infinite loops. When the limit is reached, `degxx=degxy=degyx=degyy=0`.

### DIFF
--- a/py/desispec/scripts/trace_shifts.py
+++ b/py/desispec/scripts/trace_shifts.py
@@ -235,14 +235,9 @@ def fit_trace_shifts(image,args) :
         degyx=0
         degyy=0
     
-    nloops = 5
-    while (nloops > 0) : # loop because polynomial degrees could be reduced
-
-        # Fallback to avoid infinite loops: track iterations and cut off after 10.
-        nloops -= 1
-        if nloops == 0:
-            log.warn('Hit iteration limit. Setting degxx=degxy=degyx=degyy=0')
-            degxx, degxy, degyx, degyy = 0, 0, 0, 0
+    n = 0
+    nloops = max(degxx, degyx) + max(degxy, degyy)
+    while True: # loop because polynomial degrees could be reduced
 
         # Try fitting offsets.
         log.info("polynomial fit of measured offsets with degx=(%d,%d) degy=(%d,%d)"%(degxx,degxy,degyx,degyy))
@@ -282,7 +277,7 @@ def fit_trace_shifts(image,args) :
             if merr != 100000. :
                 log.warning("max edge shift error = %4.3f pixels is too large, reducing degrees"%merr)
 
-            if degxy>0 or degyy>0 and degxy>degxx and degyy>degyx : # first along wavelength
+            if (degxy>0 or degyy>0) and (degxy>degxx or degyy>degyx): # first along wavelength
                 if degxy>0 : degxy-=1
                 if degyy>0 : degyy-=1
             else :                                                  # then along fiber
@@ -291,6 +286,11 @@ def fit_trace_shifts(image,args) :
         else :
             # error is ok, so we quit the loop
             break
+
+        # Sanity check to ensure looping is not infinite.
+        n += 1
+        if n > nloops:
+            raise RuntimeError(f'Maximum fit iterations {nloops} exceeded.')
 
     # write this for debugging
     if args.outoffsets :
--- a/py/desispec/scripts/trace_shifts.py
+++ b/py/desispec/scripts/trace_shifts.py
@@ -236,8 +236,15 @@ def fit_trace_shifts(image,args) :
         degyy=0
     
     nloops = 10
-    while(nloops > 0) : # loop because polynomial degrees could be reduced
+    while (nloops > 0) : # loop because polynomial degrees could be reduced
 
+        # Fallback to avoid infinite loops: track iterations and cut off after 10.
+        nloop -= 1
+        if nloop == 0:
+            log.warn('Hit iteration limit. Setting degxx=degxy=degyx=degyy=0')
+            degxx, degxy, degyx, degyy = 0, 0, 0, 0
+
+        # Try fitting offsets.
         log.info("polynomial fit of measured offsets with degx=(%d,%d) degy=(%d,%d)"%(degxx,degxy,degyx,degyy))
         try :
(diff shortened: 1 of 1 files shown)


py/desispec/scripts/trace_shifts.py
```diff
@@ -274,10 +282,10 @@ def fit_trace_shifts(image,args) :
             if merr != 100000. :
                 log.warning("max edge shift error = %4.3f pixels is too large, reducing degrees"%merr)
 
-            if degxy>0 and degyy>0 and degxy>degxx and degyy>degyx : # first along wavelength
+            if degxy>0 or degyy>0 and degxy>degxx and degyy>degyx : # first along wavelength
```
sbailey, line 285: Please add parentheses here to make the order of operations of evaluating the `or` and `and`s very clear.

py/desispec/scripts/trace_shifts.py
```diff
@@ -234,9 +234,17 @@ def fit_trace_shifts(image,args) :
         degxy=0
         degyx=0
         degyy=0
+    
+    nloops = 5
```
sbailey, line 238: I think the intended max number of iterations is supposed to be `args.degxx+args.degxy+args.degyx+args.degyy`, since each iteration through the loop decrements one or more of those until reaching the break condition on line 263:
```
            if degxx==0 and degxy==0 and degyx==0 and degyy==0 :
                break
```
hardcoding nloops=5 means that for larger values of `args.deg..` it could exit the loop earlier than originally intended.  i.e. I think your `and` vs. `or` bugfix below is the relevant change and should fix the known problems.  If you also want to add a belt-and-suspenders safety check to avoid the infinite loop that's ok, but then I think `nloops` should be larger than anything that is supposed to occur, and have nloops=0 trigger an exception rather than a standard `degxx, degxy, degyx, degyy = 0, 0, 0, 0` warning+proceed.

sbailey (commented):
Thanks for this fix.  For the record, please post in the comments a specific case that this fixes, and whether the `and` -> `or` change was the relevant fix, or whether there are any known cases that actually require the nloops cutoff.  I think the and/or fix should restore the code to working order with a guaranteed exit.  If anything is triggering the nloops cutoff then either we have another bug or nloops is too small (or post counter-examples if you disagree).

py/desispec/scripts/trace_shifts.py
sybenzvi: Good catch. I checked the logic more carefully and realized I'd made a mistake; it should actually be
```
if (degxy>0 or degyy>0) and (degxy>degxx or degyy>degyx)
```
to match the intended behavior of the code.

py/desispec/scripts/trace_shifts.py
sybenzvi: Thanks for catching this too. I hadn't considered that we might ever fit very high-order polynomials but it's best to future-proof everything. As written by Julien, the total number of loops as we reduce the order of the polynomials down to zero should be
```
max(degxx, degyx) + max(degxy, degyy)
```
The code now checks that the number of loops does not exceed this value and throws a `RuntimeError` if it does.

sybenzvi:
The problem this PR is meant to solve is described in desihub/nightwatch#281.

Currently at NERSC, exposures 129601-129605 from 20220410 failed to process because of an infinite loop in trace_shift.py. At KPNO, the infinite loop blocked processing of additional exposures and Nightwatch had to be manually killed.

Here is a link to Nightwatch showing the failed processing: https://nightwatch.desi.lbl.gov/20220410/exposures.html

Here is a link demonstrating the fix: https://data.desi.lbl.gov/desi/users/sybenzvi/nightwatch/20220410/exposures.html

py/desispec/scripts/trace_shifts.py
sybenzvi: Just to demonstrate that the new code works, here are two functions that reproduce the old version of the loop and the new one:
```
def reduce_order_old(degxx=2, degyx=2, degxy=2, degyy=2):
    """Old loop from master branch.
    """
    n = 0
    while True:
        print(degxx, degyx, degxy, degyy)
        
        if degxy>0 and degyy>0 and degxy>degxx and degyy>degyx: # first along wavelength
            if degxy>0 : degxy-=1
            if degyy>0 : degyy-=1
        else :                                                  # then along fiber
            if degxx>0 : degxx-=1
            if degyx>0 : degyx-=1
            
        n += 1
        if n > 10:   # For testing, stop at 10 iterations
            break
            
def reduce_order_new(degxx=2, degyx=2, degxy=2, degyy=2):
    """Updated loop from master branch.
    """
    nloops = max(degxx, degyx) + max(degxy, degyy)
    n = 0
    while True:
        print(degxx, degyx, degxy, degyy)
        
        if (degxy>0 or degyy>0) and (degxy>degxx or degyy>degyx): # first along wavelength
            if degxy>0 : degxy-=1
            if degyy>0 : degyy-=1
        else :                                                    # then along fiber
            if degxx>0 : degxx-=1
            if degyx>0 : degyx-=1
            
        if nloops == 0:
            break

        n += 1
        if n > nloops:
            break
```

When running with `degxx=degyx=degxy=degyy=2` (the default arguments) both functions return the same values:
```
reduce_order_old(2,2,2,2)     reduce_order_new(2,2,2,2)
2 2 2 2                       2 2 2 2
1 1 2 2                       1 1 2 2
1 1 1 1                       1 1 1 1
0 0 1 1                       0 0 1 1
0 0 0 0                       0 0 0 0
```
But when `degyy=0` the original loop does not properly decrease to 0 for all orders:
```
reduce_order_old(2,2,2,2)     reduce_order_new(2,2,2,2)
2 2 2 0                       2 2 2 0
1 1 2 0                       1 1 2 0
0 0 2 0                       1 1 1 0
0 0 2 0                       0 0 1 0
0 0 2 0                       0 0 0 0
0 0 2 0
...
```

sbailey (approved):

PR #1743, opened 2022-04-14
Defork proc and proc_joint_fit

marcelo-alvarez:
Removes instances of spawning subprocesses on MPI tasks in desispec.scripts.proc and desispec.scripts.proc_joint_fit, following modifications made for scale runs in the [dmargala-scale-run-wip branch](/desihub/desispec/tree/dmargala-scale-run-wip). 

Deferring testing and leaving this in a draft state until preliminary review is complete -- i.e. @dmargala please comment on whether these changes look correct to you, thanks.


py/desispec/util.py
```diff
@@ -85,22 +86,17 @@ def runcmd(cmd, args=None, inputs=[], outputs=[], clobber=False):
 
     #- run command
     if isinstance(cmd, collections.abc.Callable):
+        #- TODO: should we check outputs before returning when cmd is callable?
```
dmargala, line 89: It's worth considering whether it makes sense to return the result of callable directly or verify that the desired outputs exist like in the subprocess.call usage.

dmargala (commented):
Thanks @marcelo-alvarez . It looks like this captures all the "deforking" changes I made in my scale run branch. I would keep @sbailey's recent changes to `desispec.util.runcmd` but I think that will also require making some changes in proc/proc_joint_fit. I left suggestions for consideration in comments.

py/desispec/util.py
```diff
@@ -85,22 +86,17 @@ def runcmd(cmd, args=None, inputs=[], outputs=[], clobber=False):
...
             return cmd(*args)
-        except Exception as e:
-            import traceback
-            lines = traceback.format_exception(*sys.exc_info())
-            for line in lines:
-                line = line.strip()
-                log.error(f'{line}')
-            log.critical("FAILED {cmd=} with {args=}".format(err, cmd))
-            raise e
-
```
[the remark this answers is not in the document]
marcelo-alvarez: done and done, thanks.

py/desispec/scripts/proc.py
```diff
@@ -1271,7 +1288,14 @@ def list2str(xx) :
...
 
             inputs = [framefile, skyfile, fiberflatfile, stdfile, calibstars]
-            err = runcmd(cmd, inputs=inputs, outputs=[calibfile,])
+            try:
+                cmdargs = cmd.split()[1:]
+                fluxcal_args = desispec.scripts.fluxcalibration.parse(cmdargs)
+                err = runcmd(desispec.scripts.fluxcalibration.main, args=fluxcal_args, inputs=inputs, outputs=[calibfile,])
+            except:
+                #- Catches sys.exit from fluxcalibration.main
+                log.error('fluxcalibration.main failed for {}'.format(os.path.basename(calibfile)))
```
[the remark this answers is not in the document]
marcelo-alvarez: removed

py/desispec/util.py
marcelo-alvarez: it probably doesn't make sense, but leaving this for a future PR since it's not necessary for deforkification

marcelo-alvarez:
> Thanks @marcelo-alvarez . It looks like this captures all the "deforking" changes I made in my scale run branch. I would keep @sbailey's recent changes to `desispec.util.runcmd` but I think that will also require making some changes in proc/proc_joint_fit. I left suggestions for consideration in comments.

Thanks for @dmargala for the very helpful comments. In the end I decided not to modify `util.py` at all, since it was not necessary in order to prevent MPI tasks from spawning subprocesses. 

@sbailey this is ready for you to review.

py/desispec/scripts/proc_joint_fit.py
```diff
@@ -308,7 +308,12 @@ def main(args=None, comm=None):
                 for flat in inflats_for_camera[camera]:
                     cmd += f" {flat} "
                 num_cmd += 1
-                err = runcmd(cmd, inputs=inflats_for_camera[camera], outputs=[ofile, ])
+                cmdargs = cmd.split()[1:]
+                average_fiberflat_args = desispec.scripts.average_fiberflat.parse(cmdargs)
+                try:
+                    err = runcmd(desispec.scripts.average_fiberflat.main, args=[average_fiberflat_args], inputs=inflats_for_camera[camera], outputs=[ofile, ])
```
sbailey, line 314: Minor: Stylistically I like this form where you explicitly call `runcmd` with a function name:
```
runcmd(desispec.scripts.average_fiberflat.main, ...)
```
instead of elsewhere which has a pattern of redefining `cmd` e.g.
```
cmd = "desi_blat_foo ..."
...
cmd = desispec.scripts.procexp.main
runcmd(cmd, ...)
```
The first approach makes it clearer when reading the `runcmd` line that it is being called with a function instead of a string, and avoids redefining what `cmd` means within the local block of code.

sbailey (changes requested):
Please check every case where `desispec.scripts.proc.main` calls `runcmd`.  It looks like many cases here are still not deforked, e.g. assemble_fibermap, desi_inspect_dark, desi_compute_trace_shifts, desi_interpolate_fiber_psf, desi_compute_badcolumn_mask, desi_compute_humidity_corrected_fiberflat, desi_fit_stdstars (when not using mpistdstars), desi_select_calibstars, ...

If I understand correctly, `desispec.util.runcmd` does not check for the existence of outputs when called with a function instead of a string to spawn as a script.  Let's get that fixed here so that runcmd checks inputs and outputs consistently whether it is called with a function or a script; otherwise this deforking is a regression in pipeline robustness by not checking outputs.

py/desispec/scripts/proc_joint_fit.py
marcelo-alvarez: I've changed the relevant `runcmd` calls in this way.

marcelo-alvarez:
> Please check every case where `desispec.scripts.proc.main` calls `runcmd`. It looks like many cases here are still not deforked, e.g. assemble_fibermap, desi_inspect_dark, desi_compute_trace_shifts, desi_interpolate_fiber_psf, desi_compute_badcolumn_mask, desi_compute_humidity_corrected_fiberflat, desi_fit_stdstars (when not using mpistdstars), desi_select_calibstars, ...
> 
@sbailey I've done this, by creating wrappers with the form
```
#!/usr/bin/env python
#
# See top-level LICENSE.rst file for Copyright information
#
# -*- coding: utf-8 -*-

"""
This script does what FUNCNAME does.
"""

import desispec.scripts.FUNCNAME as FUNCNAME

if __name__ == '__main__':
    args = FUNCNAME.parse()
    FUNCNAME.main(args)
```
moving the contents to a new file `desispec/scripts/FUNCNAME.py`, where necessary, creating the following new files
```
desispec/scripts/assemble_fibermap.py
desispec/scripts/badcolumn_mask.py
desispec/scripts/inspect_dark.py
desispec/scripts/interpolate_fiber_psf.py
desispec/scripts/select_calib_stars.py
```
renaming `desispec/bin/assemble_fibermap` to `desispec/bin/desi_assemble_fibermap`, and modifying the relevane `runcmd` calls in `desispec/scripts/proc.py` accordingly.

marcelo-alvarez:
@sbailey I have verified that the cframes files produced with
```
desi_proc -n 20210601 -e 00090659 --traceshift
```
for a science exposure are identical when running this branch and master branch on cori haswell. 

Please merge only after testing independently since these changes touch almost every step of the spectroscopic pipeline.

sbailey:
@marcelo-alvarez please try running tests for each flavor of exposure, i.e. ZERO, DARK, ARC, FLAT, SCIENCE.  e.g. for an arc
```
desi_proc -n 20220401 -e 128286 --batch --traceshift
```
it ends with
```
ERROR:proc.py:1422:main: 25 processing errors; see logs above
...
FAILED: done at Mon May 2 11:35:03 PDT 2022
```
But from the log /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/defork2/run/scripts/night/20220401/arc-20220401-00128286-a0123456789-58502966.log I don't see what actually failed or why.  I suspect that it may be desi_proc mis-interpreting 0 vs. None vs. other runcmd return values which was a false-positive failure mode on a previous desi_proc+runcmd update.

marcelo-alvarez:
> ```
> ERROR:proc.py:1422:main: 25 processing errors; see logs above
> ...
> FAILED: done at Mon May 2 11:35:03 PDT 2022
> ```
> 
> But from the log /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/defork2/run/scripts/night/20220401/arc-20220401-00128286-a0123456789-58502966.log I don't see what actually failed or why. I suspect that it may be desi_proc mis-interpreting 0 vs. None vs. other runcmd return values which was a false-positive failure mode on a previous desi_proc+runcmd update.

@sbailey I found the cause of the errors (incomplete reference to the newly created callable function `desispec.scripts.interpolate_fiber_psf.main` that replaces `desi_interpolate_fiber_psf`). Arcs should run with no errors now.

py/desispec/util.py
```diff
@@ -83,45 +93,53 @@ def runcmd(cmd, args=None, inputs=[], outputs=[], clobber=False):
...
-            for line in lines:
-                line = line.strip()
-                log.error(f'{line}')
-            log.critical(f"FAILED {cmd=} with {args=}")
-            raise e
+    cmd_callable = isinstance(cmd, collections.abc.Callable)
+    if not cmd_callable and args is not None:
+        raise ValueError("Don't provide args unless cmd is a function")
+    if isinstance(args, argparse.Namespace):
+        args = (args)
```
sbailey, line 100: This should be `args = (args,)` to convert to a tuple.  Otherwise it is just interpreted as parentheses and remains a Namespace.  Note that `blat = [blat]` works to convert to a list, but `blat = (blat)` doesn't work to convert to a tuple.  Stylistically I prefer to keep the extra comma in both cases to be clearer (?) that you really mean for this to become a list or tuple.

sbailey, line 100: This should be `args = (args,)` to convert to a tuple.  Otherwise it is just interpreted as parentheses and remains a Namespace.  Note that `blat = [blat]` works to convert to a list, but `blat = (blat)` doesn't work to convert to a tuple.  Stylistically I prefer to keep the extra comma in both cases to be clearer (?) that you really mean for this to become a list or tuple.

py/desispec/scripts/proc_joint_fit.py
```diff
@@ -508,22 +530,25 @@ def main(args=None, comm=None):
...
-                        args=(cmdargs, subcomm), inputs=inputs, outputs=[stdfile]
+                        args=cmdargs, inputs=inputs, outputs=[stdfile], comm=subcomm
                     )
                 except:
                     #- Catches sys.exit from stdstars.main
                     log.error('stdstars.main failed for {}'.format(os.path.basename(stdfile)))
                     err = True
+
             if err:
                 num_err += 1
```
sbailey, line 553: For the case of `srun -n 30 ... desi_proc_joint_fit ...`:
Since `num_cmd` is only incremented for `subcomm.rank==0`, but `num_err` is incremented for every rank, if `runcmd` fails for all ranks this leads to `num_cmd==10` and `num_err==30`, which causes this later check to fail to recognize it as a failure:
```
        if num_err>0 and num_err==num_cmd:
            if rank == 0:
                log.critical('All stdstar commands failed')
            sys.exit(1)
```
and it doesn't properly exit with an error to let slurm know that the job failed.

sbailey, line 553: For the case of `srun -n 30 ... desi_proc_joint_fit ...`:
Since `num_cmd` is only incremented for `subcomm.rank==0`, but `num_err` is incremented for every rank, if `runcmd` fails for all ranks this leads to `num_cmd==10` and `num_err==30`, which causes this later check to fail to recognize it as a failure:
```
        if num_err>0 and num_err==num_cmd:
            if rank == 0:
                log.critical('All stdstar commands failed')
            sys.exit(1)
```
and it doesn't properly exit with an error to let slurm know that the job failed.

sbailey (changes requested):
We're getting closer, but not quite there.  I posted inline comments about some problems I found with my testing.

py/desispec/scripts/proc_joint_fit.py
marcelo-alvarez: duplicate

py/desispec/scripts/proc.py
```diff
@@ -1190,24 +1267,21 @@ def main(args=None, comm=None):
...
+                    log.info(f'Rank {rank=} fitting sp{sp=} stdstars with mpi')
                     err = runcmd(desispec.scripts.stdstars.main,
-                        args=(cmdargs, subcomm), inputs=inputs, outputs=[stdfile]
-                    )
-                except:
-                    #- Catches sys.exit from stdstars.main
-                    log.error('stdstars.main failed for {}'.format(os.path.basename(stdfile)))
-                    err = 1
-
+                        args=(cmdargs, subcomm), inputs=inputs, outputs=[stdfile])
```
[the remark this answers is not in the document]
marcelo-alvarez: missed this one, it's fixed in the last commit, thanks for finding it

py/desispec/util.py
marcelo-alvarez: thanks for finding this, I added the comma

py/desispec/scripts/proc_joint_fit.py
marcelo-alvarez: thanks for catching this. I moved the `num_cmd += 1` outside the if block, since an error code is always returned inside of it.

sbailey:
I updated `runcmd` in this branch to log the commands/functions being run in a cut-and-pastable manner, which involves passing in the original arguments list instead of a Namespace object, and updating the `scripts.blah.main(args)` to accept either.  

New pattern:
```
# in scripts.blah:
def parse(options=None):
    parser = argparse.ArgumentParser()
    ...
    return parser.parse_args(options)  # uses sys.argv is options is None

def main(args=None):
    if not isinstance(args, args.Namespace):
        args = parse(args)
    ....

# in scripts.proc (i.e. desi_proc)
cmdargs = ['-i', 'blat.fits', '-o', 'foo.fits']
result, success = runcmd(desispec.scripts.blat.main, args=cmdargs,
        inputs=['blat.fits'], outputs=['foo.fits'])
```

runcmd will log a message like:
```
INFO:util.py:123:runcmd: RUNNING: desispec.scripts.blat.main(['-i', 'blat.fits', '-o', 'foo.fits'])
```
The part after "RUNNING:" can be cut-and-pasted to a python terminal to run the exact function that was called, analogous to the previous behavior of logging the exact command line script that was spawned.

I also updated runcmd to return both the function return value and a separate boolean success flag.  The original design of passing the unix script return code didn't work so well for runcmd calling functions, some of which return an integer, others `None`, and others something else that isn't supposed to mean an error code.

## End Game for Merging ##

I've testing this branch with ccdcalib(zeros, dark), arc, psfnight, flat, nightlyflat, science steps, and @marcelo-alvarez also tested those steps with an earlier version of this branch.  Since this PR has touched a *lot* of files, our final test will be to use this branch for the daily prod for a night, and if that goes smoothly then merge.

one more thing...

Since this is also final testing for Himalayas, I'm also going to implement an update that changes the poststdstar job from 5 nodes to 1.  I have that in a different branch but need to do some git-fu to merge it in here and re-test.  I've discussed this with @akremin and will coordinate with him for final dailyops testing.

sbailey (approved):

akremin:
I'm seeing things in the logs like: 
`INFO:util.py:182:runcmd: SUCCESS: desispec.scripts.preproc.main(['-i', '/global/cfs/cdirs/desi/spectro/data/20220513/00134514/desi-00134514.fits.fz', '-o', '/global/cfs/cdirs/desi/spectro/redux/daily/preproc/20220513/00134514/preproc-z2-00134514.fits.gz', '--outdir', '/global/cfs/cdirs/desi/spectro/redux/daily/preproc/20220513/00134514', '--cameras', 'z2'])`

which is much better than a memory address as output, but isn't completely copy-paste ready.

My guess is that this is the best we can hope for, which I am perfectly happy with -- it is quite close to what we need. Just making the comment in case this was unexpected.

akremin:
All poststdstar jobs fail with the following:
`ImportError: cannot import name 'exposure_qa' from 'desispec.scripts' (/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/deforkproc/py/desispec/scripts/__init__.py)`

sbailey:
**exposure_qa**: oops, apologies.  I had forgotten to git add scripts/exposure_qa.py, but tests passed with my checkout because the file was there.  I added it, updated the deforkproc installation, and re-tested with the as-installed version instead of my version.  It made it through poststd exposure_qa fine (I put the redrock job itself on hold while the queue is being used for catching up from this problem from last night...)

**cut-and-paste-ability**: indeed, the new log messages are uglier and less user-friendly than the previous script commands, but they are logging exactly what was run (a function, not a script anymore).  And after doing the import statement, the actual function call is cut-and-pasteable from the log into a python prompt (not a bash prompt).  If we find that to be too irritating, we could update it to also print the equivalent command line script to run, with the minor bookkeeping hassle that e.g. `runcmd(desispec.scripts.sky.main, ...)` doesn't directly know that "desi_compute_sky" is the name of the command line script that calls the same function.

sbailey:
It looks like this was used smoothly for nightly processing on 20220514, so I'm going to proceed with merging and tagging for Himalayas.

PR #1750, opened 2022-04-21
Trace shift by default

julienguy:
Set the shift of traces a default option for `desi_proc` (it happened too many times that we forget this option when running tests).

Add option `--no-traceshift`  and print a "deprecated" warning message if `--traceshift` option is set (for backward compatibility).

Tested with the processing of a flat field frame.


sbailey:
and there was much rejoicing

PR #1753, opened 2022-04-26
desi_group_spectra --coadd option

sbailey:
This PR adds the ability for `desi_group_spectra` to directly output a coadd file instead of requiring that we first output an uncoadded spectra file and then run `desi_coadd_spectra` to coadd it.

Example usage, with example outputs in `/global/cscratch1/sd/sjbailey/coadd`, to compare with spectra and coadds in `fuji/tiles/cumulative/100/20210505` (tile based coadds) and `fuji/healpix/sv3/dark/105/10520/` (healpix based coadds):
```
cd $SCRATCH/coadd
export SPECPROD=fuji
rx=$DESI_SPECTRO_REDUX

desi_group_spectra --expfile $rx/fuji/healpix/sv3/dark/105/10520/hpixexp-sv3-dark-10520.csv \
    -o spectra-sv3-dark-10520.fits -c coadd-sv3-dark-10520.fits --healpix 10520 \
    --header SPGRP=healpix SPGRPVAL=10520 HPXPIXEL=10520 HPXNSIDE=64 HPXNEST=True SURVEY=sv3 PROGRAM=dark

desi_coadd_spectra -i spectra-sv3-dark-10520.fits -o altcoadd-sv3-dark-10520.fits

CFRAMES=$(ls $rx/fuji/exposures/20210504/00087236/cframe-[brz]0-00087236.fits $rx/fuji/exposures/20210505/00087361/cframe-[brz]0-00087361.fits)
desi_group_spectra --inframes $CFRAMES --onetile \
    -o spectra-0-100-thru20210505.fits -c coadd-0-100-thru20210505.fits \
    --header SPGRP=cumulative SPGRPVAL=20210505 NIGHT=20210505 TILEID=100 SPECTRO=0 PETAL=0

desi_coadd_spectra -i spectra-0-100-thru20210505.fits -o altcoadd-0-100-thru20210505.fits --onetile
```

Side effects of code reorganization:
* previously uncoadded spectra and coadds were written by different functions, despite the fact that they are supposed to be following the same format; now they both use `desispec.io.write_spectra`.
* both spectra and coadds now keep the masks as int32 instead of compressed image HDUs (orig spectra, didn't work well with IDL) or unsigned uint32 (orig coadds, didn't work well with `astropy.io.fits.read(..., memmap=True)`)
* a coadd from `desi_group_spectra` directly coadds the original float64 data, while writing to disk and coadding via `desi_coadd_spectra` casts via float32 and back to float64 resulting in differences at the level of 1e-5.
* original spectra files didn't have explicit `encode_table` protection for unicode str vs. bytes when writing tables; now they do.  Original coadds already had that.

### DIFF
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -159,7 +159,8 @@ def write_spectra(outfile, spec, units=None):
         scores_tbl = encode_table(spec.scores)  #- unicode -> bytes
         scores_tbl.meta['EXTNAME'] = 'SCORES'
         all_hdus.append( fits.convenience.table_to_hdu(scores_tbl) )
-        if spec.scores_comments is not None : # add comments in header
+        # add comments in header
+        if hasattr(spec, 'scores_comments') and spec.scores_comments is not None:
             hdu=all_hdus['SCORES']
             for i in range(1,999):
                 key = 'TTYPE'+str(i)
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -73,7 +73,7 @@ def write_spectra(outfile, spec, units=None):
     all_hdus.append(fits.PrimaryHDU(header=hdr))
 
     # Next is the fibermap
-    fmap = spec.fibermap.copy()
+    fmap = encode_table(spec.fibermap.copy())
     fmap.meta['EXTNAME'] = 'FIBERMAP'
     fmap.meta['LONGSTRN'] = 'OGIP 1.0'
     add_dependencies(fmap.meta)
@@ -96,7 +96,7 @@ def write_spectra(outfile, spec, units=None):
 
     # Optional: exposure-fibermap, used in coadds
     if spec.exp_fibermap is not None:
-        expfmap = spec.exp_fibermap.copy()
+        expfmap = encode_table(spec.exp_fibermap.copy())
         expfmap.meta["EXTNAME"] = "EXP_FIBERMAP"
         with warnings.catch_warnings():
             #- nanomaggies aren't an official IAU unit but don't complain
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -141,7 +141,7 @@ def write_spectra(outfile, spec, units=None):
         if spec.mask is not None:
             # hdu = fits.CompImageHDU(name="{}_MASK".format(band.upper()))
             hdu = fits.ImageHDU(name="{}_MASK".format(band.upper()))
-            hdu.data = spec.mask[band].astype(np.uint32)
+            hdu.data = spec.mask[band].astype(np.int32)
             all_hdus.append(hdu)
 
         if spec.resolution_data is not None:
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -256,7 +256,8 @@ def read_spectra(infile, single=False):
             if type == "WAVELENGTH":
                 if wave is None:
                     wave = {}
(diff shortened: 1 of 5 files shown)


akremin:
I think you made the right choices in making the outputs as widely usable as possible with minimal formatting changes. The code updates all look good to me. I still want to check the files for differences, which is why I'm not approving this quite yet. If this is urgent and you already verified they're identical (up to the reported 1e-5 deviations), then you're welcome to merge.

akremin:
In looking at extensions and columns in the coadd files, I see that the healpix based coadd and Fuji coadd match, but that the healpix based `altcoadd` is missing the following 11 columns in the fibermap table:
```
PETAL_LOC           
DEVICE_LOC          
LOCATION            
FIBER               
LAMBDA_REF          
FIBERASSIGN_X       
FIBERASSIGN_Y       
PRIORITY            
TILEID              
MEAN_FIBER_X        
MEAN_FIBER_Y
```

akremin:
I also notice that the `SCORES` HDU number has changed in the tile based spectra files from Fuji to this PR:
Fuji: `SCORES` is ext 2 (0-indexed)
This PR: `SCORES` is ext 17 (0-indexed)

People are supposed to reference extensions by name, but there are likely to be some that use numbers that may be impacted by this.

sbailey:
Missing columns: that was my error in the commands, forgetting to include the `desi_coadd_spectra --onetile` option to let it know that all these spectra are from a single tile and that it should generate the columns that only make sense if working with a single tile.  That option is not new and the pipeline currently uses it for tile-based coadds; I just forgot to include it here.  I corrected the instructions and re-ran the tile-based altcoadd and now it has those columns.

HDU number of the SCORES table: this is side effect of the previous code using two different I/O routines for writing uncoadded spectra vs. coadded spectra.  e.g. in fuji, spectra-0-100-thru20210505.fits has SCORES as HDU 2 while coadd-0-100-thru20210505.fits has SCORES as HDU 18.  This PR uses `desispec.io.spectra.write_spectra` for both, resulting in SCORES always being the last HDU.  I agree that this might break someone's code, but as you note we've been consistent about telling people they should use name not number (for reasons like this), and this PR also preserves the order for the more commonly used coadd file to minimize the impact.

FWIW, other known header differences are the new files being more consistent with having DATASUM, CHECKSUM, LONGSTRN, and ENCODING.

sbailey:
Somewhat pedantic: I double checked the underlying cause of the `desi_group_spectra --coadd ..." vs. `desi_coadd_spectra --outfile ..." coadd differences.  It's not because of float64 -> float32 -> float64 casting, but rather the other way around: the input cframe files are in float32, `desi_group_spectra` keeps them in float32 and thus coadds them in float32, while `desi_coadd_spectra` was upcasting to float64 to do the coadd calculation and then writing back out as float32, i.e. float32 -> float64 -> float32.

Normally I advocate for always doing calculations in float64 to minimize roundoff error.  In this case I don't think it matters since we are coadding values of a similar magnitude and observe differences of ~1e-5 for float32 vs. float64 coadd calculations.

Thus I updated `desi_coadd_spectra` to stay in float32 as well, partially for consistency with `desi_group_spectra --coadd`, and partially to save 2x in memory which may be more pragmatically important.

If anyone feels strongly that the coadd should happen in float64, I could revert the change and do the opposite, making `desi_group_spectra --coadd` upcast to float64 as well, at the cost of 2x more memory usage.

julienguy (approved):
I ran the test and verified for both pairs of coadds that they were identical (wave,flux,ivar,mask,fibermap).
I agree there is no point to temporarily casting from float32 to float64 just for the coaddition.

Issue #1752, opened 2022-04-26
tile_qa_plot: all fibers flagged by LOW_EFFTIME for incomplete tiles


araichoor:
Because of the recent inclusion of the `LOW_EFFTIME` flag in the `bad_qafstatus_mask`:
https://github.com/desihub/desispec/blob/aa49cfbf92dc039a2e7cc1dde87da541398b528a/py/desispec/data/qa/qa-params.yaml#L40
done in PR https://github.com/desihub/desispec/pull/1722,

when a tile is not completed yet, all fibers are flagged as bad.
this makes the `tile_qa_plot.make_tile_qa_plot()` routine crash.

For instance `/global/cfs/cdirs/desi/spectro/redux/daily/tiles/cumulative/10788/20220425/logs/tile-qa-10788-thru20220425.log`:
```
...
ERROR:desi_tile_qa:107:_wrap_make_tile_qa_plot: make_tile_qa_plot raised an exception:
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/master/bin/desi_tile_qa", line 102, in _wrap_make_tile_qa_plot
    figfile = make_tile_qa_plot(qafitsfile, specprod_dir)
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/master/py/desispec/tile_qa_plot.py", line 1294, in make_tile_qa_plot
    nqso_rr = int(zhists[sel].sum() * istracer.sum())
ValueError: cannot convert float NaN to integer
```

The crash is at the step of computing `nn_qsorr`.

A simple fix is to add a "if ... else..." checking for that case, which would just keep `nqso_rr = 0`:
```
                if istracer.sum() == 0:
                    log.warning("istracer.sum() == 0 for QSO, probably due to LOW_EFFTIME (TBC); nqso_rr = 0")
                else:
                    nqso_rr = int(zhists[sel].sum() * istracer.sum())
```
That allows the code to finish, and to produce a plot like this:
![tile-qa-10788-thru20220425](https://user-images.githubusercontent.com/61986357/165357391-fcdbafbd-f3ee-4269-84c7-e50a9f6f1c29.png)

Would we be satisfied this simple fix? (@julienguy)

[...]

PR #1754, opened 2022-04-27
bugfix: handle nqso_rr if zero valid fiber

araichoor:
This PR is a simple fix for issue https://github.com/desihub/desispec/issues/1752 in tile_qa_plot (because of recent addition of `LOW_EFFTIME` in the `bad_qafstatus_mask`, zero fiber will be valid for not-yet-completed tiles, which make the code crash).
It just keeps `nqso_rr=0` in that case.

### DIFF
--- a/py/desispec/tile_qa_plot.py
+++ b/py/desispec/tile_qa_plot.py
@@ -1292,6 +1292,7 @@ def make_tile_qa_plot(
 
             if tracer=="QSO" :
                 if istracer.sum() == 0:
+                    nqso_rr = 0
                     log.warning("istracer.sum() == 0 for QSO, probably due to LOW_EFFTIME (TBC); nqso_rr = 0")
                 else:
                     nqso_rr = int(zhists[sel].sum() * istracer.sum())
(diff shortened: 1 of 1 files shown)


py/desispec/tile_qa_plot.py
```diff
@@ -1291,7 +1291,10 @@ def make_tile_qa_plot(
             n_valid += zhists[sel].sum() * istracer.sum()
 
             if tracer=="QSO" :
-                nqso_rr = int(zhists[sel].sum() * istracer.sum())
+                if istracer.sum() == 0:
+                    log.warning("istracer.sum() == 0 for QSO, probably due to LOW_EFFTIME (TBC); nqso_rr = 0")
```
sbailey, line 1295: Minor: I suggest that you set `nqso_rr=0` here instead of relying upon that being set before entering the loop over tracers and relying upon this code only being run for tracer="QSO".  I believe the code as-is works, but it potentially fragile to future updates.

sbailey:
Thanks.  This fixes the immediate problem of the crashes, but as you note in #1752 a side-effect of LOW_EFFTIME in bad_qafstatus_mask means that incomplete tiles have less informative QA (no n(z), no meaningful efftime vs. x,y).  Let's merge this as-is and then re-evaluate whether we need to do something to improve the QA of incomplete tiles.

PR #1756, opened 2022-04-29
gzip intermediate files

sbailey:
This PR updates the pipeline to gzip the following intermediate files:

| File   |     OrigMB  |   GzipMB | MB/exp saved |
| ---- | ----------- | -------- | -------------- |
| preproc  |  195  |      166  |   870 |
| frame     |   77    |     63   |  420 |
| sframe   |   77    |     63   |  420 |
| cframe  |    77    |     63   |  420 |
| fluxcalib |  16    |    4.5  |   345 |
| sky        | 21     |   4.7   |  489 |
| stdstars  |  5.2   |    4.2  |    30 |

Adding up to a savings of ~2 GB/exposure (20%).

It additionally gzips the **spectra** files but not the more frequently used **coadd** and **redrock** files.  The size and savings depends upon the number of input exposures, but in a set of 4 sample tiles from 20220401 the savings was 200 - 500 MB/tile (times multiple flavors of redshift grouping).

For compatibility with previous productions, `desispec.io.read_blah(filename)` functions are updated so that you can provide either the gzipped or non-gzipped filename and it will use `checkgzip` to find whichever exists.  Similarly the `desispec.io.write_blah(filename)` functions now consistently use `get_tempfilename` so that the temp filename still ends with .gz (not .gz.tmp) and the writer will write gzipped or not depending upon whether the final output filename ends with .gz or not.  That ended up touching a lot of source files.

Two test prods are at `$CFS/desi/users/sjbailey/spectro/redux/gzip` and `nogzip`. I verified that the coadds (the first non-gzipped step) are identical.
**Detail**: After the initial runs, I found that this PR had a merge conflict with master due to desi_group_spectra PR #1753, so after resolving that I re-ran tile 5683, resulting in the format changes listed in that PR (SCORES HDU moved, MASK int32 instead of compressed uint32, coadd in float32 instead of float64).

I'm opening this PR for review now, but a few more items I'd like to test before final merge:
  - [x] healpix spectra+coadds+redshifts
  - [x] desi_tsnr_afterburner

@akremin I did test `desi_proc_dashboard`, but I could use some help in identifying any other pipeline runner related functions that might be looking for the existence of files to make sure they are robust to gzip or not.

### DIFF
--- a/bin/assemble_fibermap
+++ b/bin/assemble_fibermap
@@ -33,6 +33,7 @@ args = parser.parse_args()
 
 import os, sys, json
 from desispec.io.fibermap import assemble_fibermap
+from desispec.io.util import get_tempfilename
 from desispec.pixgroup import fibermap2tilepix
 from desiutil.log import get_logger
 
@@ -50,7 +51,7 @@ fibermap = assemble_fibermap(args.night, args.expid, badamps=args.badamps,
         badfibers_filename=args.badfibers, force=args.force,
         allow_svn_override=(not args.no_svn_override) )
 
-tmpfile = args.outfile+'.tmp'
+tmpfile = get_tempfilename(args.outfile)
 fibermap.writeto(tmpfile, output_verify='fix+warn', overwrite=args.overwrite, checksum=True)
 os.rename(tmpfile, args.outfile)
 log.info(f'Wrote {args.outfile}')
@@ -58,7 +59,7 @@ log.info(f'Wrote {args.outfile}')
 if args.tilepix:
     tileid = fibermap['FIBERMAP'].header['TILEID']
     tilepix = {str(tileid): fibermap2tilepix(fibermap['FIBERMAP'].data)}
-    tmpfile = args.tilepix+'.tmp'
+    tmpfile = get_tempfilename(args.tilepix)
     with open(tmpfile, 'w') as fp:
         json.dump(tilepix, fp)
     os.rename(tmpfile, args.tilepix)
--- a/bin/desi_tile_qa
+++ b/bin/desi_tile_qa
@@ -19,6 +19,7 @@ import fitsio
 
 from desiutil.log import get_logger
 from desispec.io import specprod_root,findfile,read_tile_qa,write_tile_qa
+from desispec.io.util import get_tempfilename
 from desispec.tile_qa import compute_tile_qa
 from desispec.util import parse_int_args
 
@@ -255,7 +256,7 @@ def main():
         print(table)
         print()
 
-        tmpfile = args.outfile + '.tmp'
+        tmpfile = get_tempfilename(args.outfile)
         table.write(tmpfile, overwrite=True, format='fits')
         os.rename(tmpfile, args.outfile)
         log.info("wrote {}".format(args.outfile))
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -633,7 +633,7 @@ def func(prod,night,expid,camera,recompute,alpha_only,details_dir) :
     """
     log = get_logger()
 
-    cframe_filename = '{}/exposures/{}/{:08d}/cframe-{}-{:08d}.fits'.format(prod, night, expid, camera, expid)
+    cframe_filename = findfile('cframe', night, expid, camera, specprod_dir=prod)
     if not os.path.isfile(cframe_filename) :
         return None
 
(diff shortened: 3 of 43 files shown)


akremin:
Stating for the record as to why this in complicated (Stephen is well aware of this, of course): this PR creates an bifurcation in naming convention that breaks the ability of the utility function `findfile()` to return the uniquely correct pathname for a file. Many functions use `findfile()` -- a naive search found 400+ usages in ~36 files in desispec. With the function changed to return "*.gz" versions of the file, all of those function calls need to be checked and possibly updated to be robust for existing data in the old format.

Stephen -- As I lay out above there are a lot more places we need to check in order to be confident that our "do the right thing and skip existing files" assumptions in desispec hold in this transition. 

The first places to come to mind are: 
- `check_for_outputs_on_disk` code in the pipeline https://github.com/desihub/desispec/blob/a421f192c679249400873e44230eeb44b28e9110/py/desispec/workflow/procfuncs.py#L74
- The tsnr afterburner
- nightqa (you already addressed some)
- Likely many others.

akremin:
One potential solution workaround would be to have `findfile()` itself look on disk for both the regular and gzipped version. If either exists, it returns that. If neither exists, it returns the gzipped version.

That would allow usages of `findfile()` for existing data to receive the correct filename (uncompressed or gzipped version), while retaining the new definition for functions looking for what a file should be called once it is produced in the future (gzipped only).

The upside is that this would be a lot less code edits. The downside is `findfile()` would now actually check for files rather than just being a dictionary of naming conventions, so it's feature creep on the original design.

sbailey:
@akremin good point about not wanting to break the common pattern of using findfile + os.path.exists.  I updated findfile using your suggestion to return the gzip (or not) version if the canonical location doesn't exist but the alternate one does.  Although the current case is always "canonical is gzipped and old is not", it also does the opposite check "canonical is not gzipped but I'll check for a gzipped file anyway", which gives us some future proofing at the expense of extra metadata I/O.

To somewhat the mitigate the extra metadata check load, I also added a findfile(..., return_exists) option to return (filepath, exists) .  I didn't extensively update every case of findfile + os.path.exists to use that instead, but the option is available to use to avoid a second existence check since now findfile itself already knows.

Since this is enough new moving parts, I'll run another mini test prod "gzip2" to make sure I didn't break something else.

sbailey:
An unanticipated side-effect of gzipping these files is that tile-qa now takes much longer, ~10-12 minutes per input exposure instead of ~2-3 minutes.  For a multi-exposure tile, this becomes a significant portion of the total runtime, causing jobs to timeout.

From the logs it appears that each input cframe is read 4 times: twice to read the full frame at ~9sec/read and twice to read just the fibermap at ~2sec/read, adding up to 11 minutes of just reading cframes (!).  I'd like to optimize that before merging this branch.

araichoor:
I ve not closely followed this development, but I see you mention the tile-qa: in case some of my input would be needed, please let me know.

sbailey:
@akremin this PR is ready for re-review.  Summarizing changes since your last comment:
* `findfile` now does a file existence check and will return the pre-existing version (gzipped or not) regardless of what the canonical location would be for new files.  If the file does not exist, it returns the canonical location that should be used for future files.
* I counter-balanced the slower gzip reading by
  * `exposure_qa` now opens each input file only once
  * `read_frame` now uses fitsio instead of astropy.io.fits
  * `read_fibermap` now also works with an already opened file pointer (`fitsio.FITS` or `astropy.io.fits.HDUList`) in addition to a filename.  This enables `read_frame` to use `read_fibermap` to standardize fibermap reading without having to open the gzipped file a second time.

I believe timeouts were due to fluctuations in NERSC I/O performance.  The latest version of this branch took <9m for tile 5683 cumulative redshifts.  Previously it had fluctuated to >20m and timed out, but even the daily prod for that tile with no gzipping took 19.3m.  In a separate PR I'll move the exposure QA into the poststdstar job instead of being triggered by tileqa; that will additionally help with the standardization of the timing.

I may add some more unit tests, but please re-review when you can.

sbailey:
FTR: I also updated read_spectra to use fitsio, for a 10x speedup.

akremin (approved):
I verified that `desispec.workflow.procfuncs.check_for_outputs_on_disk()` uses `findfile()` and should therefore either the gzipped or non-gzipped version of the files. The redshift files are the only case where that isn't true, but those aren't gzipped.
Once the unit test for `findfile()` is done (discussed with Stephen off line), I think we can go ahead and merge this. The `daily` prod will be a good stress test and we can resolve any corner cases as they arise.

sbailey:
Thanks for final checks @akremin .  I added unit tests for `findfile` finding pre-existing files whether or not they are the canonical gzip-or-not, and unit tests on `read_frame` and `read_spectra` finding whichever one exists.  I also grepped for our usage of glob and I'm pretty sure that we have all of the cases like `glob('spectra*.fits*')` correct.

I'll merge now and update NERSC so that tonight's prod will have consistent gzipping for the entire night.

PR #1765, opened 2022-05-06
Assess the redshift quality of DESI galaxies and quasars

rongpu:
This PR adds a script that assesses the redshift quality (and performs quasar classification) using the criteria described in the DESI target selection papers. @julienguy first suggested creating this script, and @araichoor @echaussidon and @ashleyjross helped validated the script.

### DIFF
--- a/py/desispec/validredshifts.py
+++ b/py/desispec/validredshifts.py
@@ -1,0 +1,141 @@
+import os, warnings
+import numpy as np
+from astropy.table import Table, hstack, join
+import fitsio
+
+
+# Example usage:
+# redrock_path = '/global/cfs/cdirs/desi/spectro/redux/guadalupe/tiles/cumulative/1392/20210517/redrock-5-1392-thru20210517.fits'
+# cat = validate(redrock_path, return_target_columns=True)
+
+def validate(redrock_path, fiberstatus_cut=True, return_target_columns=False, extra_columns=None, emline_path=None, ignore_emline=False):
+
+    '''
+    Validate the redshift quality with tracer-dependent criteria for redrock+afterburner results.
+
+    Args:
+       redrock_path: str, path of redrock FITS file
+
+    Options:
+       fiberstatus_cut: bool (default True), if True, impose requirements on COADD_FIBERSTATUS and ZWARN
+       return_target_columns: bool (default False), if True, include columns that indicate if the object belongs to each class of DESI targets
+       extra_columns: list of str (default None), additional columns to include in the output
+       emline_path: str (default None), specify the location of the emline file; by default the emline file is in the same directory as the redrock file
+       ignore_emline: bool (default False), if True, ignore the emline file and do not validate the ELG redshift
+    '''
+
+    if not return_target_columns:
+        output_columns = ['good_lrg', 'good_elg', 'good_qso']
+    else:
+        output_columns = ['good_lrg', 'good_elg', 'good_qso', 'LRG', 'ELG', 'QSO', 'ELG_LOP', 'ELG_HIP', 'ELG_VLO', 'BGS_ANY', 'BGS_FAINT', 'BGS_BRIGHT']
+
+    if ignore_emline:
+        output_columns.remove('good_elg')
+
+    if extra_columns is None:
+        extra_columns = ['TARGETID', 'Z', 'ZWARN', 'COADD_FIBERSTATUS']
+    output_columns = output_columns + list(np.array(extra_columns)[~np.in1d(extra_columns, output_columns)])
+
+    ############################ Load data ############################
+
+    columns_redshifts = ['TARGETID', 'CHI2', 'Z', 'ZERR', 'ZWARN', 'SPECTYPE', 'DELTACHI2']
+    columns_fibermap = ['TARGETID', 'COADD_FIBERSTATUS', 'TARGET_RA', 'TARGET_DEC', 'DESI_TARGET', 'BGS_TARGET']
+    columns_emline = ['TARGETID', 'OII_FLUX', 'OII_FLUX_IVAR']
+    columns_qso_mgii = ['TARGETID', 'IS_QSO_MGII']
+    columns_qso_qn = ['TARGETID', 'Z_NEW', 'ZERR_NEW', 'IS_QSO_QN_NEW_RR', 'C_LYA', 'C_CIV', 'C_CIII', 'C_MgII', 'C_Hbeta', 'C_Halpha']
+
+    dir_path = os.path.dirname(redrock_path)
+    qso_mgii_path = os.path.join(dir_path, os.path.basename(redrock_path).replace('redrock-', 'qso_mgii-'))
+    qso_qn_path = os.path.join(dir_path, os.path.basename(redrock_path).replace('redrock-', 'qso_qn-'))
+    if emline_path is None:
+        emline_path = os.path.join(dir_path, os.path.basename(redrock_path).replace('redrock-', 'emline-'))
+
+    tmp_redshifts = Table(fitsio.read(redrock_path, ext='REDSHIFTS', columns=columns_redshifts))
+    tmp_fibermap = Table(fitsio.read(redrock_path, ext='FIBERMAP', columns=columns_fibermap))
+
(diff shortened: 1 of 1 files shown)


julienguy:
Thanks. I am leaving this open today to give a chance for others to comment and will then merge.

py/desispec/validredshifts.py
```diff
@@ -0,0 +1,152 @@
...
+
+    Options:
+       fiberstatus_cut: bool (default True), if True, impose requirements on COADD_FIBERSTATUS and ZWARN
+       return_target_columns: bool (default False), if True, include columns that indicate if the object belongs to each class of DESI targets
+       extra_columns: list of str (default None), additional columns to include in the output
+       emline_path: str (default None), specify the location of the emline file; by default the emline file is in the same directory as the redrock file
+       ignore_emline: bool (default False), if True, ignore the emline file and do not validate the ELG redshift
+    '''
+
+    output_columns = ['good_bgs', 'good_lrg', 'good_elg', 'good_qso']
```
sbailey, line 35: elsewhere throughout the code we use UPPERCASE column names, and here if you include `return_target_columns=True` you end up with a table with a mix of UPPER and lowercase.  Let's use UPPERCASE everywhere, even if it is a bit ugly retro.

py/desispec/validredshifts.py
```diff
@@ -0,0 +1,152 @@
...
+def get_good_fiberstatus(cat, isqso=False):
+    if not isqso:
+        good_fiberstatus = cat['COADD_FIBERSTATUS']==0
+    else:
+        good_fiberstatus = (cat['COADD_FIBERSTATUS']==0) | (cat['COADD_FIBERSTATUS']==8388608) | (cat['COADD_FIBERSTATUS']==16777216)
+    good_fiberstatus &= cat['ZWARN'] & 2**9 == 0  # NO DATA flag
+    return good_fiberstatus
+
+
+def validate(redrock_path, fiberstatus_cut=True, return_target_columns=False, extra_columns=None, emline_path=None, ignore_emline=False):
```
sbailey, line 20: For flexibility, we usually try to split algoritms from I/O.  Please split this into two functions:
one function that takes the redrock_path + other options, reads the files it needs, and then calls a second function that takes tables as inputs and does the filtering without doing any I/O.

py/desispec/validredshifts.py
```diff
@@ -0,0 +1,152 @@
...
+import numpy as np
+from astropy.table import Table, hstack, join
+import fitsio
+
+
+def get_good_fiberstatus(cat, isqso=False):
+    if not isqso:
+        good_fiberstatus = cat['COADD_FIBERSTATUS']==0
+    else:
+        good_fiberstatus = (cat['COADD_FIBERSTATUS']==0) | (cat['COADD_FIBERSTATUS']==8388608) | (cat['COADD_FIBERSTATUS']==16777216)
```
sbailey, line 15: here and elsewhere, please use `desispec.maskbits.fibermask` etc. to create the masks by name instead of hardcoding magic numbers like 8388608 .

py/desispec/validredshifts.py
```diff
@@ -0,0 +1,152 @@
...
+        # Bitmask definitions: https://github.com/desihub/desitarget/blob/master/py/desitarget/data/targetmask.yaml
+        cat['LRG'] = cat['DESI_TARGET'] & 2**0 > 0
+        cat['ELG'] = cat['DESI_TARGET'] & 2**1 > 0
+        cat['QSO'] = cat['DESI_TARGET'] & 2**2 > 0
+        cat['ELG_LOP'] = cat['DESI_TARGET'] & 2**5 > 0
+        cat['ELG_HIP'] = cat['DESI_TARGET'] & 2**6 > 0
+        cat['ELG_VLO'] = cat['DESI_TARGET'] & 2**7 > 0
+        cat['BGS_ANY'] = cat['DESI_TARGET'] & 2**60 > 0
+        cat['BGS_FAINT'] = cat['BGS_TARGET'] & 2**0 > 0
+        cat['BGS_BRIGHT'] = cat['BGS_TARGET'] & 2**1 > 0
```
sbailey, line 99: Please use `desitarget.targetmask.desi_mask` and `bgs_mask` here instead of hardcoded numbers.

sbailey (changes requested):
Thanks for adding this.  I added a few cleanup requests inline.

rongpu:
@sbailey Thanks for checking the script and for the suggestions. I have made changes as you suggested. Let me know if I should make any other changes. (I simply commented out the lines with the hardcoded numbers, so I can still use the script on my laptop where I don't have desitarget and desispec installed.)

py/desispec/validredshifts.py
```diff
@@ -0,0 +1,190 @@
...
+    if not ignore_emline:
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore')
+            cat['GOOD_ELG'] = (cat['OII_FLUX']>0) & (cat['OII_FLUX_IVAR']>0)
+            cat['GOOD_ELG'] &= np.log10(cat['OII_FLUX'] * np.sqrt(cat['OII_FLUX_IVAR'])) > 0.9 - 0.2 * np.log10(cat['DELTACHI2'])
+        if fiberstatus_cut:
+            cat['GOOD_ELG'] &= get_good_fiberstatus(cat)
+
+    # QSO - adopted from the code from Edmond
+    # https://github.com/echaussidon/LSS/blob/8ca53f4c38cfa29722ee6958687e188cc894ed2b/py/LSS/qso_cat_utils.py#L282
```
sbailey, line 173: Thanks for linking to this LSS qso_cat_utils code with the same bits, but are these defined anywhere as a bitmask object, yaml/json config file, or something that maps name <-> bitnumber ?  Is there a repo that "owns" the concept of QSO_MASKBITS?  Having bit definitions buried in code is problematic for long term maintenance.

sbailey, line 173: Thanks for linking to this LSS qso_cat_utils code with the same bits, but are these defined anywhere as a bitmask object, yaml/json config file, or something that maps name <-> bitnumber ?  Is there a repo that "owns" the concept of QSO_MASKBITS?  Having bit definitions buried in code is problematic for long term maintenance.

rongpu: @echaussidon provided the script and may be able to answer this. My thinking is that QSO_MASKBITS is only an intermediate value and not part of the output, and its definition is quite simple (it's only those few lines), so perhaps it's not so critical to have it as an object or yaml/json file.

echaussidon: @rongpu, @sbailey: QSO_MASKBITS is to track from which criteria the QSO is selected. There is no yaml file for the moment. I can create one (like the desitarget maskbit) and add it in the LSS repo.

rongpu: @sbailey could you advise what should be done here? E.g., I could simply remove QSO_MASKBITS and rewrite it as a series of AND statements.

sbailey:
I kept hoping that we'd find a better solution for jointly tracking the QSO_MASKBITS here vs. LSS/qso_cat_utils.py, but "perfect is the enemy of good" so I'll just merge this so that we can include it with the Iron tags, even if it has hardcoded bit numbers that have to stay in sync with an external package.

ashleyjross:
I don't think there is even any need for anything to stay in sync (at least no more for QSO than any other tracer). As written, it is really just a convenient way to write a series of "or" statements (like Rongpu suggested). So, really, it is just like the definitions for any of the tracers. Here, we've hardcoded the criteria for selecting good redshifts, circa 2022 (for all tracers). And, this is a big improvement on what we were doing before!

geordie666:
Note that I added the infrastructure for including masks in a yaml file as part of [LSS PR #163](https://github.com/desihub/LSS/pull/63). 

The relevant yaml file is https://github.com/desihub/LSS/blob/main/py/LSS/data/masks.yaml. It would be very straightforward to include a `qso_mask` bitmask in addition to the `lrg_mask` and `elg_mask` in this yaml file. 

The formalism is borrowed from `desiutils`, so should be quite familiar, e.g.:
```
from LSS.masks import qso_mask as mx
mask_info = mx.names
mask_names = mx.names()
mask_bit_value = mx["BIT"]
is_bit_set = some_table["SOME_COLUMN"] & mx["IS_QSO_QN_NEW_RR"] != 0
```

PR #1767, opened 2022-05-10
get expids from all spectra files

araichoor:
This PR makes `tile_qa.py` and `tile_qa_plot.py` getting the exposure list from a tile from reading *all* the spectra files, not just the first one.

It fixes the issue identified in https://github.com/desihub/desisurveyops/issues/44 for tileid=42426.
For that tile:
- two exposures where made (130563 on 20220416, and 130673 on 20220417);
- but petal=0 is discarded in the second exposure 130673.
With the current master, the code infers that there is just exposure, 130563, as it reads `/global/cfs/cdirs/desi/spectro/redux/daily/tiles/cumulative/42426/20220417/spectra-0-42426-thru20220417.fits`.

The qa plot looks like this:
tile-qa-42426-thru20220417-master

And no `/global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220417/00130673/exposure-qa-00130673.fits` file is generated.

With this PR, the qa plot look like this:
tile-qa-42426-thru20220417-pr

and a `exposure-qa-00130673.fits` is generated.

### DIFF
--- a/py/desispec/tile_qa.py
+++ b/py/desispec/tile_qa.py
@@ -51,7 +51,7 @@ def compute_tile_qa(night, tileid, specprod_dir, exposure_qa_dir=None, group='cu
         log.error(f"no spectra files in {tiledir}")
         return None, None
 
-    fmap=read_fibermap(spectra_files[0])
+    fmap = vstack([read_fibermap(spectra_file) for spectra_file in spectra_files])
     expids=np.unique(fmap["EXPID"])
     lexpids=list(expids)
     log.info(f"for tile={tileid} night={night} expids={lexpids}")
--- a/py/desispec/tile_qa.py
+++ b/py/desispec/tile_qa.py
@@ -44,15 +44,67 @@ def compute_tile_qa(night, tileid, specprod_dir, exposure_qa_dir=None, group='cu
 
     log=get_logger()
 
-    # get list of exposures used for the tile
+    # tile folder
     tiledir=f"{specprod_dir}/tiles/{group}/{tileid:d}/{night}"
-    spectra_files=sorted(glob.glob(f"{tiledir}/spectra-*-{tileid:d}-*{night}.fits*"))
-    if len(spectra_files)==0 :
-        log.error(f"no spectra files in {tiledir}")
-        return None, None
 
-    fmap = vstack([read_fibermap(spectra_file) for spectra_file in spectra_files])
-    expids=np.unique(fmap["EXPID"])
+    # collect fibermaps and scores of all coadds
+    coadd_files=sorted(glob.glob(f"{tiledir}/coadd-*-{tileid:d}-*{night}.fits*"))
+
+
+    fibermaps=[]
+    scores=[]
+    redshifts=[]
+    exp_fibermaps=[] # fibermaps of all frames
+    ### zqsos=[]
+    for coadd_file in coadd_files :
+        log.info("reading {}".format(coadd_file))
+        fibermaps.append(_rm_meta_keywords(Table.read(coadd_file,"FIBERMAP")))
+        exp_fibermaps.append(_rm_meta_keywords(Table.read(coadd_file,"EXP_FIBERMAP")))
+        scores.append(_rm_meta_keywords(Table.read(coadd_file,"SCORES")))
+
+
+        redrock_file = replace_prefix(coadd_file, "coadd", "redrock")
+        extname="REDSHIFTS"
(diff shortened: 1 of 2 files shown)


py/desispec/tile_qa.py
```diff
@@ -51,7 +51,7 @@ def compute_tile_qa(night, tileid, specprod_dir, exposure_qa_dir=None, group='cu
         log.error(f"no spectra files in {tiledir}")
         return None, None
 
-    fmap=read_fibermap(spectra_files[0])
+    fmap = vstack([read_fibermap(spectra_file) for spectra_file in spectra_files])
```
sbailey, line 54: Since the spectra files are big and now gzipped, reading the fibermap from them can be slow and sensitive to current I/O performance.  Consider reading from the smaller not-gzipped coadd files instead (FIBERMAP for quantities that are independent of exposure like targeting info, or EXP_FIBERMAP for the per-exposure rows).

sbailey (commented):
I put some I/O efficiency comments inline.

I'm confused about the final result though: adding the missing exposure with lowefftime takes the tile from good to bad.  Is that intended / real?

py/desispec/tile_qa.py
araichoor: Thank for the suggestion.

Actually, with looking at the code, I realize that the coadds extensions are read + stacked further below in the code:
https://github.com/desihub/desispec/blob/7f27030dca6c2d68cea16a67cd64b993ea68d342/py/desispec/tile_qa.py#L130-L178

So I think we can just get read of those lines reading the spectra files (they are just here to get the list of expids), move the lines I highlighted (reading+stacking coadds) upper in the code, and get the list of expids from the  exp_fibermaps table.

I ll try/test that later today.

araichoor: The commit https://github.com/desihub/desispec/pull/1767/commits/cd405b4701b7679e1ee79461d7d8370268dd6196 makes the changes I mentioned.

araichoor:
The commit https://github.com/desihub/desispec/pull/1767/commits/b600492dfa99d5a3b05fafa87673258a8d4039ba addresses my last comment, with using for the per-petal tile-qa values (averaged over exposures) only exposures used for each petal.

The resulting qa plot is here, where one can see that the petal=0 now has a "regular" number for `NGOODPOS` (as the 130673 exposure is not considered anymore to compute that number):
tile-qa-42426-thru20220417-pr2


My approach is to add a boolean `"ISUSED"` column.
The `nonnull` variable defined here is probably intended to do the same:
https://github.com/desihub/desispec/blob/b600492dfa99d5a3b05fafa87673258a8d4039ba/py/desispec/tile_qa.py#L319-L322
however, it does not work for the `NGOODPOS` column, as it has a non-null value, even if the petal is discarded for the exposure.
I mean, if we are confident enough to do so, maybe we could remove the use of this `nonnull` to simplify the code.

Please let me know if I should ran more test to be sure that the changes behave as expected.

sbailey:
@araichoor In your latest plot left column, qa_efftime=0 even though you added a new exposure, compared to previously qa_efftime=69.  In the top middle the individual qa_efftimes appear to be correct (69 and 0), but perhaps the left column is only using the second exposure?

Independent of that, the second exposure has terrible positioning and adding it made most of the NGOODPOS numbers get worse.  @akremin please flag 20220417/130673 as bad due to move fibers being beyond the 30 um positioning quality cut:
image

araichoor:
Thanks for catching this qa_efftime=0 in the plot!

I ve investigated: the reported efftime there comes from the `tile-qa-TILEID-thruLASTNIGHT.fits` header.
And that value is set to zero, because:
- all fibers from expid=130673 are flagged with `BADPETALPOS`, hence have a `bad_qafstatus_mask` (related to your last message saying that expid=130673 has horrible positioning);
- the header efftime is computed from fibers which are good in all exposures; as there are no such fibers here, the code defaults to zero: https://github.com/desihub/desispec/blob/b600492dfa99d5a3b05fafa87673258a8d4039ba/py/desispec/tile_qa.py#L334-L338

So my understanding is that the code behaves "fine" (with the "qa_efftime = 0" displayed in bold red, it catches our attention).
And it s just that we want to get rid of that expid=130673 exposure, right?

sbailey:
OK, thank for checking.  I accept the explanation that it is a feature not a bug that if an exposure is included with all fibers flagged as bad, the result is qa_efftime=0 in red and the solution is to remove the bad exposure and re-process.

PR #1768, opened 2022-05-10
add temporary file when write the output in the QSO afterburners

### LINKED ISSUE #1764: qso afterburners should write via a temporary file
The qso afterburners should first write to a temporary filename, and then only upon success rename the file to the final filename.  This protects us against cases like fuji/tiles/pernight/80733/20210208/qso_qn-1-80733-20210208.fits where the job timed out after opening the file but before writing it, leaving a blank fits file behind.  We didn't catch this until recently (see desihub/desiBackup#25), because previously we were checking for the existence of files, not their validity.

the standard pipeline files follow this pattern:
```
from desispec.io.util import get_tempfilename
tmpfile = get_tempfile(filename)
table.write(tmpfile)
os.rename(tmpfile, filename)
```
so that the final correct output name will only exist if the job successfully wrote the entire file.

Note: The reason for using `get_tempfilename` instead of just `tmpfile = filename+'.tmp'` is that it creates a temporary filename with the same extension, e.g. '.fits' or '.fits.gz' so that code like `table.write(...)` can auto-derive the format to use.

echaussidon:
This is to fix #1764

I also clean a little bit the code (adding and removing space).


bin/desi_qso_mgii_afterburner
```diff
@@ -189,14 +190,20 @@ def save_dataframe_to_fits(dataframe, filename, DESI_TARGET, clobber=True):
...
-    fits = fitsio.FITS(filename, 'rw', clobber=clobber)
+    # Save file in temporary file to track when timeout error appears during the writing
+    tmpfile = get_tempfilename(filename)
+    fits = fitsio.FITS(tmpfile, 'rw', clobber=clobber)
     fits.write(data, extname='MGII')
     if clobber:
         log.info(f'OVERWRITE the file : {filename}')
     else:
         log.info(f'EXPAND the file : {filename}')
     fits.close()
```
sbailey, line 201: Since tmpfile will always be unique, this code will no longer append to a pre-existing filename and will effectively always clobber the pre-existing file with the `os.rename(tmpfile, filename)`.

What is the intended behavior here?  Do you ever really want to append another identically named "MGII" extension to a pre-existing file?  Or maybe load the pre-existing file, append to the table, and then write a new single "MGII" extension?  Or do you just want the code to stop if the filename already exists and doesn't have --clobber?

To get the original behavior (writing another extension), you'll need to do somthing like:
```
if os.path.exists(filename) and not clobber:
    os.rename(filename, tmpfile)

fits = fitios.FITS(tmpfile, ...)
...
fits.close()
os.rename(tmpfile, filename)
```

bin/desi_qso_qn_afterburner
```diff
@@ -341,14 +341,20 @@ def save_dataframe_to_fits(dataframe, filename, DESI_TARGET, clobber=True):
...
+    fits = fitsio.FITS(tmpfile, 'rw', clobber=clobber)
     fits.write(data, extname='QN_RR')
     if clobber:
         log.info(f'OVERWRITE the file : {filename}')
     else:
         log.info(f'EXPAND the file : {filename}')
     fits.close()
 
+    # Rename temporary file to the
+    os.rename(tmpfile, filename)
```
sbailey, line 355: See above comment about appending to fits files when using a temporary file; this section will need to be updated too.

bin/desi_qso_qn_afterburner
```diff
@@ -429,22 +435,22 @@ if __name__ == "__main__":
...
+            QSO_from_QN = selection_targets_with_QN(redrock, fibermap, sel_to_QN, DESI_TARGET,
+                                                    args.coadd, param_QN, param_RR, args.save_target)
 
             if QSO_from_QN.shape[0] > 0:
                 log.info(f"Number of targets saved : {QSO_from_QN.shape[0]} -- "
                          f"Selected with QN + new RR: {QSO_from_QN['IS_QSO_QN_NEW_RR'].sum()}")
                 save_dataframe_to_fits(QSO_from_QN, args.output, DESI_TARGET)
             else:
-                file = open(os.path.splitext(args.output)[0]+'.notargets', "w")
+                file = open(os.path.splitext(args.output)[0] + '.notargets', "w")
```
sbailey, line 446: While we're making changes, please update this to be '.notargets.txt' to give a hint about the filetype (like Ben did for '.misscamera.txt')

bin/desi_qso_mgii_afterburner
echaussidon: Thank for this remark !  This is now useless (put here for testing purpose), I remove the clobber option.

bin/desi_qso_qn_afterburner
```diff
@@ -27,18 +29,17 @@ from quasarnp.utils import process_preds
...
-## LOAD template to RE-RUN redrock
+# LOAD template to RE-RUN redrock
 qso_template_filename = [path for path in find_templates() if os.path.basename(path) == 'rrtemplate-qso.fits'][0]
 
-try: # check if the env variable QN_MODEL_FILE is defined, otherwise load boss_dr12/qn_train_coadd_indtrain_0_0_boss10.h5
-     model_QN_path = os.environ['QN_MODEL_FILE']
+try:  # check if the env variable QN_MODEL_FILE is defined, otherwise load boss_dr12/qn_train_coadd_indtrain_0_0_boss10.h5
+    model_QN_path = os.environ['QN_MODEL_FILE']
 except KeyError:
     model_QN_path = "/global/cfs/cdirs/desi/science/lya/qn_models/boss_dr12/qn_train_coadd_indtrain_0_0_boss10.h5"
```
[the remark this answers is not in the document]
echaussidon: I will add the logger info. I don't know why but I do not have the $QN_MODEL_FILE variable set in my environment. 

I load `source /project/projectdirs/desi/software/desi_environment.sh` in my bashrc. Do you know where QN_MODEL_FILE is set ?

echaussidon:
Hi Stephen, 

Sorry for the delay, I was on vacation last week. I do the small changes that you required.

Edmond

sbailey (approved):
Looks good, thanks.  I will make one post-merge update since this is coming from your fork and I can't directly update it: when calling `sys.exit()` after an error, we should give it a non-zero exit code so that unix / pipeline wrappers can know that the program exited due to a problem.

bin/desi_qso_qn_afterburner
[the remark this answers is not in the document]
sbailey: $QN_MODEL_FILE is set in the desispec module file, so you should have it set immediately after `source /project/projectdirs/desi/software/desi_environment.sh`.  However, if you then run `module unload desispec` because you want to put your checkout of desispec into the path instead, the module unload process will unset that variable and you will have to set it again yourself.
```
[cori05 ~] echo $QN_MODEL_FILE

[cori05 ~] source /project/projectdirs/desi/software/desi_environment.sh
[cori05 ~] echo $QN_MODEL_FILE
/global/cfs/cdirs/desi/target/catalogs/lya/qn_models/qn_train_coadd_indtrain_0_0_boss10.h5
[cori05 ~] module unload desispec
[cori05 ~] echo $QN_MODEL_FILE

```
if that doesn't answer your situation, ping me on Slack and we'll figure it out.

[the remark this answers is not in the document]
echaussidon: Okay ! 

Indeed, I load the desi_environment and then I unload desispec to add my version in the python path ... !

PR #1771, opened 2022-05-17
Fix crash in desi_run_night with non-consecutive tile obs

akremin:
This fixes a bug where desi_run_night crashes if `z_submit_types` is None and there are non-consecutive observations of a single tile. This was identified in a test run including 20211030 where a backup tile is observed twice with another tile observed in between.

The fix is simple and I verified that running the following command fails in master and succeeds in this branch:
`desi_run_night -n 20211030 --z-submit-types=None --dry-run-level=3`

I used my personal `SPECPROD` for the test, since no files were produced as output. `desi_run_night` just requires a `SPECPROD` with the corresponding exposure_table available for the night submitted, otherwise it will throw an error.

### DIFF
--- a/py/desispec/scripts/submit_night.py
+++ b/py/desispec/scripts/submit_night.py
@@ -376,7 +376,7 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
         lasttype = curtype
 
         tableng = len(ptable)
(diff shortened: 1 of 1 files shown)


sbailey (approved):
Looks good, thanks.  I'll merge after github tests finish.

PR #1772, opened 2022-05-17
set batch-friendly matplotlib backend

sbailey:
This PR fixes a new intermittent failure on Cori KNL nodes, where importing healpy imports matplotlib which sometimes crashes with:
```
ImportError: Cannot load backend 'TkAgg' which requires the 'tk' interactive framework, as 'headless' is currently running
```
It is unclear why this only occurs on KNL, and why it only occurs sometimes, or why we didn't see it with fuji (perhaps related to deforkproc causing more MPI ranks to import healpy/matplotlib instead of  spawning serial commands that loaded it).

Regardless, the workaround is to set a batch friendly backend before any other imports, using `desispec.util.set_backend` which we had previously used to work around backend issues with plotting libraries on Travis tests without $DISPLAY set.  I added this to desi_proc, desi_proc_joint_fit, desi_exposure_qa, desi_tile_qa, and desi_night_qa.  As extra belt-and-suspenders, I also defined "MPLBACKEND=agg" in the redshifts script in case any of the afterburners etc. it calls may also end up triggering a healpy or matplotlib import.

I'm testing this with himalayas jobs now which I'd like to finish before merging, but I'm opening the PR now in case @akremin or others want to look and comment.  If I don't get any comments I'll rely upon the Himalayas testing and merge in order to proceed with more nights.

### DIFF
--- a/bin/desi_night_qa
+++ b/bin/desi_night_qa
@@ -8,6 +8,9 @@
 This script generates the $DESI_ROOT/spectro/redux/nightqa/{NIGHT}/nightqa-{NIGHT}.html page, and related products, once all tile-qa*fits are done.
 """
 
+#- enforce a batch-friendly matplotlib backend
+from desispec.util import set_backend
+set_backend()
 
 import os,sys
 import numpy as np
--- a/bin/desi_tile_qa
+++ b/bin/desi_tile_qa
@@ -8,6 +8,9 @@
 This script computes QA scores per exposure, after the cframe are done
 """
 
+#- enforce a batch-friendly matplotlib backend
+from desispec.util import set_backend
+set_backend()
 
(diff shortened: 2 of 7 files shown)


akremin (approved):
Other than verifying that a tile-qa plot and night-qa page look the same with this enforced backend, I have no concerns with this PR. Have you check those outputs for this PR, or can you point me to the test run so I can check?

sbailey:
Good idea to verify that output pngs look ok with the agg backend.  You caught me though -- I used himalayas as my test run.  See himalayas/tiles/cumulative/22471/20210921/tile-qa-22471-thru20210921.png and himalayas/nightqa/20210921/nightqa-20210921.html .  They look ok to me.

PR #1780, opened 2022-05-27
Qa lya cov

araichoor:
This PR refines a bit the diagnostic plot monitoring the newly identified Ly-as in the night_qa.
It now uses a combination of the fractions of the tile coverage by 1, 2, ..., 7 tiles.
The difficult part is to decide how to set the boundary for outliers; any suggestion is welcome!

**Context:**
The current predicted Ly-as are too low for some cases, especially when ~half the tile is well-covered, and the other ~half is not (which happens often when we grow the equatorial footprint).
See for instance discussion from here: https://desisurvey.slack.com/archives/C01HNN87Y7J/p1649408325262139, prompted by the several tiles on 20220406 being above the predicted values:
Screenshot 2022-04-08 at 10 55 44

The typical tile coverage of those outliers is like this TILEID=2130:
tmp (2)

**Method:**
Elaborating on a suggestion from @ashleyjross, I did the following for all 2.5k main dark (daily) tiles:
- for each tile, I now store what fraction is covered by 1 tile, 2 tiles, ..., 7 tiles;
- I fitted the per-tile number of new Ly-as with a linear combination of those fractions, and obtain:`N_newlya_predicted = 287.7 * NTILECOV1 + 137.8 * NTILECOV2 + 58.1 * NTILECOV3 + 20.9 * NTILECOV4 + 10.5 * NTILECOV5 + 0.0 * NTILECOV6 + 0.6 * NTILECOV7`
- I then defined the y-values: `(N_newlya_observed / N_newlya_predicted) - 1`, and defined by eye the 1-2-3 sigma regions with an exponential law, reproducing as best as I can the scatter in the data so far.

The plot below shows this
newlya-daily-tilecov-upto20220526

The scatter in the high-coverage regions comes from the very low number of newly identified Ly-as there (few units).

And the following plot compares this `(N_newlya_observed / N_newlya_predicted) - 1` for:
- what s currently in the master (red)
- the initial proposition from @ashleyjross (green; `300 * NTILECOV1 + 150 * NTILECOV2 + 75 * NTILECOV3`)
- what s in this PR (black)
newlya-daily-tilecov-upto20220526-comparison
One sees the obvious shortcoming of the current master version, with an under-prediction for intermediate coverages (and over-prediction for high coverages).
The initial proposition from @ashleyjross nicely corrects that for intermediate coverages, but is biased at high coverages, as it does not include information for more than 3 pass-coverage.
This PR corrects for that.

And here is for instance the 20220406 night, with highlighting the tiles originally flagged as outliers, which now are ok:
Screen Shot 2022-05-27 at 1 02 24 AM

And here the only example (as far as I know) of a problematic night, 20211226:
Screen Shot 2022-05-27 at 12 53 12 AM

Plots for all nights are here: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/lya_cov/20220526/.
File with the coverage fractions for each tile: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/lya_cov/20220526/newlya-daily-tilecov-upto20220526.ecsv.

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -831,11 +831,11 @@ def plot_newlya(
         return (
             ntilecovs[:, 0] * 288 +
             ntilecovs[:, 1] * 138 +
-            ntilecovs[:, 2] * 57 +
-            ntilecovs[:, 3] * 23 +
-            ntilecovs[:, 4] * 11 +
-            ntilecovs[:, 5] * 7 +
-            ntilecovs[:, 6] * 0
+            ntilecovs[:, 2] * 58 +
+            ntilecovs[:, 3] * 21 +
+            ntilecovs[:, 4] * 10 +
+            ntilecovs[:, 5] * 0 +
+            ntilecovs[:, 6] * 1
         )
     # AR approximate expected regions:
     # AR - based on main dark tiles (from daily) up to May 26th 2022
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -786,57 +786,94 @@ def plot_newlya(
     tileids=None,
     title=None,
     xlabel="Number of observed tiles",
-    ylabel="Number of newly identified LYA candidates",
+    ylabel="(N_newlya_obs / N_newlya_expect) - 1",
     xlim=(0.5, 6.5),
-    ylim=(0, 400),
+    ylim=(-1, 1),
     nvalifiber_norm=3900,
 ):
     """
-    Plot the number of newly identified Ly-a vs. the tile observations coverage for a list of tiles.
+    Plot the ratio of the number of newly identified Ly-a to the expected number,
+        vs. the tile observations coverage for a list of tiles.
 
     Args:
         ax: pyplot Axes object
-        ntilecovs: list of average number of previously observed tiles (list or np.array() of floats)
-        nnewlyas: list of newly identified Ly-a (with a valid fiber) (list or np.array() of ints)
+        ntilecovs: average number of previously observed tiles for each number of pass (np.array(len(tiles), n_dark_passids))
+        nnewlyas: list of nb of newly identified Ly-a (with a valid fiber), normalized to nvalifiber_norm (list or np.array() of ints)
         tileids (optional, defaults to None): tileids (to report outliers) (list or np.array() of ints)
         title (optional, defaults to None): plot title (str)
         xlabel (optional, defaults to "Number of observed tiles"): plot xlabel (str)
         ylabel (optional, defaults to "Number of newly identified LYA candidates"): plot ylabel(str)
         xlim (optional, defaults to (0.5, 6.5)): plot xlim (tuple)
-        ylim (optional, defaults to (0, 400): plot ylim (tuple)
+        ylim (optional, defaults to (-1, 1): plot ylim (tuple)
         nvalifiber_norm (optional, defaults to 3900): number of valid fibers to normalize to, for the expected regions (int)
+
+    Notes:
+        ntilecovs[:, 0] lists the fractions of the tiles covered by 1 tile, etc.
+        The plotted y-values are: (N_newlya_observed / N_newlya_expected) - 1.
+        The expected numbers are based on all main dark tiles (from daily) up to May 26th 2022.
+        The 1-2-3-sigma regions reflect the approximate scatter of those data.
     """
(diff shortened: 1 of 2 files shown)


sbailey:
This looks good to me, but I'll defer to @schlafly and @ashleyjross for a more detailed sign-off since they were more involved in developing the ideas behind this update.

Operational question, especially for @akremin : do we want to update this today before going into the long weekend (and thus risk having to back something out if there is an unexpected case causing QA to crash) or wait until next week?

araichoor:
Thanks for having already looked at it!

My 2 cents: we can wait next week to merge (also to give people time if there is some suggestions).
In the meantime, I (or anyone else) could generate on my (their) side the night_qa petalnz-NIGHT.pdf with this branch, to double-check for the dailyops.

ashleyjross (approved):
Looks great; an obvious improvement over the status quo. I would merge ASAP as known it won't cause any crashes and if there are further improvements to make, include them in the future.

schlafly:
One suggestion: I note that the y axis is bounded at -1, but can go arbitrarily high.  And the +2sigma outliers fall off the plot at <n> = 6.  So we probably want to change the bounds of what is plotted so that we can see future high outliers?  Or one could imagine plotting something else so that the errors don't blow up, e.g., n vs n_pred, which would be more closely related to the old plots.  Sorry, I realize I suggested n/n_pred - 1, and if you don't think this is relevant, please ignore.

araichoor:
Thanks for the suggestion.
I actually tried to overcome this, with clipping the displayed y-values to the ylim, so that all points are displayed:
https://github.com/desihub/desispec/blob/80d3d1d30998098a4dc697fd1e6772533a08cab2/py/desispec/night_qa.py#L863-L864
For instance for 20220522:
Screen Shot 2022-05-27 at 11 01 26 AM

But you re right, it would be nice to have something better...
I guess a simple workaround would be to enlarge the ylim, e.g. to (-2, 2):
Screen Shot 2022-05-27 at 11 02 26 AM
That should be enough, and any tile hitting the ylim would clearly be an outlier (ie no need to know the exact y-value, as it would be above 2).

Let me know if that would be sufficient; if not I could re-write a bit the code along your suggestion to look at the delta of N_newlya, instead of the ratio.

[ps: I ll give a closer look at this TILEID=3758..]

schlafly:
Yes, that looks sufficient to me, thanks!

araichoor:
Ok great, I ll do that (I notice I wrote ylim=(-2,2), but it should have been ylim=(-2.5,2.5), as in the plots).

araichoor:
Following up on TILEID=3758 from 20220522:
I actually had a typo in `desi_night_qa` (index mixing), which was discarding the 7th-pass in the pass-coverage counting, hence under-estimating the x-axis value when there are 7-pass coverage.

I ve corrected that.
That doesn t affect the two plots with all nights displayed.
I ve updated the https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/lya_cov/20220526/petalnz* plots.

For completeness, here is the correct plot for 20220406 and 20220522:
Screen Shot 2022-05-29 at 2 20 02 PM
Screen Shot 2022-05-29 at 2 20 44 PM

TILEID=3758 from 20220522 is now ok (just on the edge of the 3-sigma region, with large error bars).

Sorry for that!

araichoor:
And a question, related to the current s observations with 9-petals only:

In principle, in `tile_qa_plot.get_tilecov()`, I could account for missing petals, with e.g. reading the `NGOODFIB` in the `TILEQA` extension of the `exposure-qa-TILEID.fits` file.

Shall I go for that?

Pros : the tile coverage should be more accurate.
Cons: it will add a dependence on the `exposure-qa-TILEID.fits` files (current code only relies on `tiles-main.ecsv` and `exposures-daily.fits`).

schlafly:
That would obviously be good, though it goes down the path of fully tracking all of the petal gaps, etc..  If I were you, I'd declare victory without tracking missing petals, but your call.

araichoor:
I was thinking to simply slice the focal plane in 10 parts (as for instance the layout in the tile-qa*png plots).
I ll see if I ve time to implement that ~tomorrow.
Otherwise, if people are happy with the current version, we could merge it now, and I can always add such a feature later.

sbailey:
This looks good to me.  I suggest merging now and save missing petals logic for a separate focused PR.

PR #1797, opened 2022-06-08
Add new redshift job dashboard

akremin:
### Overview 
This adds a new redshift job dashboard analogous to the exposure-level processing dashboard. Since the two dashboards are similar in structure and function, I ported a lot of code to a new file for import by both. I also simplified the logic where possible, and better delineated between the aggregation of information and the generation of the html.

Note: several generic functions were renamed, which touched a few files. Most of those files are unused now anyway, but they were updated regardless since removing the files was out of the scope of this pull request.

### Tests
I tested this on both the daily prod and the gaudalupe prod. The tests can be found at NERSC here: `/global/cfs/cdirs/desi/users/kremin/PRs/zdashboard`. 

#### daily prod
For the daily prod I had it running via cron job like it would be run when deployed. I also ran the normal exposure dashboard since the underlying code was refactored. Both the new redshift dashboard and the standard dashboard worked as expected and handled the incoming data appropriately. The test redshift dashboard can be found here: https://data.desi.lbl.gov/desi/users/kremin/PRs/zdashboard/daily_dashboard/zdashboard.html

An example screenshot:
image

#### guadalupe prod
Since the daily doesn't do `perexp` or `pernight` redshifts, I also tested this on `guadalupe`. The test is housed in the same directory at NERSC. The dashboard url is here: https://data.desi.lbl.gov/desi/users/kremin/PRs/zdashboard/guad_dashboard/zdashboard.html.

An example screenshot:
image


akremin:
I should add that there are known limitations to this. It will account for all redshift jobs, but it doesn't properly handle cross-night redshift jobs. Since only the cumulative redshifts cross night boundaries, those are the jobs that are potentially impacted. Issues could include incomplete `EXPID`'s and incomplete comments, and incorrect `PROCCAMWORD` and file counting if petals were flagged on a previous night but not on the current night. These will be addressed in a future PR.

sbailey (approved):
I used the prototype of this over the weekend while I was on pipeline duty and it is *great*.  Thanks.

PR #1801, opened 2022-06-14
Replace TPCORR per-fiber corrections with a model

schlafly:
This PR replaces the current per-fiber, per-camera TPCORR parameters with a model.  The TPCORR parameters are fits of the brightness of individual sky lines in a fiber's spectrum, relative to the average sky in a given camera.  This is motivated by the TPCORR parameters introducing significant differences in the sky levels between the B/R cameras, leading to cases where we get spurious redshifts when the moon is bright.  The model has the biggest impact on the B band, where TPCORR was least well measured due to the reliance on a single bright sky line.  In R and Z it has less impact, but is also applied there.

The model is that TPCORR is a cubic polynomial in a fiber's location within its patrol radius, derived from fitting past TPCORR measurements.  On top of this, there are two PCA components that describe how TPCORR varies from exposure to exposure.  The two coefficients are fit during sky computation.  Finally, there's a mean correction that reflects some difference between how the sky fills the fibers and how the flat field fills the fibers.  Note that Julien has expressed a worry here that the current approach is more robust to us, e.g., observing the white spot while the white spot is out of position.  This PR does not try to address that concern.

These features are supported by a new tpcorrparam file in desi_spectro_calib, and corresponding entries in all of the sm*-*.yaml files.

I have not investigated it, but conceivably the current model of indexing TPCORR spatial dependence within a fiber by fiber number is not adequate, and I would be better off doing something that could handle resplicing a fiber, e.g., as we did during the initial summer shutdown.  I haven't tried to do that.

Also included in this PR are the tools needed to generate the PCA eigenvectors and the spatial fits.

This PR also turns on by default the sky gradient fitting code, rather than its current off-by-default status.

This PR also adds a new extension to the sky model with the PCA components of the skycorr PCA fit, so that the dependence of these coefficients on system parameters can be more easily investigated in the future.

I haven't been careful about how we want to change the arguments to proc.py / procexp.py to turn off throughput corrections.  Currently there's a no_sky_throughput_correction option, that I might want to make default True?  Or I want to change the sign of this option, but that would break old code using this option?  Advice requested.

I removed some code from sky.py doing alternative modeling of spatial variations across the petal.  I can put this back if requested; it was too duplicative in some areas and was confusing me as I made edits, and eventually I just got annoyed and removed it.

Going forward, the TPCORR terms that get recorded but not applied are the ~residual TPCORR after the model fitting.  In order to refit the PCA and spatial variation models we would have to combine those ~residual TPCORR with the current model TPCORR.  I haven't done that yet.

Plots of the principal components are here:
https://data.desi.lbl.gov/desi/users/schlafly/sky/tpcorr-pc-new.pdf
Only the top two are used in the sky modeling.

Illustrations of the position-within-patrol-radius TPCORR trends we are fitting out are here:
https://data.desi.lbl.gov/desi/users/schlafly/sky/tpcorrperfiberguadbrz.pdf
We use only the Z band fits (most sky lines) and apply those to all bands.

I need help with the best demos for improvement.  I just reprocessed 20220611.  The nightqa files are here:
https://data.desi.lbl.gov/desi/spectro/redux/schlafly/nightqa/20220611/
vs daily
https://data.desi.lbl.gov/desi/spectro/redux/daily/nightqa/20220611/

I've interleaved the tileqa plots for the two nights here:
https://data.desi.lbl.gov/desi/users/schlafly/20220611-tileqa-old-new.pdf
The first qa plot in each group of two is the daily QA plot, and the second is with this PR and DESI_SPECTRO_CALIB set to $DESI_SPECTRO_REDUX/schlafly/desi_spectro_calib.

I would describe the new reductions as better in all cases, though only marginally.  As expected, the differences are most noticeable when the sky is bright.  Here's a fairly typical case:
daily:
image
new:
image
where one can see the reduced number of problematic redshifts at z=1.7, the reduced concentration of sky fiber redshifts at z=1.7, and the increase in n(z)/n(z)_ref.  

The worst tile of the night, 25124, is improved but still bad enough that I would say we should mark it as bad.  The last tile of the night has only 1/2 observations in my rerun, so its worse performance is expected.

I will run additional nights to more completely verify performance.  20220219 is a special very good seeing, bright moon, dark tile night where we have lots of spurious redshifts that this PR substantially improves; I'll add it as soon as the latest rerun completes.  And then I will run a recent dark and bright night.  Presently I am only imagining visually checking these QA plots, but I'd appreciate recommendations of more detailed tests.

### DIFF
--- a/py/desispec/io/__init__.py
+++ b/py/desispec/io/__init__.py
@@ -40,6 +40,7 @@ from .raw import read_raw, write_raw
 from .sky import read_sky, write_sky
 from .skycorr import (read_skycorr, write_skycorr, read_skycorr_pca, write_skycorr_pca)
 from .skygradpca import read_skygradpca, write_skygradpca
+from .tpcorrparam import read_tpcorrparam
 from .table import read_table
 from .util import (header2wave, fitsheader, native_endian, makepath,
                    write_bintable, iterfiles, healpix_degrade_fixed,
--- a/py/desispec/io/sky.py
+++ b/py/desispec/io/sky.py
@@ -24,6 +24,9 @@ def write_sky(outfile, skymodel, header=None):
             ivar : 2D inverse variance of sky flux
             mask : 2D mask for sky flux
             stat_ivar : 2D inverse variance of sky flux (statistical only)
+            dwavecoeff : 1D[ncoeff] array of PCA dwavelength coefficients
+                (optional)
+            dlsfcoeff : 1D[ncoeff] array of PCA dlsf coefficients (optional)
         header : optional fits header data (fits.Header, dict, or list)
     """
     from desiutil.depend import add_dependencies
@@ -48,12 +51,19 @@ def write_sky(outfile, skymodel, header=None):
     # hx.append( fits.CompImageHDU(skymodel.mask, name='MASK') )
     hx.append( fits.ImageHDU(skymodel.mask, name='MASK') )
     hx.append( fits.ImageHDU(skymodel.wave.astype('f4'), name='WAVELENGTH') )
+    hx[-1].header['BUNIT'] = 'Angstrom'
+
     if skymodel.stat_ivar is not None :
        hx.append( fits.ImageHDU(skymodel.stat_ivar.astype('f4'), name='STATIVAR') )
     if skymodel.throughput_corrections is not None:
         hx.append( fits.ImageHDU(skymodel.throughput_corrections.astype('f4'), name='THRPUTCORR') )
+    if skymodel.dwavecoeff is not None:
+        hx.append(fits.ImageHDU(skymodel.dwavecoeff.astype('f4'),
+                                name='DWAVECOEFF'))
+    if skymodel.dlsfcoeff is not None:
+        hx.append(fits.ImageHDU(skymodel.dlsfcoeff.astype('f4'),
+                                name='DLSFCOEFF'))
 
-    hx[-1].header['BUNIT'] = 'Angstrom'
 
     t0 = time.time()
     tmpfile = get_tempfilename(outfile)
@@ -96,11 +106,20 @@ def read_sky(filename) :
         throughput_corrections = native_endian(fx["THRPUTCORR"].data.astype('f8'))
     else :
         throughput_corrections = None
+    if "DWAVECOEFF" in fx :
+        dwavecoeff = native_endian(fx["DWAVECOEFF"].data.astype('f8'))
+    else :
+        dwavecoeff = None
+    if "THRPUTCORR" in fx :
+        dlsfcoeff = native_endian(fx["DLSFCOEFF"].data.astype('f8'))
+    else :
+        dlsfcoeff = None
     fx.close()
     duration = time.time() - t0
     log.info(iotime.format('read', filename, duration))
 
     skymodel = SkyModel(wave, skyflux, ivar, mask, header=hdr,stat_ivar=stat_ivar,\
-                        throughput_corrections=throughput_corrections)
+                        throughput_corrections=throughput_corrections,
+                        dwavecoeff=dwavecoeff, dlsfcoeff=dlsfcoeff)
 
     return skymodel
--- a/py/desispec/io/sky.py
+++ b/py/desispec/io/sky.py
@@ -19,14 +19,14 @@ def write_sky(outfile, skymodel, header=None):
     Args:
         outfile : filename or (night, expid, camera) tuple
         skymodel : SkyModel object, with the following attributes
-            wave : 1D wavelength in vacuum Angstroms
-            flux : 2D[nspec, nwave] sky flux
-            ivar : 2D inverse variance of sky flux
-            mask : 2D mask for sky flux
-            stat_ivar : 2D inverse variance of sky flux (statistical only)
-            dwavecoeff : 1D[ncoeff] array of PCA dwavelength coefficients
-                (optional)
-            dlsfcoeff : 1D[ncoeff] array of PCA dlsf coefficients (optional)
+        wave : 1D wavelength in vacuum Angstroms
+        flux : 2D[nspec, nwave] sky flux
+        ivar : 2D inverse variance of sky flux
+        mask : 2D mask for sky flux
+        stat_ivar : 2D inverse variance of sky flux (statistical only)
+        dwavecoeff : 1D[ncoeff] array of PCA dwavelength coefficients
+            (optional)
+        dlsfcoeff : 1D[ncoeff] array of PCA dlsf coefficients (optional)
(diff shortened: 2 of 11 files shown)


schlafly:
Here are the equivalent plots from 20220219, where I claim there's a big reduction in spurious redshifts.
https://data.desi.lbl.gov/desi/users/schlafly/20220219-tileqa-old-new.pdf
Again, each tileqa plot occurs twice; the first one is the current daily reduction, and the second one is the new reduction with sky gradients & a more rigid tpcorr model.

schlafly:
Last set of interleaved tileqa plots is here:
https://data.desi.lbl.gov/desi/users/schlafly/20220601-tileqa-old-new.pdf
There is very little difference between the reductions on this pristine dark night.

Here's a comparison of the n(z)/n_ref(z) for all of the images I've run.
image
This quantity is the number of good redshifts relative to expectations.  'good redshifts' means delta chi^2 > 50, no bad flags, and in the appropriate redshift range for the target (thanks to Anand for clarifying this for me).  So this is saying that we had some smaller delta chi squares before, or that spurious redshifts were pushing us out of the target range, and now we're good.  Nominally this is increasing the number of good redshifts by ~1%.  Since these are mostly additive sky errors it's possible we do better for the lower SNR regime and we'd see a bigger effect for delta chi^2 > 25, but I'm happy.

Note that this is not applying the full 'good' criteria for each target class (e.g., no OII flux cuts for the ELGs), which I could add if needed, but primarily I'm just looking for a reasonable statistic to say that things have improved, which this should satisfy?

Here's a histogram of the ratio new/old of n(z)/n_ref(z):
image
Clearly the largest impact is on the bright nights (20220219, dark tiles on a bright night in very good seeing; 20220611, bright tiles near full moon), while the impact on the pristine dark night 20220601 is muted.

schlafly:
I've improved the documentation.  Let's hash out what we want to save for tpcorr a bit more.

The major remaining issue with this PR is that it changes ~nature of tpcorr and sframe.  That's sort of fine and intentional.  Before most of the signal in the sframe files came from unmodeled sky gradients and tpcorrparam residuals.  And before much of the signal in the tpcorr measurements came from the spatial model and the PCA components.  Now those are removed and only the residual is included.

That's obviously great, except that it means I am not sure how I would repeat this analysis to update the tpcorr fits or the sky gradient fits.  If I apply the existing code to our upcoming h-specprod, I'll get very different results, and replacing the current files with the results there would be a major regression.

On the skygradpca side, the current analysis code would need updating to add the sky model to the sframe files and then to subtract a median sky for each exposure.  That's reasonable, but I don't have a large library of skies to do that analysis on, so I wouldn't feel good about it.

On the tpcorr side, it's less clear to me what I would want.  I could add code to output the model tpcorrs and add them as an additional field somewhere.  Then the analysis code could be updated to use tpcorr_residual * tpcorr_model or something.  Or I could update the tpcorr's currently output to be this product.  Or I could change nothing about the output but I could derive the tpcorr_model by something like the ratio of the sky for a particular fiber ratioed against median sky in the camera, and then multiply that with the residual tpcorrs that are saved.

julienguy:
- For the skygradpca , one would have to add back the coeff*component to the sky residuals and probably remove an average. That should be doable because you are saving the coefficients.
- For TPCORR , I might indeed be useful to save in 2 different rows the model values that have been applied and the residual values that have been measured but not applied.

julienguy:
2 additional minor requests after reading the new sky.py 
 - please provide some explanation  as comments in the code about the tpcorr pca fit line 386 (if tpcorrparam is not None: ...). (about the fixed and non fixed skytpcorr ...)
 - remove unused 'from astropy.stats import mad_std'

schlafly:
Okay, good, I'll store the model tpcorr somewhere.  I also _do not_ think I currently save the sky gradient PCA coefficients, so I'll make sure to store them and look at making their application more convenient.  I still also need to condense compute_sky and compute_uniform_sky together.

schlafly:
@ashleyjross , I mentioned during the LSS session at the collaboration meeting today that it would be good to demonstrate as well as possible that our pipeline changes lead to spectroscopic success rate improvements.  This PR is the kind of change I have in mind, where we have changed the sky subtraction algorithm.

Here so far I just rely on dchisq > 50 and redshifts in expected range for each tracer.  It would be good to incorporate the best measurements of success rates we have.  It is reasonably clear in this case that we have made an improvement, but making the metrics more similar to the ultimate science metrics might help avoid unpleasant future surprises.

ashleyjross:
Absolutely; what @rongpu coded up recently might already be exactly what is needed? If no, I'm sure we can quickly produce a script to evaluate these runs that includes our current best criteria.

From: Eddie Schlafly ***@***.***>
Date: Tuesday, June 21, 2022 at 12:00 PM
To: desihub/desispec ***@***.***>
Cc: ashleyjross ***@***.***>, Mention ***@***.***>
Subject: Re: [desihub/desispec] Replace TPCORR per-fiber corrections with a model (PR #1801)

@ashleyjross<https://github.com/ashleyjross> , I mentioned during the LSS session at the collaboration meeting today that it would be good to demonstrate as well as possible that our pipeline changes lead to spectroscopic success rate improvements. This PR is the kind of change I have in mind, where we have changed the sky subtraction algorithm.

Here so far I just rely on dchisq > 50 and redshifts in expected range for each tracer. It would be good to incorporate the best measurements of success rates we have. It is reasonably clear in this case that we have made an improvement, but making the metrics more similar to the ultimate science metrics might help avoid unpleasant future surprises.

schlafly:
I believe I have addressed the remaining issues.  In particular, I find very good agreement between the daily sky files and the new sky files, after dividing out the model tpcorr and subtracting the model gradient.  So this is all now being well tracked.  I'm running a night of data tonight, and then I'll clear out all of the test nights and re-run the original sample to make sure I haven't introduced any problems, and then I would be ready to merge.

schlafly:
I am now happy with this PR.  Here's a gallery of before/after QA plots; they look just like they used to, as expected.  I'm happy to do some more tests as requested.
https://data.desi.lbl.gov/desi/users/schlafly/tileqa-newsky-20220219-20220601-20220611.pdf
@sbailey, @julienguy , anything else you want to see, or want to run tests yourselves?

julienguy (approved):
I verified desi_compute_sky runs fine with the most recent updates to the code. The parameters saved in the sky files are THRPUTCORR, THRPUTCORR_MOD, DWAVECOEFF, DLSFCOEFF, SKYGRADPCACOEFF. I checked the sky file io. I checked desi_process_exposure runs with/without option --apply-sky-throughput-correction and that the difference is indeed the values of THRPUTCORR. I am not going to rerun nights of data because the tests presented above are sufficient.

PR #1803, opened 2022-06-16
Retry with regular QOS for failed Slurm jobs

marcelo-alvarez:
A workaround on Perlmutter when using the realtime queue that changes submissions to the regular queues when sbatch fails.


akremin (changes requested):
I think this is used in the current daily pipeline where we still want to submit to the realtime queue. Could you have it do `max_attempts` normal submits and then one regular queue submit attempt afterward if the previous attempts all fail?

I acknowledge it's inefficient for your needs but I think it still does what you want while letting the daily pipeline on cori remain robust.

Happy to iterate solutions if what I propose wouldn't work.

marcelo-alvarez:
Closing this because a better workaround was found.

PR #1805, opened 2022-07-07
afterburners update for new qso templates

echaussidon:
Hi, 

I update the qso afterburners to deal with the new qso templates. 

It circumvent some problems in redrock that we should improve. This is a quick solution enable us to perform the analysis of the new templates without waiting new redrock development. 

Edmond


bin/desi_qso_qn_afterburner
```diff
@@ -162,39 +158,61 @@ def collect_redshift_with_new_RR_run(spectra_name, targetid, z_prior, param_RR):
...
+    for template_filename in param_RR['templates_filename']:
+        filename_priors = param_RR['filename_priors']
+        filename_output_rerun_RR = param_RR['filename_output_rerun_RR']
+        filename_redrock_rerun_RR = param_RR['filename_redrock_rerun_RR']
+
+        # find redshift range of the template
+        redshift_template = fitsio.FITS(template_filename)['REDSHIFTS'][:]
+        zmin, zmax = np.min(redshift_template), np.max(redshift_template)
+        sel = (z_prior >= zmin) & (z_prior <= zmax)
+
```
sbailey, line 182: What happens if no targets pass the selection?  I was expecting something like
```python
if np.all(sel == False):
    contintue
```
instead of proceeding with a redrock fit of no targets, which may work but might also produce messy log error messages.

echaussidon: Good  point .. I just experiment this, I had a failure when I pass the specific targetid in redrock. I add the selection with `sel.sum() != 0` (do it before reading your comment.

py/desispec/mgii_afterburner.py
```diff
@@ -1,17 +1,18 @@
...
+Please see https://desi.lbl.gov/trac/attachment/wiki/TargetSelectionWG/TSTeleconMinutes/DESITS210113/QSO_completeness_Ravoux_13_01_2021_v2.pdf for details about the MgII fitter
+
+the "optimal"/chosen parameters are (can be modify when desi_qso_mgii_afterburner is called):
+lambda_width = 250
+max_sigma = 200
+min_sigma = 10
+min_deltachi2 = 16
+min_signifiance_A = 3
+min_A = 0
+"""
```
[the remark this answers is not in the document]
echaussidon: Yes, only style formatting trying to follow the flake8

PR #1808, opened 2022-07-22
require PHOTSYS when matching to get DR9 photometry

moustakas:
This PR has a partial-line change in `desispec.io.photo._gather_tractorphot_onebrick` which is needed to fix https://github.com/desihub/fastspecfit/issues/75. This function is not used anywhere else except by FastSpecFit and to generate the photometric VAC, so I'm going to self-merge momentarily since the issue this addresses is a blocking factor.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -79,6 +79,9 @@ Full desispec API Reference
 .. automodule:: desispec.io.params
     :members:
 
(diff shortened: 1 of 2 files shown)


weaverba137:
The failure was caused by a temporary network outage, it's not a problem with the code itself. Merge when ready.

PR #1817, opened 2022-08-03
stdstar correct sky and flat fiber; scale ivar

### LINKED ISSUE #1810: stdstars using wrong fiberflat and sky
Getting this into a ticket so I don't lose track of it --

Snippets from scripts/stdstars.py:
```python
for frame,sky in zip(frames[cam],skies[cam]) :
    frame.flux = frame.flux[starindices]
    frame.ivar = frame.ivar[starindices]
    ...
    for star in range(frame.flux.shape[0]) :
        ok=np.where((frame.ivar[star]>0)&(flat.fiberflat[star]!=0))[0]
        if ok.size > 0 :
            frame.flux[star] = frame.flux[star]/flat.fiberflat[star] - sky.flux[star]
```
Note that the `frame.flux` was pre-filtered down to just the `starindices` entries, but the `flat` and `sky` were not, meaning that the final line is using the wrong fiber for the fiberflat and sky subtraction.  This may not matter much because of a later median filter on the uncalibrated star spectra.

I tried to fix this while addressing issue #1785 and #1789 for stdstar memory issues (which also involves trimming down arrays to save memory), but I'm getting quite different answers on the final stdstar fits so now I'm back tracking to determine if I introduced a different indexing bug (likely) or whether this sky/flat bug is actually a bigger deal than I thought.

sbailey:
This PR fixes #1810 for standard star fitting:
  * use the correct fiber when applying the fiberflat and sky subtraction
  * when scaling the flux by the fiberflat, also scale the ivar to match

Testing on all exposures of 20210531, this results in a distinct improvement in the scatter of the data-model G-R colors:
image

The test outputs are in `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/stdstar`, which used guadalupe inputs and re-ran just the stdstar fitting step for each tile.

### DIFF
--- a/py/desispec/fiberbitmasking.py
+++ b/py/desispec/fiberbitmasking.py
@@ -91,13 +91,12 @@ def get_fiberbitmasked_frame_arrays(frame,bitmask=None,ivar_framemask=True,retur
             bad |= bit
 
     # find if any fibers have an intersection with the bad bits
-    badfibers = fmap['FIBER'][ (fmap['FIBERSTATUS'] & bad) > 0 ].data
-    badfibers = badfibers % 500
+    badfibers = (fmap['FIBERSTATUS'] & bad) > 0
     # For the bad fibers, loop through and nullify them
-    for fiber in badfibers:
-        mask[fiber] |= specmask.BADFIBER
+    for i in np.where(badfibers)[0]:
+        mask[i] |= specmask.BADFIBER
         if ivar_framemask :
-            ivar[fiber] = 0.
+            ivar[i] = 0.
 
     if return_mask:
         return ivar,mask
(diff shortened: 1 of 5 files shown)


julienguy (changes requested):
Thank you for the bug fix. It does indeed improve the fit of standard stars.
I see the original code and this new version are missing the transformation to convert fiberflat wavelength from the Kitt Peak frame to the solar system barycenter. It is included in the routine ```fiberflat.apply_fiberflat(frame, fiberflat)```. It would be better to use directly this function here.

sbailey:
I updated to use `apply_fiberflat` and `subtract_sky`, which led to other changes for SkyModel object slicing and updating the tests and fixing a bug in fiberbitmasking when working on a sliced subset of a full frame, but in the end the improvements are similar:
image

sbailey:
For the record: the reason why the results are different when using `apply_fiberflat` and `subtract_sky` is that those also propagate the error model from the fiberflat and sky models, which was previously ignored when doing `flux/fiberflat - sky` and `ivar*fiberflat**2`.

julienguy (approved):
Great. I checked the code and successfully ran std star fit for one tile and got slightly reduced rms.

PR #1818, opened 2022-08-08
Add desi spectro dark

Waelthus:
This PR is to allow generating new DARK models via a `make_biweekly_darks` script that are stored in a new DESI_SPECTRO_DARK directory.
The functions in `preproc.py` are updated such that the most recent such model produced before the to-be-processed exposure will be used for pre-processing. We fall back to the default models from DESI_SPECTRO_CALIB if none of the new ones are found.
Also adds a `compare_dark` function which is simailar to `compare_bias` but runs on fully preprocessed frames (instead of raw-overscan_per_amp).
Code used for creating/updating tables for dark exposures, evaluating the `compare_dark` outputs to make decisions about which dark to use and producing qa plots will be added as well.


Waelthus:
ok, re automated tests: currently the code changes in here are not fully applied in testing. I.e. if darkfinder cannot find DESI_SPECTRO_DARK as an environment variable it will just fall back to calibfinder (and that env is not currently defined in the tests)

julienguy:
Thanks Michael for this promising improvement.

I successfully tested the code with and without the env. variable DESI_SPECTRO_DARK=${DESI_ROOT}/spectro/desi_spectro_dark. If the variable is missing, it falls back to the standard process. If the variable is set, it uses the most recent past dark in the directory. In my test case, it picked /global/cfs/cdirs/desi/spectro/desi_spectro_dark/dark_frames/dark-sm10-b1-20220119.fits.gz for exposure 20220121/00119663.

- Please add a README in /global/cfs/cdirs/desi/spectro/desi_spectro_dark describing the content of this directory where there are many files of various formats, some probably temporary, some cleanup may also be possible, I am not sure. The README file should give the path to the essential files in this directory, in particular the list of dark observations and the list of hardware changes. Also the README should tell the purpose of the directory, which env variable to set to use it ...

- Please add figures to this PR illustrating the improvement over the whole period since we started the main survey (May 14, 2021).

- A comment about the code : adding a specific class DarkFinder is invasive in the rest of the code. Isn't it possible to simply add the logic to switch from the darks in the calib repo to the ones in $DESI_SPECTRO_DARK inside the CalibFinder class directly? That would allow us to avoid any change in preproc.py.

Waelthus:
Thanks for the comments. Your description is indeed the expected behaviour of the code currently.

I'm doing the processing for the timespan of the main survey, and for the moment keep a lot of temporary files lying around (also to allow debugging if things go wrong). I'll clean those up when things are more complete (and probably create a script for the cleanup tasks) and then will also update the README (added a simple one for the moment).

Regarding the DarkFinder class: While it is true that the logic could be added to CalibFinder, personally I found it easier to read/write this when seperated (it also is more explicit and thus potentially more transparent what happens in preproc that way). But in the end, I'm happy to do either.
Note also, that a change of preproc can only be avoided when using a modified CalibFinder if we do not want to be able to choose to use an older dark or configure the selection process in any other way.

Waelthus:
[dark_time_series_mad_fuji.pdf](https://github.com/desihub/desispec/files/9557851/dark_time_series_mad_fuji.pdf)
[dark_time_series_mad_guadalupe.pdf](https://github.com/desihub/desispec/files/9557852/dark_time_series_mad_guadalupe.pdf)
[dark_time_series_mad_himalayas.pdf](https://github.com/desihub/desispec/files/9558036/dark_time_series_mad_himalayas.pdf)

The plots attached show the evolution of preprocessing when using new darks based on 1 month of observations with a 2 week cadence where possible and less where not enough darks have benn taken to do this (or some errors have been preventing me from running the dark analysis). 
Panels are ordered such that each row is one spectrograph, each column one of the b,r,z chennels.
- The x-axis always shows the date of observation
- The y-axis always shows the max(median in patches of the preprocessed 300s dark), i.e. the output of [compare_dark](https://github.com/desihub/desispec/blob/ff38a759cdb921b3c807e4cbe5a173ed8c3aae0d/py/desispec/ccdcalib.py#L668)

- black vertical dotted curves show the cutoff date where a new dark was created (in light grey cases only a new bias was created and the dark failed for that camera, e.g. because there have not been enough dark observations or some job crashed)
- red vertical dashed lines show changes of the spectrograph config
- cyan vertical lines should show changes of the master darks/biases that have no associated config change

- blue curve is using the existing fuji/guadalupe/himalayas preprocs, i.e. using the nightly bias where that was deemed good and the svn master bias elsewise and using the svn master darks
- orange curve uses the same nightly biases and master darks (and bias when nightly was bad) from $DESI_SPECTRO_DARK

Overall it looks like the orange curve mostly lies below the blue curve (or is basically the same) for the time currently shown (the 2022 data is coming up as well, but not yet in the plot). Spiky features are probably bias fluctuations and appear in both at the same nights (I guess that this is when nightly biases were deemed bad, but need to check).
Most of the time a new dark does not seem to strongly change the result wrt the dark 2 weeks earlier. So I'd suspect that running things every month would be ~ as good.
Particular features for some individual detectors need more investigation.

Waelthus:
Added nightqa plots for the preprocessed 300s darks. See here [for daily since 2022 restart](https://data.desi.lbl.gov/desi/spectro/desi_spectro_dark/daily/nightqa/), [for fuji](https://data.desi.lbl.gov/desi/spectro/desi_spectro_dark/fuji/nightqa/), [for guadalupe](https://data.desi.lbl.gov/desi/spectro/desi_spectro_dark/guadalupe/nightqa/), [for himalayas](https://data.desi.lbl.gov/desi/spectro/desi_spectro_dark/himalayas/nightqa/) (not all have been run yet, but will be filled)

Waelthus:
attached are now also time evolutions of the number of bad columns found per detector per night based on the preprocessed 300s darks using either the new dark models or the stored preproc with the current dark models based on `desi_inspect_dark`.

It looks like most differences concern only individual nights, but it looks like for z7 we currently have some bad columns, that are vanishing with the new dark models. See images [here (old)](https://data.desi.lbl.gov/desi/spectro/redux/daily/nightqa/20220918/dark-20220918.pdf) vs [here (new)](https://data.desi.lbl.gov/desi/spectro/desi_spectro_dark/daily/nightqa/20220918/00142875/20220918-00142875-dark.pdf).

[dark_badcol_time_series_guadalupe.pdf](https://github.com/desihub/desispec/files/9635107/dark_badcol_time_series_guadalupe.pdf)
[dark_badcol_time_series_fuji.pdf](https://github.com/desihub/desispec/files/9635108/dark_badcol_time_series_fuji.pdf)
[dark_badcol_time_series_daily.pdf](https://github.com/desihub/desispec/files/9635109/dark_badcol_time_series_daily.pdf)
[dark_badcol_time_series_himalayas.pdf](https://github.com/desihub/desispec/files/9635110/dark_badcol_time_series_himalayas.pdf)

Waelthus:
Ok, will put all changes to calibfinder.
And start versioning things inside DESI_SPECTRO_DARKS in case of future changes.
Also MAD plots have a bug that needs fixing before deciding between 2 week and 1 month cadence

Waelthus:
Ok, the darkfinder is gone now and replaced by an additional class function of calibfinder that searches for the most recent dark with matching setup.
In a small test run based on data from this week, those seem to be correctly picked up and are written into the fibermap headers, i.e.: `f["FIBERMAP"].read_header()["DEPVER03"]=='/global/cfs/cdirs/desi/spectro/desi_spectro_dark/v2209//dark_frames/dark-sm4-b0-20220913.fits.gz'`

I recomputed the statistics for the MAP comparison and redid the plotting (there was a bug in the vertical lines, but I'm not sure where the repetitions came from) and don't see the biweekly repetitions anymore.

Updated figs (for fuji the full file system prevented 042021 to be written out, so that month is missing in this update, guadelupe not yet recalculated for the same reason):
[dark_time_series_mad_fuji.pdf](https://github.com/desihub/desispec/files/9657410/dark_time_series_mad_fuji.pdf)
[dark_time_series_mad_himalayas.pdf](https://github.com/desihub/desispec/files/9657411/dark_time_series_mad_himalayas.pdf)
[dark_time_series_mad_daily.pdf](https://github.com/desihub/desispec/files/9657409/dark_time_series_mad_daily.pdf)

[dark_badcol_time_series_fuji.pdf](https://github.com/desihub/desispec/files/9657478/dark_badcol_time_series_fuji.pdf)
[dark_badcol_time_series_himalayas.pdf](https://github.com/desihub/desispec/files/9657476/dark_badcol_time_series_himalayas.pdf)
[dark_badcol_time_series_daily.pdf](https://github.com/desihub/desispec/files/9657479/dark_badcol_time_series_daily.pdf)

julienguy (changes requested):
* The usage in DESI for environment variables defining directories is to have a single env. variable pointing directly to the versioned directory rather than having 2 environment variables.
so DESI_SPECTRO_DARK=/global/cfs/cdirs/desi/spectro/desi_spectro_dark/v2209 instead of DESI_SPECTRO_DARK=/global/cfs/cdirs/desi/spectro/desi_spectro_dark and DESI_DARK_VERSION=v2209.

* I see that in the constructor of the calib finder, the dark directory content is read, and some fits files headers are opened. I am worried that this could generate more IO issues (this would happened at least 30 times per exposure, simultaneously). It will also be CPU costly as the files are gzipped (they have to be decompressed to read the header content). I would suggest we add an ascii cvs table in the directory with the list of files and all the information needed (NIGHT CAMERA DETECTOR CCDCFG ...). The calibfinder constructor would then read a single file (need to add script that actually reads all the headers and writes the table).

* Consider renaming desi_make_biweekly_darks to something more generic and adding a cadence option because we decided today to recompute darks only monthly (and we may again change our mind in the future).

* Same for function make_biweekly_darks in ccdcalib.py

Waelthus:
Addressed the two last points.
There is no need for a cadence option here, as the script is just using a month of data (or something specified in the `window` option). It would just need to be started with a given cadence either manually or via cron-job (with additional cleanup needed after).
The first issue can be easily fixed as well.

I'll have a look at the second issue now to avoid opening the fits files of the darks potentially a gazillion times (incl e.g. automated updating of dark lists when new darks are there and checks for existance if something is in the list, but actually missing).

Waelthus:
Have such tables now, but code changes for this still tbd tomorrow.

Waelthus:
@julienguy I addressed your comments, i.e. now using a csv table to check which dark should be right, using only one environment variable for directory + version of darks instead of two and having "biweekly" renamed to regular.
If there's anything else that should be addressed: please tell.

julienguy (changes requested):
Need to remove test of DESI_DARK_VERSION in routine find_darks_in_desi_spectro_dark.
Except for that it looks great. I tested the code successfully with a dummy value of DESI_DARK_VERSION which is not used (except for the test). The code picked up the correct dark. I also tested that without the env variable DESI_SPECTRO_DARK we fall back to the default calib darks.

Waelthus:
Ah, missed the test yesterday, sorry for that. Should be fine now.

julienguy (approved):
Good for me.Thank you for all the work.

julienguy:
I let @sbailey  or @akremin have a look and then merge, but for me it's ready to go.

sbailey:
Thanks for the work and the review.  Let's wait for Monday to do the merge to minimize changes going into the weekend.

PR #1819, opened 2022-08-11
EDR redshift database

weaverba137:
This PR supports redshift databases for `fuji` and `guadalupe`.  A complete test-load of `fuji` was successful, and a test-load of `guadalupe` should be finished by the time this PR is fully reviewed.

In addition the requirements and test suite are updated.

[Access instructions](https://desi.lbl.gov/trac/wiki/DESIProductionDatabase).

I have been performing the loading by submitting a batch job to the `bigmem` queue on cori. This is needed to process the `targetphot-potential-specprod.fits` file.  A load of all tables takes several hours, but not days or weeks.  Roughly 8-10 hours, but I haven't formally timed it.

Installing the database as a Spin service is beyond the scope of this PR.  However, we can create `pg_dump` files that could be easily loaded into such a database, once it is set up. Technical note: the database needs [q3c](https://github.com/segasai/q3c), and it is an open question whether this has to be built into the image, or if it can be loaded into a stock PostgreSQL image via initialization files.

Remaining questions:

- [x] The fiberassign files are currently loaded from trunk. If trunk corresponds *exactly* to any tag that would have been made for edr/fuji or dr1, then that's fine.  If not, we should test-load from at least the tag corresponding to fuji.
- [x] Light-curve (`LC_*`) columns are excluded since these are large, array-valued columns. If there are strong opinions that those should be included, we can discuss.
- [x] There is some column duplication, especially between the `Target` table and the `Ztile` and `Zpix` tables. I think the duplication should be examined in a separate issue/PR, but we should discuss.
- [x] There is some code cruft left over from when this database was loaded off of simulations. Is there any need to keep that?
- [x] What do we *officially* call this database?  This could have some effect on, *e.g.* selecting a DNS name for easy access to a future Spin service.


weaverba137:
For the record, there is much less column duplication between the target and redshift tables than I had (mis)remembered.  For example, in the [zpix data model](https://desidatamodel.readthedocs.io/en/latest/DESI_SPECTRO_REDUX/SPECPROD/zcatalog/zpix-SURVEY-PROGRAM.html), columns in between `COADD_FIBERSTATUS` and `COADD_NUMEXP` are excluded.  Similarly, photometric columns are excluded from the fiberassign table.

weaverba137:
Breaking down the remaining tasks listed above:

1. Version of fiberassign files. This is not a blocking factor for this PR, address in a separate issue.
2. Dropping light-curve columns. This is not a blocking factor for this PR, address in a separate issue if needed.
3. Column (de)duplication.  See previous comment.
4. Code cruft. I'll remove the large commented-out sections for the purposes of this PR, but leave other, possibly old, code intact.
5. Official name. This is not a blocking factor for this PR, address in a separate issue.

sbailey (approved):
I'm fine with merging this as-is as a step in the right direction even if we're not done with the DB.

Issue #1813, opened 2022-08-01
stdstars output fibermap sorted differently than other HDUs


sbailey:
In the stdstars files, it appears that most HDUs are sorted by descending blue S/N, except for the FIBERMAP HDU which is still sorted by fiber number.  This makes it a pain to cross match models and METADATA fit info with FIBERMAP entries, especially since the METADATA HDU has neither FIBER nor TARGETID columns to verify what order it is in.

For the future it would be convenient if the FIBERMAP HDU was sorted to match the other HDUs (or perhaps sort the other HDUs by fiber number like the fibermap), and if the METADATA column included TARGETID and/or FIBER as a confirmation of its sort order.

In the meantime, I think the post-factor recipe to line them up is:
```python
with fitsio.FITS(filename) as fx:
    metadata = Table(fx['METADATA'].read())
    fibers = fx['FIBERS'].read()
    fibermap = Table(fx['FIBERMAP'].read())

metadata['FIBER'] = fibers
metadata.sort('FIBER')
assert np.all( metadata['FIBER'] == fibermap['FIBER'] )
```

sbailey:
Implemented in PR #1820; closing.

Issue #1789, opened 2022-06-02
use stdstars --maxstdstars 30


sbailey:
desi_fit_stdstars has an option --maxstdstars with a default of 30 per petal, but a subsequent code change ignored this option and hardcoded it to 50 instead, which leads to job timeouts on KNL.  This impacts some bright tiles and many backup tiles when processed on KNL.

Go back to using args.maxstdstars, and use 30, which is 3x the fiberassign stdstar budget needed for good calibration (10 per petal).  I'd rather limit the max number of stdstars instead of increase the requested job runtime, which would impact how quickly the jobs can get through the queue.

Example bright tile timeouts from Himalayas:
```
20220202 - 121156 (tile 24809), 121157 (tile 24787), 121158 (tile 24780)
20220208 - 121845 (tile 23227)
20220209 - 121960 (tile 23229)
20220211 - 122193 (tile 24791)
20220213 - 122422 (tile 22160), ...
```

sbailey:
Implemented in PR #1820; closing.

Issue #1785, opened 2022-06-02
stdstar out of memory with 6 exposures


sbailey:
Tile 82266 on night 20210928 was observed 6 times (expids 102103-102108).  The stdstar joint fit across all these exposures runs out of memory:
```
srun -N 1 -n 30 -c 4 --cpu-bind=cores  desi_proc_joint_fit   --obstype science --cameras a0123456789 -n 20210928 -e 102103,102104,102105,102106,102107,102108 --mpi --mpistdstars
...
slurmstepd: error: Detected 1 oom-kill event(s) in StepId=59371624.0. Some of your processes may have been killed by the cgroup out-of-memory handler.
srun: error: nid10540: tasks 3,13,15,23,25: Out Of Memory
srun: launch/slurm: _step_signal: Terminating StepId=59371624.0
FAILED: done at Tue May 24 18:19:57 PDT 2022
```

Unmerged PR #1735 also identified this, but addressed this by increasing the number of nodes for all stdstar jobs even though that isn't needed for the majority of tiles.  Let's try to fix this by adjusting the actual amount of memory used, e.g. by trimming the input frames to just the standard stars instead of keeping the xentire frame in memory.  This would require some bookkeeping updates to make sure that the output stdstar indices are correct relative to the original input frames and not just the trimmed frames.

Also consider reviving old PR #1284 to do I/O on a single rank before broadcasting, which might be an even better place to trim the frames down.

sbailey:
night 20220324 tile 82636 with 8 exposures is another case of this stdstar memory problem.

sbailey:
Implemented in PR #1820; closing.

Issue #1781, opened 2022-05-31
stdstar job "succeeds" even if some petals fail


sbailey:
stdstar jobs exit with code=0 (success) even if individual petals fail their processing.  e.g. from /global/cfs/cdirs/desi/spectro/redux/daily/run/scripts/night/20220527/stdstarfit-20220527-00136980-a0123456789-59579365.log:
```
...
CRITICAL:util.py:149:runcmd: FAILED rank 0 exception
...
ERROR:proc_joint_fit.py:543:main: stdstars.main failed for stdstars-3-00136980.fits.gz
...
ERROR:proc_joint_fit.py:555:main: 3/30 stdstar commands failed
...
INFO:proc_joint_fit.py:629:main: All done at Sat May 28 01:04:32 2022
SUCCESS: done at Sat May 28 01:04:33 PDT 2022
```
`desi_proc_joint_fit` needs to intercept these individual failures and exit with an overall failure code instead of success.

sbailey:
The logic appears to be purposeful, so documenting it for discussion before overriding:
From [proc_joint_fit.py L554-561](https://github.com/desihub/desispec/blob/b7fbeba8a8bb518d814e887da1e648ba669143b9/py/desispec/scripts/proc_joint_fit.py#L554-L561):
```python
        if rank == 0 and num_err > 0:
            log.error(f'{num_err}/{num_cmd} stdstar commands failed')

        sys.stdout.flush()
        if num_err>0 and num_err==num_cmd:
            if rank == 0:
                log.critical('All stdstar commands failed')
            sys.exit(1)
```

i.e. individual failures log an error, but it requires all petals to fail before the command itself exits with an error thus triggering the job to fail and canceling dependencies.  I suspect this was originally motivated by wanting the daily pipeline to proceed with whatever it could.  In practice we've been slowly undoing that choice in other cases and I suspect we want to do that here as well.

sbailey:
Implemented in PR #1820; closing.

PR #1820, opened 2022-08-11
stdstar memory optimization

sbailey:
This PR fixes issues
* #1785: examples of stdstars running out of memory for tiles with many exposures
* #1789: go back to using a max of 30 stdstars instead of 50; make that an option not a hardcode
* #1813: stdstars FIBERMAP HDU is now in the same sort order as the other HDUs
* #1781: stdstar jobs would claim to "succeed" (at the slurm level) even if some petals had failed; now if any petal fails the job is flagged as failing

Test runs are in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/stdmem (this branch) and stdmem-main (current master), testing
```
time srun -N 1 -n 30 -c 4 --cpu-bind=cores desi_proc_joint_fit --obstype science --mpi --mpistdstars \
    --cameras a0123456789 -n 20210928 -e 102103,102104,102105,102106,102107,102108 
time srun -N 1 -n 30 -c 4 --cpu-bind=cores desi_proc_joint_fit --obstype science --mpi --mpistdstars \
    --cameras a0123456789 -n 20220324 -e 127343,127344,127345,127346,127347,127348,127349,127350 
```
On Cori KNL, these run in 9-10 minutes using this branch, and run out of memory on current master.  Switching to `-n 15` on current master doesn't run out of memory but takes ~20 minutes to run.  The script stdmem/compare_stdstars.py confirms that they produce the same answer except for the following differences:
* the FIBERMAP HDU is now sorted to match the other HDUs.  If both new/old are sorted by FIBER, they are an idetical match
* METADATA HDU has two more columns: TARGETID and FIBER (see #1813 for the incantations on how to align the previous METADATA HDU with the differently sorted FIBERMAP)
* A side effect of handling the integer datatypes for TARGETID and FIBER resulting in the DATA_G-R column becoming float32 instead of float64.  The previous code was accidentally upcasting it to float64 when writing out, so the actual results are still the same and pass `np.all(new32 == old64)`.

Unfortunately it still runs out of memory with 64 ranks on Cori KNL due to each rank needing a full copy of the stdstar templates, so the batch config still throttles this step to 32 ranks.  I also confirmed that this works on Cori Haswell, Perlmutter CPU (1.8x faster than haswell), and Perlmutter GPU (3.3x faster than haswell).

## Details ##

The main memory saving changes were:
1. trim each frame to just the standard stars while reading them, instead of reading all N>>1 frames and then filtering them down.  Also do this for the sky and fiberflat data.
2. convert the resolution data to the sparse Resolution object only once per fiber instead of doing it on-the-fly every time it was needed (this also saves time)
3. when evaluating the stdstar models from the templates + coefficients, re-use the same memory buffer instead of re-allocating multiple times.
4. The final evaluated stdstar models are written by only a subset of ranks so allocate the memory buffers only for the ranks that need them.

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -143,14 +143,14 @@ try:
 except TypeError:
     hc = 1.9864458241717586e-08
 
-def resample_template(data_wave_per_camera,resolution_data_per_camera,template_wave,template_flux,template_id) :
+def resample_template(data_wave_per_camera,resolution_obj_per_camera,template_wave,template_flux,template_id) :
     """Resample a spectral template on the data wavelength grid. Then convolve the spectra by the resolution
     for each camera. Also returns the result of applySmoothingFilter. This routine is used internally in
     a call to multiprocessing.Pool.
 
     Args:
         data_wave_per_camera : A dictionary of 1D array of vacuum wavelengths [Angstroms], one entry per camera and exposure.
-        resolution_data_per_camera :  A dictionary of resolution corresponding for the fiber, one entry per camera and exposure.
+        resolution_obj_per_camera :  A dictionary of Resolution objects corresponding for the fiber, one entry per camera and exposure.
         template_wave : 1D array, input spectral template wavelength [Angstroms] (arbitrary spacing).
         template_flux : 1D array, input spectral template flux density.
         template_id   : int, template identification index, used to ensure matching of input/output after a multiprocessing run.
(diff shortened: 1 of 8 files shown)


akremin (approved):
This is essential for re-processing tiles with many input exposures, and is a welcome improvement to the memory management of the standard star fitting routines. 

I checked output files for sp2 and sp7 of night 20210928 and exposures 102103-102108 from the outputs provided at `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/stdmem` and `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/stdmem-main`. The  `FLUX`, `WAVELENGTH`, `FIBERS`, `COEFF`, and `INPUT_FRAMES` HDU's matched. The `METADATA` HDU had the two additional columns and no other changes. The `FIBERMAP` HDU was sorted differently, but had no other changes. Since the two differences were reported and justified as useful features/changes compared with main, I see no issues in the comparison.

Both the code and data look good, so I'm happy to approve and merge this.

Issue #1823, opened 2022-08-16
Improve 'sector' fit for CTE correction in sky subtraction


julienguy:
The 'sector' fit during sky subtraction does not work well to remove the offsets due to CTE issues in one test exposure (the first one I am looking at).

```
export DESI_SPECTRO_CALIB=/global/cfs/cdirs/desi/users/jguy/teststand/desi_spectro_calib
export SPECPROD=jguy-cte
desi_proc -n 20211014 -e 104277 --cam r1  --nostdstarfit --nofluxcalib --batch --system-name cori-haswell
```

The 'sector' fit did work but a negligible offset is found (abs value >20 expected):
```
grep sector $DESI_SPECTRO_REDUX/jguy-cte/run/scripts/night/20211014/prestdstar-20211014-00104277-r1-61941174.log
INFO:sky.py:141:get_sector_masks: Adding CCD sector in amp D with offset: [2064, 4128, 2057, 3715]
INFO:sky.py:152:get_sector_masks: Using PSF /global/cfs/cdirs/desi/spectro/redux/jguy-cte/exposures/20211014/00104277/psf-r1-00104277.fits
INFO:sky.py:385:compute_sky_linear: sectors: 1 sectors fit, offsets -0.03879413973541599
```

The following image of the sky subtracted sframe file shows that large residuals are still present:
```
/global/cfs/cdirs/desi/users/jguy/cte-for-DR1/plot-sframe.py /global/cfs/cdirs/desi/spectro/redux/jguy-cte/exposures/20211014/00104277/sframe-r1-00104277.fits.gz 
```

![Screenshot from 2022-08-16 14-54-15](https://user-images.githubusercontent.com/5192160/184992134-321e2a66-bc2f-4a87-acba-ea0482fd9b0f.png)

julienguy:
For the record, the sframe image after the code fixes :
![Screenshot from 2022-08-17 12-55-00](https://user-images.githubusercontent.com/5192160/185230749-4578a075-3a04-4dbf-b511-b5b643cfb0eb.png)

PR #1825, opened 2022-08-17
Restore sector offsets

schlafly:
The tpcorr modeling code broke the sector implementation.  This PR addresses fixes a few bugs with the sector fitting code:
- Check all amps for sectors
- Report correct sector amplitudes
- Apply correct sector amplitudes
- Add sectors to modeled sky
- Divide sectors by flat field (sectors are in detector counts; frame fluxes are flattened by dividing by flat).
This fixes https://github.com/desihub/desispec/issues/1823 identified by Julien.

### DIFF
--- a/py/desispec/sky.py
+++ b/py/desispec/sky.py
@@ -126,20 +126,20 @@ def get_sector_masks(frame):
         yb = sec[0].start
         ye = sec[0].stop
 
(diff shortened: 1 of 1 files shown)


julienguy:
I checked the code and verified it solves issue #1823.

PR #1831, opened 2022-08-19
Fix bug in find_overscan_cosmic_trails when CCD read with 2 amps

julienguy:


julienguy:
Self merging this simple PR which solves issue #1831

PR #1833, opened 2022-08-19
iotime.parse_logfile dont crash on logs without iotime messages

sbailey:
This PR fixes a bug where `desispec.io.iotime.parse_logfile(filename)` would crash if given a logfile that didn't have any iotime messages in it.  Now it returns `None` instead of crashing.


akremin (approved):
Changes looks good and bonus points for adding a unit test for the identified failure mode

PR #1837, opened 2022-08-22
Catch exceptions in tilenight and revise batch time limits on Perlmutter

marcelo-alvarez:
Modifies `scripts.proc_tilenight.main` to catch and log exceptions from `scripts.proc.main` and `scripts.proc_joint_fit.main`. This improves logging of errors and allows other exposures to be processed prior to terminating early.

Increases `tilenight` and `flat` batch job time limits as specified in `workflow.desi_proc_funcs.determine_resources`, based on running the equivalent of Guadalupe on Perlmutter. 

Fixes a minor bug where `internal_id` was not being incremented in `workflow.profcuncs.submit_tilenight`.

Changes to desispec in preparation to run the pipeline in production mode on Perlmutter are ongoing; this is an incremental update to simplify the integration and documentation of changes to the main branch.

@sbailey please review merge if the changes look good (particularly the tilenight exception catching).

### DIFF
--- a/py/desispec/scripts/proc_tilenight.py
+++ b/py/desispec/scripts/proc_tilenight.py
@@ -147,7 +147,9 @@ def main(args=None, comm=None):
         prestdstar_args_parsed = proc.parse(prestdstar_args.split())
         if not args.dryrun:
             try:
-                error_count += proc.main(prestdstar_args_parsed,comm)
+                errcode = proc.main(prestdstar_args_parsed,comm)
+                if errcode != 0:
+                    error_count += 1
             except (BaseException, Exception) as e:
                 import traceback
                 lines = traceback.format_exception(*sys.exc_info())
@@ -178,7 +180,9 @@ def main(args=None, comm=None):
         stdstar_args_parsed = proc_joint_fit.parse(stdstar_args.split())
         if not args.dryrun:
             try:
-                error_count += proc_joint_fit.main(stdstar_args_parsed, comm)
+                errcode = proc_joint_fit.main(stdstar_args_parsed, comm)
+                if errcode != 0:
+                    error_count += 1
             except (BaseException, Exception) as e:
                 import traceback
                 lines = traceback.format_exception(*sys.exc_info())
@@ -210,7 +214,9 @@ def main(args=None, comm=None):
             poststdstar_args_parsed = proc.parse(poststdstar_args.split())
             if not args.dryrun:
                 try:
-                    error_count += proc.main(poststdstar_args_parsed, comm)
+                    errcode = proc.main(poststdstar_args_parsed, comm)
+                    if errcode != 0:
+                        error_count += 1
                 except (BaseException, Exception) as e:
                     import traceback
                     lines = traceback.format_exception(*sys.exc_info())
--- a/py/desispec/scripts/proc_tilenight.py
+++ b/py/desispec/scripts/proc_tilenight.py
@@ -144,30 +144,79 @@ def main(args=None, comm=None):
             prestdstar_args += f' --badamps {badamps[expid]}'
         if rank==0:
             log.info(f'running desi_proc {prestdstar_args}')
-        prestdstar_args = proc.parse(prestdstar_args.split())
+        prestdstar_args_parsed = proc.parse(prestdstar_args.split())
         if not args.dryrun:
-            error_count += proc.main(prestdstar_args,comm)
+            try:
+                error_count += proc.main(prestdstar_args_parsed,comm)
+            except (BaseException, Exception) as e:
+                import traceback
+                lines = traceback.format_exception(*sys.exc_info())
+                log.error(f"prestdstar step using desi_proc with args {prestdstar_args} raised an exception:")
+                print("".join(lines))
+                log.warning(f'continuing to next prestdstar exposure, if any')
+                error_count += 1
 
-    #- run joint stdstar fit using all exp for this tile night
-    stdstar_args  = common_args + mpi_args
-    stdstar_args += f' --obstype science --expids {",".join(map(str, stdstar_expids))} --cameras {joint_camwords}'
-    if rank==0:
-        log.info(f'running desi_proc_joint_fit {stdstar_args}')
-    stdstar_args = proc_joint_fit.parse(stdstar_args.split())
-    if not args.dryrun:
-        error_count += proc_joint_fit.main(stdstar_args, comm)   
+    #- collect post prestdstar error count
+    if comm is not None:
+        all_error_counts = comm.gather(error_count, root=0)
+        error_count_pre = int(comm.bcast(np.sum(all_error_counts), root=0))
+    else:
+        error_count_pre = error_count
 
-    #- run desiproc poststdstar over exps
-    for expid in poststdstar_expids:
-        poststdstar_args  = common_args
-        poststdstar_args += f' --nostdstarfit --noprestdstarfit --expid {expid} --cameras {camwords[expid]}'
-        if len(badamps[expid]) > 0:
-            poststdstar_args += f' --badamps {badamps[expid]}'
+    continue_tilenight=True
+    if error_count_pre > 0:
+        if rank == 0:
+            log.info(f'prestdstar failed with {error_count_pre} errors, skipping rest of tilenight steps')
+        continue_tilenight=False
+
+    if continue_tilenight:
+        #- run joint stdstar fit using all exp for this tile night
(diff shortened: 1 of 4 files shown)


py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -330,8 +330,13 @@ def determine_resources(ncameras, jobdesc, queue, nexps=1, forced_runtime=None,
     if jobdesc in ('ARC', 'TESTARC'):
         ncores          = 20 * (10*(ncameras+1)//20) # lowest multiple of 20 exceeding 10 per camera
         ncores, runtime = ncores + 1, 45             # + 1 for worflow.schedule scheduler proc
-    elif jobdesc in ('FLAT', 'TESTFLAT'):
+    elif jobdesc in ('FLAT', 'TESTFLAT', 'TILENIGHT'):
         ncores, runtime = 20 * nspectro, 25
```
sbailey, line 334: TILENIGHT is enough different from FLAT that I think these should be in separate `elif` blocks to give flexibility to adjust their resource requests independently.

Previously the TILENIGHT case would use 1 node on Perlmutter which I think is what we still want.  Now this is requesting 20*30 cores = 5 Perlmutter nodes here and then later (line 370) setting it back to 1 node, which seems like a fragile code flow.

py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -330,8 +330,13 @@ def determine_resources(ncameras, jobdesc, queue, nexps=1, forced_runtime=None,
...
+        if system_name[0:10] == 'perlmutter':
+            # ~150 frames per node hour plus
+            # 20-minute overhead for a single node
+            # perlmutter-[cpu,gpu] have timefactor=[0.7,0.5]
+            runtime = 20 + int(60. / 140. * ncameras * nexps)
```
sbailey, line 339: Originally the intention was that we would request the same number of cores across any system for any given step, and use `timefactor` to adjust the runtime needed to achieve that step on that system.  Here you have a system-specific runtime calculation separate from `timefactor`, and it appears that the resulting resources requested for TILENIGHT on cori would just be wrong (runtime=25 min).

If TILENIGHT won't work on cori, let's be explicit about refusing to schedule it, or otherwise have scheduling logic that works across all systems, leveraging `timefactor` as much as possible to encode the system differences in the yaml file rather than in the code.

py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -354,18 +359,12 @@ def determine_resources(ncameras, jobdesc, queue, nexps=1, forced_runtime=None,
...
     else:
         msg = 'unknown jobdesc={}'.format(jobdesc)
         log.critical(msg)
         raise ValueError(msg)
 
-    if jobdesc in ('FLAT', 'TESTFLAT', 'SKY',
+    if jobdesc in ('FLAT', 'TESTFLAT', 'SKY', 'TILENIGHT',
         'TWILIGHT', 'SCIENCE','PRESTDSTAR','POSTSTDSTAR'):
         if system_name[0:10] == 'perlmutter':
             ncores, nodes = config['cores_per_node'], 1
```
sbailey, line 370: Let's drop this section and encode these choices earlier in the flow so that the `ncores`, `nodes`, and `runtime` are set only once for each jobdesc instead of being set in one place and then possibly reset here.

sbailey (changes requested):
Overall looks good but please see comments inline for some small requested changes.

py/desispec/scripts/proc_tilenight.py
```diff
@@ -144,30 +144,79 @@ def main(args=None, comm=None):
...
-        error_count += proc_joint_fit.main(stdstar_args, comm)   
-
-    #- run desiproc poststdstar over exps
-    for expid in poststdstar_expids:
-        poststdstar_args  = common_args
-        poststdstar_args += f' --nostdstarfit --noprestdstarfit --expid {expid} --cameras {camwords[expid]}'
-        if len(badamps[expid]) > 0:
-            poststdstar_args += f' --badamps {badamps[expid]}'
+            try:
+                error_count += proc.main(prestdstar_args_parsed,comm)
```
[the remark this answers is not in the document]
marcelo-alvarez: Done.

py/desispec/workflow/desi_proc_funcs.py
marcelo-alvarez: Done. @sbailey please have a look and see if this works.

py/desispec/workflow/desi_proc_funcs.py
marcelo-alvarez: Done. @sbailey please have a look and see if this works.

marcelo-alvarez:
@sbailey I addressed your comments above, thanks for the review. Please have a look and merge if appropriate.

py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -350,26 +366,19 @@ def determine_resources(ncameras, jobdesc, queue, nexps=1, forced_runtime=None,
         #- new version using MPI on one node
         ncores, runtime = ncameras, (8+2*nexps) #ncameras, 10
     elif jobdesc == 'POSTSTDSTAR':
-        ncores, runtime = ncameras, 10
+        runtime = 10
+        if system_name[0:10] == 'perlmutter':
+            ncores = config['cores_per_node']
+        else:
+            ncores = ncameras
```
sbailey, line 373: In other "if perlmutter" blocks, you are throttling `ncores = config['cores_per_node']` which is less than it otherwise would have requested in order to keep the job on a single Perlmutter node.

In this case for POSTDSTAR, `config['cores_per_node']` is considerably more than otherwise would have been requested (ncameras=30) on other systems.  Is that purposeful and demonstrated to be beneficial?  e.g. most (all?) of the poststdstar code isn't parallelized beyond the camera-level, so if we're only going to be using 30 ranks then we shouldn't bother doing slow python imports on 128 ranks.  Do we need a "if perlmutter" special case here?

py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -331,9 +331,25 @@ def determine_resources(ncameras, jobdesc, queue, nexps=1, forced_runtime=None,
         ncores          = 20 * (10*(ncameras+1)//20) # lowest multiple of 20 exceeding 10 per camera
         ncores, runtime = ncores + 1, 45             # + 1 for worflow.schedule scheduler proc
     elif jobdesc in ('FLAT', 'TESTFLAT'):
-        ncores, runtime = 20 * nspectro, 25
+        runtime = 25
+        if system_name[0:10] == 'perlmutter':
+            ncores = config['cores_per_node']
+        else:
+            ncores = 20 * nspectro
+    elif jobdesc in ('TILENIGHT'):
```
sbailey, line 339: Minor: this was probably intended to be `('TILENIGHT',)` to be a tuple, rather than `('TILENIGHT')` where the parentheses are just interpreted as operator precedence not a tuple.  It happens that `'TILENIGHT' in 'TILENIGHT'` evaluates to `True` anyway so it works with or without that trailing comma inside the parentheses, but that is a lucky accident of how strings work with "in".

In contrast, this doesn't work with other types, e.g, `some_integer in (1)` raises a TypeError, while `some_integer in (1,)` works, and it is arguably preferable to use `if some_integer == 1` if you really just have a single element to prevent confusion about how those parentheses and trailing commas are parsed (which has bitten us in other cases which is why I noticed it here and wrote a paragraph about it...)

sbailey (commented):
Thanks.  I put one pedagogical/stylistic comment and one substantive question about the intended ncores for the POSTSTDSTAR case.

Since this has morphed beyond what you originally had and tested, please use this branch to re-test on Perlmutter with a single night to make sure that all flavors of jobs are still properly allocated and flowing through end-to-end.  Thanks.

py/desispec/workflow/desi_proc_funcs.py
marcelo-alvarez: @sbailey thanks for catching this. We do not need the special case so I removed it.

py/desispec/workflow/desi_proc_funcs.py
marcelo-alvarez: @sbailey I changed all the `in (x)` to `== x` in this block.

marcelo-alvarez:
Now runs a single night up to redrock with no errors on Perlmutter. As agreed previously with @sbailey, I will now merge.

Issue #1839, opened 2022-08-23
Add existence of biasnight files to nightly QA


sbailey:
Currently the nightqa webpages "calibnight" section includes checks that the psfnight and fiberflatnight files exist.  Please add to that a check that the "biasnight-CAMERA-NIGHT.fits.gz" files exist.

Context: these should exist if we took the standard calib ZEROs, but that is occasionally skipped when there is an instrument problem at the beginning of the night and we are rushed to get on sky.  That's ok, but we want to be aware that it happened and not accidentally go a week without calib ZEROs without noticing.

@araichoor could you add this?  Thanks.

araichoor:
yes, sure, will do.

sbailey:
Implemented in PR #1841.  Closing.

PR #1841, opened 2022-08-24
Nightqa v13

araichoor:
This PR addresses https://github.com/desihub/desispec/issues/1839, with adding in the night qa page a check on the existence of the `biasnight` files.

With adding this third set of files (in addition to `psfnight` and `fiberflatnight`), the table gets too big, so I ve removed from the reported name the night.
In case of a link to another file, the linked basename (with the night) will/should be reported.

Besides, I also correct a small (but embarassing...) bug: because of an indentation error, if a file was missing, it was not printed out; now it is.

Before:
Screen Shot 2022-08-23 at 6 33 01 PM

Now:
Screen Shot 2022-08-23 at 6 33 20 PM

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -1625,7 +1625,7 @@ def write_nightqa_html(outfns, night, prod, css, surveys=None, nexp=None, ntile=
     html.write("</div>\n")
     html.write("\n")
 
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks for the quick update.  Looks good, merging now.

PR #1843, opened 2022-08-26
desi_run_night only runs completed tiles by default

akremin:
#### Overview
As first outlined in [[desi-survey 3824] ](https://desi.lbl.gov/trac/tracmailman/browser/private/desi-survey/2022-August/003823.html), this PR adds the logic to desi_run_night so that it only submit completed tiles for processing. After discussion in that email thread and in a telecon, the final selection was chosen to be the OR of the following:

- `ZDONE == 'true'`   # selects main archived tiles and SV tiles with enough time
- `(EFFTIME_SPEC > 0.1*GOALTIME)  &  SURVEY != 'main'`  # selects non-main (including special) tiles with enough time
- `(EFFTIME_SPEC > 0.1*GOALTIME)  &  FAPRGRM == 'backup'`  # selects any backup tiles with enough time

The email proposal specified using `MINTFRAC` (typically 0.85) rather than the final 0.1 reflected in this PR. The choice applies to non-main special, unknown, SV, and cmx tiles; as well as backup tiles in both SV and the main survey.

This change adds 58 additional tiles in the non-main non-backup sample and 89 backup tiles that are now considered "complete" enough to be processed. 9 non-main non-backup are "unknown" tiles that will typically be flagged by other selections in a normal production.

3 nonzero backup tiles and 28 non-main non-backup tiles remain incomplete. 10 non-main non-backup are "unknown" tiles that will typically be flagged by other selections in a normal production.

#### Code changes
This adds a function `desispec.scripts.submit_night.get_completed_tiles()` that takes an optional argument to specify the path to a `specstatus` file. The function returns the list of tiles from that `specstatus` file that pass the above criteria.

This also adds two new flags to `desi_run_night`: `--all-tiles` that disables the subselection of tiles that pass the above selection, and `--specstatus-path` to define a pathname to a non-default `specstatus` file. If **not** given the flag `--all-tiles` then tiles on a night are compared with the list of completed tiles returned in `get_completed_tiles()` and only the exposures associated with tiles in the list of completed tiles are processed.

EDIT: This PR used to refer to `--only-completed-tiles` which enabled the selection of only tiles that pass the above selection. The option is now flipped to opt-out with `all-tiles` following discussion with @sbailey.

#### Testing

I ran the following tests that all appear to produce the correct filtering output:
`desi_run_night -n 20210422 --only-completed-tiles --dry-run-level=3` 
`desi_run_night -n 20210512 --only-completed-tiles --dry-run-level=3` 
`desi_run_night -n 20220522 --only-completed-tiles --dry-run-level=3`

### DIFF
--- a/bin/desi_run_night
+++ b/bin/desi_run_night
@@ -74,8 +74,8 @@ def parse_args():  # options=None):
                              "are pruned using PROCCAMWORD to only process the remaining cameras not found to exist.")
     parser.add_argument("--use-tilenight", action="store_true",
                         help="Use desi_proc_tilenight for prestdstar, stdstar, and poststdstar steps.")
-    parser.add_argument("--only-completed-tiles", action="store_true",
-                        help="Set to restrict to completed tiles as defined by the table pointed to by specstatus_path.")
+    parser.add_argument("--all-tiles", action="store_true",
+                        help="Set to NOT restrict to completed tiles as defined by the table pointed to by specstatus_path.")
     parser.add_argument("--specstatus-path", type=str, required=False, default=None,
                         help="Location of the surveyops specstatus table. Default is "+
                              "$DESI_SURVEYOPS/ops/tiles-specstatus.ecsv")
--- a/bin/desi_run_night
+++ b/bin/desi_run_night
@@ -74,6 +74,11 @@ def parse_args():  # options=None):
                              "are pruned using PROCCAMWORD to only process the remaining cameras not found to exist.")
     parser.add_argument("--use-tilenight", action="store_true",
                         help="Use desi_proc_tilenight for prestdstar, stdstar, and poststdstar steps.")
+    parser.add_argument("--only-completed-tiles", action="store_true",
+                        help="Set to restrict to completed tiles as defined by the table pointed to by specstatus_path.")
+    parser.add_argument("--specstatus-path", type=str, required=False, default=None,
+                        help="Location of the surveyops specstatus table. Default is "+
+                             "$DESI_SURVEYOPS/ops/tiles-specstatus.ecsv")
     args = parser.parse_args()
 
     # convert str lists to actual lists
--- a/py/desispec/scripts/submit_night.py
+++ b/py/desispec/scripts/submit_night.py
@@ -28,7 +28,7 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
                  append_to_proc_table=False, ignore_proc_table_failures = False,
                  dont_check_job_outputs=False, dont_resubmit_partial_jobs=False,
                  tiles=None, surveys=None, laststeps=None, use_tilenight=False,
-                 only_completed_tiles=False, specstatus_path=None):
+                 all_tiles=False, specstatus_path=None):
     """
     Creates a processing table and an unprocessed table from a fully populated exposure table and submits those
     jobs for processing (unless dry_run is set).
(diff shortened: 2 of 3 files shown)


py/desispec/scripts/submit_night.py
```diff
@@ -444,3 +461,46 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
...
+                    specstatus['GOALTIME'] * enoughfraction)
+    ## only take the approved QA tiles in main
+    goodtiles = iszdone
+    ## not all special and cmx/SV tiles have zdone set, so also pass those with enough time
+    goodtiles |= (isenoughtime & isnotmain)
+    ## main backup also don't have zdone set, so also pass those with enough time
+    goodtiles |= (isenoughtime & (specstatus['FAPRGRM'] == 'backup'))
+
+    specstatus = specstatus[goodtiles]
+    return np.array(specstatus['TILEID'])
```
sbailey, line 506: minor optimization: consider `np.array(spectstatus['TILEID'][goodtiles])` instead of slicing entire table just to return one column of it.

py/desispec/scripts/submit_night.py
```diff
@@ -444,3 +461,46 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
             write_table(ptable, tablename=proc_table_pathname)
 
     print(f"Completed submission of exposures for night {night}.", '\n\n\n')
+
+
+def get_completed_tiles(specstatus_path=None):
```
sbailey, line 466: Please add a unit test for this function, that would write a temporary specstatus file covering all of the different cases and then calls this function with that file and confirms that it produces the expected answer.

bin/desi_run_night
```diff
@@ -74,6 +74,11 @@ def parse_args():  # options=None):
                              "are pruned using PROCCAMWORD to only process the remaining cameras not found to exist.")
     parser.add_argument("--use-tilenight", action="store_true",
                         help="Use desi_proc_tilenight for prestdstar, stdstar, and poststdstar steps.")
+    parser.add_argument("--only-completed-tiles", action="store_true",
+                        help="Set to restrict to completed tiles as defined by the table pointed to by specstatus_path.")
```
sbailey, line 78: In the spirit of `desi_run_night` being primarily for production runs and wanting default options to be the most likely "do the right thing", let's switch the sense of this option, i.e. make the default the equivalent of only-completed-tiles, and make the option something like `--all-tiles`.

sbailey (commented):
Overall looks good, see inline for requested changes.  If I'm unavailable when these are implemented, ok to merge yourself.

bin/desi_run_night
akremin: Resolved in 652e9dc

py/desispec/scripts/submit_night.py
akremin: Resolved in 652e9dc

py/desispec/scripts/submit_night.py
akremin: Resolved in 982afa5

akremin:
I independently checked that the new unit test passes, in addition to the github checks. I retested the three nights in the original test with the following commands:
```
desi_run_night -n 20210422 --all-tiles --dry-run-level=3
desi_run_night -n 20210512 --all-tiles --dry-run-level=3
desi_run_night -n 20220522 --all-tiles --dry-run-level=3
desi_run_night -n 20210422 --dry-run-level=3
desi_run_night -n 20210512 --dry-run-level=3
desi_run_night -n 20220522 --dry-run-level=3
```

All give the expected output. @sbailey is away so self-merging.

Issue #1786, opened 2022-06-02
exposure dashboard color code night summary rows


sbailey:
desi_proc_dashboard output color codes the rows of individual exposures to make it easy to see when something is missing, but the collapsed nightly rows do not have any color coding, making it easy to miss when one night is missing some files, e.g.
![image](https://user-images.githubusercontent.com/218471/171734970-b06426c3-2376-45c0-8578-7f805d6987d7.png)
It would be helpful if 20210928 was flagged more prominently as missing files.

This is optional for Himalayas, but it sure would be nice...

akremin:
This was completed with PR #1845 1845

PR #1845, opened 2022-08-31
Add color coding to the nightly and monthly buttons in the dashboards

akremin:
### Overview
This enhances the current exposure and redshift dashboards to include meaningful coloring of nightly and monthly drop down menus, which resolves issue #1786. Note the implementation means that any missing or extra data will change the color of the day and month, so the coloring is rather aggressive, but I think that is what we want so we can catch every issue.

### Details
The (color-blind) color scheme is somewhat the same as the cell color meanings.

Cell Color Meaning:

- `NULL` - coloring when no data expected-  (gray background, dark gray text)
- `BAD` - coloring if all end products missing - (red background, black text)
- `INCOMPLETE` - coloring if some end products missing -  (orange background, black text)
- `GOOD` - coloring if all end products present -  (blue background, black text)
- `GOODNULL` - coloring if no products present and none expected  - (blue background, gray text)
- `OVERFULL` - coloring if more end products present than expected -  (purple background, black text)

Menu Color Meaning:
- `DEFAULT` - default coloring - (gray background, black text)	
- `GOOD` - coloring if all non-null cells/rows are `GOOD` -  (blue background, black text)
- `BAD` - coloring if any cells/rows under menu are `BAD` - (red background, black text)
- `INCOMPLETE` - coloring if any cells/rows under menu are `INCOMPLETE` and none are `BAD` -  (orange background, black text)
- `OVERFULL` - coloring if any cells/rows under menu are `OVERFULL` and none are `INCOMPLETE` or `BAD`  - (purple background, gray text)

### Testing
I tested the desi_proc_dashboard on `daily`, `fuji`, `guadalupe`, and `himalayas` and they all run fine and the coloring works. `daily` fails for early nights where the exposure tables are the wrong format, but that is known and expected behavior (a production maintains the files used for the processing, even if formats change later).

I also tested this with desi_proc_dashboard on `fuji` and `guadalupe` and it worked without issue.

An example of an exposure dashboard is below.

image


akremin:
I was given permission to self merge when ready and after testing. Doing so now.

Issue #1847, opened 2022-09-09
desi_run_night no longer numerically sorting science exposures


akremin:
I recently came upon this when reprocessing a science exposure for `daily` but using `desi_run_night`. It found the three exposures I wanted reprocessed corresponding to a tile, but it submitted them in reverse numeric order (which is also reverse chronological order). This isn't important except the joint `stdstar` script was also named using the largest `EXPID`, rather than the smallest as it is supposed to be. This creates ambiguity in file names, which is not good.

Fix so that exposures within a single tile are sorted properly.

akremin:
This doesn't always seem to sort in reverse order, which is even worse than simply changing the default order that we used to have. 

I am able to reproduce the issue in a fresh prod using only the exposure table from svn. I.e. this doesn't have to do with the previous state of the prod, edits to the exposure_table, or the existence of the processing table.

with the exposure table for 20220113 present in the current `SPECPROD`, this issue can be reproduced with the following:
`desi_run_night -n 20220113 --tiles="4403" --z-submit-types=cumulative --laststeps="all,skysub,fluxcal" --all-tiles --dry-run-level=3`

gives among other outputs showing the reversed order:
`INFO:procfuncs.py:184:check_for_outputs_on_disk: stdstarfit job with exposure(s) [118525 118524] has no existing stdstars's. Submitting full camword=a0123456789.`

This is true even if we don't allow skysub exposures.
`desi_run_night -n 20220113 --tiles="4403" --z-submit-types=cumulative --all-tiles --dry-run-level=3`
produces the same behavior.

akremin:
I've identified the culprit as https://github.com/desihub/desispec/blob/c496e36e29beedead08bf7a4be33376fff9be788/py/desispec/scripts/submit_night.py#L175, which sorts on `TILEID` only. `astropy` appears to be inconsistent at the ordering of `EXPID`'s in this case, possibly sorting on one of the other columns. In a toy example with only two columns the original code does provide the correct sorting in each case I tried.

akremin:
Resolved with PR #1848

PR #1848, opened 2022-09-10
Sort on EXPID as well as TILEID in desi_run_night

akremin:
This solves issue #1847 with a one line fix. I tested that it produces the correct output. Instead of sorting on `TILEID` alone, which appears to implicitly sort on other columns, it now sorts on `TILEID`, then `EXPID`. This should now always produce the correct numeric ordering of `EXPID`'s.

Now running: `desi_run_night -n 20220113 --tiles="4403" --z-submit-types=cumulative --laststeps="all,skysub,fluxcal" --all-tiles --dry-run-level=3`

shows the correct ordering of `EXPID`'s, as shown in this snippet:
`INFO:procfuncs.py:184:check_for_outputs_on_disk: stdstarfit job with exposure(s) [118524 118525]`

### DIFF
--- a/py/desispec/scripts/submit_night.py
+++ b/py/desispec/scripts/submit_night.py
@@ -172,7 +172,7 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
 
     ## Sort science exposures by TILEID
     sciexps = (etable['OBSTYPE']=='science')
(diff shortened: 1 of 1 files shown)


akremin:
The above unit test failures have to do with numpy and don't seem related to this one line PR. I am self-merging so I can proceed with re-running some tiles.

Issue #1851, opened 2022-09-14
Add emlinefit to the Redshift Scripts


akremin:
The `emlinefit`'s were processed separately in Fuji and Guadalupe, but are now mature enough to be integrated into the redshift code alongside the other afterburners.

Here is an example usage from Anand:
```
RRFN=/global/cfs/cdirs/desi/spectro/redux/guadalupe/tiles/cumulative/1000/20210517/redrock-0-1000-thru20210517.fits
COFN=/global/cfs/cdirs/desi/spectro/redux/guadalupe/tiles/cumulative/1000/20210517/coadd-0-1000-thru20210517.fits
OUTFN=/global/cfs/cdirs/desi/spectro/redux/guadalupe/tiles/cumulative/1000/20210517/emline-0-1000-thru20210517.fits
LOGFN=/global/cfs/cdirs/desi/spectro/redux/guadalupe/tiles/cumulative/1000/20210517/logs/emline-0-1000-thru20210517.log

desi_emlinefit_afterburner --redrock $RRFN --coadd $COFN --output $OUTFN > $LOGFN 2>&1
```

sbailey:
implemented in PR #1852.  Closing.

PR #1852, opened 2022-09-15
Add emlinefit afterburner to redshift scripts

akremin:
### Overview
This addresses issue #1851 . `emlinefit` was run by-hand in fuji and guadalupe but is now ready to be incorporated into the pipeline. The implementation directly mimics those of the QSO afterburners, which were already included in the redshift script.

### Testing
I tested this on tile 22469 from night 20220514. The scripts and results can be found here at: `/global/cfs/cdirs/desi/users/kremin/PRs/integrate_emlinefit`

The test included `perexp`, `pernight`, and `cumulative` redshifts. The tile had two exposures to test combining spectra (though none of that code was modified).

The `emlinefit` code was called for all spectrographs and the outputs and logs were written to the proper locations in all flavors of the redshifts tested.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -58,6 +58,7 @@ jobs:
                 python-version: ['3.9'] # fuji+guadalupe version
                 astropy-version: ['==5.0']  # fuji+guadalupe version
                 fitsio-version: ['==1.1.6']  # fuji+guadalupe version
+                numpy-version: ['<1.23']  # to keep asscalar, used by astropy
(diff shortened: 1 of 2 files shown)


sbailey:
Updates for emlinefit look good.  I added the github test hooks to pin the numpy version for the coverage tests (unrelated to this PR itself, but needed for the tests to pass due to astropy 5.0 being incompatible with the latest numpy that removed `asscalar`).

sbailey (approved):

Issue #1836, opened 2022-08-22
coadd-redshifts script cframes glob getting temp files too


sbailey:
The $CFRAMES glob in the coadd-redshifts*.slurm scripts is e.g.
```
CFRAMES=$(ls exposures/20201220/00069042/cframe-[brz]$SPECTRO-00069042.fits*)
```
That was intended to work with either gzipped or not input cframes, but it is also picking up temporary files leftover from crashes, e.g.
```
cframe-b1-00069042.fits_tmp42870.gz
```

Two items:
* Make a more specific glob
* Update the `desispec.io.util.get_tempfilename` to special-case the `.gz` extension so that this tempfile would have been `cframe-b1-00069042_tmp42870.fits.gz` instead which would not have passed the original glob.
  * context: `get_tempfilename` is creating a temporary filename in the same directory while preserving the extension so that code like `Table.write(filename)` can still auto-derive that this is a fits file.  In this case the extension is effectively `.fits.gz` not just `.gz`.

akremin:
I am not a glob expert, but I am able to get `CFRAMES=$(ls exposures/20201220/00069042/cframe-[brz]$SPECTRO-00069042.fits{,.gz} 2>/dev/null)` to work reliably on the command line. I don't like that I need to redirect the error outputs, but that is required with this solution since one of the two curly bracket evaluations will always fail.

Let me know if that is acceptable or if you can come up with something better and I can implement this @sbailey .

sbailey:
@akremin good solution, go for it, thanks

sbailey:
implemented in PR #1853 .  Closing issue.

PR #1853, opened 2022-09-19
Improve cframe file globbing and tempfilename

akremin:
### Overview
This solves issue #1836 where the globbing in the redshift scripts is picking up temporary files that are labeled as `*.fits_temp??????.gz` when we only want it finding `*.fits` and `*.fits.gz`. This resolves that using a "{" brackets to only check those two specific file endings. This produces an error for one or the other (either `cframes` were all gzipped or not so one of the two `ls` evaluations always fails to find any files). That means we have to catch the error and then we need a new mechanism to check whether the glob succeeded. I use the number of glob templates compared to the identified `cframes` as a test of glob success instead. It is a lower limit, but I can't be more restrictive, since a given glob could correspond to anywhere from one to three exposures (depending on the number of cameras).

This also solves the real problem which is that `get_tempfilename` was naming `.fits.gz` files as `fits_temp??????.gz` instead of `_temp??????.fits.gz`. The code now generically supports any `*.ext.fz` and `*.ext.gz` files, and adds new unit test cases to check that it is working.

### Testing
I ran this for one example tile, 8934, on one example night, 20220225. The test files can be found here:
`/global/cfs/cdirs/desi/users/kremin/PRs/betterglobbing`

Specifically:
`/global/cfs/cdirs/desi/users/kremin/PRs/betterglobbing/run/scripts/tiles/cumulative/8934/20220225/coadd-redshifts-8934-thru20220225.slurm`  shows the updated globbing is written properly. Then
`/global/cfs/cdirs/desi/users/kremin/PRs/betterglobbing/run/scripts/tiles/cumulative/8934/20220225/coadd-redshifts-8934-thru20220225-62848639.log` and `/global/cfs/cdirs/desi/users/kremin/PRs/betterglobbing/tiles/cumulative/8934/20220225/logs/spectra-0-8934-thru20220225.log` show that the globbing is finding six cframes, as expected for two exposures.

### DIFF
--- a/py/desispec/io/util.py
+++ b/py/desispec/io/util.py
@@ -787,11 +787,9 @@ def get_tempfilename(filename):
     """
     pid = os.getpid()
     if filename.endswith('.gz'):
-        second_ext = '.gz'
-        filename = filename.rstrip(second_ext)
+        filename, second_ext = os.path.splitext(filename)
     elif filename.endswith('.fz'):
-        second_ext = '.fz'
-        filename = filename.rstrip(second_ext)
+        filename, second_ext = os.path.splitext(filename)
     else:
         second_ext = ''
     base, extension = os.path.splitext(filename)
--- a/py/desispec/io/util.py
+++ b/py/desispec/io/util.py
@@ -785,9 +785,17 @@ def get_tempfilename(filename):
     race condition protection (the last one do do os.rename wins, but at least
     different processes won't corrupt each other's files).
     """
-    base, extension = os.path.splitext(filename)
     pid = os.getpid()
-    tempfile = f'{base}_tmp{pid}{extension}'
+    if filename.endswith('.gz'):
+        second_ext = '.gz'
+        filename = filename.rstrip(second_ext)
+    elif filename.endswith('.fz'):
+        second_ext = '.fz'
+        filename = filename.rstrip(second_ext)
+    else:
+        second_ext = ''
+    base, extension = os.path.splitext(filename)
+    tempfile = f'{base}_tmp{pid}{extension}{second_ext}'
     return tempfile
 
 def addkeys(hdr1, hdr2, skipkeys=None):
--- a/py/desispec/io/util.py
+++ b/py/desispec/io/util.py
@@ -786,9 +786,7 @@ def get_tempfilename(filename):
     different processes won't corrupt each other's files).
     """
     pid = os.getpid()
-    if filename.endswith('.gz'):
-        filename, second_ext = os.path.splitext(filename)
-    elif filename.endswith('.fz'):
+    if filename.endswith(('.gz', '.fz')):
         filename, second_ext = os.path.splitext(filename)
     else:
         second_ext = ''
--- a/py/desispec/scripts/tile_redshifts.py
+++ b/py/desispec/scripts/tile_redshifts.py
@@ -405,11 +405,11 @@ for SPECTRO in {spectro_string}; do
         echo $(basename $spectra) already exists, skipping grouping
     else
         # Check if any input frames exist
-        # Use echo instead of ls because it doesn't raise an error for missing entries
-        CFRAMES=$(echo {frame_glob})
+        # Use either .fits or .fits.gz search will fail and throw error, so catch them
+        CFRAMES=$(ls {frame_glob} 2>/dev/null)
         NUM_TERMS=$(echo "{frame_glob}" | wc -w)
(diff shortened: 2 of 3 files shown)


py/desispec/io/util.py
```diff
@@ -785,9 +785,15 @@ def get_tempfilename(filename):
     race condition protection (the last one do do os.rename wins, but at least
     different processes won't corrupt each other's files).
     """
-    base, extension = os.path.splitext(filename)
     pid = os.getpid()
-    tempfile = f'{base}_tmp{pid}{extension}'
+    if filename.endswith('.gz'):
+        filename, second_ext = os.path.splitext(filename)
+    elif filename.endswith('.fz'):
+        filename, second_ext = os.path.splitext(filename)
```
sbailey, line 792: Minor:  `filename.endswith( ('.gz', '.fz') )` would also work and avoids duplicating the logic of what to do when the file ends with .fz or .gz.  Note: `.endswith` works with str or tuple of str, but not list of str.

py/desispec/test/test_io.py
```diff
@@ -1202,11 +1202,25 @@ def test_get_tempfilename(self):
...
+        tempfile = get_tempfilename(filename)
+        self.assertNotEqual(filename, tempfile)
+        self.assertTrue(tempfile.endswith('.fits.gz'))
+        self.assertTrue('/a/b/blat' in tempfile)
+
+        filename = '/a/b/blat.fits.fz'
+        tempfile = get_tempfilename(filename)
+        self.assertNotEqual(filename, tempfile)
+        self.assertTrue(tempfile.endswith('.fits.fz'))
+        self.assertTrue('/a/b/blat' in tempfile)
```
sbailey, line 1217: Your updated `get_tempfilename` conveniently "does the right thing" even if there is a single `.gz` extension, i.e. `blat.gz` -> `blat_tmp14074.gz` and `blat.fits.gz` -> `blat_tmp14074.fits.gz`.  Let's add a test to explicitly check that since we don't want to accidentally break it in the future and some other implementation might crash with a single extension `.gz` .

py/desispec/scripts/tile_redshifts.py
```diff
@@ -404,11 +405,12 @@ def write_redshift_script(batchscript, outdir,
...
-        CFRAMES=$(ls {frame_glob})
-        MISSING_CFRAMES=$?
+        # Use either .fits or .fits.gz search will fail and throw error, so catch them
+        CFRAMES=$(ls {frame_glob} 2>/dev/null)
+        NUM_TERMS=$(echo "{frame_glob}" | wc -w)
         NUM_CFRAMES=$(echo $CFRAMES | wc -w)
-        if [ $MISSING_CFRAMES -ne 0 ] && [ $NUM_CFRAMES -gt 0 ]; then
-            echo ERROR: some expected cframes missing for spectrograph $SPECTRO but proceeding anyway
+        if [ $NUM_TERMS -gt $NUM_CFRAMES ]; then
+            echo ERROR: some expected cframes missing for spectrograph $SPECTRO with $NUM_TERMS search terms but only $NUM_CFRAMES found. Proceeding anyway
```
sbailey, line 413: After writing and deleting 3 draft comments, I'll discuss this with you in person.

sbailey (approved):
This addresses the core issue, thanks.  I put a few minor comments inline, and one I want to discuss with you in person about how to track number of expected cframes and what we even mean by that.

py/desispec/test/test_io.py
akremin: This should be resolved in 638e35f

py/desispec/io/util.py
akremin: This should be resolved in 638e35f

py/desispec/scripts/tile_redshifts.py
akremin: As we discussed I have explicitly defined the number of exposures rather than having bash count globs. I replaced the ERROR message here with an informative message used for all spectrographs (not under an `if` statement). I did retain a WARNING statement that just helps guide the eye (or a `grep`) if the number of `cframes` is fewer than the number of exposures. That can genuinely happen since the code isn't smart enough to figure out how many it should have, but I think a WARNING is useful even if there are false alarms. Let me know if you disagree and we can iterate.

akremin:
I re-tested the script submission and it still works with the updates generated from the comments. I also verified that the scripts run without unexpected issues.

PR #1855, opened 2022-09-22
fix psferr option in desi_extract_spectra

dmargala:
This PR updates `desi_extract_spectra` to pass the value of its `--psferr` command line argument through to `specter.extract.ex2d`.
To preserve consistency with fuji and guadalupe reductions, this PR also removes the `--psferr 0.1` added to the `desi_extract_spectra` command in `desi_proc` since that was ignored anyway.

The default value for the `--psferr` command line argument in  `desi_extract_spectra` is `None`. Both specter and gpu_specter handle the case of `psferr is None` in the same way: they use the value of the "PSFERR" keyword in the PSF header and, if that is not defined, they use a default value of `0.01`.

`desi_extract_spectra` is already passing `psferr` through when using gpu_specter so no action is needed there. (Comparing extractions using desi_proc with specter/gpu_specter revealed this issue, see desihub/gpu_specter#75)

From what I can tell in git history, `psferr` used to be forwarded to specter from `desi_extract_spectra` for the non-mpi code path but that was never added to the mpi code path. At some point (> 2 yrs ago), those code paths were mostly merged but passing `psferr` to `specter.extract.ex2d` did not survive.

### DIFF
--- a/py/desispec/scripts/extract.py
+++ b/py/desispec/scripts/extract.py
@@ -747,7 +747,7 @@ def _extract_and_save(img, psf, bspecmin, bnspec, specmin, wave, raw_wave, fiber
     results = ex2d(img.pix, img.ivar*((img.mask & extractmaskval)==0), psf, bspecmin,
                    bnspec, wave, regularize=args.regularize, ndecorr=args.decorrelate_fibers,
                    bundlesize=bundlesize, wavesize=args.nwavestep, verbose=args.verbose,
-                   full_output=True, nsubbundles=args.nsubbundles,
+                   full_output=True, nsubbundles=args.nsubbundles, psferr=args.psferr,
                    wavepad_frac=args.wavepad_frac, pixpad_frac=args.pixpad_frac)
 
     flux = results['flux']
--- a/py/desispec/scripts/proc.py
+++ b/py/desispec/scripts/proc.py
@@ -774,6 +774,7 @@ def main(args=None, comm=None):
                 cmd += ' -i {}'.format(preprocfile)
                 cmd += ' -p {}'.format(psffile)
                 cmd += ' -o {}'.format(framefile)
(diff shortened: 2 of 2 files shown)


sbailey:
Thanks @dmargala for tracking this down and the simple bug fix.  Now that desi_extract_spectra correctly passes the psferr option to specter, I suggest that proc.py should continue to set `--psferr` option when calling `desi_extract_spectra`, but just make it 0.01 instead of dropping it and relying upon both specter and gpu_specter to have the same default 0.01 (which they currently do).  I think this is still compatible with fuji / guadalupe, because those were ignoring the option and using 0.01, and now it would be respecting the option and using 0.01.  i.e. I'd rather have desispec be the "owner" of what default psferr we use for productions, rather than specter/gpu_specter.

Since this PR came from your fork rather than a desihub/desispec branch, I don't have write permission to make that update myself.

dmargala:
Okay, that sounds reasonable to be. One thing I didn't check is if the "PSFERR" keyword is ever set in the header of the PSF input files and if its value is different than 0.01 in fuji / guadalupe. If that were ever the case, then fixing psferr option and supplying the value of 0.01 on the command line would introduce a change from the fuji / guadalupe.

I can update this PR with your suggestion. (This PR is from desihub/desispec, not a fork. The branch name is `dmargala/fix-psferr-option` so maybe it looks like it came from a fork. I don't recall where I picked up that habit but it helps me remember which branches I care about)

sbailey:
Thanks.  Knowing that this is a branch with a "/" in the name got my past my git syntax confusion and I was able to push the update I wanted.

I spot checked fugu PSFs and we don't set PSFERR in the header, so this should be backwards compatible with that too.  If we ever do start setting PSFERR in the PSF file headers, we could then revisit whether to force it in extractions (probably not).

Issue #1864, opened 2022-10-04
sky model failures on 20210529


sbailey:
In the Himalayas run we had multiple failures of the new PCA sky model code for exposures that worked with Fuji.  An example is

```
desi_compute_sky -i /global/cfs/cdirs/desi/spectro/redux/himalayas/exposures/20210529/00090224/frame-r3-00090224.fits.gz --fiberflat /global/cfs/cdirs/desi/spectro/redux/himalayas/exposures/20210529/00090224/fiberflatexp-r3-00090224.fits.gz -o /global/cfs/cdirs/desi/users/$USER/temp/sky-r3-00090224.fits.gz --adjust-wavelength --adjust-lsf --pca-corr /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm6/skycorr-pca-sm6-r3.fits --fit-offsets --skygradpca /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm6/skygradpca-sm6-r3.fits --tpcorrparam /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm6/tpcorrparam-sm6-r3.fits
...
Intel MKL ERROR: Parameter 4 was incorrect on entry to DGELSD.
Traceback (most recent call last):
  File "/global/common/software/desi/perlmutter/desiconda/20220119-2.0.1/code/desispec/main/bin/desi_compute_sky", line 17, in <module>
    sys.exit(sky.main(args))
  File "/global/common/software/desi/perlmutter/desiconda/20220119-2.0.1/code/desispec/main/py/desispec/scripts/sky.py", line 109, in main
    skymodel = compute_sky(frame,add_variance=(not args.no_extra_variance),\
  File "/global/common/software/desi/perlmutter/desiconda/20220119-2.0.1/code/desispec/main/py/desispec/sky.py", line 564, in compute_sky
    res = compute_sky_linear(
  File "/global/common/software/desi/perlmutter/desiconda/20220119-2.0.1/code/desispec/main/py/desispec/sky.py", line 373, in compute_sky_linear
    skytpcorrcoeff = np.linalg.lstsq(
  File "<__array_function__ internals>", line 5, in lstsq
  File "/global/common/software/desi/perlmutter/desiconda/20220119-2.0.1/conda/lib/python3.9/site-packages/numpy/linalg/linalg.py", line 2306, in lstsq
[...]

PR #1865, opened 2022-10-04
Sanitize x/y inputs to skytpcorr, skygradpca.

schlafly:
This PR replaces NaNs in FIBERASSIGN_X/Y in skygradpca with 0s.  It also replaces NaNs and 0s in skytpcorr with the nominal locations of the fibers.

This addresses https://github.com/desihub/desispec/issues/1864.  This is unlikely to change any tile that previously succeeded, since any NaNs in FIBERASSIGN_X/Y would lead to all NaNs in the corresponding skygradpca columns due to the removal of the mean.  It's conceptually possible that there were other cases of FIBER_X/Y = 0 without FIBERASSIGN_X/Y = NaN that would now be set to default nominal values in skytpcorr; i.e., this would be a ~non-backward compatible change.  I don't know if any such tiles exist, and they'd clearly be problematic.

Ideally I'd set all NaNs / 0s to the nominal centers of the fibers, but I don't know how to conveniently access that from skygradpca land.  The sky gradients are focal-plane wide so it's not too bad to put stuff at the middle.  For skytpcorr, a polynomial is being computed that is centered on the nominal center of the patrol disc of the positioner, so I really don't want to set X/Y to zero for those... that could lead to a 400 being passed into a cubic function where 6 is the normal maximum.  Thus the more careful treatment in skytpcorr, where I do conveniently store nominal locations for all of the fibers.

### DIFF
--- a/py/desispec/skygradpca.py
+++ b/py/desispec/skygradpca.py
@@ -51,6 +51,11 @@ def configure_for_xyr(skygradpca, x, y, R, skyfibers=None):
         R: list of resolution matrices for each fiber (unitless)
         skyfibers: (optional) indices of sky fibers in list of all fibers
     """
+    x = x.copy()
+    y = y.copy()
+    m = ~np.isfinite(x) | ~np.isfinite(y)
+    x[m] = 0
+    y[m] = 0
     skygradpca.x = x
     skygradpca.y = y
     skygradpca.dx = x - np.mean(x)
--- a/py/desispec/tpcorrparam.py
+++ b/py/desispec/tpcorrparam.py
@@ -76,6 +76,11 @@ def tpcorrmodel(tpcorrparam, xfib, yfib, pcacoeff=None):
         Throughput model evaluated at xfib, yfib with pca coefficients
         pcacoeff.
     """
+    xfib = xfib.copy()
+    yfib = yfib.copy()
+    m = ~np.isfinite(xfib) | ~np.isfinite(yfib) | (xfib == 0) | (yfib == 0)
+    xfib[m] = tpcorrparam.spatial['X'][m]
+    yfib[m] = tpcorrparam.spatial['Y'][m]
     res = tpcorrparam.mean + tpcorrspatialmodel(
         tpcorrparam.spatial, xfib, yfib) - 1
     if pcacoeff is not None:


schlafly:
It seems unlikely to me that I've injected any bugs here, but out of an abundance of caution I'd like to rerun an existing daily sky.  I see in the logs messages like
```
INFO:util.py:123:runcmd: RUNNING: desispec.scripts.sky.main(['-i', '/global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220924/00144230/frame-r2-00144230.fits.gz', '--fiberflat', '/global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220924/00144230/fiberflatexp-r2-00144230.fits.gz', '-o', '/global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220924/00144230/sky-r2-00144230.fits.gz', '--adjust-wavelength', '--adjust-lsf', '--pca-corr', '/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm5/skycorr-pca-sm5-r2.fits', '--fit-offsets', '--skygradpca', '/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm5/skygradpca-sm5-r2.fits', '--tpcorrparam', '/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm5/tpcorrparam-sm5-r2.fits'])
```
is that as close as I can come to exactly what I need to invoke to reproduce a particular daily run?

sbailey:
@schlafly yes, as a side-effect of Perlmutter not letting us spawn actual commands from MPI-parallelized code, the logs now print what actually happens (a function call) rather than the more user-friendly in-principle-equivalent command line that it used to spawn.  i.e. you can test that from a python prompt with

```python
import desispec.scripts.sky
desispec.scripts.sky.main(['-i', '/global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220924/00144230/frame-r2-00144230.fits.gz', '--fiberflat', '/global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220924/00144230/fiberflatexp-r2-00144230.fits.gz', '-o', '/global/cfs/cdirs/desi/spectro/redux/daily/exposures/20220924/00144230/sky-r2-00144230.fits.gz', '--adjust-wavelength', '--adjust-lsf', '--pca-corr', '/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm5/skycorr-pca-sm5-r2.fits', '--fit-offsets', '--skygradpca', '/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm5/skygradpca-sm5-r2.fits', '--tpcorrparam', '/global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm5/tpcorrparam-sm5-r2.fits'])
```

schlafly:
Thanks.  It might still be nice to print out the corresponding shell invocation, but you're right that it was plenty easy to just do what you described.  Thanks.

For my one test sky, fitsdiff reports no differences in values.  A lot of different checksums, which maybe include the header keywords?  All of the DATASUMs are the same, except for the comments that have the times in them.  Some other different modest unrelated keywords (SPECPROD, DEPVER00 having a trailing slash, FIBERFLT and IN_FRAME having some SPECPROD expansion in one case and not the other.  So I think I'm happy that I didn't mess up anything important in this PR.

So this is good to merge from my perspective; let me know if I should run more tests.

sbailey:
Thanks for the quick fix.  While you were testing+typing I was doing the same.  The code looks good, I confirmed that it fixes the failing case, and for a succeeding case (r1 on same exposure) it produces bitwise identical data outputs and keywords only differ by harmless things like timestamps.

Converting the function call into the command line for the succeeding case because yeah, those function-call cut-and-pastes are ugly and a pain:
```
desi_compute_sky -i /global/cfs/cdirs/desi/spectro/redux/himalayas/exposures/20210529/00090224/frame-r1-00090224.fits.gz --fiberflat /global/cfs/cdirs/desi/spectro/redux/himalayas/exposures/20210529/00090224/fiberflatexp-r1-00090224.fits.gz -o /global/cfs/cdirs/desi/users/$USER/temp/sky-r1-00090224.fits.gz --adjust-wavelength --adjust-lsf --pca-corr /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm10/skycorr-pca-sm10-r1.fits --fit-offsets --skygradpca /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm10/skygradpca-sm10-r1.fits --tpcorrparam /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm10/tpcorrparam-sm10-r1.fits
fitsdiff /global/cfs/cdirs/desi/spectro/redux/himalayas/exposures/20210529/00090224/sky-r1-00090224.fits.gz /global/cfs/cdirs/desi/users/sjbailey/temp/sky-r1-00090224.fits.gz
```

Side note: the logging is a bit of a hassle because `desi_proc` knows the name of the function it is calling with what arguments, but there isn't a guaranteed algorithmic mapping of which command line script is the wrapper for that function.  So to print the equivalent command line, something would need to track the reverse mapping of which bin/ scripts wrap which desispec.scripts.blah.main functions.  Possible, but not done, and potentially fragile to maintain.

schlafly:
Thanks, should I go ahead and merge?  Re logging, maybe just printing the ' '.join(args) would be largely adequate; I'm willing to add in desi_compute_sky, but the copy/pasting of the arguments is more annoying.  :-)

Issue #1812, opened 2022-07-29
coadd-redshift grouping in slurm scripts


marcelo-alvarez:
Currently multiple sets of independent coadd-redshift jobs, each one for a given spectrograph, are grouped together in a single slurm script and run in the background, usually as a single 10-node job. This is not necessary, since the jobs are actually independent of one another, and can delay processing overall depending on constraints from the queue.

We should find a more optimal way of packaging the jobs. This can be accomplished by submitting one job per spectrograph, on a single node. But there may be other ways that are more efficient, so this should be discussed before making the changes. Mentioning @sbailey @akremin to keep them in the loop.

sbailey:
I suggest one job for all 10 spectrographs, but only using 1 node and looping serially over petals.  The QA step requires information from all 10 petals after redrock is done, so orchestrating that will be easiest by keeping it all within one job, and that also avoids inflating the number of jobs.

This is related to desihub/redrock#217 that proposes a multi-petal wrapper around rrdesi so that the python process would only wakeup+import once (~60 seconds), and then loop over petals serially (~40 seconds each).  We can also experiment with running 2 petals in parallel using 2 GPUs each, etc.

Other steps like afterburners may also benefit from MPI wrappers that could could process N petals in parallel on a single node instead of the current bash-orchestrated launching 10 sruns to 10 nodes using a single process each.

marcelo-alvarez:
This is being addressed in [desispec/desi_zproc](https://github.com/desihub/desispec/tree/desi_zproc).

sbailey:
Implemented in PR #1870.   Closing ticket.

PR #1870, opened 2022-10-06
Single-node redrock slurm scripts on Perlmutter 

marcelo-alvarez:
Sets the number of nodes to one on Perlmutter for redshift slurm scripts. Partially addresses #1812, but still results in a full round of python imports for each spectrograph.

@akremin please review and verify that this works as expected for your single-node pipeline testing on Perlmutter.


akremin:
Thanks Marcelo. I think we will also need to adjust the run time, do you have a good sense of the scaling and whether that is also an easy patch?

marcelo-alvarez:
I think 0.5 minutes per spectrograph  plus 5 minutes overhead would be a good first guess, so 10 minutes for the typical tile. Then we could tighten that up depending on the performance.

sbailey:
This is superseded by #1904 .  Closing.

PR #1872, opened 2022-10-11
fix stdstar indexing crash

### LINKED ISSUE #1871: stdstar crash tile 82406 20211118/00109351
desi_fit_stdstars crashes on backup tile 82406 on 20211118/00109351 due to a bookkeeping inconsistency
about what fibers are standard stars:

```
cd /global/cfs/cdirs/desi/spectro/redux/himalayas
desi_fit_stdstars --delta-color 0.1 \
  --frames exposures/20211118/00109351/frame-?0-00109351.fits.gz \
  --skymodels exposures/20211118/00109351/sky-?0-00109351.fits.gz \
  --fiberflats calibnight/20211118/fiberflatnight-?0-20211118.fits \
  --outfile $SCRATCH/stdstars-0-00109351.fits.gz
...
INFO:stdstars.py:506:main: No Legacy survey photometry identified in fibermap
INFO:stdstars.py:516:main: computing model mags for ['GAIA-BP', 'GAIA-G', 'GAIA-RP']
INFO:stdstars.py:520:main: done computing model mags
INFO:stdstars.py:608:main: rank 0: finding best model for observed star #0
INFO:stdstars.py:644:main: star#0 fiber #452, GAIA-BP-RP = 0.308243, number of pre-selected models = 180/1491
INFO:stdstars.py:663:main: Star Fiber: 452; TEFF: 7000.000; LOGG: 5.000; FEH: 0.371; Redshift: -3.99988e-05; Chisq/dof: 0.500
INFO:stdstars.py:680:main: Applying MW dust extinction to star 0 with EBV = 1.0251508951187134
INFO:stdstars.py:713:main: scaling GAIA-G mag 8.072 to 17.422 using scale 0.00018189888269967697
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/main/bin/desi_fit_stdstars", line 21, in <module>
    sys.exit(stdstars.main(args))
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/main/py/desispec/scripts/stdstars.py", line 741, in main
    assert np.all(fibermap['FIBER'] == starfibers[fitted_stars])
AssertionError
```

Note that it is using Gaia standards and not Legacy Survey standards.  If that turns out to be related to the problem, let's get unit tests in place that cover whatever this case is, since Gaia vs. LS has been a historic point of maintenance fragility, and our typical test tiles (e.g. sv3 deep tiles) don't use Gaia standards.

sbailey:
This PR fixes #1871 where an indexing bug lead to standard star crashes.  It happened to be on a backup tile with Gaia standards, but the actual bug was not directly related to that.  It was a problem with a pre-filter that the standard stars have valid photometry and fragility with multiple ways of filtering: slicing frames, tracking lists of indices, tracking lists of booleans.  This update simplifies the tracking by dropping the independent (and inconsistent in this case) tracking of indices, treating slicing frames as the primary method.  It does still keep the `starfibers` array which is redundant with `frame.fibermap['FIBER']` but convenient for logging and convenient since there are N>1 frames but they all have the same fiber selections (which is explicitly verified).

Command that crashes on main and now runs:
```
cd $CFS/desi/spectro/redux/himalayas
export SP=0 && export EXPID=00109351 && desi_fit_stdstars --delta-color 0.1 \
  --frames exposures/20211118/${EXPID}/frame-?${SP}-${EXPID}.fits.gz \
  --skymodels exposures/20211118/${EXPID}/sky-?${SP}-${EXPID}.fits.gz \
  --fiberflats calibnight/20211118/fiberflatnight-?${SP}-20211118.fits \
  --outfile $SCRATCH/stdstars-${SP}-${EXPID}.fits.gz
```

Detail 1: with the correct indexing, none of the standard stars pass SNR(B) cuts and it exists with an informative error message about that but not a crash like before.  i.e. this PR doesn't recover that exposure.

Detail 2: this expid/spectrograph did "succeed" in daily, why not now?  That turns out to be due to some bugs fixed in PR #1817.  Previously (at the time daily was run) stdstar was not correctly scaling ivar to account for the fiberflat and sky, and using the wrong indices too, which led to over-optimistic SNR and having stars "pass" cuts.  With the correct ivar propagation, they no longer pass.  I suspect we will discover more classes like this before Himalayas/Iron are done.

Cross check: For spectrographs that do succeed, this PR produces the same output as current main (himalayas), e.g.
```
export SP=1 && export EXPID=00109351 && desi_fit_stdstars --delta-color 0.1 \
  --frames exposures/20211118/${EXPID}/frame-?${SP}-${EXPID}.fits.gz \
  --skymodels exposures/20211118/${EXPID}/sky-?${SP}-${EXPID}.fits.gz \
  --fiberflats calibnight/20211118/fiberflatnight-?${SP}-20211118.fits \
  --outfile $SCRATCH/stdstars-${SP}-${EXPID}.fits.gz

export SP=2 && export EXPID=00109340 && desi_fit_stdstars --delta-color 0.1 \
  --frames exposures/20211118/${EXPID}/frame-?${SP}-${EXPID}.fits.gz \
  --skymodels exposures/20211118/${EXPID}/sky-?${SP}-${EXPID}.fits.gz \
  --fiberflats calibnight/20211118/fiberflatnight-?${SP}-20211118.fits \
  --outfile $SCRATCH/stdstars-${SP}-${EXPID}.fits.gz

fitsdiff -k CHECKSUM,DEP\* -c DATASUM,CHECKSUM $CFS/desi/spectro/redux/himalayas/exposures/20211118/00109351/stdstars-1-00109351.fits.gz $SCRATCH/stdstars-1-00109351.fits.gz
fitsdiff -k CHECKSUM,DEP\* -c DATASUM,CHECKSUM $CFS/desi/spectro/redux/himalayas/exposures/20211118/00109340/stdstars-2-00109340.fits.gz $SCRATCH/stdstars-2-00109340.fits.gz
```
(SCRATCH=/global/cscratch1/sd/sjbailey if you want to look at the files without re-running yourself).

Caveats: bitwise identical if on KNL, otherwise close but not identical.  Header keywords differences are due to pipeline being run with MPI thus including an mpi4py version, and harmless timestamps and development checkout versions vs. installed versions.

Came along for the ride: `io.read_frame` wasn't setting `frame.spectrograph`, now it is (which also required test updates which were generating frame files with inconsistent spectrographs given their fibers).

### DIFF
--- a/py/desispec/scripts/stdstars.py
+++ b/py/desispec/scripts/stdstars.py
@@ -203,7 +203,6 @@ def main(args=None, comm=None) :
 
     spectrograph=None
     starfibers=None
-    starindices=None
     fibermap=None
     # For each unique expid,spec pair, get the logical OR of the FIBERSTATUS for all
     # cameras and then proceed with extracting the frame information
@@ -223,42 +222,30 @@ def main(args=None, comm=None) :
             frame = get_fiberbitmasked_frame(frame,bitmask='stdstars',ivar_framemask=True)
             frame_fibermap = frame.fibermap
 
-            # frame was filtered to just stdstars upon reading, so initial list of starindices is full range
-            frame_starindices = np.arange(len(frame.fibermap), dtype=int)
-
             #- Confirm that all fluxes have entries but trust targeting bits
             #- to get basic magnitude range correct
-            keep_legacy = np.ones(len(frame_starindices), dtype=bool)
-
+            keep_legacy = np.ones(len(frame_fibermap), dtype=bool)
             for colname in ['FLUX_G', 'FLUX_R', 'FLUX_Z']:  #- and W1 and W2?
-                keep_legacy &= frame_fibermap[colname][frame_starindices] > 10**((22.5-30)/2.5)
-                keep_legacy &= frame_fibermap[colname][frame_starindices] < 10**((22.5-0)/2.5)
-            keep_gaia = np.ones(len(frame_starindices), dtype=bool)
+                keep_legacy &= frame_fibermap[colname] > 10**((22.5-30)/2.5)
+                keep_legacy &= frame_fibermap[colname] < 10**((22.5-0)/2.5)
 
-            for colname in ['G', 'BP', 'RP']:  #- and W1 and W2?
-                keep_gaia &= frame_fibermap['GAIA_PHOT_'+colname+'_MEAN_MAG'][frame_starindices] > 10
-                keep_gaia &= frame_fibermap['GAIA_PHOT_'+colname+'_MEAN_MAG'][frame_starindices] < 20
+            keep_gaia = np.ones(len(frame_fibermap), dtype=bool)
+            for colname in ['G', 'BP', 'RP']:
+                keep_gaia &= frame_fibermap['GAIA_PHOT_'+colname+'_MEAN_MAG'] > 10
+                keep_gaia &= frame_fibermap['GAIA_PHOT_'+colname+'_MEAN_MAG'] < 20
+
             n_legacy_std = keep_legacy.sum()
             n_gaia_std = keep_gaia.sum()
-            keep = keep_legacy | keep_gaia
-            # accept both types of standards for the time being
-
-            # keep the indices for gaia/legacy subsets
-            is_gaia_std = keep_gaia[keep]
-            is_legacy_std = keep_legacy[keep]
-
-            frame_starindices = frame_starindices[keep]
 
             if spectrograph is None :
                 spectrograph = frame.spectrograph
                 fibermap = frame_fibermap
-                starindices=frame_starindices
-                starfibers=np.asarray(fibermap["FIBER"][starindices])
+                starfibers=np.asarray(fibermap["FIBER"])
 
             elif spectrograph != frame.spectrograph :
                 log.error("incompatible spectrographs {} != {}".format(spectrograph,frame.spectrograph))
                 raise ValueError("incompatible spectrographs {} != {}".format(spectrograph,frame.spectrograph))
-            elif starindices.size != frame_starindices.size or np.sum(starindices!=frame_starindices)>0 :
+            elif len(fibermap) != len(frame_fibermap) or np.any(fibermap['FIBER'] != frame_fibermap['FIBER']):
                 log.error("incompatible fibermap")
                 raise ValueError("incompatible fibermap")
 
@@ -309,10 +296,10 @@ def main(args=None, comm=None) :
         if n_gaia_std == 0 and gaia_color:
             raise Exception('Specified gaia color, but no gaia stds')
 
-    if starindices.size == 0:
+    if starfibers.size == 0:
         log.error("no STD star found in fibermap")
         raise ValueError("no STD star found in fibermap")
-    log.info("found %d STD stars" % starindices.size)
+    log.info("found %d STD stars" % starfibers.size)
 
     if n_legacy_std == 0:
         gaia_std = True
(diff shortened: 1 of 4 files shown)


py/desispec/scripts/stdstars.py
```diff
@@ -223,42 +222,30 @@ def main(args=None, comm=None) :
...
             frame_fibermap = frame.fibermap
 
-            # frame was filtered to just stdstars upon reading, so initial list of starindices is full range
-            frame_starindices = np.arange(len(frame.fibermap), dtype=int)
-
             #- Confirm that all fluxes have entries but trust targeting bits
             #- to get basic magnitude range correct
-            keep_legacy = np.ones(len(frame_starindices), dtype=bool)
-
+            keep_legacy = np.ones(len(frame_fibermap), dtype=bool)
```
akremin, line 227: `keep_legacy`, `keep_gaia`, `n_legacy_std`, and `n_gaia_std` all seem to be redefined each loop here over both exposures (outer loop) and cameras (inner loop), but then used generically later in the code outside of any loops. Is there an assumption made here that allows that to be true?

akremin (commented):
A much appreciated change to simplify the selection logic from using both indices and booleans to just using booleans (mostly). I do have a question about re-initializing those boolean arrays inline that I would like the answer to before I approve this PR. I will also run the code on a random exposure to ensure that it does the right thing (but I haven't done that yet).

py/desispec/scripts/stdstars.py
sbailey: `scripts.stdstars` requires all input frames to be from the same spectrograph of the same tile, so if given correct inputs then this calculation should be redundant.  This block of code at the end of the loop confirms that the selected stars from each frame are consistent with the ones cached from the first frame, which is the core assumption made later in the code that needs to be checked before proceeding:
```python
            keep = keep_legacy | keep_gaia
            ...
            frame_starindices = frame_starindices[keep]

            if spectrograph is None :
                spectrograph = frame.spectrograph
                fibermap = frame_fibermap
                starindices=frame_starindices
                starfibers=np.asarray(fibermap["FIBER"][starindices])
            elif spectrograph != frame.spectrograph :
                log.error("incompatible spectrographs {} != {}".format(spectrograph,frame.spectrograph))
                raise ValueError("incompatible spectrographs {} != {}".format(spectrograph,frame.spectrograph))
            elif starindices.size != frame_starindices.size or np.sum(starindices!=frame_starindices)>0 :
                log.error("incompatible fibermap")
                raise ValueError("incompatible fibermap")
```
the most common user error would be providing frames from multiple spectrographs or multiple tiles, in which case simply checking FIBER and TARGETID values would be sufficient, but this loop adds additional checks on the consistency.  The calculation is ~0.005 sec per frame so the runtime is trivial.  There is in theory a pathological corner case where `keep_legacy` and `keep_gaia` could be different across frames while leaving `keep = keep_legacy | keep_gaia` the same, which this check would not catch.  It also wouldn't catch if the selected stars had "valid" but different photometry while still having the same fibers.

The structure of the loop pre-dates this PR and could be refactored for arguably better maintainability by separating out the concept of "keep" (on one frame) from checking the consistency of TARGETID, FIBER, FLUX* , GAIA_PHOT_* etc. across all frames.  But unless you think the current code is actively confusing or especially fragile, I suggest leaving it as is.

akremin: Thannk you for the explanation. I now understand the flow of this and also agree that any inefficiency is trivial in the grand scheme of things.

Just for the record, the reference snippet above is the current `main` version of the code before your changes, which removed `keep`, `starindices`, and redefined  `starfibers`. I believe your points still stand, however, so I'll approve this once I test an example.

akremin (approved):
I tested the new code with a random example from a recent daily bright tile-exposure (Exp 146748, Night 20221008):
```
cd $CFS/desi/spectro/redux/daily
export SP=6; export EXPID=00146748; export NIGHT=20221008
desi_fit_stdstars --delta-color 0.1   \
     --frames exposures/${NIGHT}/${EXPID}/frame-?${SP}-${EXPID}.fits.gz  \
     --skymodels exposures/${NIGHT}/${EXPID}/sky-?${SP}-${EXPID}.fits.gz   \
     --fiberflats calibnight/${NIGHT}/fiberflatnight-?${SP}-${NIGHT}.fits   \
    --outfile $SCRATCH/stdstars-${SP}-${EXPID}.fits.gz
fitsdiff -k CHECKSUM,DEP\* -c DATASUM,CHECKSUM $CFS/desi/spectro/redux/daily/exposures/$NIGHT/$EXP/stdstars-${SP}-${EXP}.fits.gz $SCRATCH/stdstars-${SP}-${EXP}.fits.gz
```

The outputs were identical except for non-consequential header keywords.

PR #1875, opened 2022-10-18
desi_run_night with no zeros

sbailey:
This PR fixes cases like night 20211128 which had no good zeros.  With current main, the nightly bias job is created, but it exits with a non-zero error code because it can't find any zeros.  This prevents the rest of the job dependencies from running because they think the nightly bias failed when the intended behavior was just to fall back to using the default biases.  This PR updates the logic such that:
  1. if there are zeros and a dark (normal situation), it runs the ccdcalib job which does both the nightly bias and bad column identification.
  2. if there are no zeros and no dark (like 20211128), it skips ccdcalib and starts with the arcs
  3. if there are zeros but not a dark, it runs a nightlybias job then proceeds with arcs
  4. if there is a dark but no zeros, it runs a new category "badcol" job to look for bad columns, then proceeds with arcs

This PR was tested in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/nozeros using
```
desi_run_night --dry-run-level 1 -n $NIGHT &> run/logs/$NIGHT-submit.log
```

with exposure tables for 4 nights edited to create 4 cases:

1. night 20211127 - normal night with ZEROs, DARK, ARCs, FLATs, SCIENCE:
   Submits ccdcalib, arc, psfnight, flat, nightlyflat, science jobs as normal
2. night 20211128 - original failing case with no good ZEROs or DARK, but still ARCs FLATs and SCIENCE:
   Skips ccdcalib and goes directly to arc, psfnight, flat, nightlyflat, science jobs
3. night 20211129 - ZEROs but no DARK:
   Runs nightlybias job instead of ccdcalib, then proceeds with arcs et al
4. night 20211130 - DARK but no ZEROs:
   Runs badcol job instead of ccdcalib, then proceeds with arcs et al

I launched the nightlybias (case 3) and badcol (case 4) jobs to confirm they would work, and verified that the job dependencies in the processing table were correct for all 4 cases.  I did not explicitly test that the arc jobs could proceed even if missing biasnight and/or bacol files though I'll submit a single job for each of those cases now just in case.  I also did not test desi_daily_proc_manager for these cases (mainly because I'm not familiar with how to best test that).

@akremin please review and let's iterate on whether you think more testing is needed.


py/desispec/workflow/procfuncs.py
```diff
@@ -96,7 +96,9 @@ def check_for_outputs_on_disk(prow, resubmit_partial_complete=True):
             'stdstarfit': 'stdstars',
             'poststdstar': 'cframe',
             'nightlybias': 'biasnight',
+            'zero': 'biasnight',
```
akremin, line 99: Should this be `'bias': 'biasnight'`?

akremin (commented):
The changes look good except one in-line question. I would request 2 extra tests just to make absolutely sure:
1) dark+zeros at the end of the night, after the arcs+flats+sciences. This tests that the exposure shuffling is working as expected.
2) Only a dark at the end of the night. Similar to above except also testing the new badcol scenario.

If your original tests were end of night, then I would request the beginning of the night as an additional test just to make extra sure we have all cases covered. Hopefully these are not too hard to achieve either artificially or with appropriate real nights.

For the record: this does not implement `badcol` or no `biasnight` submission for the daily pipeline, but it does setup the tools to do so. I tested the daily processing code in this branch and it still runs without issue from the new pieces. I tested this with the most recent night of data using `desi_daily_proc_manager --override-night=20221016 --dry-run-level=1`.

py/desispec/workflow/procfuncs.py
sbailey: good question.  It turns out this was an extraneous addition.  The jobdesc I intended to cover was already the `'nightlybias': 'biasnight'` entry.  In the end `bias` is a special-case obstype (meaning process all of the zeros into a nightly bias), and `nightlybias` (no dark) or `ccdcalib` (with dark) are the resulting jobdesc entries.  I removed this line and re-ran all the tests to confirm it didn't break anything.

sbailey:
I added two more hand-edited nights to test those additional cases that you suggested:

* 20211201 - zeros+dark at end of night: correctly moves them to the beginning and runs a ccdcalib job before continuing with arcs.
* 20211202 - no zeros but a single dark at the end of night: correctly moves that to the beginning of the processing with a badcol job before continuing with arcs.

If it is trivial to add `desi_daily_proc_manager` support to this PR before merging, we could do that, otherwise let's open a ticket for that while merging this so that we can use it to unblock himalayas.

akremin:
Thanks for the comments and answer. Implenting the daily processing is straight-forward, but likely requires a bit more logic than the re-processing script. Once the commit removing the `"zero": biasnight` line is in I can merge this and open a ticket for the daily functionality.

Issue #1876, opened 2022-10-20
emlinefit 1+z factors


sbailey:
Double check 1+z factors in `desispec.emlinefit.emlines_gaussfit()` and correct as needed; desihub/fastspecfit#86 says that it is wrong (numerator vs. denominator leading to being off by (1+z)^2).

We need this for the Iron run, and ideally ASAP so that it can be included in the Himalayas redshift runs that we are launching soon (maybe even tomorrow?).

araichoor:
the fix is simple.
however, I m waiting for @moustakas investigation result on that.

moustakas:
Here's an update on my code, so `emlinefit` will need the same fix. This was a really dumb bug on my part, sorry.
  https://github.com/desihub/fastspecfit/issues/86#issuecomment-1286300278

PR #1878, opened 2022-10-21
bugfix: correct (1+z) factor in equivalent width

araichoor:
This PR corrects a bug in `emlinefit.emlines_gaussfit()` for the equivalent width measurement.
It fixes https://github.com/desihub/desispec/issues/1876.

### DIFF
--- a/py/desispec/emlinefit.py
+++ b/py/desispec/emlinefit.py
@@ -251,7 +251,7 @@ def emlines_gaussfit(
                     emdict["FLUX"] = popt[1]
                     emdict["FLUX_IVAR"] = diag[1] ** -1
                     # AR rest-frame equivalent width
-                    factor = (1 + zspec) / emdict["CONT"]
+                    factor = 1 / ((1 + zspec) * emdict["CONT"])
                     emdict["EW"] = emdict["FLUX"] * factor
                     emdict["EW_IVAR"] = emdict["FLUX_IVAR"] / factor ** 2
     # AR fitted waves / fluxes / ivars / models


moustakas (approved):

moustakas:
I double-checked by measuring the EWs of some strong lines in a couple spectra by hand and the fix in this PR (and an equivalent one in FastSpecFit) does indeed address the underlying bug.

moustakas:
@araichoor please update the change log and then you should be able to merge.

araichoor:
thanks for all the quick checks!

just, what is the change log?
do you mean `doc/changes.rst` ?
if so, I think the desispec procedure is to update it when creating a tag.

moustakas:
> just, what is the change log?
> do you mean doc/changes.rst ?
> if so, I think the desispec procedure is to update it when creating a tag.

I'm not aware of a change in procedure. Generally every PR should include an appropriate line in `doc/changes.rst` under the "unreleased" version number so that when we go to make a tag we don't have to hunt back through the various merged PR to figure out which ones came after the previous tag. 

That said, looking back through the last many PRs, it doesn't look like this procedure has been followed. So I guess I'm wrong!

araichoor:
I let @sbailey decide here :)

sbailey:
Yes, every PR should eventually be reflected by an entry in doc/changes.rst.  However, we often have multiple simultaneous PRs going on in desispec and making the updates prior to merging causes unnecessary merge conflicts.  What we should be doing is adding the entry immediately after merging, but in practice it usually piles up and then I fill in a bunch of entries at once just before making a new tag.

I'm about to make one of those big "update doc/changes.rst from all the recent PRs" updates, so I'll take care of this one too.

Issue #1881, opened 2022-10-24
desi_run_night default CPU+GPU support not using GPUs?


sbailey:
desi_run_night CPU vs. GPU job selection doesn't quite work yet, I think:

`desi_run_night --system-name perlmutter-gpu ...` will submit all jobs to GPU nodes, even jobs that can't yet use a GPU, e.g. arcs.  For GPU-enabled jobs, it includes `desi_proc / desi_proc_tilenight --gpuspecter --gpuextract` and `redrock --gpu ...` so that the GPUs are used.  The non-GPU jobs on GPU nodes still work, but it isn't optimal.

`desi_run_night` without specifying a system name submits arcs, ccdcalib, and nightlybias to CPU nodes and other jobs to GPU nodes.   However, it doesn't instruct the GPU jobs to actually use the GPUs and thus it appears that they would still run in CPU-only mode.  I haven't actually tried letting these jobs run to avoid wasting GPU resources, so it is possible that I'm mis-interpreting what would happen if they ran.

Semi-related: even on Perlmutter CPU nodes `desispec.workflow.batch.default_system` returns "perlmutter-gpu".  Since calling `desi_proc` without `--system-name` triggers a call to `default_system`, this results in confusing log messages about guessing perlmutter-gpu even though the job is on a CPU node and couldn't use a GPU even if it wanted to.

Review what behavior is treated as default and update such that jobs submitted to GPU nodes will use GPUs and jobs that can't use a GPU get submitted to a CPU node by default.

While we're at it: in addition to arc, ccdcalib, and nightlybias, other jobdesc that can't use GPUs yet are nightlyflat, psfnight, and badcol.

Examples for testing:
```
# all job scripts have --constraint=gpu (even arcs) and desi_proc jobs have --gpuspecter --gpuextract
desi_run_night -n 20210220 --z-submit-types pernight --use-tilenight --system-name perlmutter-gpu -q regular --dry-run-level 1

# some jobs --constraints=cpu and others --constraint=gpu, but gpu jobs don't have --gpuspecter --gpuextract
# or --system-name so it appears the would run in CPU mode instead of GPU mode
desi_run_night -n 20210221 --z-submit-types pernight --use-tilenight -q regular --dry-run-level 1
```

akremin:
[...]

PR #1883, opened 2022-10-24
add --use-specter to desi_run_night

marcelo-alvarez:
Currently the choice of specter vs gpu_specter when running desi_run_night is dependent on system_name, such that gpu_specter is only used for system_name == 'perlmutter-gpu'. 

This PR makes gpu_specter the default for any value of system_name with desi_run_night. Specter is only used if --use-specter is specified. GPUs will be used with gpu_specter if system_name == 'perlmutter-gpu'.

I have checked that the slurm files are as expected by running desi_run_night --dry-run-level 1 ... on cori and perlmutter.

@sbailey and @akremin please review for style and unintended consequences. 

<s>
We can either (1) merge this PR as is or (2) change the default to gpu_specter by using --use-specter, instead, since eventually gpu_specter will be the default. 
</s>

### DIFF
--- a/py/desispec/workflow/procfuncs.py
+++ b/py/desispec/workflow/procfuncs.py
@@ -239,7 +239,7 @@ def create_and_submit(prow, queue='realtime', reservation=None, dry_run=0, joint
         prow['SCRIPTNAME'] = orig_prow['SCRIPTNAME']
     return prow
 
-def desi_proc_command(prow, queue=None):
+def desi_proc_command(prow, queue=None, system_name, use_specter):
     """
     Wrapper script that takes a processing table row (or dictionary with NIGHT, EXPID, OBSTYPE, JOBDESC, PROCCAMWORD defined)
     and determines the proper command line call to process the data defined by the input row/dict.
@@ -247,6 +247,8 @@ def desi_proc_command(prow, queue=None):
     Args:
         prow, Table.Row or dict. Must include keyword accessible definitions for 'NIGHT', 'EXPID', 'JOBDESC', and 'PROCCAMWORD'.
         queue, str. The name of the NERSC Slurm queue to submit to. Default is None (which leaves it to the desi_proc default).
+        system_name: batch system name, e.g. cori-haswell, cori-knl, perlmutter-gpu
+        use_specter, bool, optional. Default is False. If True, use specter, otherwise use gpu_specter by default.
 
     Returns:
         cmd, str. The proper command to be submitted to desi_proc to process the job defined by the prow values.
@@ -263,6 +265,10 @@ def desi_proc_command(prow, queue=None):
             cmd += ' --noprestdstarfit --nostdstarfit'
     elif prow['JOBDESC'] in ['nightlybias', 'ccdcalib']:
         cmd += ' --nightlybias'
+    elif prow['JOBDESC'] in ['flat'] and not use_specter:
+        cmd += ' --gpuspecter'
+        if system_name=="perlmutter-gpu":
+            cmd += ' --gpuextract'
     pcamw = str(prow['PROCCAMWORD'])
     cmd += f" --cameras={pcamw} -n {prow['NIGHT']}"
     if len(prow['EXPID']) > 0:
@@ -358,7 +364,7 @@ def create_batch_script(prow, queue='realtime', dry_run=0, joint=False, system_n
             if joint:
                 cmd = desi_proc_joint_fit_command(prow, queue=queue)
             else:
-                cmd = desi_proc_command(prow, queue=queue)
+                cmd = desi_proc_command(prow, queue=queue, system_name, use_specter)
         if dry_run > 1:
             scriptpathname = batch_script_name(prow)
             log.info("Output file would have been: {}".format(scriptpathname))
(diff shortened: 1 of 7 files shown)


py/desispec/workflow/procfuncs.py
```diff
@@ -1158,7 +1164,7 @@ def submit_tilenight(ptable, prows, calibjobs, internal_id, queue, reservation,
 def science_joint_fit(ptable, sciences, internal_id, queue='realtime', reservation=None,
                       z_submit_types=None, dry_run=0, strictly_successful=False,
                       check_for_outputs=True, resubmit_partial_complete=True,
-                      system_name=None):
+                      system_name=None,use_gpuspecter=False):
```
akremin, line 1167: This doesn't seem to be passed on to the `joint_fit` function here (this is just a wrapper). You did add the hooks to `joint_fit`, so you just need to add the argument in the function call on line 1178.

akremin (commented):
Thanks Marcelo. I made one comment inline where I think an additional argument propagation is needed. 

Given that we hope to be running with `gpuspecter` quite soon (ideally for Iron), I think it would be best to change this to opt-in to specter and have `gpuspecter` as default. I.e. option (2) with `--use-specter`. Apologies for making you adjust all your changes but I think it is a short timescale before we will be using `gpuspecter` and we generally want the default to be no new arguments.

py/desispec/workflow/procfuncs.py
marcelo-alvarez: Thanks for catching this. As I understand it, jobs processing multiple exposures (other than tilenights) don't involve the extraction step. So does anything called by `checkfor_and_submit_joint_job` even need to know about what kind of specter to use?  If not, I'll just get rid of `use_gpuspecter` in the functions `checkfor_and_submit_joint_job`, `science_joint_fit`, and `joint_fit`.

marcelo-alvarez:
> Thanks Marcelo. I made one comment inline where I think an additional argument propagation is needed.
> 
> Given that we hope to be running with `gpuspecter` quite soon (ideally for Iron), I think it would be best to change this to opt-in to specter and have `gpuspecter` as default. I.e. option (2) with `--use-specter`. Apologies for making you adjust all your changes but I think it is a short timescale before we will be using `gpuspecter` and we generally want the default to be no new arguments.

@akremin No worries. I figured this would be how we end up doing it but wanted to get the ball rolling on the PR with the still-specter-as-default version I'm using for testing gpu_specter end-to-end on cori and perlmutter-cpu, in particular to get the review of passing the use_[gpu]specter boolean from function to calling function out of the way. I will switch `use_gpuspecter` to `use_specter` after I finish the last gpu_specter validation tests, in this PR, probably today or tomorrow.

py/desispec/workflow/procfuncs.py
marcelo-alvarez: @akremin I went ahead and got rid of `use_specter` in the functions for "joint" fits and jobs, since they don't involve extractions. Please check this to make sure we are not missing it anywhere after this change.

marcelo-alvarez:
@sbailey I have checked the latest commit with and without --use-specter with `desi_run_night --dry-run-level 1` on cori-haswell, perlmutter-cpu and perlmutter-gpu, with results as expected. Feel free to test yourself and/or merge.

akremin:
Thank Marcelo. It is hard to parse the git diff here to verify it covers everything, but what I see looks good. I'll check some scripts tomorrow and ideally run an example or two to verify functionality.

sbailey:
In `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/p8r` I tried running
```
desi_run_night -n 20210222 --z-submit-types pernight --use-tilenight -q regular --dry-run-level 1 --system-name perlmutter-gpu &> run/logs/20210222.log
```
it looks like the resulting tilenight jobs are good, with `desi_proc_tilenight ... --gpuspecter --gpuextract`, but the flat jobs run `desi_proc` without requesting `--gpuspecter --gpuextract`.

@marcelo-alvarez please check on that; the `desi_proc` calls for flats should use the same gpuspecter options as for science exposures.

Note: let's keep the scope of this PR to cover the cases when `desi_run_night --system-name` is specified.  After that is working, I'll rebase my branch and submit another PR to handle the cases where `--system-name` is not specified and it is auto-picking based upon job type (issue #1881).

sbailey:
@marcelo-alvarez please also add `desi_daily_proc_manager --use-specter` option so that we can merge this but still continue to use specter for daily processing until the validation runs are completed and documented.

Otherwise the code updates and generated scripts look good, though I'm also doing a paranoia check of also running them for a night.

marcelo-alvarez:
@sbailey flat jobs now run with `--gpuspecter` and `--gpuextract` where expected, and `desi_daily_proc_manager --use-specter` option has been added. We should make sure to test the latter in a realistic way, given that it could affect the daily processing.

akremin:
I ran the `desi_daily_proc_manager` with both `--use-specter` and without, and it did the correct thing for all job types in both circumstances. Eventually we will want to change `--gpu-specter` in the lower level functions to also use (or not use) `--use-specter`, but that can be left for a future PR.

akremin:
Since @marcelo-alvarez and @sbailey have tested the actually running of `specter` vs. `gpu-specter` and the `desi_run_night` implementation in this PR, I will now merge.

PR #1885, opened 2022-10-26
Corrections for sims

julienguy:
This PR adds one option to desi_proc (--no-barycentric-correction) and fixes issues/bugs when running with option '--apply-sky-throughput-correction'.
The latter option was considered only when writing the sframe files when it is also applicable for the standard star fit, the flux calibration code, and when writing the cframe files (any time we subtract the sky). 

Both changes are need to run the current version of the pixel-level simulations. They have no impact on real data processing (we don't use the options).


sbailey:
looks good, thanks.

PR #1886, opened 2022-10-26
Adjust sky line amplitude

julienguy:
The TPCORR model used for sky subtraction and described in PR #1801 is providing a more precise model of the sky continuum for each fiber than in the previous version of the code (fuji/guadalupe). This is because it is based on a model of the variations of throughput instead of a (noisy) fit of the sky line amplitudes. 
But this approach requires a precise fiber flat field and a good knowledge of the position of the fiber for the science exposure (and assumed at the same position for all fiber flat exposures). Also for the sky lines (not the continuum), such a model cannot outperform the actual measurement of the sky lines amplitude.

So in this PR, for protection against flat field error, we apply the empirical sky line amplitude correction to the sky model **on the sky lines only**. 

We keep the fit of the correction to sky line amplitudes (the TPCORR) as is in the current version of the code. The modification is to apply it by default to the sky lines. The only new algorithm is to separate the continuum for the lines in the sky model. An example is shown in the following figure.

skylines

There are now 3 different ways to subtract  the sky: no correction at all (beyond the TPCORR model), empirical correction of sky lines , or empirical correction of both sky lines and continuum.

Examples on simulations where the flat field is wrong by design are shown below.
```
desi_process_exposure -i 00000001/frame-z0-00000001.fits.gz --sky 00000001/sky-z0-00000001.fits.gz -o sframe-z0-00000001-corr-on-sky-lines-only.fits --no-tsnr
desi_process_exposure -i 00000001/frame-z0-00000001.fits.gz --sky 00000001/sky-z0-00000001.fits.gz -o sframe-z0-00000001-fullcorr.fits --no-tsnr --apply-sky-throughput-correction
desi_process_exposure -i 00000001/frame-z0-00000001.fits.gz --sky 00000001/sky-z0-00000001.fits.gz -o sframe-z0-00000001-nocorr.fits --no-tsnr --no-sky-line-throughput-correction

ds9 sframe-z0-00000001-nocorr.fits sframe-z0-00000001-corr-on-sky-lines-only.fits sframe-z0-00000001-fullcorr.fits &
plot_frame -i sframe-z0-00000001-nocorr.fits sframe-z0-00000001-corr-on-sky-lines-only.fits sframe-z0-00000001-fullcorr.fits --fiber 289 --legend
```

Screenshot from 2022-10-26 11-18-01
spectrum
spectrum-zoom

On the last figure one can see on the orange curve that the sky line residuals have disappeared by we keep the same continuum as the uncorrected version (blue curve).

I am now running several tiles to see if there are any issues.


sbailey:
I have gotten lost in the terminology of what the different correction methods mean and what options map onto what corrections, but in the meantime a basic request: please make the default of no options do what you recommend we should do, and have the options only for enabling alternatives for comparison testing.

If you haven't already covered these in the tests that you are running, let's make sure we check:
* a dark night with low sky
* a bright night with lots of sky but likely uniform
* a case where we know the sky is non-uniform
* if we know of an example, a night where the fiberflat is questionable (IIUC, the case where this is supposed to help)

py/desispec/sky.py
```diff
@@ -982,25 +982,63 @@ def subtract_sky(frame, skymodel, apply_throughput_correction = False, zero_ivar
...
+        # sky spectrum
+        if apply_throughput_correction  :
+            for fiber in range(frame.flux.shape[0]) :
+                # apply this correction to the sky model even if we have not fit it (default can be 1 or 0)
+                skymodel.flux[fiber] *= skymodel.throughput_corrections[fiber]
+
+        elif apply_throughput_correction_to_lines :
+
+            # we first identify bright lines on the sky model
+
```
schlafly, line 998: Do you really want to ~discover the line list here, versus using some line list like this one?
https://github.com/desihub/desispec/blob/main/py/desispec/sky.py#L1027-L1040
I guess likely the only difference is for nebular MW emission, though.  I don't think we ever established what the difference was between 'continuum' tpcorr and 'sky line tpcorr', so maybe one could imagine that nebular MW emission would also be better modeled here.

julienguy: Agreed. I was exactly doing this change (fixed list of lines :-) ), after finding that my current code was not robust enough. Nebular MW emission is another story.

py/desispec/sky.py
```diff
@@ -982,25 +982,63 @@ def subtract_sky(frame, skymodel, apply_throughput_correction = False, zero_ivar
...
+            minima  = np.where(minima)[0]
+            if minima.size > 0 :
+                cont    = np.interp(skymodel.wave,skymodel.wave[minima],msky[minima])
+                # second minima
+                minima2 = np.zeros(msky.shape)
+                minima2[1:-1] = (cont[1:-1]<cont[:-2])&(cont[1:-1]<cont[2:])&(cont[1:-1]>-20)&(cont[1:-1]<0.5*np.max(cont))
+                minima2  = np.where(minima2)[0]
+                cont    = np.interp(skymodel.wave,skymodel.wave[minima2],msky[minima2])
+                in_cont = np.where(np.abs(msky-cont)<threshold)[0] # keep all pixels without n sigma of cont for debiasing estimate
+
```
schlafly, line 1016: FWIW, I had trouble following the logic here.  Sometimes minimum_filter is nice for this kind of thing; I might have imagined defining minima as minima = msky == minimum_filter(msky, size=10) (a pixel is a 'minimum pixel' if it's the lowest within 10 pixels).  But what you're doing may well be better---I just am having problems thinking about it!

py/desispec/sky.py
```diff
@@ -982,25 +982,63 @@ def subtract_sky(frame, skymodel, apply_throughput_correction = False, zero_ivar
...
+            if in_cont.size > 0 :
+                # apply this correction to the sky lines only
+                for fiber in range(frame.flux.shape[0]) :
+
+                    # estimate and subtract continuum for this fiber specifically
+
+                    cont = np.interp(skymodel.wave,skymodel.wave[in_cont],skymodel.flux[fiber][in_cont])
+                    skylines = skymodel.flux[fiber] - cont
+                    skylines[skylines<threshold] = 0
```
schlafly, line 1029: I was originally imagining having a sky line template rather than figuring it out on the fly.  One feature there is that one could imagine handling the wings of the line a little more gracefully than setting them to zero.

py/desispec/sky.py
julienguy: this part of the code has been removed

py/desispec/sky.py
julienguy: part of code removed

julienguy:
I will commit changes to the code I am working on after a validation based on several tiles/exposures of night 20221013 with BRIGHT and DARK tiles.

schlafly:
I left a few comments in the code, but maybe bringing one up more broadly here... for cases other than b this is probably a strict improvement.  For the b spectrograph it's less clear to me---since there is really basically only the one line this kind of just nulls it plus any possible signal there, which is a little different from improving the sky subtraction there.  But maybe the noise is just so much higher there that one can do whatever violence one wants to that part of the spectrum without paying much of a price.

julienguy:
Test of several tiles compared to main:

tile=3377 (dark) efftime=981.7 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=3415/3413=1.0006
tile=4974 (dark) efftime=1142.6 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=3404/3406=0.9994
tile=20709 (bright) efftime=213.6 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=3897/3898=0.9997
tile=22259 (bright) efftime=216.2 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=3840/3840=1.0000
tile=22428 (bright) efftime=214.6 frac(|dz|/(1+z)>0.003)=0.0002 nz(new)/nz(old)=4127/4128=0.9998
tile=23331 (bright) efftime=197.7 frac(|dz|/(1+z)>0.003)=0.0003 nz(new)/nz(old)=3933/3937=0.9990
tile=25373 (bright) efftime=204.5 frac(|dz|/(1+z)>0.003)=0.0005 nz(new)/nz(old)=3851/3851=1.0000
tile=25374 (bright) efftime=225.0 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=3948/3949=0.9997
tile=25375 (bright) efftime=189.1 frac(|dz|/(1+z)>0.003)=0.0051 nz(new)/nz(old)=3384/3408=0.9930
tile=25397 (bright) efftime=208.6 frac(|dz|/(1+z)>0.003)=0.0003 nz(new)/nz(old)=3902/3901=1.0003
tile=26165 (bright) efftime=210.5 frac(|dz|/(1+z)>0.003)=0.0008 nz(new)/nz(old)=3843/3849=0.9984

No difference in terms of performance. That is expected.
There are however true differences between the processing. Difference of sframe files (new - main):
sframe-z3-diff

julienguy:
Answering Stephen's comments above: 
- The default option is the wanted behavior in prod: correct amplitude of sky lines (in R and Z but not B) but not the continuum. 
- The test I ran last week had several dark and bright tiles. 
- What I am missing is a tile or exposure where we know the flat is not great.

I also think the changes I made to the code address all of Eddie's points.

schlafly:
I don't have any good guesses for when the flat wasn't great.  It looks to me like we introduced the "park positioners" button on 2021-06-01.  So any date before that should be somewhat worse as we weren't routinely parking at the end of the night?

julienguy:
I tested tile 1224 night=20210531 expid=90516 to test Eddie's suggestion of a possible incorrect flat due to positioners not being parked. Because it is not a recent night, I ran this exposure twice, once with the main code and once with this branch (adjust-sky-line-amplitude).
I do not find a significant difference in the redshift results:
tile=1224 (dark) efftime=1027.2 frac(|dz|/(1+z)>0.003)=0.0006 nz(new)/nz(old)=3531/3525=1.0017
So this PR is really not doing much!

sbailey:
Clarifying one point:

> We keep the fit of the correction to sky line amplitudes (the TPCORR) as is in the current version of the code. The modification is to apply it by default to the sky lines. The only new algorithm is to separate the continuum for the lines in the sky model.

This PR includes an expanded set of sky lines for the correction, and also uses that expanded set for the TPCORR fit, so the values of TPCORR also change, not just how they are used (or not) for correcting the sky model.  The tests indicate that this isn't a big deal, but I wanted that to be clear to avoid confusion about what changed.

Merging now and we'll include it in the Himalayas run of SV data.

PR #1888, opened 2022-10-28
fix relative symlinks to $DESI_ROOT_READONLY files

### LINKED ISSUE #1873: /global/cfs vs. /dvs_ro/cfs and relative links
The spectro pipeline creates relative links for files like `exposures/20220412/00130002/fiberflatexp-r0-00130002.fits.gz` to `calibnight/20220412/fiberflatnight-r0-20220412.fits`.  It uses `os.path.relpath` to convert two absolute paths to the appropriate relative path.

However, if one of those is a readonly `/dvs_ro/cfs/cdirs...` and the other is a writable `/global/cfs/cdirs...` it creates a valid-but-not-intended relative link like 
```
lrwxrwxrwx 1 desi desi     127 Oct 13 15:26 fiberflatexp-r0-00130002.fits.gz -> ../../../../../../../../../../dvs_ro/cfs/cdirs/desi/spectro/redux/himalayas/calibnight/20220412/fiberflatnight-r0-20220412.fits
```

Find all uses of `os.symlink` and update them for this special case, probably by providing a new `desispec.io.util.something` utility function so that we can special-case this logic once and not N>>1 times.

sbailey:
This PR fixes #1873 by adding `desispec.io.util.relsymlink(src, dst)` which creates a relative symlink from `dst` -> `src`, while treating $DESI_ROOT and $DESI_ROOT_READONLY as the same root.  i.e. this symlink in the daily prod:

```
lrwxrwxrwx 1 desi desi 123 Oct 26 21:27 exposures/20221025/00150146/fiberflatexp-r0-00150146.fits.gz -> ../../../../../../../../../../dvs_ro/cfs/cdirs/desi/spectro/redux/daily/calibnight/20221025/fiberflatnight-r0-20221025.fits
```
becomes this link in the test prod `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/symlinks`:
```
lrwxrwxrwx 1 sjbailey desi 60 Oct 27 20:19 exposures/20221025/00150146/fiberflatexp-r0-00150146.fits.gz -> ../../../calibnight/20221025/fiberflatnight-r0-20221025.fits
```

i.e. the symlinks become portable if the directories are copies to non-NERSC sites, and also work on cori login nodes (which don't mount /dvs_ro/cfs/cdirs).

@julienguy this fixes the problem that you tripped across.

### DIFF
--- a/bin/copyprod
+++ b/bin/copyprod
@@ -57,6 +57,7 @@ log.info(f'Linking {num_expids} exposures on {num_nights} nights')
 #- What type of copy or link are we making?
 if opts.fullcopy:
     link = shutil.copy2
+    opts.abspath = True
     log.debug('Performing full copy of selected NIGHT/EXPID')
 else:
     link = os.symlink
--- a/py/desispec/io/util.py
+++ b/py/desispec/io/util.py
@@ -840,3 +840,33 @@ def is_svn_current(dirname):
         log.error(msg)
         raise ValueError(msg)
 
+def relsymlink(src, dst):
+    """
+    Create a relative symlink from dst -> src, while also handling
+    $DESI_ROOT vs $DESI_ROOT_READONLY
+
+    Args:
+        src (str): the pre-existing file to point to
+        dst (str): the symlink file to create
+    """
+    #- Standardize path
+    src = os.path.normpath(os.path.abspath(src))
+    dst = os.path.normpath(os.path.abspath(dst))
+
+    #- handle DESI_ROOT (required) vs. DESI_ROOT_READONLY (optional)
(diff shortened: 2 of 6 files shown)


julienguy:
The code looks good. I will run a small test prod to verify this works correctly.

akremin:
It is outside the scope of this PR but it may be worth going back and re-creating the existing symlinks in `daily`. Nights from 20220923 to the present have these links that are valid but appear broken on cori login nodes.

julienguy:
I tested running and exposure from main and this branch. I confirm that the links to the nightly fiberflat are broken in main and fixed in this branch.
I'm merging this PR.

PR #1891, opened 2022-10-31
bugfix: stdstars and relsymlink

### LINKED ISSUE #1890: relsymlink broke multi-exp stdstar fit
I have a fix that I'm testing, but I'm filing a ticket to minimize confusion in the meantime in case others trip on this:

`relsymlink(src, dst)` from PR #1888 didn't work if `src` was already a relative path.  This broke multi-exposure stdstar fitting which passes in a relative link to the stdstar file when creating the links from other exposures.

I'll open a PR with a fix when my test jobs pass (likely Monday morning, not tonight).  In the meantime, multi-exposure stdstar fitting is broken on main, apologies.

sbailey:
This PR fixes #1890, i.e. a bug introduced with relative symlinks in PR #1888 that broke the symlinks creation for multi-exposure stdstar fitting.  This PR adds two redundant fixes:
1. `relsymlink(src, dst)` now accepts relative paths for `src` instead of requiring two absolute paths and deriving the relative path.  If `src` is a relative path, it is relative to `dirname(dst)`, not relative to the current directory (same behavior as `os.path.symlink`).  I added tests to confirm this behavior.
2. `desi_proc_joint_fit` no longer derives the relative path, leaving it to `relsymlink` instead.  In this case `desi_proc_joint_fit` was deriving the correct path anyway so fix (1) would have been sufficient, but it has the possible future fragility of /dvs_ro/cfs vs. /global/cfs, so I also updated it to let `relsymlink` standardize how relative paths are computed, since `os.path.relpath` isn't sufficient for all cases due to read-only vs. read-write mounts of the same directories.

Extra: I also updated some `log.debug(...)` messages to use `log.debug("format string %s", blat)` instead of `log.debug(f"format string {blat}")` so that it doesn't spend time evaluating the string if it isn't going to log it.  That has mattered in the past where debug statements were in loops and ended up taking a significant fraction of the compute time.  In this case it probably doesn't matter, but since I was updating a log.debug message anyway for the link variable name, I also updated the formatting method and then did the same for several others nearby.

Test run in `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/p8r` night 20201214, in particular tile 80605 exposures 00067710, 00067711, 00067712, 00067713.

### DIFF
--- a/py/desispec/io/util.py
+++ b/py/desispec/io/util.py
@@ -855,19 +855,23 @@ def relsymlink(src, dst, pathonly=False):
     Returns:
         the releative path from dst -> src
     """
-    #- Standardize path
-    src = os.path.normpath(os.path.abspath(src))
-    dst = os.path.normpath(os.path.abspath(dst))
+    #- if src is already a relative path, just use that
+    if src.startswith('..'):
+        relpath = src
+    else:
+        #- Standardize path
+        src = os.path.normpath(os.path.abspath(src))
+        dst = os.path.normpath(os.path.abspath(dst))
 
-    #- handle DESI_ROOT (required) vs. DESI_ROOT_READONLY (optional)
-    if 'DESI_ROOT_READONLY' in os.environ:
-        ro_root = os.path.normpath(os.environ['DESI_ROOT_READONLY'])
-        rw_root = os.path.normpath(os.environ['DESI_ROOT'])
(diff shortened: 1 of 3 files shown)


akremin:
The changes look good thanks. There was a failure in cholesky_solve, unrelated to this PR. I am rerunning the test now then I can merge.

PR #1893, opened 2022-11-01
copyprod --tiles option

sbailey:
This PR adds a `copyprod --tiles ...` option to filter a night by specific tiles.  I find that easier than creating a separate file of NIGHT EXPIDs to include via the `--explist` option (which is more flexible, but requires more steps for most of the cases where I'm making a mini prod for testing).


akremin (approved):
A useful addition to copyprod. The changes look good, thanks.

Issue #1892, opened 2022-10-31
stdstar crash due to photsys=G targets


sbailey:
This stdstar fit crashes:
```
desi_fit_stdstars --delta-color 0.1 \
--frames /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/p8r/exposures/20210222/00077579/frame-[brz]9-00077579.fits.gz \
--skymodels /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/p8r/exposures/20210222/00077579/sky-[brz]9-00077579.fits.gz \
--fiberflats /global/cfs/cdirs/desi/spectro/desi_spectro_calib/trunk/spec/sm3/fiberflatnight-b9-20201214.fits \
  /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/p8r/calibnight/20210222/fiberflatnight-r9-20210222.fits \
  /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/p8r/calibnight/20210222/fiberflatnight-z9-20210222.fits \
--starmodels /global/cfs/cdirs/desi/spectro/templates/basis_templates/v3.2/stdstar_templates_v2.2.fits  \
--outfile $SCRATCH/stdstars-9-00077579.fits.gz
...
Traceback (most recent call last):
  File "/global/common/software/desi/users/sjbailey/desispec2/bin/desi_fit_stdstars", line 21, in <module>
    sys.exit(stdstars.main(args))
  File "/global/common/software/desi/users/sjbailey/desispec2/py/desispec/scripts/stdstars.py", line 543, in main
    r_band = extinction_total_to_selective_ratio(band , photsys) # dimensionless
  File "/global/common/software/desi/perlmutter/desiconda/20220119-2.0.1/code/desiutil/main/py/desiutil/dust.py", line 92, in extinction_total_to_selective_ratio
    return R["{}_{}".format(band.upper(),photsys.upper())]
KeyError: 'R_G'
```

I'm not sure how we didn't trip on this before, but this comes from line 537+ of scripts/stdstars.py:
```
    photometric_systems = np.unique(fibermap['PHOTSYS'])
    if not gaia_std:
        for band in ['G', 'R', 'Z']:
            star_mags[band] = 22.5 - 2.5 * np.log10(fibermap['FLUX_'+band])
            star_unextincted_mags[band] = np.zeros(star_mags[band].shape)
            for photsys in  photometric_systems :
[...]

PR #1894, opened 2022-11-02
fix indexing bug in fibermap when mix of LEGACY and GAIA.

julienguy:
This PR fixes an indexing bug in fibermap when there is a mix of LEGACY and GAIA stars among the standard stars.
The down selection of the LEGACY stars was not considered for the fibermap (only for starfibers  and the frames).
This fixes issue #1892. But I wouldn't mind a second pair of eyes on this! (because we have made several indexing mistakes for the selection of std stars in the past).


sbailey:
Wow, that was a closer near-miss than I thought it would be.  Indeed, another indexing bug.  I just added some more asserts, both where you caught this bug and also later in the code after another round of filtering.

sbailey (approved):

PR #1895, opened 2022-11-03
Sky subtraction bug fix

julienguy:
This PR fixes a bug introduced recently in PR #1886 where the amplitude correction was applied twice when cosmic_nsig>0 .

The changes are :
  - modify and subtract a copy of the skymodel.flux array in subtract_sky
  - do not run twice the sky subtraction when flagging cosmic rays because this is not necessary. this was justified when the throughput corrections were computed during the sky subtraction but it is not the case any more.

I will rerun the tiles of PR #1886 to verify this does not cause new problems.

### DIFF
--- a/py/desispec/sky.py
+++ b/py/desispec/sky.py
@@ -982,15 +982,18 @@ def subtract_sky(frame, skymodel, apply_throughput_correction_to_lines = True, a
         log.error(message)
         raise ValueError(message)
 
+
+    skymodel_flux = skymodel.flux.copy() # always use a copy to avoid overwriting model
+
     if skymodel.throughput_corrections is not None :
         # a multiplicative factor + background around
         # each of the bright sky lines has been fit.
         # here we apply this correction to the emission lines only or to the whole
         # sky spectrum
+
         if apply_throughput_correction  :
-            for fiber in range(frame.flux.shape[0]) :
-                # apply this correction to the sky model even if we have not fit it (default can be 1 or 0)
-                skymodel.flux[fiber] *= skymodel.throughput_corrections[fiber]
+
+            skymodel_flux *= skymodel.throughput_corrections[:,None]
 
         elif apply_throughput_correction_to_lines :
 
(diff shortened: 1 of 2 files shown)


julienguy:
Tested a few tiles (bright and dark) showing no surprise (below comparison with daily):
tile=3377 (dark) efftime=981.7 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=3415/3413=1.0006
tile=4974 (dark) efftime=1142.6 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=3400/3406=0.9982
tile=20709 (bright) efftime=213.6 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=3898/3898=1.0000
tile=22259 (bright) efftime=216.2 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=3840/3840=1.0000
tile=22428 (bright) efftime=214.6 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=4128/4128=1.0000
tile=23331 (bright) efftime=197.7 frac(|dz|/(1+z)>0.003)=0.0000 nz(new)/nz(old)=3935/3937=0.9995

Nearly identifical redshift performance. Slightly improved best fit redrock chi2 (here difference between best fit redrock chi2 with this PR compared to daily):
chi2

py/desispec/sky.py
```diff
@@ -1014,11 +1017,11 @@ def subtract_sky(frame, skymodel, apply_throughput_correction_to_lines = True, a
...
                         skylines[skylines<0]=0
                         # apply correction to the sky lines only
-                        skymodel.flux[fiber] += (skymodel.throughput_corrections[fiber]-1.)*skylines
+                        skymodel_flux[fiber] += (skymodel.throughput_corrections[fiber]-1.)*skylines
                 else :
                     log.warning("Could not determine sky continuum, do not apply throughput correction on sky lines")
 
-    frame.flux -= skymodel.flux
+    frame.flux -= skymodel_flux
     frame.ivar = util.combine_ivar(frame.ivar, skymodel.ivar)
```
sbailey, line 1025: minor (and pre-dating this PR): the skymodel_flux is adjusted for throughput corrections, but the propagated ivar is not.  I think the corrections are small enough that this doesn't matter, but please confirm that you don't care (or add adjustments for ivar too just for mathematical consistency / completeness)

sbailey (approved):
Understood and agreed about needing to make a copy of skymodel.flux.  Good catch.

I think I understand and agree why we don't need to make a copy of the frame anymore and do the two-step sky-subtraction + cosmics mask + re-skysubtract, but I haven't reviewed that as closely.

I put a minor comment inline about ivar propagation, which pre-dates the fix of this PR itself.  I don't think it matters, but please confirm.  Most corrections are small enough to not matter, but I'm not sure if we every have large outliers where we would be making large (likely incorrect) changes to the flux without corresponding changes to the ivar that might be needed to meaningfully reject it downstream.

py/desispec/sky.py
julienguy: yes corrections are at the level of 1% so this is negligible. I decided not to change this to save CPU and memory.

PR #1896, opened 2022-11-03
desi_job_graph makes job dependency and status graph webpage

sbailey:
This PR adds a script that makes a webpage with the job dependency graph and status for a given night, e.g.

image
image

Clicking on a box takes you to the log for that job.  This graph is based upon the processing table and SLURM job return codes; it does not check for missing job outputs like the dashboard and if any steps were run outside of the pipeline scripts that update the processing table, this won't know about them.  Overall the dashboard is better for getting the big picture across many nights and for checking actual file outputs, while I find this view to be easier to identify what steps went wrong when there is a problem in a night, and for distinguishing "crashed" from "in progress".

Currently this is a single script for simplicity.  If we find that we want to call it from other code (e.g. a spin webapp), it could be factored out into a separate module of desispec.

```
usage: desi_job_graph [-h] -n NIGHT [-o OUTPUT] [-s SPECPROD]

optional arguments:
  -n NIGHT, --night NIGHT
                        night to process
  -o OUTPUT, --output OUTPUT
                        output HTML file (default run/jobgraph/jobgraph-NIGHT.html in reduxdir)
  -s SPECPROD, --specprod SPECPROD
                        override $SPECPROD, or full path to override $DESI_SPECTRO_REDUX/$SPECPROD
```


bin/desi_job_graph
```diff
@@ -0,0 +1,147 @@
...
+    e1 = int(expid.split(',')[0])  #- the first expid
+    tileid = row['TILEID']
+    proccamword = row['PROCCAMWORD']
+
+    if qid in jobinfo:
+        state = jobinfo[qid]['STATE'].split()[0]
+    else:
+        state = 'UNKNOWN'
+
+    if jobdesc == 'tilenight':
```
akremin, line 102: I think this if-else will fail for redshift jobs that aren't stored under` scripts/night/` and have a different naming scheme.

bin/desi_job_graph
```diff
@@ -0,0 +1,147 @@
...
+        logfile = f'{reduxurl}/run/scripts/night/{args.night}/{jobdesc}-{args.night}-{tileid}-{qid}.log'
+    else:
+        logfile = f'{reduxurl}/run/scripts/night/{args.night}/{jobdesc}-{args.night}-{e1:08d}-{proccamword}-{qid}.log'
+
+    if jobdesc in ('tilenight', 'pernight', 'perexp', 'cumulative'):
```
akremin, line 107: For perexp perhaps we also want the description to be expid instead of tileid?

bin/desi_job_graph
```diff
@@ -0,0 +1,147 @@
...
+        help='output HTML file (default run/jobgraph/jobgraph-NIGHT.html in reduxdir)')
+p.add_argument('-s', '--specprod', type=str, required=False,
+        help=('override $SPECPROD, or full path '
+              'to override $DESI_SPECTRO_REDUX/$SPECPROD'))
+
+args = p.parse_args()
+
+#- Default output file into specprod directory
+if args.specprod is not None and os.path.isdir(args.specprod):
+    reduxdir = args.specprod
```
akremin, line 31: Could we title this `specproddir` or `specdir` if it includes the `specprod`?

akremin (commented):
This is a handy script for visualizing a night of processing. I am already imagining additional features but I will hold back on that now. I have some comments in-line that would be nice to have addressed before merging. Then we can build up the feature set with future PR's.

bin/desi_job_graph
sbailey: indeed, good catch; I updated this using `desispec.scripts.tile_redshifts.get_tile_redshift_script_pathname` (plus slurm -> log, adding qid...).

bin/desi_job_graph
sbailey: added

bin/desi_job_graph
sbailey: done

sbailey:
Thanks for the good comments; I believe all are addressed, and a slipped in one more feature of adding a link to the specproddir instead of just the test of the specproddir.

akremin (approved):
All comments have been resolved. The script looks good and my test run succeeded. I'll approve and merge this now.

Issue #1897, opened 2022-11-04
CUDA_ERROR_OUT_OF_MEMORY on tilenight stdstar step


sbailey:
desi_proc_tilenight runs out of GPU memory during the desi_fit_stdstars step.  This especially occurs when it also ran GPU extractions earlier in the same script invocation, indicating the cupy isn't releasing memory from the extraction step.   If you re-run desi_proc_tilenight a second time so that it skips over the already completed extractions, it succeeds on some petals but still runs out of memory on others (See note 2 below).

Using only 40 MPI ranks instead of 64 on a second desi_proc_tilenight invocation avoids the problem, but it still runs out of memory using 40 ranks if starting from the beginning with extractions.

@dmargala could you take a look at this and advise (or even fix?).  @marcelo-alvarez @craigwarner-ufastro heads up too.

### Steps to reproduce

```
source /global/cfs/cdirs/desi/software/desi_environment.sh main

# cd somewhere, then...
export DESI_SPECTRO_REDUX=$(pwd)
export SPECPROD=gpu_mem
copyprod $DESI_ROOT/spectro/redux/daily $SPECPROD -n 20211220 -t 22111

cd $SPECPROD
rm exposures/20211220/00114851/*.*

salloc -N 1 -C gpu -A desi_g --gpus-per-node=4 -t 01:00:00 -q interactive

# runs out of memory, hangs; a second invocation will get further but still hang
time srun -N 1 -n 64 -c 2 --cpu-bind=cores desi_mps_wrapper \
    desi_proc_tilenight -n 20211220 -t 22111 --gpuspecter --gpuextract --mpi --mpistdstars &> tilenight-1.log
```

### Note 1

See https://docs.cupy.dev/en/stable/user_guide/memory.html for discussion about cupy memory management.  `cupy.get_default_memory_pool().free_all_blocks()` can force it to release memory, but it's unclear to me whether every MPI rank should do that, or one per GPU, or something else.  Something along this line would probably be helpful between major steps of the pipeline, e.g. to ensure that all extraction GPU memory is freed before moving on.

### Note 2

[...]

PR #1899, opened 2022-11-07
GPU memory and rank allocation

sbailey:
This PR fixes the GPU memory and rank:GPU allocation problems reported in issue #1897, using the methods suggested by @dmargala in that issue.  This PR provides a new `desispec.gpu` module with helper functions
* `is_gpu_available()`: standardizing if a GPU+cupy+numba.cuda is available (will be used more in a future PR that will remove the other places we do the same sorts of calculations)
* `free_gpu_memory()`: free unused GPU memory if possible, no-op if we're not using GPUs
* `redistribute_gpu_ranks(comm)`: redistribute MPI rank:GPU mapping, handling case of `comm=None` and the case of no GPUs.

Testing was bogged down by:
* typos resulting in NameError and ValueError exceptions that were getting swallowed by stdout buffering before the job was killed
* a surprising feature/bug where `sys.stdout.flush` immediately after reassigning the GPUs resulted in an error `srun: error: eio_handle_mainloop: Abandoning IO 60 secs after job shutdown initiated`, but flushing before reassigning did not cause this.
  * This remains surprising enough that I'm not entirely convinced that that was the actual relevant feature, vs. just tweaking the timing of some other race condition that I wasn't understanding while trying to debug where my stdout messages were going.  This current branch doesn't have the debugging `sys.stdout.flush()` commands, and hasn't failed in multiple test runs.  I'm mainly documenting this for the record in case the error pops up again with larger test runs.

### DIFF
--- a/py/desispec/scripts/proc_tilenight.py
+++ b/py/desispec/scripts/proc_tilenight.py
@@ -183,7 +183,6 @@ def main(args=None, comm=None):
         continue_tilenight=False
 
     #- Cleanup GPU memory and rank assignments before continuing
-    desispec.gpu.free_gpu_memory()
     desispec.gpu.redistribute_gpu_ranks(comm)
 
     if continue_tilenight:
@@ -218,7 +217,6 @@ def main(args=None, comm=None):
             continue_tilenight=False
 
     #- Cleanup GPU memory and rank assignments before continuing
-    desispec.gpu.free_gpu_memory()
     desispec.gpu.redistribute_gpu_ranks(comm)
 
     if continue_tilenight:
--- a/py/desispec/scripts/proc_tilenight.py
+++ b/py/desispec/scripts/proc_tilenight.py
@@ -20,6 +20,8 @@ import desispec.scripts.proc_joint_fit as proc_joint_fit
 from desispec.workflow.desi_proc_funcs import assign_mpi, get_desi_proc_tilenight_parser
 from desispec.workflow.desi_proc_funcs import update_args_with_headers, create_desi_proc_tilenight_batch_script
 
+import desispec.gpu
+
 #########################################
 ######## Begin Body of the Code #########
 #########################################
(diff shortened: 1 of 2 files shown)


sbailey:
Recently discovered: the logs from this PR test run have messages like
```
[6] This process is not using the same device it did during gtl_init.
 Current device_id 2 original device_id 0.
 Use of IPC is disabled from this point. IPC PUT protocol may be used instead 
```
This does not seem to be causing any tangible problems, but also doesn't seem good.

py/desispec/scripts/proc_tilenight.py
```diff
@@ -180,6 +182,10 @@ def main(args=None, comm=None):
             log.info(f'prestdstar failed with {error_count_pre} errors, skipping rest of tilenight steps')
         continue_tilenight=False
 
+    #- Cleanup GPU memory and rank assignments before continuing
+    desispec.gpu.free_gpu_memory()
+    desispec.gpu.redistribute_gpu_ranks(comm)
```
dmargala, line 187: There is a risk to clearing the wrong device memory pool if these functions are called in opposite order. Perhaps it would be worth adding an opt-out option like `desispec.gpu.redistribute_gpu_ranks(..., free_gpu_memory=True)` to call to `desispec.gpu.free_gpu_memory()` before switching to a new device ?

dmargala (commented):
This looks good overall, just one major issue with the contiguous rank/device mapping method.

I'm also a bit puzzled by the `sys.stdout.flush` issue and the GTL messages. It doesn't seem like they are blocking for now but I'll see if I can figure out what is going on there.

sbailey:
@dmargala thanks for catching the method='contiguous' bug.  I have fixed that, including handling the multi-node case, albeit with some assumptions about MPI rank distribution documented in the docstring.  Please double check me.

Good point about freeing memory before re-allocating ranks to GPUs.  I couldn't think of a reason why we would want to re-allocate rank:gpu without freeing the unused memory, so I added a call to `free_gpu_memory` inside `redistribute_gpu_ranks`, without an option to toggle that behavior.  If we find we need to reallocate ranks without freeing memory, we can add an option for that in the future.  Or if you know of a reason why we need that now, let me know and I can add it.  In the meantime, the default does the right thing (free memory before re-distributing) and avoids options that we may never use.

dmargala:
@sbailey the method='contiguous' logic looks good now. 

The main reason to not free the memory pool would be performance considerations depending on the situation. I think it's fine to leave as is until a performance study reveals an opportunity there.

Issue #1900, opened 2022-11-07
improve fiberflat_vs_humidity I/O efficiency


sbailey:
desi_compute_fiberflat_vs_humidity appears to be more sensitive to NERSC I/O performance than other steps, sometimes leading to job timeouts.  Some of this is worse than it used to be due to the gzipped frame files now being inefficient to read headers.  From a quick look, some ideas to improve this:

* in scripts/proc.py, the gzipped frame file header is read but apparently not used in the "desi_compute_fiberflat_vs_humidity" block, and then re-read later in the sky subtraction block.
* in scripts.humidity_corrected_fiberflat.py, the header is read again separately, even though it is also read as part of `read_frame` (it is the `frame.meta`)
* I don't think fiberflat_vs_humidity uses the resolution matrix; if that's the case the I/O could also be faster using `read_frame(filename, skip_resolution=True)`

I'm flagging this for Iron since fluctuations in the timing of this step in test runs led to job timeouts (we'll also increase the job time to give more robustness).

Issue #1799, opened 2022-06-13
standardize GPU options


sbailey:
We're accreting a variety of methods to trigger GPU usage, e.g. `desi_fit_stdstars --use-gpu`, `desi_proc --gpuextract` (while apparently missing an option to enable stdstar GPU usage from desi_proc), and `desispec.fluxcalibration.use_gpu` auto-deciding.  I suggest that we standardize GPU options and usage as:

* if a GPU is available, it is used by default without having to specify options
* if a GPU is not available, the code uses CPU by default without having to specify options
* a `--no-gpu` option would opt-out of using the GPU even if available (mainly for testing).  On a CPU node, `--no-gpu` is not required, and if specified it is just a no-op.
* For high-level pipeline wrappers (e.g. desi_run_night), the `--system-name` (e.g. "perlmutter-gpu, perlmutter-cpu") triggers whether the jobs are resourced and submitted to GPU or non-GPU nodes without requiring additional GPU-specific options from the user.
* For high-level pipeline wrappers, I don't think we need to add per-step options (e.g. extract with GPUs but don't use them for stdstars).  We can add that level of granularity later if we find we really need it.
* Low-level algorithmic steps (extraction, stdstars, fluxcalibration) should provide `--no-gpu` hooks too to be able to turn them off when running that single step (and that is what the high-level wrappers would pass forward from their `--no-gpu` option).

i.e. I'm suggesting that the default should be to automatically use a GPU if available but also work on CPU nodes without requiring options either way, while still providing a way to turn off GPU use if needed for debugging.

Let's take the auto-decide logic from `desispec.fluxcalibration.use_gpu` and wrap that in a utility function for all steps to use.

Note: specter and redrock may be a special case since they are intended to be DESI-agnostic toolkits that shouldn't depend upon DESI code, so the automatic GPU usage selection might get replicated there, or they may have a different explicit opt-in/opt-out.  TBD.

marcelo-alvarez:
Most if not all of these should be resolved by #1901.

PR #1901, opened 2022-11-08
GPU vs. CPU production running

### LINKED ISSUE #1881: desi_run_night default CPU+GPU support not using GPUs?

sbailey:
This PR contains a number of fixes in preparation for running productions on Perlmutter:
* Standardize GPU/CPU defaults following suggestions in issue #1799.  Basically GPU usage is now opt-out: if the code can find a GPU it will use it unless `--no-gpu` (single script) or by setting $DESI_NO_GPU (global opt-out).  This replaces `--use-gpu` (stdstars) and `--gpuextract` (extractions).
* gpu_specter is default everywhere, instead of `--use-specter` at some levels and `--gpuspecter` at others.
* `--system-name` is used by pipeline wrappers to know what kinds of batch jobs to generate and where to send them, but once the code wakes up it uses whatever resources it finds
* On Perlmutter, not specifying `--system-name` automatically picks between CPU and GPU for each job type, and for the jobs that are sent to GPU nodes they correctly use GPUs (fixes #1881)
* From PR #1899 which was spun off of this work, `desispec.gpu.is_gpu_available()` standardizes the logic for identifying if a GPU is available and should be used.

Some other changes that came along for the ride:
* `desi_run_night` defaults to regular queue instead of realtime, since it's primary usage is production runs using the regular queue.
* `--use-specter` extractions default to `--nsubbundles=5` instead of 6 for consistency with the gpu_specter default (which can't support 6).
* partial implementation of fiberflat_vs_humidity I/O improvements in #1900 
* tilenight jobs are giving a little more timing overhead to be more robust to NERSC I/O performance fluctuations
* --dry-run sleeps are shorter for faster iterations (but less stdout movie-like)

Test cases that I checked:
  * `desi_run_night ...` on Perlmutter without specifying `--system-name`: jobs are correctly sent to CPU vs. GPU nodes, and the GPU jobs actually use the GPUs
  * `desi_run_night --system-name perlmutter-cpu ...`: jobs are sent only to CPU nodes, and those jobs don't trip over the lack of GPUs
  * `desi_run_night --system-name cori-knl ...` : this PR doesn't break production running on KNL (though I hope we never need to use that again)
  * `desi_run_night --use-specter`: correctly uses specter instead of gpu_specter

TODO (doesn't work yet):
  * `desi_daily_proc_manager --dry-run-level 1 --use-specter ...`: daily proc on haswell with specter generates the correct jobs scripts (didn't try fully running).

Test productions using various iterations of this branch are in `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/cpugpu-*` (various test cases listed above) and `h1` (7 full nights from sv1 and sv3).

@akremin and @marcelo-alvarez please take a look; also heads up @dmargala .

### DIFF
--- a/py/desispec/scripts/proc.py
+++ b/py/desispec/scripts/proc.py
@@ -64,6 +64,7 @@ import desispec.scripts.fluxcalibration
 import desispec.scripts.procexp
 import desispec.scripts.nightly_bias
 from desispec.maskbits import ccdmask
+from desispec.gpu import is_gpu_available
 
 from desitarget.targetmask import desi_mask
 
@@ -169,6 +170,16 @@ def main(args=None, comm=None):
 
     timer.start('mpi_connect', starttime=start_mpi_connect)
     timer.stop('mpi_connect', stoptime=stop_mpi_connect)
+
+    #- Use GPUs?
+    if is_gpu_available():
+        if args.no_gpu:
+            log.warning("GPUs are available but not using them due to --no-gpu")
+            use_gpu = False
+        else:
+            use_gpu = True
+    else:
+        use_gpu = False
 
     #- Freeze IERS after parsing args so that it doesn't bother if only --help
     timer.start('freeze_iers')
@@ -759,14 +770,14 @@ def main(args=None, comm=None):
                 cmd += ' -o {}'.format(framefile)
                 cmd += ' --psferr 0.01'
 
-                if args.gpuspecter:
-                    cmd += ' --gpu-specter'
+                if args.use_specter:
+                    cmd += ' --use-specter'
                     #- default for CPU is nsubbundles=6 but gpu_specter only allows 1, 5, or 25
                     cmd += ' --nsubbundles 5'
                     cmd += ' --mpi'
 
-                if args.gpuextract:
-                    cmd += ' --use-gpu'
+                if not use_gpu:
+                    cmd += ' --no-gpu'
 
                 if args.obstype == 'SCIENCE' or args.obstype == 'SKY' :
                     if not args.no_barycentric_correction :
@@ -797,7 +808,7 @@ def main(args=None, comm=None):
             inputs = comm.bcast(inputs, root=0)
             outputs = comm.bcast(outputs, root=0)
 
-            if args.gpuextract:
+            if use_gpu and (not args.use_specter):
                 import cupy as cp
                 ngpus = cp.cuda.runtime.getDeviceCount()
                 if rank == 0 and len(cmds)>0:
@@ -806,21 +817,21 @@ def main(args=None, comm=None):
             #- Set extraction subcomm group size
             extract_subcomm_size = args.extract_subcomm_size
             if extract_subcomm_size is None:
-                if args.gpuextract:
-                    #- GPU extraction with gpu_specter uses
-                    #- 5 ranks per GPU plus 2 for IO.
-                    extract_subcomm_size = 2 + 5 * ngpus
-                elif args.gpuspecter:
-                    #- CPU extraction with gpu_specter uses
-                    #- 16 ranks.
-                    extract_subcomm_size = 16
-                else:
+                if args.use_specter:
                     #- CPU extraction with specter uses
                     #- 20 ranks.
                     extract_subcomm_size = 20
+                elif use_gpu:
+                    #- GPU extraction with gpu_specter uses
+                    #- 5 ranks per GPU plus 2 for IO.
+                    extract_subcomm_size = 2 + 5 * ngpus
+                else:
+                    #- CPU extraction with gpu_specter uses
+                    #- 16 ranks.
+                    extract_subcomm_size = 16
 
             #- Create list of ranks that will perform extraction
-            if args.gpuextract:
+            if use_gpu:
                 #- GPU extraction uses only one extraction group
                 extract_group      = 0
                 num_extract_groups = 1
@@ -831,7 +842,7 @@ def main(args=None, comm=None):
             extract_ranks = list(range(num_extract_groups*extract_subcomm_size))
 
             #- Create subcomm groups
-            if args.gpuextract and len(cmds)>0:
+            if use_gpu and len(cmds)>0:
                 if rank in extract_ranks:
                     #- GPU extraction
                     extract_incl = comm.group.Incl(extract_ranks)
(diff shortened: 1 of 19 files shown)


sbailey:
I've flagged this as WIP until I can sort out the tests (gpu_specter is more restrictive about bundlesize, nsubbundles, and nspec) and fix the `desi_proc_daily_manager --use-specter` case, but I think this can be usefully reviewed in the meantime anyway.

py/desispec/workflow/desi_proc_funcs.py
```diff
@@ -82,10 +82,10 @@ def get_shared_desi_proc_parser():
     parser.add_argument("--starttime", type=str, help='start time; use "--starttime `date +%%s`"')
     parser.add_argument("--timingfile", type=str, help='save runtime info to this json file; augment if pre-existing')
     parser.add_argument("--no-xtalk", action="store_true", help='disable fiber crosstalk correction')
-    parser.add_argument("--system-name", type=str, default=batch.default_system(), help='Batch system name (cori-haswell, perlmutter-gpu, ...)')
+    parser.add_argument("--system-name", type=str, help='Batch system name (cori-haswell, perlmutter-gpu, ...)')
```
akremin, line 85: I'm happy to remove the definition of the system name here, but should we explicitly set `default=None`?

py/desispec/workflow/procfuncs.py
```diff
@@ -239,7 +239,7 @@ def create_and_submit(prow, queue='realtime', reservation=None, dry_run=0, joint
         prow['SCRIPTNAME'] = orig_prow['SCRIPTNAME']
     return prow
 
-def desi_proc_command(prow, system_name, use_specter, queue=None):
+def desi_proc_command(prow, system_name, use_specter=False, queue=None):
```
akremin, line 242: Do we want/need to propagate `no_gpu` here?

py/desispec/workflow/procfuncs.py
```diff
@@ -481,7 +480,7 @@ def submit_batch_script(prow, dry_run=0, reservation=None, strictly_successful=F
     if dry_run:
         ## in dry_run, mock Slurm ID's are generated using CPU seconds. Wait one second so we have unique ID's
         current_qid = int(time.time() - 1.6e9)
-        time.sleep(1)
+        time.sleep(0.1)
```
akremin, line 483: As the comment mentions, this needs to wait one second because the mock qid's are tied to a clock (convenient way to get monotonically increasing and unique values). 

The other change in workflow/utils is welcomed to speed up the tests, but this particular line is a limitation with the current means of simulating submissions and needs to remain.

akremin (commented):
This is a welcome cleanup that will simplify things and make them more suitable for the new normal of perlmutter+gpus+gpu_specter. I have three inline questions I would like to get responses about and one trivial fix to ensure that the `dry-run` testing infrastructure continues to work properly.

I have not yet tested the code. I will let Marcelo help with that as part of a larger perlmutter test. I will also do a few more sanity checks on output scripts before approving.

py/desispec/scripts/proc.py
```diff
@@ -759,14 +770,14 @@ def main(args=None, comm=None):
                 cmd += ' -o {}'.format(framefile)
                 cmd += ' --psferr 0.01'
 
-                if args.gpuspecter:
-                    cmd += ' --gpu-specter'
+                if args.use_specter:
+                    cmd += ' --use-specter'
                     #- default for CPU is nsubbundles=6 but gpu_specter only allows 1, 5, or 25
                     cmd += ' --nsubbundles 5'
```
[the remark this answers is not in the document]
sbailey: Good point.  Since desi_extract_spectra now has a default of --nsubbundles=5 for both specter and gpu_specter, we don't have to explicitly set it here in desi_proc for either case.  I'll remove this.

Correcting previous misstatements I've made: at the library level, both `gpu_specter.core.extract_frame` and `specter.extract.ex2d` use nsubbundles=1 as default (i.e. no subbundles), not 5 and 6 as I had previously thought.  So to use subbundles we do have to set it somewhere, and the desi_extract_spectra options is that place (not desi_proc).

py/desispec/workflow/desi_proc_funcs.py
sbailey: default=None is the default for any argparse argument regardless of type, so we don't have to explicitly specify that.  There are multiple other places where we also rely on the argparse default being None without explicitly specifying it, and in general I prefer to not specify a parameter unless we are actually changing what it would otherwise be, or concerned that that default might change in the future outside of our control.

py/desispec/workflow/procfuncs.py
sbailey: good catch.  I clearly made the update without reading the comment two lines above.  :)  I'll revert.

py/desispec/workflow/procfuncs.py
sbailey: Documenting the reasoning for not including no_gpu here:  desi_proc_command is only used by higher level pipeline wrappers (create_batch_script called by desi_run_night and desi_daily_proc_manager) and those higher level scripts don't provide `--no-gpu` either.  They either auto-derive which jobs should be sent to GPU vs. CPU nodes, or use `--system-name` to send all jobs to one or the other.  `--no-gpu` is provided to lower level scripts like desi_proc just in case, but even then it is questionable why you would want to be on a GPU node but not use the GPU, vs. just doing the same testing on a CPU node.

If you really really need to be on a GPU node but not use the GPUs, you can also set $DESI_NO_GPU for a global opt-out, and if that becomes a common use-case then we can add a `--no-gpu` option at a higher level in the pipeline options, but in the meantime I haven't added and option that I don't think we'll use.

akremin (approved):
I'll go ahead and approve this since my comments were satisfied, but a production-style test and subsequent checking of scripts and logs would be useful before merging.

sbailey:
I have fixed the --use-specter case, updated the unit tests, and verified that `desi_daily_proc_manager` generates the correct scripts and I spot checks submitting some of them to run.  When Perlmutter is back, I'll do one more end-to-end test with the latest branch before merging.

sbailey:
I ran another end-to-end test successfully.  Two more changes for the record:
  * added py/desispec/test/test_gpu.py unit tests which I had previously forgotten to commit
  * updated tilenight to be consistent with non-tilenight and only process LASTSTEP=all by default instead of also trying to process LASTSTEP=skysub.  `desi_run_night --laststeps ...` can still be used to override that.

PR #1904, opened 2022-11-11
Add a python script for spectral grouping and redshifting

akremin:
### Overview

This implements a new python script, `desi_zproc`, to process tile based redshifts. Previously such jobs were performed by first generating a bash script with commands that called individual python scripts to do the spectral grouping, coadding, redshift fitting, `mtl` file generation, and running of the afterburners. At present the three afterburners are `quasarnet`, `qsomgii`, and `emlinefit`. All the steps above are now performed as function calls (via `runcmd`) within `desi_zproc`. The script is modeled after `desi_proc` and has similar features including being able to call it from the command line within an interactive node or calling it with `--batch` to have it write and submit a batch script of itself to run on compute nodes.

`desi_zproc` is mpi-enabled and supports the use of gpus. The implementation at present is as follows:

* The underlying spectral grouping and coadding codes are not mpi-enabled, so it uses 1 rank for each spectrograph. All spectrographs are run in parallel.
* After spectral grouping, all ranks are passed to `redrock` for each spectrograph one-at-a-time, since `redrock` is mpi-enabled and capable of using all 4 gpus on perlmutter. Spectrographs are running serially (but quickly because each `redrock` call has up to 64 ranks depending on the system).
* `Tileqa` is then run with rank 0 as it is not mpi enabled.
* Next, `zmtl` is run with one rank per spectrograph as it is not mpi enabled. All spectrographs are run in parallel.
* Finally, the three afterburners are run with 1 rank each, but all 30 are run in parallel. 

For better log parsing and to maintain the same logging scheme as the previous bash script, each function call has its outputs redirected to a log file for that step and petal using `desispec.parallel.stdouterr_redirected`, while the log messages for `desi_zproc` itself only pertain to startup, submission and results of the functions, and shutdown.

## Additional changes in this PR:

* The `qsoqn`, `qsomgii`, `emlinefit`, and `zmtl` files had to be refactoring into functions and standardized. The scripts now call functional versions of the original codes. The codes now reside under the `/py/desispec/scripts` subdirectory.
* `quasarnet`, which calls `redrock`, was using multiprocessing in `redrock` by default since it didn't pass in an mpi communicator. That broke on perlmutter when called from the mpi-enabled `desi_zproc`, so it was updated to pass in a subcommunicator of only a single rank to prevent redrock from trying to use multiprocessing.
* The helper function `findfile` was updated to include a new argument, `logfile`, which is a boolean. When set for a specific set of filetypes it will return where the log for that file type will be rather than the files themselves. This is only implemented for the `desi_zproc` filetypes as the desi_proc commands are not currently redirected to separate logs. If a type is requested that isn't implemented a NotYetImplementedError will be raised.
* Added `qso_mgii_*`, `qso_qn_*`, and `emline_*` were added to `findfile`, where "*" is either "hp", "tile", or "single" for `healpix`, `cumulative`, `pernight`, and `perexp` redshift jobs.
* In the process of adding new file types, an optimization was made of the mapping of potential redshift related filenames to the actual names for a given set of inputs. I modified the code so that the code will find any filetypes that end with "tile" or "hp" or "single" and output the correct name for all such keys depending on whether `cumulative`, `pernight`, `perexp`, or `healpix` was requested. Previously the keys were hard-coded, which temporarily caused my new filetypes not to be found and led to the new implementation.
* I added a new helper utilities to take `CAMWORD` and `BADCAMWORD` and return the `GOODCAMWORD`. I also added unit tests for the function.
* `desispec.parallel.stdouterr_redirected` had a bug when redirecting to `/dev/null` with mpi. That is now fixed.
* `desi_tile_redshifts` was renamed to `desi_tile_redshifts_bash` to make room for a new version of `desi_tile_redshifts` that now uses `desi_zproc`. 
* Some functions in `desi_tile_redshifts` were updated or altered to be used with the new version of the code, but were backward compatible enough that `desi_tile_redshifts_bash` is able to call them. Such functions are imported by `desi_tile_redshifts_bash` to reduce redundancies in the code.
* `desi_daily_proc_manager` and `desi_run_night` were updated to generate scripts calling `desi_zproc` instead of the bash script generation. This has the downside that on cori scripts will be submitted with 5 nodes and use the 32x5 ranks very inefficiently. We could change this setting to run with only 1 node on cori as well. This will take much longer to run than the previous 10 node job, but 10 jobs can run at the same time, and the small jobs can better pack into the 12 node allocation.

### Caveats
 Things to keep in mind about this PR:
* This PR will impact daily operations on cori, as it is almost assuredly inefficient on cori when submitting with 5 nodes. 32x5 ranks can't be effectively utilized in the current implementation. The number of cori nodes should be changed to 1, but with that change the 1 node jobs it will take much longer to run than the previous 10 node jobs (though 10 can run at the same time, and the jobs better pack into the 12 node allocation).
* I have not tested this on cori with GPUs enabled. It likely wouldn't work as the GPU logic is currently setup for perlmutter.
* There is still a failing docstring that I didn't have time to track down.

### Things we should consider doing before merging could be done 

* Changing the default number of nodes for cori redshift jobs to 1 node.
* More exhaustive testing of the `cumulative`, `pernight`, and `perexp` logic for different sets of `thrunight`, `nights`, `expids`, and `tileids` could be done. This would ideally be expanded even further to different systems and with both `no_gpu=True` and `no_gpu=False`. I have done some testing but would have preferred to do more.
* Timing tests with different numbers of ranks on perlmutter and cori (and different numbers of nodes on cori). 
* Fold this code in with the parallel PR standardizing gpu and system name calls. As of now I think this would request and try to use GPUs even on a system without them unless the user disables it with `--no-gpu`. It would be better to have a system check and checks for environment variables to properly determine if gpus can be used.

### Future Improvements that could be made:
* This doesn't implement healpix based redshifts. Eventually that should be done.
* Logic could be added to identify when multiple nodes are available and run one redrock job per node in parallel with subcommunicators. This would improve performance on multinode jobs but still has inefficiencies because most ranks will be idle in the non-mpi steps (i.e. all steps other than redrock).
* Work could be done to mpi-enable the spectral grouping and afterburners, but these are relatively fast already so it would have to be determined on a code by code basis whether tangible speedups are possible.
* More complex logic could be used to send more than one rank to each `quasarnet` instance with a subcommunicator such that the rerunning of `redrock` within `quasarnet` can use multiple ranks. On perlmutter there are 64 ranks, so after assigning 10 ranks to `qsomgii` and 10 to `emlinefit` there are 44 remaining that could be devoted to `quasarnet`. With each new afterburner this optimization would become less useful, however, as fewer idle ranks would be available.
* The `findfile` addition for log files could be spun off into a distinct function that calls `findfile` and then modifies the result. This is how it is implemented in the current code, except within `findfile` with a boolean flag to alter the output. A similar function could be created for getting script names too.


sbailey (commented):
This PR is a huge improvement over the current bash redshift jobs.  In addition to improved flexibility, maintainability, and runcmd input/output checking, this new structure is ~7x faster (!) per-node-hour by removing the multiple srun and python import overheads and better use of single-node parallelism.

Prior to this PR I tested it as a standalone script for perexp, pernight, and cumulative redshifts; when Perlmutter is back I'll test the script generation and interaction with the full pipeline.  I don't have any substantive comments on the code, though I'll document a few smaller changes I'll make:
  * promote `_log_timer` to non-hidden `log_timer` since it is now used by multiple modules
  * add hooks to backup logfiles to blat.log.N instead of overwriting them on every rerun

sbailey:
The pipelining aspects of this work well.  Comparing to main, the spectra files are the same, but something is different about the coadds, leading to differences downstream in the redshifts.  I think this predates this PR, but before merging and putting into production, track down the difference between:

main: `desi_group_spectra` (cframes -> spectra) + `desi_coadd_spectra` (spectra -> coadd)
this PR: `desi_group_spectra --outfile ... --coaddfile ...` (cframes -> spectra and coadd in a single call)

Those are supposed to produce the same coadd.

sbailey:
I have tested this for a number of cases including
* running `desi_zproc` individually
* running `desi_tile_redshifts` as a wrapper for launching multiple `desi_zproc` jobs
* running `desi_run_night` to submit a workflow that includes `desi_zproc`
* running `desi_daily_proc_manager` to submit jobs including `desi_zproc`

I'm sure there is more testing that could be done, but at this point managing N>1 PRs and the difference between main and this branch is becoming a barrier to further testing, so I'm going to merge now and we can continue with any additional features in smaller more targeted PRs.

The primary thing I have not tested is running this on Cori, where 1 node is probably not sufficient to finish in a reasonable amount of time.  Daily processing on Cori is currently protected by running out of the "daily" branch, so merging this won't break daily ops on Cori.

PR #1905, opened 2022-11-11
Cosmic ray rejection parameter in yaml file

julienguy:
In this PR the preprocessing code looks for the keyword COSMICS_C2 in the calibration yaml file config if the argument is not set.
This is needed to process the new MOSAIC CCD (500 micron-tick device).

### DIFF
--- a/bin/desi_compute_pixel_flatfield
+++ b/bin/desi_compute_pixel_flatfield
@@ -68,7 +68,7 @@ def clipped_mean_image(image_filenames) :
     smimage2=np.sum(mimage**2)
     for i in range(len(images)) :
         a=np.sum(images[i]*mimage)/smimage2
-        log.debug("scale %d = %f"%(i,a))
+        log.debug("scale %d = %f", i,a)
         if a<=0 :
             raise ValueError("scale = %f for image %s"%(a,image_filenames[i]))
         images[i] /= a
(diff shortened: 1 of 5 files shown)


bin/desi_compute_pixel_flatfield
```diff
@@ -236,21 +236,21 @@ def amplifier_matching(image) :
     vband_a = np.median(image[c0-lw:c0,c1-w:c1],axis=1)
     vband_b = np.median(image[c0-lw:c0,c1::c1+w],axis=1)
     rab = np.sqrt( np.median(vband_a/vband_b) / np.median(vband_b/vband_a) ) # to debias
-    log.debug("a/b=",rab)
+    log.debug("a/b={}".format(rab))
```
sbailey, line 239: In this case you are fixing a logging bug, and this statement isn't buried in a loop so performance doesn't matter much, but in general we should get in the habit of formatting log.debug statements with
```
log.debug("a/b=%f", rab)
```
The difference is that if the logger knows it won't print the message, it skips evaluating the format string, whereas the way it is currently written the string is evaluated whether or not it will be printed.  In the past that has mattered when we accidentally had a formatted debug string inside a nested loop and ended up taking a lot of time just on formatting a debug statement that was normally never printed.

bin/desi_compute_pixel_flatfield
```diff
@@ -458,20 +462,28 @@ n1=image.shape[1]
...
-    medspec=np.median(band,axis=1)
-    for i in range(band.shape[1]) :
-        band[:,i] *= np.median(band[:,i]/medspec)
-    medspec=np.median(band,axis=1)
+
+if np.any(np.isnan(band)) :
+    log.error("band image has nan values")
+    sys.exit(12)
+
+if 1 :
```
sbailey, line 470: Leftover debugging `if 1`?  i.e. let's either make this an option that we can genuinely turn on and off, or drop the `if 1` and indentation.

sbailey (approved):
The actual algorithmic update looks good, including how you handle the default parameter, setting the parameter, or getting it from the config file.

I put two minor stylistic comments inline for the record, but I can make the actual cleanup changes.

Issue #1908, opened 2022-11-15
desi_daily_proc_manager add tilenight support


sbailey:
Add support for tilenight to `desi_daily_proc_manager` (currently it is only supported by `desi_run_night` via `--use-tilenight` opt-in).  This will be important for efficient Perlmutter daily processing.

Consider switching `--use-tilenight` to be a boolean option:
* if not specified (None), use False on Cori and True on Perlmutter
* otherwise require True or False to opt in/out of the default on that system

i.e. I keep forgetting to specify `--use-tilenight` on Perlmutter, and only realize it when it submits 3x more jobs than I intended.

sbailey:
One key thing that desi_daily_proc_manager does that desi_run_night does not is the creation of the exposure tables: desi_run_night requires them to pre-exist, while desi_daily_proc_manager creates them on-the-fly.  In the specific current case of sneakernet with all data arriving at once, we can circumvent this with desi_create_exposure_tables + desi_run_night (I think; I've personally never done that), but I think desi_daily_proc_manager is the only thing that can monitor for new data, update the exposure tables on the fly, and submit jobs as needed.

marcelo-alvarez:
@sbailey thanks for pointing out this crucial difference. 

My understanding there are two cases we need for Perlmutter:
1. data for a night is transferred all at once, as it is now
2. data is transferred in real time throughout a night

For (1) we should be able to run a new night either (a) by using desi_create_exposure_tables + desi_run_night or (b) by adding --use-tilenight to desi_daily_proc_manager. 

For (2) adding --use-tilenight to desi_daily_proc_manager does not work on Perlmutter, since there is no way to leave it running indefinitely. A scrontab job that wakes up periodically to update exposure tables and run a modified version of desi_run_night seems the simplest approach going forwared, but this could be beyond the slope of this issue.

Returning to (1) it seems preferable to use option (1a), since option (1b) would take more time and may not be useful for (2).

sbailey:
I took another look at desi_run_night -> desispec.scripts.submit_night, and agree that it isn't a trivial update to add the same logic to desi_daily_proc_manager.  Followup comments:

[...]

Issue #1907, opened 2022-11-15
desi_job_graph "Maximum text size in diagram exceeded"


sbailey:
`desi_job_graph -n 20221111` results in a page with "Maximum text size in diagram exceeded".  That might be because of one of the hover tips "nightlyflat Exp 152947,152948,152949,152952,152953,152954,152957,152958,152959,152962,152963,152964 Job 64303321 CANCELLED", though that doesn't seem particularly longer than other successful cases of nightlyflats.  Or maybe the diagram itself has a maximum text length limit that we are hitting due to the super-long non-relative links like https://data.desi.lbl.gov/desi/spectro/redux/daily/run/scripts/night/20221111/psfnight-20221111-00152931-a0123456789-64303272.log .  Investigate and fix.

PR #1909, opened 2022-11-15
Jobgraph size

sbailey:
This PR partially addresses issue #1907 where the jobgraph mermaid diagrams were bigger than it will process by default.  This lowers the file size by using relative instead of absolute URL links to the log files (which is good to do anyway).  It will also help when processing runs use tilenight by default to have ~3x fewer jobs, i.e. 3x fewer nodes to represent in the diagrams; see #1908 for adding tilenight support to desi_daily_proc_manager.

This PR is sufficient for jobgraphs kept in run/jobgraph of a prod, though the file size limit is still exceed if using `desi_job_graph -o ...` to create output in a separate location (e.g. to monitor a production that your user doesn't own).  The "correct" solution would be to increase the mermaid maxFileSize limit, but I haven't figured out the javascript syntax for that.  See https://stackoverflow.com/questions/61870481/maximum-text-size-in-diagram-exceeded for an example that no longer works.  Help from javascript savvy folks would be appreciated.

image


sbailey:
canceling this PR since I branched off of main instead of daily.  I'll resubmit with a rebased branch.

PR #1926, opened 2022-12-01
create_sframesky_pdf(): display IVAR=0 pixels in yellow

araichoor:
This PR highlights IVAR=0 pixels in the sframesky-NIGHT.pdf plots in the the night qa.

I picked the yellow color, so that it does not attract too much the eye.

For instance for EXPID=154146 on 20221121, where b8 is flagged:

Current:
Screenshot 2022-12-01 at 11 38 11 AM

With this PR:
Screenshot 2022-12-01 at 11 38 25 AM

I am wondering if I should go for something a bit more fancier, with rather overplotting with transparency the yellow thing, so that one would still see the masked features (here b8), at the risk of losing in readibility...

any suggestion is welcome!

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -682,6 +682,12 @@ def _read_sframesky(night, prod, expid):
                         sel = fibermap["OBJTYPE"] == "SKY"
                         fibermap = fibermap[sel]
                         flux = h["FLUX"].read()[sel]
+                        # AR set flux=np.nan for ivar=0 pixels
+                        # AR so that those are displayed with a special
+                        # AR color in the sframesky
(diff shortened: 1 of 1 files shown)


araichoor:
mmhh some test fails, on something unrelated to this PR, I think:

```
FAILED py/desispec/test/test_linalg.py::TestLinalg::test_cholesky_invert - AssertionError: 1.2767227175480552e-07 != 0.0 within 7 places (1.2767227175480552e-07 difference)
```

sbailey:
Thanks @araichoor .  Overplotting with transparency to see the features that were masked could be neat but might take some time and tuning to get right, so let's take this improvement as-is and save further tuning for the future if we need it.

PR #1936, opened 2022-12-14
check for type of the dark option in preprocessing

Waelthus:
Currently whenever a dark file is specified by the `--dark` option the setting will be overruled by entries from DESI_SPECTRO_CALIB or DESI_SPECTRO_DARK.
This fix should allow to specify a specific dark used in the preprocessing.


julienguy (approved):
It used to work that way. Thanks for restoring this.

PR #1938, opened 2022-12-16
create_ctdet_pdf(): display the known offsets

araichoor:
This PR adds in the nightqa ctedet plot the range of columns with known offsets (for each amplifier).
So that we stop bugging @julienguy with known issues..

This parses the `$DESI_SPECTRO_CALIB/spec/sm*/sm*-?.yaml` files through the `desispec.calibfinder.CalibFinder()` function.
Thanks to @julienguy for the guidelines!

Examples for 20221215:
Screenshot 2022-12-16 at 11 32 31 AM
Screenshot 2022-12-16 at 11 32 47 AM

The color scheme follows what s displays (ie AMPA and AMPB in orange, AMPC and AMPD in blue).
Hopefully, it should work for all cases (in addition to 20221215, I ve tested it on 20210708, where AMPA and AMPD have some known columns offsets).

From a code-implementation point-of-view, I ve added a dictionary as a default value for the new argument `yoffcols={"A" : -2.5, "B" : -3.5, "C" : -2.5, "D" : -3.5}`; I don t know if this is a bad habit; if so, I m happy to implement any better suggestion!

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -23,6 +23,7 @@ from desitarget.geomask import match_to
 # AR desispec
 from desispec.fiberbitmasking import get_skysub_fiberbitmask_val
 from desispec.io import findfile
+from desispec.calibfinder import CalibFinder
 from desispec.tile_qa_plot import get_tilecov
 # AR matplotlib
 import matplotlib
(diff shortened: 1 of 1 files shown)


julienguy (approved):
Very useful, thanks.

PR #1941, opened 2022-12-20
Do not set ivar=0 to UNASSIGNED fibers by default.

julienguy:
Do not set ivar=0 to UNASSIGNED fibers by default because they might be useful for some analysis.
This addresses part of desisurveyops issue https://github.com/desihub/desisurveyops/issues/76 .
(The other part being the counterintuitive default values of redrock redshifts for spectra with zero valid data)


julienguy:
arg. This PR should be towards the daily branch not main. I am closing it.

PR #1942, opened 2022-12-20
do not set ivar=0 to UNASSIGNED fibers by default

julienguy:
Do not set ivar=0 to UNASSIGNED fibers by default because they might be useful for some analysis.
This addresses part of desisurveyops issue desihub/desisurveyops#76 .
(The other part being the counterintuitive default values of redrock redshifts for spectra with zero valid data)

### DIFF
--- a/py/desispec/fiberbitmasking.py
+++ b/py/desispec/fiberbitmasking.py
@@ -158,7 +158,7 @@ def get_all_nonamp_fiberbitmask_val():
     Also does not include POORPOSITION which is bad for stdstars
     but not necessarily fatal for otherwise processing a normal fiber.
     """
-    return (fmsk.UNASSIGNED | fmsk.BROKENFIBER | fmsk.MISSINGPOSITION | \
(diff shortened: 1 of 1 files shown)


akremin:
A simple change that will partially solve the referenced issue. This does impact the masking for all pipeline steps, as it was implemented in a lower-level function. I believe this includes coadditions, so if we still want to mask these at the coadd/redshift level then we will have to do that in a future PR. I verified that this shouldn't impact sky modeling, because those are only selected among `SKY` targets, and this affects `UNASSIGNED` targets.

sbailey:
For the record: redrock uses the fiberstatus UNASSIGNED, BADPOSITION, and BROKENFIBER bits to set the zwarn NODATA bit, so this PR allows the UNASSIGNED fibers to be fit but they will still receive a zwarn flag, which I think is what we want.  This is similar to fitting redshifts for SKY fibers but setting the zwarn SKY bit.

Arguably it would be better to set the zwarn BAD_TARGET bit for unassigned fibers, but at least something is being set.

PR #1944, opened 2022-12-20
Night qa : create_sframesky_pdf(): transparency for IVAR=0 pixels

araichoor:
This PR is (hopefully) a small upgrade of https://github.com/desihub/desispec/pull/1926.
In the night_qa sframesky plots, the `IVAR=0` pixels are now displayed as before, but with a transparent overlay.

I struggled a bit with how to do that, probably because of my lack of familiarity with the `ax.imshow()` routine...

Example before this PR (where the `IVAR=0` pixels are just displayed in yellow, with no indication of the `FLUX` values):
Screenshot 2022-12-19 at 6 24 03 PM

and with this PR:
Screenshot 2022-12-19 at 6 24 16 PM

I had hard times to decide the transparency scale, i.e. what is the right balance between seeing the `FLUX` values and seeing if it has `IVAR=0`.
And I picked the "green" color.

All that can be changed of course; @schlafly : any suggestion is welcome!

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -709,12 +709,7 @@ def _read_sframesky(night, prod, expid):
                         sel = fibermap["OBJTYPE"] == "SKY"
                         fibermap = fibermap[sel]
                         flux = h["FLUX"].read()[sel]
-                        # AR set flux=np.nan for ivar=0 pixels
-                        # AR so that those are displayed with a special
-                        # AR color in the sframesky
-                        # AR ivar is not used further, so we do not propagate it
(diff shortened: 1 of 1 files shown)


schlafly:
That looks good!  I think this is a good middle ground between wanting to see the data but knowing it's marked.  In contrast to the existing version, it's less clear that the isolated cosmic rays are marked---but I much prefer that because before it was harder to see faint amp offsets through the colored noise of masked regions.  Fine with me to merge!

schlafly:
Let's merge this!

PR #1947, opened 2022-12-20
Night qa: add 1200s morning dark

araichoor:
This PR adds the 1200s morning dark (temporary) processing + plot to the night_qa.
It is a follow-up of this PR https://github.com/desihub/desispec/pull/1945 (also see this slack thread: https://desisurvey.slack.com/archives/C01HNN87Y7J/p1670887277610869).

The code looks for the latest 1200s dark image of the night.
As the 1200s morning dark are not currently processed by the daily pipeline, this PR script does the processing on-the-fly in a temporary folder, so that night_qa can generate the plots.
On a cori-haswell interactive node, it runs in 2min, so it should not make the `desi_night_qa` script too long to run.

The call to `desi_preproc` is here:
https://github.com/desihub/desispec/blob/b584b5556e66c2ffaf605ae4f1b2e97ada3112b0/py/desispec/night_qa.py#L481-L488

@julienguy suggested: "_Make sure to have the env. variable `SPECPROD=daily` set in order to use the nightly bias (and because of that this has to be run after the daily prod)._"
I am not super-certain of how the environment variables are propagated in such a call with `os.system(cmd)`.
Please let me know if there is a better way to proceed here.

This is the test case for EXPID=157181 on 20221209:
Screenshot 2022-12-20 at 3 11 22 PM

All petals, camera here: [morningdark-20221209.pdf](https://github.com/desihub/desispec/files/10273264/morningdark-20221209.pdf)

I then also added a "Morning DARK" section in the `nightqa-NIGHT.html` page; see for instance here:
https://data.desi.lbl.gov/desi/users/raichoor/tmpdir/20221209/nightqa-20221209.html

### DIFF
--- a/bin/desi_night_qa
+++ b/bin/desi_night_qa
@@ -23,6 +23,7 @@ from desispec.night_qa import (
     get_nightqa_outfns,
     get_surveys_night_expids,
     get_dark_night_expid,
+    get_morning_dark_night_expid,
     get_ctedet_night_expid,
     create_dark_pdf,
     create_badcol_png,
@@ -137,6 +138,9 @@ def main():
     if np.in1d(["dark", "badcol"], steps_tbd).sum() > 0:
         dark_expid = get_dark_night_expid(args.night, args.prod)
 
+    if "morningdark" in steps_tbd:
+        morningdark_expid = get_morning_dark_night_expid(args.night, args.prod)
+
     # AR CTE detector expid
     if "ctedet" in steps_tbd:
         ctedet_expid = get_ctedet_night_expid(args.night, args.prod)
@@ -145,6 +149,11 @@ def main():
     if "dark" in steps_tbd:
         if dark_expid is not None:
             create_dark_pdf(outfns["dark"], args.night, args.prod, dark_expid, args.nproc)
+
+    # AR morning dark expid
+    if "morningdark" in steps_tbd:
+        if morningdark_expid is not None:
+            create_dark_pdf(outfns["morningdark"], args.night, args.prod, morningdark_expid, args.nproc)
 
     # AR badcolumn
     if "badcol" in steps_tbd:
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -1843,6 +1843,7 @@ def write_nightqa_html(outfns, night, prod, css, surveys=None, nexp=None, ntile=
 
     # AR various tabs:
     # AR - dark
+    # AR - morningdark
     # AR - badcol
     # AR - ctedet
     # AR - sframesky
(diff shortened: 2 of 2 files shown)


araichoor:
@akremin : with running a sanity test on 20210514, where there are no morning dark, I realize that the 300s dark exposure is not processed for that night (expid=88312).

I am wondering:
In the current code, in `night_qa.get_dark_night_expid()`:
https://github.com/desihub/desispec/blob/b584b5556e66c2ffaf605ae4f1b2e97ada3112b0/py/desispec/night_qa.py#L136-L181
I relied on the `processing_tables` to say if I find something or not; ie here for 20210514, I consider there is no 300s dark image.

However, I could implement there the same logic that I did for the `night_qa.get_morning_dark_night_expid()` function:
https://github.com/desihub/desispec/blob/b584b5556e66c2ffaf605ae4f1b2e97ada3112b0/py/desispec/night_qa.py#L184-L234
ie browse the `exposure_tables` instead to identify the dark image, and if it is not processed, then the code would process it on-the-fly.

Would that make sense?

julienguy:
This looks great Anand. I wasn't sure but apparently the trick to set os.environ["SPECPROD"] works:
```
In [5]: os.environ["SPECPROD"]="tata" ; os.system('echo $SPECPROD')                                                                                                                               
tata
```

sbailey:
A few changes that I'll take care of:

We should not hardcode a specific SPECPROD, even "daily".  The standard night_qa runs with SPECPROD=daily already, and any testing in other productions should setup symlinks etc. (e.g. with copyprod) to have it pickup the correct nightly bias.  I know that we've tripped on this before (running preproc outside of daily and being surprised that it was different due to getting the bias from $DESI_SPECTRO_CALIB instead of the nightlybias that doesn't exist in the test prod), but I think hardcoding a production name is even worse.

Let's directly call `desispec.scripts.preproc.main` instead of spawning a system call to `desi_preproc`.  This will give us flexibility in the future to wrap multiple night_qa calls in an MPI script on Perlmutter, which doesn't fully support MPI+os.system.

Let's *not* have it also spawn processing for the afternoon 300s dark.  In that case, the pipeline is supposed to have done it already, and if it really is missing I'd rather have QA tell us that rather than try to correct for it.  This doesn't help us for past nights before the pipeline was processing the 300s dark by default, but in general I want to avoid having QA spawn significant processing that should be the responsibility of the pipeline.

Post-Iron we can further update our workflow to have the regular pipeline process the last dark by default, but this is a pragmatic solution for now.

sbailey:
test failures are due to "Big-endian buffer not supported on little-endian compiler" fixed in main via desimodel update.  I re-ran a test night of nightqa with latest version of branch on perlmutter and it generated the morningdark*.pdf fine.  Merging.

akremin:
Sorry I missed this Anand. For the record I agree with what Stephen said above. It's unfortunate for older nights, but we should leave it to the pipeline to do most/all of the processing. Otherwise this script will become too long and computationally intensive.

PR #1948, opened 2022-12-21
Merge current daily branch into main to pick up changes for perlmutter

akremin:
Merge recent updates to nightqa and fiberbitmasks used in daily processing back into `main`. This ensures that any daily processing done on perlmutter (typically using `main`) will also pick up those changes.


marcelo-alvarez:
This looks good @akremin, thanks for taking care of this. FYI #1949 tracks whether this is necessary in the other direction, main -> daily. Hopefully not!

Issue #1951, opened 2023-01-03
tile_qa_plots.get_expids_efftimes I/O efficiency


sbailey:
`desispec.tile_qa_plots.get_expids_efftimes` can be quite slow for sv1 tiles with lots of exposures.  e.g. tile 80617 takes 6m50s to generate tile-qa*.png, compared to redrock running in <1 minute.  This can lead to job timeouts when the I/O system is bogged down from other users.

A major factor is that it reads nexp*10*3 cframe files to get the SCORES table.  It would be more I/O efficient to read these from the SCORES table of the spectra files, since that file is already being opened to read the FIBERMAP and it already contains the concatenation of all of the SCORES tables from the input cframe files (row-matched with the concatenated FIBERMAP table).  Even that is still pretty slow because the spectra files are gzipped (sorry, Julien) and fitsio unzips the entire file to read the tables, but it is still better to read 10 files instead of nexp*10*3 files.

If for some reason it is still necessary to read from the original cframe files instead of the spectra files, another I/O speedup would be to read the EXP_FIBERMAP from the coadd*.fits files since it appears to only be used to get the NIGHT and EXPIDs and the non-compressed smaller coadd files are much faster to read than the spectra files.

Detail: the coadd files also have coadded SCORES, which would be much faster to read than the spectra SCORES, but I think the tile-qa plots need the non-coadded SCORES values.

A more invasive optimization would be to add optional MPI support to also read the 10 spectra files in parallel.  Unfortunately this would need to be MPI not just multiprocessing, since the redshift jobs now bundle all the steps in an MPI wrapper and MPI+multiprocessing doesn't work well together on Perlmutter.

araichoor:
sorry for that :)

as you ve understood, the goal is to get the per-exposure efftime_spec (mimicking what s done for the `exposures-PROD.fits` file, which does not exist at this point), so it cannot be done from the `coadd` files.

I ve given a quick look a simple optimization, along your suggestion of only using the `spectra` files: for this `TILEID=80617` example, it reduces the computing time by ~2/3 (from 780s to 250s).

would that be good enough? shall I go ahead and make a PR?

ps: when exploring the files I realized that the list of (EXPID, NIGHT) can also be inferred from the `INFIL???` keywords in the `spectra` files header; this allows one instantaneously get that list; though, one still needs to read the per-camera `TSNR2` columns, i.e. open the `spectra` files.

sbailey:
[...]

PR #1953, opened 2023-01-04
tile_qa_plot: get_expids_efftimes(): speed-up, with only reading spectra files

araichoor:
This PR addresses https://github.com/desihub/desispec/issues/1951.

It is a speed-up in the `desispec.tile_qa_plot.get_expids_efftimes()` function.
Instead of opening both all the `spectra` and the `cframe` files, this PR only opens the `spectra` files.

The output should be identical.
I verified this with running the function on ~100 random exposures of the himalayas redux.

For the problematic case (TILEID=80617 in himalayas), it reduces the computation time by 50% (or more, it depends).


araichoor:
argh, I requested merging in the "main".. sorry; I ll close it and open another one requesting merging in the "daily".

PR #1950, opened 2023-01-03
Fix tsnr afterburner crash when no cframes on a night

akremin:
### Overview
This fixes #1345, which is a long-standing bug in the tsnr afterburner (`desi_tsnr_afterburner`) that causes it to crash if no cframes are available on a night. This is generally not an issue for reprocessing since we don't process or care about nights without valid science data, but for daily it matters since the crash prevents the exposures and tiles from being listed in the output files and therefore the survey operations don't have a way to propagate the 0 efftime for those exposures.

This fix is imperfect, as it only works if you supply an existing `exposures-${SPECPROD}.fits` file. This solves all existing issues, however, since we can generate that file with any other night in the production that has cframes prior to running on a problematic night. 

### Testing
I tested on all nights linked to that issue: 20210710 , 20210922, 20211213, 20211226

The output files can be found here: `/global/cfs/cdirs/desi/users/kremin/PRs/tsnr_nocframe/`
And include: `tiles-daily.csv `, `exposures-daily.csv`, and `exposures-daily.fits`.

#### Example: Night 20211213

Running out of the `daily` branch
```
kremin@cori09:/global/cfs/cdirs/desi/users/kremin/PRs/tsnr_nocframe> desi_tsnr_afterburner -o /global/cfs/cdirs/desi/users/kremin/PRs/tsnr_nocframe/exposures-daily.fits --prod daily --tile-completeness /global/cfs/cdirs/desi/users/kremin/PRs/tsnr_nocframe/tiles-daily.csv --aux /global/cfs/cdirs/desi/survey/observations/SV1/sv1-tiles.fits --gfa-proc-dir /global/cfs/cdirs/desi/survey/GFA/ --compute-skymags --update --nproc 8 --nights 20211213 --add-badexp
```

Yields: 
```
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/main/bin/desi_tsnr_afterburner", line 1211, in <module>
    sys.exit(main())
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/code/desispec/main/bin/desi_tsnr_afterburner", line 1045, in main
    tsnr2_expid_table = vstack(tsnr2_expid)
  File "/global/common/software/desi/cori/desiconda/20211217-2.0.0/conda/lib/python3.9/site-packages/astropy/table/operations.py", line 646, in vstack
    tables = _get_list_of_tables(tables)  # validates input
[...]

PR #1954, opened 2023-01-04
Tile_qa_plot: get_expids_efftimes(): speed-up, with only reading spectra files

araichoor:
[re-opening a PR requesting to merge in "daily" this time]

This PR addresses https://github.com/desihub/desispec/issues/1951.

It is a speed-up in the desispec.tile_qa_plot.get_expids_efftimes() function.
Instead of opening both all the spectra and the cframe files, this PR only opens the spectra files.

The output should be identical.
I verified this with running the function on ~100 random exposures of the himalayas redux.

For the problematic case (TILEID=80617 in himalayas), it reduces the computation time by 50% (or more, it depends).

### DIFF
--- a/py/desispec/tile_qa_plot.py
+++ b/py/desispec/tile_qa_plot.py
@@ -891,13 +891,30 @@ def get_expids_efftimes(tileqafits, prod):
                 )
                 tmpstr = os.path.join(tiledir, f'spectra-*-{tileid}-*{night}.fits*')
                 spectra_fns = sorted(glob(tmpstr))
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks.  The code looks good.  I'll trust your reproducibility testing.  The unreleated github unit test failures are due to some new incompatibility with scipy 1.10.x; in PR #1950 Anthony fixed this by pinning scipy==1.7.3 (same as NERSC).  To avoid merge conflicts I won't do the same here, but I'll investigate a long term fix to support scipy 1.10.x (it might involve updating desimodel instead of desispec).

Issue #1952, opened 2023-01-04
track cumulative redshift jobs in processing tables


sbailey:
For Himalayas, we submitted cumulative redshift jobs post-facto using 
```
desi_tile_redshifts -t TILEID -g cumulative --run_zmtl
```
after first processing all exposures through cframes.  This is tedious and error prone, especially since these jobs are not tracked via the processing tables to make them visible to desi_proc_dashboard, desi_job_graph, and desi_resubmit_queue_failures.

Two options to improve get these into the processing tables:
1. desi_run_night already allows submitting cumulative redshifts and adding them to the processing table, but it doesn't know about whether this is the last night for a tile and whether a cumulative redshift job should be submitted.  Add hooks to look at future exposure tables to decide which tiles should get cumulative redshifts for this night.
2. add hooks to desi_tile_redshifts to also update the processing tables.

After discussions with @akremin, we think we want both, but should start with (1) since that would allow us to submit cumulative redshift jobs at the same time as everything else.  The new desi_zproc will handle cross-night dependencies and exit with an error if a required previous night isn't finished yet.  This isn't quite as clean as full cross-night job dependencies, but it gives the flexibility of submitting nights in any order and then cleaning up failures with desi_resubmit_queue_failures.

This isn't strictly needed for Iron, but it would make that production go much more smoothly so I'm going to add it to the Iron project.  The alternative is "be more careful in the first place and resubmit jobs by hand as needed", which will likely be more work than implementing this feature in the first place.

sbailey:
Part 1 is implemented (desi_run_night updates) by PR #1955.  Part 2 is still do to, but isn't critical for Iron so I will flag it as done for that project while leaving the ticket open.

akremin:
The correct way to do this is using `desi_proc_night`. `desi_tile_redshifts` is useful explicitly because it runs outside of some of the pipeline infrastructure. Therefore adding the processing table hooks there wouldn't be beneficial. `desi_proc_night` allows just submitting missing redshift jobs now too.

Issue #1943, opened 2022-12-20
Some tiles cross Y1-Y2 boundary and would be incompletely included in Y1 processing


akremin:
I discovered a subtlety in how we determine if tiles are "completed" or not, as implemented in PR #1843, while investigating  the unrelated issue #1914 for tile 41257. That PR impacts which tiles are submitted for processing when using `desi_run_night`.

The code uses the "specstatus" file in the survey operations svn, which at NERSC is located at `/global/cfs/cdirs/desi/survey/ops/surveyops/trunk/ops/tiles-specstatus.ecsv`. That file is updated daily and contains up-to-date tile information about all observations. 

However, a production like Himalayas is only run on a subset of the data, e.g. year 1 only. Some tiles were observed in Y1 and then again in Y2. This leads to issues when we want to process complete tiles, but the year 1 portion of observations are incomplete for a given tile.

This was identified when looking at an unrelated issue for tile 41257, since that tile was observed twice in Y1 and once in Y2. Half of its effective time came in the final exposure taken in Oct 2022, after the Y1 cutoff.

Doing a sweep over exposure tables in daily looking for tiles with valid observations made both before and after 20220914 gives the table below.

```
Tiles observed across the Y1 boundary:
Tile  2946: survey=main, faprgrm=dark  ,  faflavor=maindark  , Completed=True,   with nights=[20220327 20221130]
Tile  7454: survey=main, faprgrm=dark  ,  faflavor=maindark  , Completed=True,   with nights=[20211222 20220915]
Tile  9623: survey=main, faprgrm=dark  ,  faflavor=maindark  , Completed=True,   with nights=[20220611 20220915]
Tile 23108: survey=main, faprgrm=bright,  faflavor=mainbright, Completed=False,  with nights=[20220406 20221024 20221025]
Tile 23445: survey=main, faprgrm=bright,  faflavor=mainbright, Completed=True,   with nights=[20211208 20220923]
Tile 24447: survey=main, faprgrm=bright,  faflavor=mainbright, Completed=True,   with nights=[20220315 20221019]
Tile 25744: survey=main, faprgrm=bright,  faflavor=mainbright, Completed=True,   with nights=[20220127 20220922]
Tile 40151: survey=main, faprgrm=backup,  faflavor=mainbackup, Completed=True,   with nights=[20220613 20220910 20220918]
Tile 40299: survey=main, faprgrm=backup,  faflavor=mainbackup, Completed=True,   with nights=[20220514 20220917]
[...]

PR #1955, opened 2023-01-04
submit cumulative redshifts only for lastnight of each tile

sbailey:
This PR adds support to desi_run_night to submit cumulative redshifts only for tiles for which the current night is the last night that they were observed.  This allows them to be tracked in the processing tables (see issue #1952) while not submitting cumulative redshift jobs for nights earlier than the LASTNIGHT for each tile.

I did *not* implement an option to override this and submit cumulative redshifts for every tile, even if they have future observations (current behavior).  The code is already complex from supporting tilenight, non-tilenight, productions, and daily-like "is this the last exposure for this tile or the end of the night?" style logic, and I didn't want to add additional options that we may never use and then feel obligated to support those in any future refactors.  If we find we need that functionality, we can add it later.

## Tests ##

I tested on the following nights, to cover several cases:
  * 20211216 - ends with a science exposure
  * 20220602 - ends with a full set of arcs+flats+zeros+darks
  * 20220603 - ends with zeros+darks but the science exposures are the last that would be processed

```
export DESI_SPECTRO_REDUX=/global/cfs/cdirs/desi/users/sjbailey/spectro/redux
export SPECPROD=zproc
cd $DESI_SPECTRO_REDUX/$SPECPROD
svn checkout https://desi.lbl.gov/svn/data/exposure_tables

desi_run_night --dry-run-level=2 -n 20211216 --z-submit-types pernight,cumulative &> 20211216-submit.log
desi_job_graph -n 20211216
...
```

The resulting jobgraphs in https://data.desi.lbl.gov/desi/users/sjbailey/spectro/redux/zproc/run/jobgraph/ show that all tiles get perexp redshift jobs, but only some tiles get cumulative redshift jobs.  I used code like the following to identify which tiles are supposed to get cumulative jobs:
```python
from desispec.workflow.redshifts import read_minimal_exptables_columns
exps = read_minimal_exptables_columns()

night = 20220603
tiles = np.unique(exps["TILEID"][exps["NIGHT"] == night])
for t in tiles:
    jj = exps["TILEID"] == t
    lastnight = np.max(exps["NIGHT"][jj])
    if lastnight == night:
        print(f"tile {t} - ok to submit")
    else:
        print(f"tile {t} {lastnight=}")
```

I repeated with `desi_run_night --no-use-tilenight ...` to make sure that variant of the logic works too.

### DIFF
--- a/bin/desi_run_night
+++ b/bin/desi_run_night
@@ -12,7 +12,7 @@ def parse_args():  # options=None):
     """
     parser = argparse.ArgumentParser(description="Submit a one past night of data for processing with the DESI data pipeline.")
 
-    parser.add_argument("-n","--night", type=str, required=True, help="The night you want processed.")
+    parser.add_argument("-n","--night", type=int, required=True, help="The night you want processed.")
     parser.add_argument("--proc-obstypes", type=str, default=None, required=False,
                         help="The basic data obstypes to submit for processing. " +
                              "E.g. science, dark, twilight, flat, arc, zero.")
--- a/py/desispec/scripts/submit_night.py
+++ b/py/desispec/scripts/submit_night.py
@@ -19,6 +19,7 @@ from desispec.workflow.procfuncs import parse_previous_tables, get_type_and_tile
                                         checkfor_and_submit_joint_job, submit_tilenight_and_redshifts
 from desispec.workflow.queue import update_from_queue, any_jobs_not_complete
 from desispec.workflow.desi_proc_funcs import get_desi_proc_batch_file_path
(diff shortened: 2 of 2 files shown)


akremin (approved):
The code looks great and the tests are thorough. This will be susceptible to the year-boundary issue outlined in #1943, but will be automatically accounted for when that is resolved, so nothing needs to be done with that here. I will do more testing once this and the PR resolving #1943 are both implemented to ensure we don't have any corner cases that have snuck through. For now I will approve and merge this.

PR #1957, opened 2023-01-05
use desimodel/0.18.0 for testing

sbailey:
This PR updates desispec to use desimodel/0.18.0 for testing.  This version of desimodel includes an endianness fix for compatibility with scipy/1.10.0 (see desihub/desimodel#164).  This is the underlying cause of the test failures in recent PRs (e.g. #1950, which temporarily fixed it by pinning to an older scipy; also #1954 and #1955).

I plan to self-merge as long as tests do indeed use scipy/1.10.0 and pass.  After this PR I'll also merge daily -> main to pickup this and other recent updates to daily.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -23,7 +23,7 @@ jobs:
                 numpy-version: ['<1.23']  # to keep asscalar, used by astropy
         env:
             DESIUTIL_VERSION: 3.2.5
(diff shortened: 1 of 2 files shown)


sbailey:
Test logs confirm that it used scipy/1.10.0 and desimodel/0.18.0 and passed.  Merging.

PR #1966, opened 2023-01-12
re-allow redrock+GPU on large inputs

sbailey:
Redrock+GPU used to run out of memory on input files with >1k targets (e.g. large healpix), which we had worked around in zproc by checking the number of targets and falling back to CPU-only on big inputs.  desihub/redrock#211 fixed the problem, so this PR removes the ntargets check from zproc.  I tested by rerunning himalayas sv3 dark healpix 10152 with 2192 targets and confirming that zproc ran redrock with GPU and that it completed fine.

This PR also fixes a bug where `zproc --batch ...` was accidentally forcing `--no-gpu` unless `--mpi` was also specified.


akremin (approved):
The changes look good, and good catch on the bug fix as well. I will trust your test as representative and approve+merge.

PR #1982, opened 2023-01-26
Add calibnight jobs to exposure dashboard

akremin:
Short PR that adds some information about the calibnight jobs into the exposure dashboard. I didn't change the columns, so the `ccdcalib` job is uninformative (like the `zero` and `dark` jobs currently are), but the `psfnight` and `nightlyflat` jobs now show the number of files.

After this is merged I will build upon it to include up-to-date queue information on the jobs, which will make the inclusion of the `ccdcalib` row slightly more useful/meaningful.

I have run a test in `/global/cfs/cdirs/desi/users/kremin/PRs/dash_calibnights/dashboard`. A snapshot of a single nights dashboard is below showing the newly included `ccdcalib`, `psfnight`, and `nightlyflat` rows properly populated.

image

### DIFF
--- a/py/desispec/scripts/procdashboard.py
+++ b/py/desispec/scripts/procdashboard.py
@@ -162,19 +162,23 @@ def populate_night_info(night, check_on_disk=False,
     ## Note that the following list should be in order of processing. I.e. the first filetype given should be the
     ## first file type generated. This is assumed for the automated "terminal step" determination that follows
     expected_by_type = dict()
+    expected_by_type['zero'] =     {'psf': 0,    'frame': 0, 'ff': 0,
+                                    'sframe': 0, 'std': 0,   'cframe': 0}
     expected_by_type['arc'] =      {'psf': 1,    'frame': 0, 'ff': 0,
                                     'sframe': 0, 'std': 0,   'cframe': 0}
     expected_by_type['cteflat'] =  {'psf': 1,    'frame': 1, 'ff': 0,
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks; it is great to have psfnight and nightlyflat included.

I ran this on iron, with output at https://data.desi.lbl.gov/desi/users/sjbailey/dev/dashboard/dashboard.html , which highlights some corner cases for when we don't have all arcs:

20220207: one exposure has a bad camera but the others are fine and the pipeline successfully created all 30 psfnight files.  However, the dashboard incorrectly thinks this is an overfull case, expecting only 29:

image

On the flip side, I think this PR catches a genuine bug in iron:  On 20220104 two arc exposures had bad r2, but three others had r2 and it should have been able to create an r2 psfnight but it didn't:

image

20210613 is the inverse of 20220207: there was only one exposure missing z3, but in this case, the dashboard correctly derived how many PSFs it could create, and the pipeline is incorrectly missing psfnight-z3, whereas on 20220207 it was able to handle missing one camera.

image

i.e. I think 20220707 is an overfull bug in the dashboard, and 20220104 and 20210613 are bugs in the pipeline, though it's not yet clear to me why the pipeline succeeded in making 20220207-r4 but failed to make 20210613-z3 when in both cases they had 4 out of 5 input PSFs.

sbailey:
The updated logic counts the PSFs for all of these cases correctly.  Thanks!  Merging now.

Issue #2000, opened 2023-02-09
Functionality to mask or override sky fibers lists when computing sky


segasai:
Hi, 

for various M31 related reasons (@arjundey), we'd like to be able to compute custom skymodels while either 
* removing some targetids from being treated as sky
* or completely override the sky target list
And ideally we'd prefer to be able to do that without having overwrite the OBJTYPE in the fibermap.

Is it okay to submit the patch like below that implements that ? (obviously it would have no use for regular pipeline runs, but would be useful for our project 309)
It would add a couple of options to desi_compute_sky (something like --override_sky_targetids and --exclude-sky-targetids

```diff
diff --git a/py/desispec/sky.py b/py/desispec/sky.py
index 44f095af..2c0436c9 100644
--- a/py/desispec/sky.py
+++ b/py/desispec/sky.py
@@ -23,7 +23,7 @@ from desispec.fiberbitmasking import get_fiberbitmasked_frame_arrays, get_fiberb
 import scipy.ndimage
 from desispec.maskbits import specmask
 
-def compute_sky(frame, nsig_clipping=4.,max_iterations=100,model_ivar=False,add_variance=True,angular_variation_deg=0,chromatic_variation_deg=0,adjust_wavelength=False,adjust_lsf=False) :
+def compute_sky(frame, nsig_clipping=4.,max_iterations=100,model_ivar=False,add_variance=True,angular_variation_deg=0,chromatic_variation_deg=0,adjust_wavelength=False,adjust_lsf=False, exclude_sky_targetids=None, override_sky_targetids=None):
     """Compute a sky model.
 
     Input flux are expected to be flatfielded!
@@ -51,7 +51,8 @@ def compute_sky(frame, nsig_clipping=4.,max_iterations=100,model_ivar=False,add_
     """
     if angular_variation_deg == 0 :
         skymodel = compute_uniform_sky(frame, nsig_clipping=nsig_clipping,max_iterations=max_iterations,\
-                                       model_ivar=model_ivar,add_variance=add_variance,adjust_wavelength=adjust_wavelength,adjust_lsf=adjust_lsf)
+                                       model_ivar=model_ivar,add_variance=add_variance,adjust_wavelength=adjust_wavelength,adjust_lsf=adjust_lsf
+                                       , exclude_sky_targetids=exclude_sky_targetids, override_sky_targetids=override_sky_targetids)
     else :
 
         if adjust_wavelength :
[...]

PR #2001, opened 2023-02-09
Allow to override the targetids of sky fibers when computing sky models

segasai:
Hi, 

This is a proposal patch to to be able to compute custom skymodels with desi_compute_sky while either

* removing some targetids from being treated as sky
* or completely override the sky target list
 without having overwrite the OBJTYPE in the fibermap.

It adds a couple of options to desi_compute_sky --override_sky_targetids and --exclude-sky-targetids

(see #2000  )
It is helpful for custom coadds while removing sky fibers contaminated by interstellar emission lines for example.

### DIFF
--- a/py/desispec/io/sky.py
+++ b/py/desispec/io/sky.py
@@ -68,6 +68,9 @@ def write_sky(outfile, skymodel, header=None):
     if skymodel.skygradpcacoeff is not None:
         hx.append(fits.ImageHDU(skymodel.skygradpcacoeff.astype('f4'),
                                 name='SKYGRADPCACOEFF'))
+    if skymodel.skytargetid is not None:
+        hx.append(fits.ImageHDU(skymodel.skytargetid,
+                                name='SKYTARGETID'))
 
 
     t0 = time.time()
--- a/py/desispec/scripts/sky.py
+++ b/py/desispec/scripts/sky.py
@@ -62,7 +62,7 @@ def parse(options=None):
     parser.add_argument('--exclude-sky-targetids', type = str, default = None, required = False,
                         help = 'List of TARGETIDs to exclude from sky calculation')
     parser.add_argument('--override-sky-targetids', type = str, default = None, required = False,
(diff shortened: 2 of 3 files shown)


sbailey:
Thanks!  This looks really useful and a lot less hassle than having to edit the FIBERMAP in an input file every time.  A few comments / questions:

In ticket #2000 the patch was in three functions; this PR just has it in one.  Even if we're not using all of the variations for sky modeling for standard processing, it would be unpleasantly surprising if the override list was silently ignored just because some other non-default sky model method was requested.  I suggest abstracting your new code into a function like
```python
skyfibers = get_sky_fibers(fibermap, override_sky_targetids=None, exclude_sky_targetids=None)
```
and then use that everywhere in place of the current
```
skyfibers = np.where(frame.fibermap['OBJTYPE'] == 'SKY')[0]
```

Please also add the new options to the docstrings of the functions where you add them.

It appears that the output files don't record which sky fibers were actually used, which is important to track now that it has the possibility to be different from the fibermap.  I suggest adding an additional HDU to the sky files that contain either the FIBER numbers or the TARGETIDs used.

segasai:
Thanks for having a look Stephen!

Regarding the change in only one function vs three  it's because when I wrote #2000 I was incorrectly comparing against an old desispec version.
In the current desispec the skys are fetched in one single place: https://github.com/desihub/desispec/blob/d9f7cd833452088bbeed2791a000516c37e00da9/py/desispec/sky.py#L501 

But I agree it's better to put the fiber selection in a function. I'll do that. 

I agree on recording which fibers were used. If I'd done that I'd probably wrote that in the header, but if you think the HDU is better I'll do that instead.

sbailey:
Thanks, looks good.  Since the list of TARGETIDs can be long, I think a new HDU is better than header keyword(s).  I updated the docstrings and added some logging, but will merge after tests pass again.

segasai:
Thanks for merging/updating, Stephen!

PR #2008, opened 2023-02-13
iron-necessary updates to io.photo

moustakas:
[WIP]

In generating the *targetphot* and *tractorphot* catalogs for **iron**, `desispec.io.photo` crashed on tile `82401` because of the following issue:
```
source /global/cfs/cdirs/desi/software/desi_environment.sh main
from desispec.io.photo import gather_targetdirs
gather_targetdirs(82401)
array(['/global/cfs/cdirs/desi/survey/ops/surveyops/trunk/mtl/main/ToO/ToO.ecsv',
       '/global/cscratch1/sd/adamyers/gaiadr2/1.3.0.dev5218/targets/main/resolve/backup/'],
      dtype='<U80')

fitsheader /global/cfs/cdirs/desi/target/fiberassign/tiles/trunk/082/fiberassign-082401.fits.gz
[snip]
DESIROOT= '/global/cfs/cdirs/desi'
GFA     = 'DESIROOT/target/catalogs/dr9/1.1.1/gfas'
MTL     = 'DESIROOT/survey/ops/surveyops/trunk/mtl/main/backup'
SKY     = 'DESIROOT/target/catalogs/dr9/1.1.1/skies'
SKYSUPP = 'DESIROOT/target/catalogs/gaiadr2/1.1.1/skies-supp'
TARG    = '/global/cscratch1/sd/adamyers/gaiadr2/1.3.0.dev5218/targets/main/re&'CONTINUE  'solve/backup/'
[snip]
```

@geordie666 (and @araichoor confirmed) this issue is due to the following:
```
These tiles (as I recall) were experimental tiles in preparation for developing the 
actual backup program. As the code was under development (1.3.0.dev5218) I 
wouldn't have retained the target files. If we're lucky, the targets will be the same 
as for the actual backup program, which was processed with the next tagged code 
for backup targets (2.2.0). If so, then using $TARG_DIR/gaiadr2/2.2.0/targets/main/resolve/backup 
will work. If we aren't lucky, I'd be tempted to drop targeting information for any specialbackup 
files with a note that the targets don't exist. I'd be surprised if the spectra on these tiles are 
any good — unless they were cleaned-up post-facto using the knowledge gained from these 
experimental specialbackup programs. The relevant specialbackup tiles are 
82401,82402,82403,82404,82405,82406,82407,82408 and 82409.
```

So this PR includes a few-line change to `io.photo.gather_targetdirs` to handle this corner case.

In addition, I discovered another corner-case crash in the *tractorphot* step were there are precisely zero *brick_primary==True* targets on a given brick, e.g., 
https://www.legacysurvey.org/viewer/?ra=236.81242889&dec=-2.6225938&bricks


geordie666:
A note in case future people rediscover the information in this PR:

We _were_ lucky and version `2.2.0` of the backup target files can be safely used in place of `1.3.0.dev5218`. Anand reminded me that this was actually by design, rather than being luck. Or, at least, we already knew that the `1.3.0.dev5218` -> `2.2.0` substitution was safe, and I'd forgotten that point.

araichoor:
hi @moustakas,

just a very quick comment/suggestion on https://github.com/desihub/desispec/pull/2008/commits/2bad6b68644984d86af1edc3097f6bade418298f:
- note that it can also happen to have some `TOO2` header keywords (e.g., for calibration tiles, where we provide a custom ToO + the regular ToO); see e.g. https://github.com/desihub/fiberassign/pull/437#issuecomment-1300844918
- for SCND* and TOO*: I don t know if that will be of any use in the future, but instead to explore up to `TARG6`, `SCND6`, couldn t you just brutally explore, say, 2, 3, 4, ..., 999, and then you ll be sure you ll handle any future cases?

moustakas:
@araichoor how about something like this?
```
    # Gather the targeting directories.
    targetdirs = [fahdr['TARG']]
    for _targ in (np.arange(99)+2).astype(str):
        moretarg = 'TARG'+_targ
        if moretarg in fahdr and fahdr[moretarg].strip() != '-':
            targetdirs += [fahdr[moretarg]]
        else:
            break

    # Any secondary targets or ToOs?
    if 'SCND' in fahdr:
        targetdirs += [fahdr['SCND']]
        for _scnd in (np.arange(99)+2).astype(str):
            morescnd = 'SCND'+_scnd
            if morescnd in fahdr and fahdr[morescnd].strip() != '-':
                targetdirs += [fahdr[morescnd]]
            else:
                break
```

py/desispec/io/photo.py
```diff
@@ -52,29 +52,16 @@ def gather_targetdirs(tileid, fiberassign_dir=None):
...
-        
-    if 'TOO' in fahdr:
-        TOOfile = fahdr['TOO']
-        # can be a KPNO directory!
-        if 'DESIROOT' in TOOfile:
-            TOOfile = os.path.join(desi_root, TOOfile.replace('DESIROOT/', ''))
-        if TOOfile[:6] == '/data/': # fragile
-            TOOfile = os.path.join(desi_root, TOOfile.replace('/data/', ''))
-        if 'afternoon_planning' in TOOfile:
-            TOOfile = TOOfile.replace('afternoon_planning/surveyops', 'survey/ops/surveyops') # fragile!
```
araichoor, line 73: quick remark (I ve not checked the rest of the code): wouldn t you want to keep those kpno-related lines?

araichoor:
I ve added a small question on the last commit.

more generally (again, I ve not checked the rest of your code, sorry...): this function seems to be tackling the same issues as this `fiberassign.fba_patch_io.get_static_desitarget_fns()` function, no?
https://github.com/desihub/fiberassign/blob/309d30a630ed92d979b7e9f3fe313544944d0c17/py/fiberassign/fba_patch_io.py#L169

I don t know if we d want to "merge" them, ie to have a single place with this function (in the fiberassign one?); but maybe it s not worth the trouble...
you know better!

moustakas:
Thanks for the suggestion @araichoor. However, I use the functions in `desispec.io.photo` in fastspecfit, and I don't want to have to bring in the fiberassign package as an unnecessary dependence.

Nevertheless, I tried using your code as a test and I found it to be very slow (perhaps because of the multiple loops over 100??), so the latest committed version of `gather_targetdirs` is a hybrid approach. Take a look if you're inspired, but there's no urgency to do so.

I'm testing on fuji now and so far I'm getting identical results to the VACs I produced previously, so I'm going to run on iron next.

py/desispec/io/photo.py
```diff
@@ -46,93 +53,109 @@ def gather_targetdirs(tileid, fiberassign_dir=None):
...
+        fn = fn.replace("/data/target", os.getenv("DESI_TARGET"))
+        fn = fn.replace(
+            "/data/afternoon_planning/surveyops/trunk", os.getenv("DESI_SURVEYOPS")
+        )
+        fn = fn.replace(
+            "/global/cscratch1/sd/adamyers/gaiadr2/1.3.0.dev5218/targets/main/resolve/backup",
+            "{}/catalogs/gaiadr2/2.2.0/targets/main/resolve/backup".format(
+                os.getenv("DESI_TARGET")
+            ),
+        )
```
sbailey, line 143: for special cases like this, please add a comment for our future selves along the lines of "experimental tile 82401 for developing the backup program used gaiadr2/1.3.0.dev5218 inputs that no longer but are equivalent to gaiadr2/2.2.0"

sbailey (approved):
I put some minor code maintainability comment requests inline, but will trust the algorithmic tests/review of @moustakas @araichoor and @geordie666 .

moustakas:
@sbailey I know you strive to avoid tag churn, but I'm going to need a new tag in order to finalize the **iron** *targetphot* and *fastspecfit* VACs (in terms of reproducibility and what version of `desispec` is written to the file headers).

PR #2014, opened 2023-03-03
consider the --cameras option in desi_proc_tilenight 

### LINKED ISSUE #1992: desi_proc_tilenight ignores its --cameras option
`desi_proc_tilenight` has a `--cameras` option, but it appears to ignore that option and use the exposure tables instead to derive what can be done.  Since there can be some subtleties when a camera is good on one exposure but bad on another, I suggest that the code should treat `--cameras` as the superset of what to consider.  e.g.

```
TILEID EXPID BADCAMWORD
123    100   r1
123    101   --
```
with `--cameras a012` it would process petals 0, 1, and 2 (but not trip on r1 missing on one exposure), but also not process petals 3-9 (which it currently does).

This would be useful for rapid testing and custom processing, e.g. M31 tile 80715 where we can fully process 3 petals through cframes but 7 petals can only make it through sframes due to lack of stdstars.

marcelo-alvarez:
Fixes #1992 by considering the intersection of current camwords for each step with those of the args.cameras option, effectively making the argument to --cameras the superset for all steps resulting from the desi_proc_tilenight command.

@sbailey please test and/or provide a test case to try. Thanks.

### DIFF
--- a/py/desispec/scripts/proc_tilenight.py
+++ b/py/desispec/scripts/proc_tilenight.py
@@ -76,6 +76,11 @@ def main(args=None, comm=None):
     if args.laststeps is not None:
         keep &= np.isin(exptable['LASTSTEP'].astype(str), args.laststeps)
 
+    #- Camera superset from args.cameras; if None default to all
+    camera_superset = 'a0123456789'
+    if args.cameras is not None:
+        camera_superset = args.cameras
+
     exptable = exptable[keep]
 
     if len(exptable) == 0:
@@ -98,7 +103,7 @@ def main(args=None, comm=None):
             prestd_camwords[expids[i]] = difference_camwords(camword,badcamword,suppress_logging=True)
         else:
             prestd_camwords[expids[i]] = camword
-        prestd_camwords[expids[i]] = camword_intersection([prestd_camwords[expids[i]],args.cameras])
+        prestd_camwords[expids[i]] = camword_intersection([prestd_camwords[expids[i]],camera_superset])
 
         laststep = str(exptable['LASTSTEP'][i]).lower()
         if laststep in ('all', 'fluxcalib', 'skysub'):
@@ -108,11 +113,11 @@ def main(args=None, comm=None):
             poststdstar_expids.append(expid)
 
     joint_camwords = camword_union(list(prestd_camwords.values()), full_spectros_only=True)
-    joint_camwords = camword_intersection([joint_camwords, args.cameras])
+    joint_camwords = camword_intersection([joint_camwords, camera_superset])
 
     poststd_camwords = dict()
     for expid, camword in prestd_camwords.items():
-        poststd_camwords[expid] = camword_intersection([joint_camwords, camword, args.cameras])
+        poststd_camwords[expid] = camword_intersection([joint_camwords, camword, camera_superset])
 
     #-------------------------------------------------------------------------
     #- Create and submit a batch job if requested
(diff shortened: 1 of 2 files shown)


marcelo-alvarez:
@sbailey After my most recent commits,
```
desi_proc_tilenight --cameras a678 --tile 80694 --night 20210306 --batch --nosubmit
```
will generate a batch file like $CFS/desi/users/malvarez/spectro/redux/tilenight-cameras/run/scripts/night/20210306/tilenight-20210306-80694.slurm that does propagate the `--cameras a678` option to the batch script.

Please do more testing with an actual run to see if there are any other gotchas and feel free to merge if it looks good. Thanks.

sbailey:
Running `desi_proc_tilenight --cameras a678 --tile 80694 --night 20210306 --batch` correctly processes only cameras a7 and a8 because a6 is missing from all exposures of that tile, and it properly skips exposure 20210306/00079571 z8 since that is flagged as bad, but it still successfully processes 20210306/00079570 z8 and thus is able to fit standard stars and flux calibration a8.  Good, that is the desired behavior: process the intersection of what is requested and what is possible.

One thing left to fix: `desi_proc_tilenight --tile 80694 --night 20210306 --batch --nosubmit` without specifying `--cameras ...` crashes.  Please update it so that 
* if `--cameras` is specified it is propagated to the batch job and used
* if `--cameras` is not specified it doesn't crash, doesn't propagate an option to the batch up, and let the invoked `desi_proc_tilenight` default to using all the cameras that it can.

log:
```
$> desi_proc_tilenight --tile 80694 --night 20210306 --batch --nosubmit
...
  File "/global/common/software/desi/users/sjbailey/desispec/bin/desi_proc_tilenight", line 8, in <module>
    sys.exit(proc_tilenight.main(args))
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/scripts/proc_tilenight.py", line 101, in main
    prestd_camwords[expids[i]] = camword_intersection([prestd_camwords[expids[i]],args.cameras])
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/io/util.py", line 618, in camword_intersection
    cameras &= set(decode_camword(cw))
  File "/global/common/software/desi/users/sjbailey/desispec/py/desispec/io/util.py", line 430, in decode_camword
    while len(searchstr) > 1:
TypeError: object of type 'NoneType' has no len()
```

marcelo-alvarez:
@sbailey thanks for catching this. The latest commit does *almost* exactly what you asked. Its behavior corresponds to changing your second bullet point from
- if `--cameras` is not specified it doesn't crash, doesn't propagate an option to the batch up, and let the invoked desi_proc_tilenight default to using all the cameras that it can.

to

- if `--cameras` is not specified it doesn't crash and propagates the option `--cameras a0123456789`, so that the invoked desi_proc_tilenight defaults to using all the cameras that it can.

Given that a `--cameras` option is defined multiple times in desispec, explicitly indicating its value in the slurm script is the safer, more transparent route. Should we develop a clearer way of grouping user options and function arguments internally (e.g. #2021 ), so that the `--cameras` option is only defined once, we could revisit this. 

When running
```
$> desi_proc_tilenight --tile 80694 --night 20210306 --batch --nosubmit
```
and 
```
$> desi_proc_tilenight --tile 80694 --night 20210306 --batch --nosubmit --cameras a678
```
I obtain tilenight slurm scripts with arguments to desi_proc_tilenight of `--cameras  a0123456789` and `--cameras a678`, respectively. If this looks good please go ahead and merge.

PR #2016, opened 2023-03-04
Night_qa: create_dark_pdf(): handle missing cameras in desi_preproc

araichoor:
This PR addresses the case where some cameras are missing in the dark processing.
`desi_night_qa` processes the dark on-the-fly if those are not processed yet (which currently is the case for the morning darks).

For simplicity, I just parse the header extnames of the dark `desi-EXPID.fits.fz` file.

Here an example for 20230303, where the spectro. 3 is missing: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/nightqa_v20/morningdark-20230303.pdf

Calling sequence (from an interactive node):
```
desi_night_qa -p /global/cfs/cdirs/desi/spectro/redux/daily -n 20230303 -g cumulative -o /global/cfs/cdirs/desi/users/raichoor/nightqa_dev/nightqa_v20 --steps morningdark --nproc 8
```

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -478,8 +478,19 @@ def create_dark_pdf(outpdf, night, prod, dark_expid, nproc, binning=4):
             specprod_dir = tempfile.mkdtemp()
             outdir = os.path.join(specprod_dir, "preproc", str(night), "{:08d}".format(dark_expid))
             os.makedirs(outdir, exist_ok=True)
-            cmd = "desi_preproc -n {} -e {} --outdir {} --ncpu {}".format(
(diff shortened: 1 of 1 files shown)


araichoor:
thanks, I totally forgot about this one!

akremin (commented):
FWIW there are functions to do this is `desispec.workflow.desi_proc_func.cameras_from_raw_data` takes a filename or filehandle and gets the cameras, and `desispec.io.util.create_camword` can take the list of cameras and produce a camword.

I.e. this could be a one-liner:
`campets = create_camword(cameras_from_raw_data(raw_fn))`

PR #2019, opened 2023-03-06
fix documentation tests on github

sbailey:
Recent PRs are failing documentation build tests because of missing pytz, which is apparently new.  This PR starts with a minor formatting change to try to reproduce the problem independent of the code changes in other PRs, before I proceed with attempting to fix it.  If I don't succeed quickly, I'll call for more expert help, and apologies in advance for any github email churn from attempts on this.


sbailey:
It worked to add `pytz` to the `autodoc_mock_imports` list so that documentation can be built even if pytz isn't installed.  I didn't do the archeology of why this wasn't a problem before.

PR #2022, opened 2023-03-10
Don't crash submit_night if no laststep=all exposures for a tile

akremin:
There is a bug in `desi_run_night` when wanting to process a `LASTSTEP=skysub` exposure that has no `LASTSTEP=all` observations. This is in the new logic that only submits cumulative redshift jobs for tiles in which the night being submitted is the last night the tile was observed.

The behavior wasn't seen in Iron because we didn't process skysub exposures in Iron.

Directory `/global/cfs/cdirs/desi/users/kremin/PRs/runnight_maxnight` has log outputs when running `desi_run_night` on night 20230309. The main branch crashes (`daily_20230309-main.log`), while this branch (`daily_20230309-runnight_maxnight.log`) succeeds.


abhi0395:
Thanks @akremin and @sbailey

PR #2031, opened 2023-03-31
Dark no expdir

sbailey:
This PR updates desi_proc (desispec.scripts.proc.main) to *not* create an output exposures/NIGHT/EXPID directory for DARK and ZEROs exposures, since they only create preproc/NIGHT/EXPID output but don't have any spectral extractions to put in exposures/NIGHT/EXPID .  I also enforced UPPERCASE obstype values since that is how it was used in all but one place.

This was motivated by previous production runs having empty exposures/NIGHT/EXPID directories for the nightly 300 sec darks, leading to bookkeeping confusion about why they existed with no contents and whether that was a processing crash or harmless.

I tested with a ccdcalib job in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/temp (admittedly before the second commit standardizing UPPERCASE).  ccdcalib ran, produced expected preproc, and calibnight output, but didn't create the unnecessary exposures directory.

### DIFF
--- a/py/desispec/scripts/proc.py
+++ b/py/desispec/scripts/proc.py
@@ -159,6 +159,9 @@ def main(args=None, comm=None):
         hdr = comm.bcast(hdr, root=0)
         camhdr = comm.bcast(camhdr, root=0)
 
+    if args.obstype is not None:
+        args.obstype = args.obstype.upper()
+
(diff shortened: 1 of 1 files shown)


akremin (approved):
The changes look good and thank you for identifying the lowercase 'dark' `obstype` typo. I verified that everywhere else in this code `obstype` is expected to be uppercase.

PR #2033, opened 2023-04-17
Change CTE fitting to be offset + ramp instead of just offset.

schlafly:
This PR changes the sector offsets to instead be offsets plus ramps to address the worsening CTE on z3.

We discussed on the survey-ops call that this should instead be smarter about only doing the ramp components of the fits when the CTE is large, but I haven't tried to do the bookkeeping around that.

This does appear to improve things but it's not obviously a slam dunk and I need to push through a whole night.
image

The CTE frames make much more obvious that we need to do an offset plus a ramp:
image
Here I'm taking 20230414/176333 and doing a simple fiber flat & sky subtraction just by doing medians.  The flat fielding is from the median of the first 1200 pix in each spectrum (away from the CTE effect), and the "sky subtraction" is from the median of each wavelength after applying the flat.  Note that for the smooth flat field spectrum life is much easier and this should be fine?  Top to bottom is no CTE correction, an offset, and an offset plus a ramp.  I haven't really tried to understand the slit block pattern; presumably first/last spectra are less affected by CTE because of the absence of neighbors?  Interestingly there's an indication of that in the science data as well.

Anyway, this clearly isn't ready to merge, but I wanted to start a discussion.

### DIFF
--- a/py/desispec/sky.py
+++ b/py/desispec/sky.py
@@ -142,7 +142,7 @@ def get_sector_masks(frame):
                     amp,sector))
 
     if len(sectors) == 0:
-        return []
+        return [], [[]]
 
     psf_filename = findfile('psf', meta["NIGHT"], meta["EXPID"],
                             meta["CAMERA"])
@@ -232,7 +232,8 @@ def compute_sky_linear(
         skytpcorrfixed = skytpcorrfixed[skyfibers]
 
     skytpcorr = skytpcorrfixed.copy()
-    sectors, sectemplates = sectors
+    if sectors is not None:
+        sectors, sectemplates = sectors
 
     for iteration in range(max_iterations) :
         # the matrix A is 1/2 of the second derivative of the chi2 with respect to the parameters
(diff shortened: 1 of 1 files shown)


julienguy:
This looks promising. I think the slit bloc effect could be explained by a different response of the extracted PSF flux to a change in background level for edge fiber (less overlap with neighboring fibers). This response could be evaluated numerically as used as a guide for each fiber and wavelength. This response is however dependent on the signal level which changes the pixel weights in the extraction due to the Poisson noise. So the empirical approach you propose might be a better fit. An hybrid approach (using a deterministic response and then an empirical fit) could be used.

schlafly:
I have run the following additional nights:
20210514
20211020
20220528
20221121
20230326
20230414
These are all in $DESI_SPECTRO_REDUX/schlafly/....  @ashleyjross , it sounded like you're pretty well set up to see if something is badly broken; would you mind checking these?  Note that I am only running nightqa for them now, so if you require looking at that output, it will be an hour or two.  This also means that I haven't looked at the output at all yet!

ashleyjross:
I looked at all except for 20210514 (it crashed because it didn't find 'FAPRGRM' in the associated exposures file). I find nothing worrying. Compared to daily, BGS and LRG redshift success is improved for the older nights and for more recent nights the changes look consistent with random variations (rms ~0.1% in terms of redshift success per petal ).

On 20230326, petal 1 improves from 98.4% to 99.7% success (the 99.7% is exceptionally high).

schlafly:
@julienguy , I'm inclined to merge this as resolving the ongoing z1 issue.  Then we would want to reprocess all nights starting with 20230414, repeat QA briefly, archive, and update the MTLs.  I think we can improve things further here, but I think we want to look at those improvements with less urgency.  Do you agree?

PR #2034, opened 2023-05-02
Only add skysub exposures if doing tilenight in daily proc

akremin:
This is a one-line change to the recently merge `desi_daily_proc_manager` modifications to add tilenight support. It restricts the appending of `skysub`-only exposures to when `use_tilenight` is True to maintain the old behavior for non-tilenight runs.

I've run a short test on 20230416, which had a tile observed with two exposures but the second of which was a skysub only exposure. I've run it with current main, this PR, and this PR using `--use-tilenight` to show that I didn't break any of the new functionality. They show the desired outcome -- `use_tilenight=False` doesn't append the skysub exposure, but the `use_tilenight=True` does.

output files are in: `/global/cfs/cdirs/desi/users/kremin/PRs/tilenight_dailyproc_cori`


sbailey:
Looks good, thanks for catching this; I'm trusting your testing and merging now.

PR #2043, opened 2023-05-16
Night_qa: dark: add possibility of plots with --bkgsub-for-science

araichoor:
This PR adds the option to additionally display the dark/morningdark processed with the `--bkgsub-for-science` option for a set of cameras.
- this has come up with the follow-up of the new b8 ccd, which is the previously retired b4, which can have significant features, but such features should be corrected in the science exposures with the `--bkgsub-for-science` option;
- this would be used as a visual, qualitative test to assess that it works as expected;
- the current default behavior is to display that for the b-camera (so a row for a petal would have 4 panels: b, b with bkgsub, r, and z).

Examples generated with the current version of this PR: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/nightqa_v21/v0.
For instance for 20230512:
https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/nightqa_v21/v0/dark-20230512.pdf
Screenshot 2023-05-15 at 10 23 22 PM

**Code comments:**
- I ve added a `--dark_bkgsub_science_cameras` argument to `desi_night_qa`, which is a comma-separated list of cameras (or None); the current default is `b`, but one can for instance run with `b,r,z`:
https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/nightqa_v21/v0/dark-20230512-brz.pdf
Screenshot 2023-05-15 at 10 27 43 PM

- `night_qa` generate on-the-fly those with:
https://github.com/desihub/desispec/blob/c28428aa3ced495bf6025169e03df127f1c6ca86/py/desispec/night_qa.py#L567-L568
- the code throws an error, but keeps running:
```
ERROR:calibfinder.py:490:find_darks_in_desi_spectro_dark: Didn't find matching b8 calibration darks in $DESI_SPECTRO_DARK using default from $DESI_SPECTRO_CALIB instead
```
(if I understood correctly, it may be what @schlafly raised during today s telecon, ie that the code there is missing some calibration files; but as @julienguy said, it falls back to some default files, which should be good enough for this test);
- I re-shuffled a bit the code in `night_qa.create_dark_pdf()`, but the changes are not that significant.

**bkgsub-for-science comments:**
I ve not closely followed the development of this `--bkgsub-for-science` option, so the following may just be already known, or me mis-using or mis-interpreting; still e.g. with looking at https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/nightqa_v21/v0/dark-20230512-brz.pdf, a couple of remarks:
- I notice that the flux values are systematically offset to some positive values (see side histograms);
- the code does a very nice job to correct the features in b8; though it can add artificially new features (see e.g. r1 or r4).

### DIFF
--- a/bin/desi_night_qa
+++ b/bin/desi_night_qa
@@ -68,7 +68,7 @@ def parse(options=None):
     else:
         args = parser.parse_args(options)
     # AR handle None...
-    if args.dark_bkgsub_science_cameras == "None":
+    if args.dark_bkgsub_science_cameras.lower() == "none":
         args.dark_bkgsub_science_cameras = None
     return args
 
--- a/bin/desi_night_qa
+++ b/bin/desi_night_qa
@@ -59,7 +59,7 @@ def parse(options=None):
                         help = "comma-separated list of steps to execute (default={})".format(",".join(steps_all)))
     parser.add_argument("--nproc", type = int, default = 1, required = False,
                         help="number of parallel processes for create_dark_pdf(), create_ctedet_pdf() and create_sframesky_pdf() (default=1)")
-    parser.add_argument("--dark_bkgsub_science_cameras", type = str, default = "b", required = False,
+    parser.add_argument("--dark-bkgsub-science-cameras", type = str, default = "b", required = False,
                         help="for the dark/morningdark, comma-separated list of the cameras to be additionally processed with the --bkgsub-for-science argument (default=b)")
 
     args = None
--- a/bin/desi_night_qa
+++ b/bin/desi_night_qa
@@ -59,12 +59,17 @@ def parse(options=None):
                         help = "comma-separated list of steps to execute (default={})".format(",".join(steps_all)))
     parser.add_argument("--nproc", type = int, default = 1, required = False,
                         help="number of parallel processes for create_dark_pdf(), create_ctedet_pdf() and create_sframesky_pdf() (default=1)")
+    parser.add_argument("--dark_bkgsub_science_cameras", type = str, default = "b", required = False,
+                        help="for the dark/morningdark, comma-separated list of the cameras to be additionally processed with the --bkgsub-for-science argument (default=b)")
 
     args = None
     if options is None:
         args = parser.parse_args()
     else:
         args = parser.parse_args(options)
+    # AR handle None...
+    if args.dark_bkgsub_science_cameras == "None":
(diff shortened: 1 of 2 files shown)


bin/desi_night_qa
```diff
@@ -59,12 +59,17 @@ def parse(options=None):
                         help = "comma-separated list of steps to execute (default={})".format(",".join(steps_all)))
     parser.add_argument("--nproc", type = int, default = 1, required = False,
                         help="number of parallel processes for create_dark_pdf(), create_ctedet_pdf() and create_sframesky_pdf() (default=1)")
+    parser.add_argument("--dark_bkgsub_science_cameras", type = str, default = "b", required = False,
```
akremin, line 62: Please change the underscores to hyphens ("_" -> "-") in the command line flag.

bin/desi_night_qa
```diff
@@ -59,12 +59,17 @@ def parse(options=None):
...
+                        help="for the dark/morningdark, comma-separated list of the cameras to be additionally processed with the --bkgsub-for-science argument (default=b)")
 
     args = None
     if options is None:
         args = parser.parse_args()
     else:
         args = parser.parse_args(options)
+    # AR handle None...
+    if args.dark_bkgsub_science_cameras == "None":
```
akremin, line 71: May I suggest the following instead to be more robust?
`if args.dark_bkgsub_science_cameras.lower() == "none":`

akremin (commented):
I've made two superficial requests to make the command line interfacing a bit smoother. I'll trust your testing for the code, as this is only used for monitoring and not for any pipeline data products or decisions. Happy to approve this after the small changes are made.

akremin:
Hi @araichoor , pinging you here. I had two minor requests before merging this.

bin/desi_night_qa
araichoor: done.

araichoor:
@akremin : sorry for the delay here...
thanks for the review and the suggestions, which I ve just implemented.

note: I see there s also this PR on the night_qa: https://github.com/desihub/desispec/pull/2066
I ve not checked if there s any conflict between those...

akremin:
Thank you! In looking at that PR I don't think it will be an issue. There may be a merge conflict in the function declaration, but it should be an easy resolution.

I'll merge this now.

Issue #2041, opened 2023-05-11
add retries for submit_batch_script -> queue_info_from_qids -> sacct


sbailey:
Due to a bug in slurm job dependency tracking, currently `desispec.workflow.procfuncs.submit_batch_script` calls `desispec.workflow.queue.queue_info_from_qids` to filter already-completed jobs from the job dependency list before calling sbatch.

Sometimes that fails due to transient sacct failures, which causes `desi_daily_proc_manager` to crash.  I suggest that we try twice before giving up (similar to retries for sbatch).  I think that's a simple addition but since I don't have time to implement it right now I'm filing a ticket as a place holder.

PR #2044, opened 2023-05-18
Restore dry-run and add sacct retries

### LINKED ISSUE #2038: Dry run mode functionality for pipeline scripts
A recent change to query the Slurm scheduler prior to job submission will break the ability to run `desi_daily_proc_manager` and `desi_run_night` on a machine that doesn't have a Slurm scheduler.

The simple solution is to propagate `dry_run` to `queue_info_from_qids`, but then that would require updating the `dry_run` mode for that function to only return the requested set of columns, and with the requested Slurm jobid's. Currently, it returns a copied table from several years ago with outdated jobid's and the default set of columns.

This raises the large question of what we want `dry_run` to do.  It used to operate in the strict sense of not interacting with Slurm. But perhaps we do want it to get information from Slurm so that its mock actions are realistic to the current state of jobs, rather than being more self-contained but less realistic.

marcelo-alvarez:
Fixes #2038 and #2041. Two ongoing tests started on my own account initiated with 
```
source /global/cfs/cdirs/desi/users/malvarez/testscripts/sacct-dryrun.sh 1
```
and
```
source /global/cfs/cdirs/desi/users/malvarez/testscripts/sacct-dryrun.sh
```
to run the daily processing in separate SPECPROD directories,
```
/global/cfs/cdirs/desi/spectro/redux/sacct-dryrun-dryrunlevel-1
```
and
```
/global/cfs/cdirs/desi/spectro/redux/sacct-dryrun-dryrunlevel-0
```
with and without the `--dry-run-level 1` argument added to the `desi_daily_proc_manager` command, respectively, are proceeding normally on tonight's data, with the ccdcalib job (jobid 8951789) submitted as of this posting.

@sbailey please do any additional checks and merge when ready. Thanks.

### DIFF
--- a/py/desispec/workflow/procfuncs.py
+++ b/py/desispec/workflow/procfuncs.py
@@ -448,7 +448,7 @@ def submit_batch_script(prow, dry_run=0, reservation=None, strictly_successful=F
 
     # workaround for sbatch --dependency bug not tracking completed jobs correctly
     # see NERSC TICKET INC0203024
-    if len(dep_qids) > 0:
+    if len(dep_qids) > 0 and not dry_run:
         dep_table = queue_info_from_qids(np.asarray(dep_qids), columns='jobid,state')
         for row in dep_table:
             if row['STATE'] == 'COMPLETED':
--- a/py/desispec/workflow/queue.py
+++ b/py/desispec/workflow/queue.py
@@ -8,7 +8,7 @@ import numpy as np
 from astropy.table import Table
 import subprocess
 from desiutil.log import get_logger
-import time
+import time, datetime
 
 
 def get_resubmission_states():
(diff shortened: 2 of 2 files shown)


marcelo-alvarez:
@sbailey tests ran successfully overnight including calib processing in
```
/global/cfs/cdirs/desi/users/malvarez/spectro/redux/desispec-tests/sacct-dryrun-dryrunlevel-0
```
The log shows sacct was called 156 times over the night, which is the expected number given the 300-second sleep time between checking for new exposures and the amount of time it was running. The test with `--dry-run-level 1` generated all the expected slurm files without any error, i.e.:
```
% ls -rlt /global/cfs/cdirs/desi/users/malvarez/spectro/redux/desispec-tests/sacct-dryrun-dryrunlevel-1/run/scripts/night/20230517/*.slurm | wc -l
21
% ls -rlt /global/cfs/cdirs/desi/spectro/redux/daily/run/scripts/night/20230517/*.slurm | wc -l
21
```
The changes appear correct to me and have passed these tests.

sbailey (approved):

Issue #2051, opened 2023-05-23
desi_run_night for test production


julienguy:
I would like to run one tile to test a desispec branch.
I made symbolic links to calibnight exposure_tables directories  in my SPECPROD directory pointing to daily.
I then run on perlmutter:
```
desi_run_night --tiles 7471 --night 20230512 --all-tiles
```
and gets:
```
Processing the following obstypes: ['bias', 'dark', 'arc', 'flat', 'science', 'twilight', 'sci', 'dither']
Redshift fitting with redshift group types: ['perexp', 'pernight']
Processing exposures with the following LASTSTEP's: ['all']
INFO:tableio.py:334:load_table: Found table: /global/cfs/cdirs/desi/spectro/redux/jguy/exposure_tables/202305/exposure_table_20230512.csv
INFO:tableio.py:337:load_table: Table /global/cfs/cdirs/desi/spectro/redux/jguy/processing_tables/processing_table_jguy-20230512.csv not found, creating new table of type proctable
INFO:submit_night.py:190:submit_night: Filtering by tiles=[7471]


##################### 180299 #########################

Found: {'EXPID': 180299, 'OBSTYPE': 'science', 'TILEID': 7471, 'LASTSTEP': 'all', 'CAMWORD': 'a0123456789', 'BADCAMWORD': '', 'BADAMPS': '', 'EXPTIME': 1073.4472, 'EFFTIME_ETC': 1009.187195, 'SURVEY': 'main', 'FA_SURV': 'main', 'FAPRGRM': 'dark', 'GOALTIME': 1000.0, 'GOALTYPE': 'dark', 'EBVFAC': 1.07626653729032, 'AIRMASS': 1.059487, 'SPEED': 1.2048918743252015, 'TARGTRA': 237.013808, 'TARGTDEC': 15.93793, 'SEQNUM': 1, 'SEQTOT': 1, 'PROGRAM': 'dark', 'PURPOSE': 'main survey', 'MJD-OBS': 60077.360032021, 'NIGHT': 20230512, 'HEADERERR': array([], dtype=object), 'EXPFLAG': array([], dtype=object), 'COMMENTS': array([], dtype=object)}


Sleeping 1s to slow down the queue submission rate
Resuming...


Completed submission of exposures for night 20230512. 
```

But no job is actually started or even written.

akremin:
I can help debug this but before I do can you run the following test?

First please remove any processing_table for the night you're running in your test SPECPROD.

[...]

PR #2056, opened 2023-05-31
Perform final processing check in desi_run_night when processing table is empty

### LINKED ISSUE #1889: desi_run_night --use-tilenight doesn't work if only one science exposure
`desi_run_night --use-tilenight ...` doesn't work if a night has a single science exposure with no calibrations preceeding it.  This is admittedly an artificial case that I don't think we have in real data, but it is useful for testing (and I think would be triggered if trying to run a single tile while exclusing cals).  Examples in test prod `/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/tilenight`, dry-run "submitted" with
```
for NIGHT in 20221021 20221025 20221026; do
    desi_run_night --use-tilenight -q regular --z-submit-types None --all-tiles --dry-run-level 1 -n $NIGHT &> run/logs/$NIGHT.log
done
```

Cases:
1. 20221021 : single science exposure, preceded by cals<br>**result**: that tile is submitted
2. 20221025: single science exposure and nothing else<br>**result**: that tile is *not* submitted
3. 20221026: two science exposures and nothing else<br>**result**: both are submitted

The second case (single tile not getting submitted) is because desispec.scripts.submit_night only goes though a single iteration of the etable loop and line 349 never gets a chance to have `lasttype` not None:
```
        if lasttype is not None and ((curtype != lasttype) or (curtile != lasttile)):
            ....
            submit_tilenight_and_redshifts(...)
```

I haven't figured out case 3 yet (given that case 2 was dropping a tile, I was expecting case 3 to drop one tile and submit the other...).

@marcelo-alvarez @akremin please take a look.  I'm assigning this to Himalayas and Iron in case it is a symptom of a more significant corner case that could affect real data.  If we confirm that it only applies to tests cases, we could remove it from those projects (but I'd still like to get it fixed sooner than later to support test runs of a single tile).

marcelo-alvarez:
Fixes #1889 and #2051 by performing the final processing check even if processing table is empty, in order to process single exposures / tiles on a given night, for which there are no previous entries in the processing table, by definition.

I have checked this for the tile and night from #2051, respectively, using
```
desi_run_night --night 20230512 --dry-run-level 1 --z-submit-types=cumulative --append-to-proc-table --all-tiles --tiles=7471
```
and
```
desi_run_night --night 20230512 --dry-run-level 1 --z-submit-types=cumulative --append-to-proc-table --all-tiles
```
with results in
```
/global/cfs/cdirs/desi/users/malvarez/spectro/redux/desispec-2051/single-tile-20230512-7471
```
and
```
/global/cfs/cdirs/desi/users/malvarez/spectro/redux/desispec-2051/single-tile-20230512-all
```
In particular, the former contains a tilenight script for tile=7471 (showing the problem is fixed), while the latter directory contains one less flat (i.e. missing the 1-second flat that is only done by default for daily) and the same number of arc, psfnight, nightlyflat, ccdcalib, and tilenight slurm files as in daily for that night (indicating that the change does not affect job submission for a full night). Thanks @julienguy for posting #2051 which reminded me about this.

@sbailey please take a look and merge if appropriate, thanks.

### DIFF
--- a/py/desispec/scripts/submit_night.py
+++ b/py/desispec/scripts/submit_night.py
@@ -490,7 +490,7 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
         sys.stdout.flush()
         sys.stderr.flush()
 
-    if tableng > 0:
+    if tableng >= 0:
         ## No more data coming in, so do bottleneck steps if any apply
 
         # don't submit cumulative redshifts for lasttile if it isn't in tiles_cumulative
--- a/py/desispec/scripts/submit_night.py
+++ b/py/desispec/scripts/submit_night.py
@@ -490,7 +490,7 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
         sys.stdout.flush()
         sys.stderr.flush()
 
-    if tableng >= 0:
+    if tableng > 0 or (use_tilenight and len(sciences)>0):
         ## No more data coming in, so do bottleneck steps if any apply
 
         # don't submit cumulative redshifts for lasttile if it isn't in tiles_cumulative


py/desispec/scripts/submit_night.py
```diff
@@ -490,7 +490,7 @@ def submit_night(night, proc_obstypes=None, z_submit_types=None, queue='realtime
         sys.stdout.flush()
         sys.stderr.flush()
 
-    if tableng > 0:
+    if tableng >= 0:
```
akremin, line 493: This is equivalent to removing the `if` statement, which is there to protect against cases where there genuinely is no data on a given night. 

I'd recommend modifying the `if` statement instead:

`if tableng > 0:`   -->  `if tableng > 0 or (use_tilenight and len(sciences)>0):`

An alternate solution would be to ensure all calling functions are robust to empty arguments (empty `sciences`  passed to `checkfor_and_submit_joint_job`, no entries in `ptable` passed to `update_from_queue`, etc.

The third option would be to throw an error there if there is no good data rather than having the code crash. I slightly disfavor this, but am open to discussion.

Personally, I think the first option is the least invasive until we officially retire the non tilenight code in the script rewrite.

akremin (changes requested):
Thanks Marcelo. While this resolves the issue in question, it will cause the code to crash in other instances. I have a proposed alternative in my inline comment. I'm happy to discuss this more if you would like.

For context to others: This portion of the code was originally meant to perform joint fits. The issue being addressed arose in the tilenight version because the science exposures aren't immediately submitted and therefore aren't registered in the processing table prior to this step.

py/desispec/scripts/submit_night.py
marcelo-alvarez: 1 done

akremin (approved):
Thank you, the changes look good.

Issue #1980, opened 2023-01-24
arc jobs not cleanly recovering from mid-job failures


sbailey:
Iron 20211001 arc expid 102547 initially failed due to a transient I/O failure reading input preproc files; see
iron/run/scripts/night/20211001/arc-20211001-00102547-a0123456789-4852456.log, e.g.
```
ERROR:specex.py:338:fitframe: FAILED: MPI ranks 161-180 on camera z1
ERROR:specex.py:340:fitframe: FAILED: desi_compute_psf --input-image /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/preproc/20211001/00102547/preproc-z1-00102547.fits.gz --input-psf /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/exposures/20211001/00102547/shifted-input-psf-z1-00102547.fits --output-psf /global/cfs/cdirs/desi/spectro/redux/iron/exposures/20211001/00102547/fit-psf-z1-00102547.fits
ERROR:specex.py:341:fitframe: [Errno 5] Input/output error: '/dvs_ro/cfs/cdirs/desi/spectro/redux/iron/preproc/20211001/00102547/preproc-z1-00102547.fits.gz'
```

Resubmitting the job didn't cleanly recover, and failed with different errors in the second log run/scripts/night/20211001/arc-20211001-00102547-a0123456789-4868394.log, e.g.
```
ERROR:specex.py:323:fitframe: FAILED: commands for camera r6 not found for MPI group rank 15 and world rank 276
```

It appears that all of the successful PSFs from the first iteration confused the MPI rank bookkeeping of the second submission, causing it to report failure even though it had succeeded in writing the previously failing PSFs.

For the purposes of Iron I need to cleanup the data files and re-run, but the above logs will still be there.

sbailey:
@marcelo-alvarez I think the test would be:
1. take the outputs from a previously successful arc exposure processing
2. remove one of the psf files
3. rerun the arc*.slurm script

My understanding of the previous failure was that when the script woke up in step (3) and allocated work to ranks, but the presence of a partial set of successful PSFs confused the bookkeeping leading to a fake failure.

marcelo-alvarez:
@sbailey thanks, it's clear now.

PR #2059, opened 2023-06-01
Do not report failure when skipping previously-written PSF fits

marcelo-alvarez:
Solves #1980 by not increasing the error count on ranks corresponding to already-completed PSF fits written by a previous job. Instead, a message is written to the log noting that there is nothing to do for each such rank and the arc job returns the same error code as it would have if those ranks had completed those fits.

I tested this by:
1. taking the outputs from a previously successful arc exposure (EXPID) processing
2. removing one of the psf files (i.e. fit-psf-z0-EXPID.fits)
3. rerunning the arc*.slurm script

The resulting log files are in 
```
/global/cfs/cdirs/desi/users/malvarez/spectro/redux/desispec-spxmidjobfail/specex-midjobfail-20230512-all/run/scripts/night/20230512/arc-20230512-00180239-a0123456789-9829453.log
```
and
```
/global/cfs/cdirs/desi/users/malvarez/spectro/redux/desispec-spxmidjobfail/main-20230512-all/run/scripts/night/20230512/arc-20230512-00180239-a0123456789-9829449.log
```
for this branch and main, respectively. Both main and this branch successfully generated the missing psf file, but only the latter returned success.

@sbailey please have a look and merge if appropriate, thanks.

### DIFF
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -270,9 +270,9 @@ def run(comm,cmds,cameras):
                  values being the 'desi_compute_psf ...' string that one would run
                  on the command line.
         cameras: list of camera strings identifying the entries in cmds to be run
-                 as jobs in parallel jobs, one entry per ccd image to be fit; entries
-                 not in the dictionary will be logged as an error while still
-                 continuing with the others and not crashing.
+                 as jobs in parallel jobs, one entry per ccd image to be fit.
+                 Processes assigned to cameras not present as keys in cmds will
+                 write a message to the log instead of running a PSF fit.
 
     The function first defines the procedure to call specex for a given ccd image
     with the "fitframe" inline function, passes the fitframe function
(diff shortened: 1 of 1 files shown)


akremin (commented):
Thanks Marcelo, the changes look good and thank you for the tests. 

Can you comment more on why it used to throw an error? Is it ever true that an error can occur such that cmd won't have a camera in it's keys, and run() should be logging an error?

Following the git history it looks like this "if camera not in cmds" was introduced in a refactor you did, which is why I ask. If the original error was a misunderstanding, and there is never an error, then I'm happy with this solution to the current issue.

marcelo-alvarez:
Hi @akremin, thanks for the comment and question. I don't know about whether the original error was a misunderstanding or if it is every true that an error can occur such that cmd won't have a camera in it's keys. @sbailey or @julienguy  might be able to answer that.

marcelo-alvarez:
@sbailey do you have any insight on the question of whether the original "if camera not in cmds" error I introduced in a refactor was due to a misunderstanding because there is never an error? I don't have enough context on the overall use cases for specex to say whether this was a misunderstanding because such an error never happens or rather just because it caused the situation that this PR attempts to fix. Thanks.

sbailey:
I think the previous behavior was just a misunderstanding.  Although it could be reasonable to consider it an error if you ask the code to do something that is already done, elsewhere in desispec we treat "already done" as a non-error so that the code just skips anything that is already done and completes anything that is left to do.  This makes it easier to recover from system-induced failures by just re-running the identical command again without having to edit it down to only the cameras left to do (a very early version of the pipeline required editing commands to recover from failures and that was a pain...).  Merging now.

PR #2060, opened 2023-06-02
use the latest secondary target catalog in gather_targetdirs

moustakas:
[WIP]

In `sv1`, the secondary target directory recorded in the fiberassign headers occasionally does not match what was used to generate the actual fiberassign catalogs, e.g., `/global/cfs/cdirs/desi/target/fiberassign/tiles/trunk/080/fiberassign-080734.fits.gz` records `/global/cfs/cdirs/desi/target/catalogs/dr9/0.50.0/targets/sv1/resolve/bright`, but the file in this directory does not have any *grz* fluxes.

In fact, the correct directory is `/global/cfs/cdirs/desi/target/catalogs/dr9/0.52.0/targets/sv1/resolve/bright`.

This PR captures this corner case and addresses an inconsistency for 1532 targets in the `lsdr9-photometry` VAC for Fuji.

### DIFF
--- a/py/desispec/io/photo.py
+++ b/py/desispec/io/photo.py
@@ -114,6 +114,9 @@ def gather_targetdirs(tileid, fiberassign_dir=None, verbose=False):
         fn = fn.replace(
             "/data/afternoon_planning/surveyops/trunk", os.getenv("DESI_SURVEYOPS")
         )
+        # use the last / latest version of the catalogs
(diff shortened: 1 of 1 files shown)


moustakas:
Per @araichoor's recommendation, we're not going to merge this PR. The differences I was seeing was due to row-mixing issue which was patched in later (post-Fuji) versions of the fiberassign catalogs.

Issue #2055, opened 2023-05-30
STD_FIBER_RA needs to be multiplied by cos(dec), not divided by it


stephjuneau:
Sergey Koposov noticed that STD_FIBER_RA should be multiplied by cos(dec) and not divided by it. The screenshot demonstrates that indeed the values are as expected when multiplied by `np.cos(np.radians(MEAN_FIBER_DEC))` (i.e., comparable to STD_FIBER_DEC). The current implementation results in artificially larger STD_FIBER_RA values as shown in the bottom panel.

Question: does this arise elsewhere besides in `desispec.coaddition.coadd_fibermap()`?

<img width="444" alt="image" src="https://github.com/desihub/desispec/assets/26126576/6bb62674-5596-406f-9a6e-220c16f44f69">

Issue #2054, opened 2023-05-30
Aggregate quantities from coadd_fibermap() not based on "good_coadds"


stephjuneau:
While desispec.coaddition.coadd_fibermap() correctly identifies the subset of `good_coadds`, the aggregate quantities (MEAN, STD, RMS) are computed over all exposures and not just the subset of good ones. An example is `TARGETID==39628525115805322` (screenshot).

<img width="1526" alt="image" src="https://github.com/desihub/desispec/assets/26126576/26cccdeb-27c6-4e24-a97a-5b36eae3c988">

Issue #2046, opened 2023-05-19
MEAN_FIBER_RA,MEAN_FIBER_DEC: mean computed with some non-valid values


araichoor:
Copy-pasting https://desisurvey.slack.com/archives/C2AT2NEHY/p1684528715042649:

the issue is that in some exposures, `FIBER_{RA,DEC}=0` for whatever reason, and apparently `MEAN_FIBER_{RA,DEC}` uses those in the computation, hence ending with ~non-sense values..
example with one fuji file:
```
fn = "/global/cfs/cdirs/desi/spectro/redux/fuji/zcatalog/zpix-sv3-dark.fits"
d, e = Table.read(fn, "ZCATALOG"), Table.read(fn, "EXP_FIBERMAP")
sel = (d["COADD_FIBERSTATUS"] == 0) & (np.abs(d["MEAN_FIBER_RA"] - d["TARGET_RA"]) > 1)
sel.sum()
d["TARGETID", "COADD_FIBERSTATUS", "TARGET_RA", "MEAN_FIBER_RA", "COADD_NUMEXP"][sel][:5]
```

returns:

```
4096
     TARGETID     COADD_FIBERSTATUS     TARGET_RA        MEAN_FIBER_RA    COADD_NUMEXP
      int64             int32            float64            float64          int16    
----------------- ----------------- ------------------ ------------------ ------------
39627805364850503                 0 150.17318166142363 137.65874163244868           12
39627811404647604                 0 150.15253184951817 131.38347189564968            8
39627805369044515                 0 150.40749740772065 138.83769495536356           13
39627811408841765                 0 150.39695082153526 125.33080368119363            6
39627805369042455                 0  150.3238465769502 120.25908829660841            5
```

and if I take as example the first targetid `39627805364850503`, `expid=87354` is the culprit.
the ra is 11/12*150~137.
```
tid = d["TARGETID"][sel][0]
e[e["TARGETID"] == tid]["TARGETID", "FIBERSTATUS", "PLATE_RA", "FIBER_RA", "EXPID", "NIGHT", "TILEID", "FIBER"]

     TARGETID     FIBERSTATUS      PLATE_RA           FIBER_RA      EXPID  NIGHT   TILEID FIBER
      int64          int32         float64            float64       int32  int32   int32  int32
----------------- ----------- ------------------ ------------------ ----- -------- ------ -----
39627805364850503           0 150.17318166142363 150.17318168112703 83714 20210406      1   153
[...]

PR #2065, opened 2023-06-06
fixed cos(dec), good_coadds in aggregate columns, missing fiber coords

stephjuneau:
This PR fixes three bugs/issues in coadd_fibermap():

- https://github.com/desihub/desispec/issues/2055 cos(dec) term is now multiplied;
- https://github.com/desihub/desispec/issues/2054 the aggregate quantities `{MEAN|STD}_FIBER_{RA|DEC}, {MEAN|RMS}_DELTA_{X|Y}, MEAN_PSF_TO_FIBER_SPECFLUX` are now computed over only the `good_coadds` when available (len>0), otherwise over the "bad" coadds;
- https://github.com/desihub/desispec/issues/2046 the `{MEAN|STD}_FIBER_{RA|DEC}` are now computed on the subset with "good" coordinates (i.e., avoiding missing cases with `FIBER_RA=FIBER_DEC=0`

Example test cases for Healpix coadd and for Tile (cumulative) coadd: 
```
import numpy as np
from astropy.table import Table, setdiff
from desispec import coaddition

# Select here which type of coadd to test (remainder is identical)
test_type = 'healpix'
#test_type = 'cumulative'

if test_type=='healpix':

    # Example test case for Bug Fixes
    survey = 'sv1'
    program = 'bright'
    healpix = 22960
    hpxgrp = healpix//100
    
    # Example TARGETID to print some test results below
    test_id = 39627798029009428

    path = f"/global/cfs/cdirs/desi/spectro/redux/fuji/healpix/{survey}/{program}/{hpxgrp}/{healpix}/"

    specfile = path+f"spectra-{survey}-{program}-{healpix}.fits"
    coaddfile = path+f"coadd-{survey}-{program}-{healpix}.fits"

if test_type=='cumulative':

    # Example test case for Bug Fixes
    survey = 'sv1'
    program = 'bright' #dark
    tileid = 80639
    night = 20210224
    petal = 0
    
    # Example TARGETID to print some test results below
    test_id = 39627731578654162

    path = f"/global/cfs/cdirs/desi/spectro/redux/fuji/tiles/cumulative/{tileid}/{night}/"

    specfile = path+f"spectra-{petal}-{tileid}-thru{night}.fits"
    coaddfile = path+f"coadd-{petal}-{tileid}-thru{night}.fits"

# Input fibermap
fibermap_input = Table.read(specfile, hdu=1)

# check if this example has missing coordinates!
missing_coords = (fibermap_input['FIBER_RA']==0)&(fibermap_input['FIBER_DEC']==0)

print(f"N(total) = {len(fibermap_input)}")
print(f"N(missing FIBER coordinates) = {len(fibermap_input[missing_coords])}")
print(f"N(good FIBER coordinates) = {len(fibermap_input[~missing_coords])}")

# Run fixed version of coadd_fibermap
tfmap, exp_fibermap = coaddition.coadd_fibermap(fibermap_input)

# coadded fibermap (on disk)
c_fm = Table.read(coaddfile, hdu=1)
# exp_fibermap (on disk)
e_fm = Table.read(coaddfile, hdu=2)

# The EXP_FIBERMAP should be identical to the extension already on disk
test = setdiff(exp_fibermap, e_fm)

# Check that length is 0!
print(f"Check that EXP_FIBERMAP is identical: {len(test)==0}")

# Example TARGETID selected to showcase at least one missing FIBER coordinates
tid = test_id

exp_cols = ['TARGETID','FIBERSTATUS','FIBER_RA','FIBER_DEC','DELTA_X','DELTA_Y', 'PSF_TO_FIBER_SPECFLUX']
cols = ['TARGETID','MEAN_FIBER_RA', 'STD_FIBER_RA', 'MEAN_FIBER_DEC', 'STD_FIBER_DEC',\
        'MEAN_DELTA_X', 'RMS_DELTA_X', 'MEAN_DELTA_Y', 'RMS_DELTA_Y', 'MEAN_PSF_TO_FIBER_SPECFLUX']
        
print(fibermap_input[exp_cols][fibermap_input['TARGETID']==tid])

# From coadd file on disk (Fuji) with *WRONG VALUES* due to using all fiberstatuses
print(c_fm[cols][c_fm['TARGETID']==tid])

# New coadd fibermap with *CORRECT VALUES* from bug fixes
print(tfmap[cols][tfmap['TARGETID']==tid])

```

### DIFF
--- a/py/desispec/coaddition.py
+++ b/py/desispec/coaddition.py
@@ -181,6 +181,9 @@ def coadd_fibermap(fibermap, onetile=False):
         'FIBER_RA', 'FIBER_DEC',
         'PSF_TO_FIBER_SPECFLUX',
         ]
+    # Note: treat the fiber coordinates separately because of missing coordinate problem
+    # that require an additional "good_coords" condition relative to other mean cols
+
     if onetile:
         mean_cols.extend(['FIBER_X', 'FIBER_Y'])
 
@@ -236,13 +239,16 @@ def coadd_fibermap(fibermap, onetile=False):
     else :
          raise KeyError("no FIBERSTATUS nor COADD_FIBERSTATUS column in fibermap")
 
+    # SJ: for verbose testing (commenting out)
+    #print(f"Will loop over N={len(targets)} targets")
+
     for i,tid in enumerate(targets) :
         jj = fibermap["TARGETID"]==tid
 
         #- coadded FIBERSTATUS = bitwise AND of input FIBERSTATUS
         tfmap['COADD_FIBERSTATUS'][i] = np.bitwise_and.reduce(fibermap[fiberstatus_key][jj])
 
-        #- Only FIBERSTATUS=0 were included in the coadd
+        #- Only a subset of "good" FIBERSTATUS flags are included in the coadd
         fiberstatus_nonamp_bits = get_all_nonamp_fiberbitmask_val()
         fiberstatus_amp_bits = get_justamps_fiberbitmask()
         targ_fibstatuses = fibermap[fiberstatus_key][jj]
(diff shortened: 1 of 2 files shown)


py/desispec/coaddition.py
```diff
@@ -236,13 +239,16 @@ def coadd_fibermap(fibermap, onetile=False):
     else :
          raise KeyError("no FIBERSTATUS nor COADD_FIBERSTATUS column in fibermap")
 
+    # SJ: for verbose testing (commenting out)
+    #print(f"Will loop over N={len(targets)} targets")
```
sbailey, line 243: Please remove these debugging prints before we merge.

py/desispec/coaddition.py
```diff
@@ -251,41 +257,64 @@ def coadd_fibermap(fibermap, onetile=False):
         good_coadds = np.bitwise_not( nonamp_fiberstatus_flagged | allamps_flagged )
         tfmap['COADD_NUMEXP'][i] = np.count_nonzero(good_coadds)
 
+        #- For FIBER_RA/DEC quantities, only average over good coordinates
+        #  There is a bug that some "missing" coordinates were set to FIBER_RA=FIBER_DEC=0
+        #  (we are assuming there are not valid targets at exactly 0,0; only missing coords)
+        if 'FIBER_RA' in fibermap.colnames:
+            good_coords = (fibermap['FIBER_RA'][jj]!=0)|(fibermap['FIBER_RA'][jj]!=0)
```
sbailey, line 264: typo: FIBER_RA listed twice instead of FIBER_RA and FIBER_DEC

py/desispec/coaddition.py
```diff
@@ -297,10 +326,13 @@ def coadd_fibermap(fibermap, onetile=False):
         #         tfmap['LAST_'+k][i] = np.max(vals)
         #         tfmap['NUM_'+k][i] = np.unique(vals).size
 
+        #SJ: Not the correct formula for IVAR (should be inverse of sum of variances)
+        # because these are not for inverse variance weighted quantities.
+        # Those values are not used right now but should be fixed if used later
```
sbailey, line 331: Good catch!  even though we don't currently have these FIBER_RA/DEC_IVAR and DELTA_X/Y_IVAR columns, let's go ahead and fix this now as well.  I think the correction would be:
```python
tfmap[k][i] = 1/np.sum(1/fibermap[k][jj][compute_coadds])
```
Please double check that.  If one of the input ivars was 0 (infinite uncertainty), this will result in a numpy inf warning and an output ivar of 0, but I think that is what we want.  Do you agree, or otherwise what do you suggest?

sbailey (commented):
Thanks @stephjuneau.  I'd like to do some more unit-testing on various corner cases, but I put some initial comments inline.

py/desispec/coaddition.py
stephjuneau: I guess if all IVARs are zero, I'd leave zero as the combined value (all invalid/masked). And if some are non-zero, I'd apply the formula to that subset. In an ideal world, the same subset with non-zero IVAR would also be combined (MEAN, STD) but by default one could make an assumption that those IVAR=0 cases would **not** be part of the `good_coadds` condition.

stephjuneau: I'll test it in case it already works with the formula before trying to implement something more sophisticated.

py/desispec/coaddition.py
stephjuneau: Done.

py/desispec/coaddition.py
stephjuneau: Fixed.

py/desispec/coaddition.py
stephjuneau: I updated the comments and the formula to use the MEAN because I'm assuming that this would be error propagation on the mean values of those columns. I tested that one or more of the IVAR values being zero results in the output to be zero (with a dividing by zero warning).

stephjuneau:
@sbailey I made a new commit that fixes your three comments from your first review.

sbailey:
Thanks for the updates.  I added a variety of unit tests to check cases of coadding fibermaps with various FIBERSTATUS bits set.  Many of those tests fail on main but now pass on this branch.  Great!  The one failure I found was the pathological case where all inputs have TARGET_RA=TARGET_DEC=FIBER_RA=FIBER_DEC=0, which results in an index error:
```
import numpy as np
from desispec.coaddition import coadd_fibermap

nspec = 10
fm = Table()
fm['TARGETID'] = 111 * np.ones(nspec, dtype=int)
fm['DESI_TARGET'] = 4 * np.ones(nspec, dtype=int)
fm['TILEID'] = np.ones(nspec, dtype=int)
fm['NIGHT'] = 20201220 * np.ones(nspec, dtype=int)
fm['EXPID'] = np.arange(nspec, dtype=int)
fm['FIBER'] = np.ones(nspec, dtype=int)
fm['FIBERSTATUS'] = np.zeros(nspec, dtype=int)
fm['TARGET_RA'] = np.zeros(nspec, dtype=float)
fm['TARGET_DEC'] = np.zeros(nspec, dtype=float)
fm['FIBER_RA'] = np.zeros(nspec, dtype=float)
fm['FIBER_DEC'] = np.zeros(nspec, dtype=float)

cofm, expfm = coadd_fibermap(fm, onetile=True)

...
File /global/common/software/desi/users/sjbailey/desispec/py/desispec/coaddition.py:299, in coadd_fibermap(fibermap, onetile)
    296 #SJ: Check STD_FIBER_MAP with +360 value; doesn't do the wrap-around correctly
    297 #- STD of FIBER_RA, FIBER_DEC in arcsec, handling cos(dec) and RA wrap
    298 if 'FIBER_RA' in fibermap.colnames:
--> 299     dec = fibermap['TARGET_DEC'][jj][compute_coadds&good_coords][0]
    300     vals = fibermap['FIBER_RA'][jj][compute_coadds&good_coords]
    301     std = np.std(vals+360.0) * 3600 * np.cos(np.radians(dec))

File astropy/table/_column_mixins.pyx:81, in astropy.table._column_mixins._ColumnGetitemShim.__getitem__()

File astropy/table/_column_mixins.pyx:64, in astropy.table._column_mixins.base_getitem()

File astropy/table/_column_mixins.pyx:76, in astropy.table._column_mixins.column_getitem()

IndexError: index 0 is out of bounds for axis 0 with size 0
```

While that is pathological input, we should do something more graceful than crash (or at least crash with an informative log message instead of an IndexError).

I think you need to add something for `good_coords` equivalent to what you did for `good_coadds`:
```
        if np.count_nonzero(good_coadds)>0:
            compute_coadds = good_coadds
        else:
            compute_coadds = ~good_coadds
```
please investigate and add some patch for the case of all inputs being bad coordinates.

stephjuneau:
@sbailey  Alright, this is no longer throwing an error when all FIBER coords are bad for an object. I tested it by using a real input Fibermap but then setting all the FIBER_RA.DEC values to 0.

```
# Fake Stephen's scenario will all bad coords for the test_id
fibermap_input['FIBER_RA'][fibermap_input['TARGETID']==test_id] = 0
fibermap_input['FIBER_DEC'][fibermap_input['TARGETID']==test_id] = 0
```

It returns that MEAN_FIBER_RA/DEC = 0.0 for that object in the output coadded fibermap.

sbailey:
Looks good, thanks!  Merging now.

PR #2066, opened 2023-06-06
Update night_qa.py

Waelthus:
This will keep the default behaviour of nightqa as is, but allows creating the dark_pdfs without re-preprocessing even if no proctable is there. I'd like this behaviour for plotting dark frames based on new dark models before they are added to the main reduction.


akremin:
Sorry Michael, but there was another PR of nightqa that I just merged that caused a conflict with this new PR. Would you be willing to do one of the following?

Your PR is a simple enough that you could either make them again off of master in a new PR or do the conflict resolution. Just please be careful not to negate anything from the current main branch if you decide to resolve them here.

araichoor:
thanks @Waelthus.
looks good to me.
one minor comment: could you add in the docstr this new `run_preproc` argument?

araichoor:
thanks @sbailey; on my side, it s good to merge!

PR #2073, opened 2023-06-20
Force QA string field to have a minimum length.

schlafly:
This PR ~hacks the QA column to be 20 characters long on read in in desi_tile_vi.  This is intended to resolve issues like
https://github.com/desihub/desisurveyops/issues/125
where all the max length QA status was 4 characters long, making new unsure tiles get truncated down to 'unsu' tiles.

This PR's fix is pretty marginal, reading the file in with Table.read(...) as usual and then pulling out the 'QA' column and changing its type to be longer.

It feels to me like the "correct" solution would involve telling the ecsv file to use a longer datatype for the string.  But AFAICT the ecsv string data type is just string, and can handle arbitrary length on read-in and write out.  The problem is that it then gets read into a fixed length structure by astropy/python.

I'm surprised that we don't get bitten by this more often.  e.g., I think I'd expect similar issues if we made a new FAPRGRM that needed to get added to this file?  So maybe I'm missing something obvious.  Anyway, this addresses the immediate problem but I am interested in better solutions.

### DIFF
--- a/bin/desi_tile_vi
+++ b/bin/desi_tile_vi
@@ -44,10 +44,10 @@ def parse(options=None):
                         help = 'your name or initials')
     parser.add_argument('--viewer', type = str, default = "eog", required=False,
                         help = 'image viewer (default is eog)')
-    parser.add_argument('--new', action = 'store_true', required=False,
-                        help = 'only inspect new tiles (equivalent to --qastatus none)')
-    parser.add_argument('--qastatus', type=str, default=None, required=False,
-                        help = 'only inspect tiles with this QA status (e.g. "unsure")')
+    parser.add_argument('--qastatus', type=str, default='none', required=False,
+                        help = ('only inspect tiles with this QA status '
+                                '(e.g. "unsure").  Default of none shows only new '
+                                'tiles.'))
     parser.add_argument('--survey', type = str, default = 'main', required=False,
                         help = 'look only at tiles from this survey')
     parser.add_argument('--program', type=str, nargs='+', required=False,
@@ -142,10 +142,6 @@ def main():
 
     args=parse()
 
-    if args.new and args.qastatus is not None:
-        log.error('Specify --new or --qastatus QA but not both')
-        sys.exit(1)
-
     if args.prod is None:
         args.prod = specprod_root()
     elif args.prod.find("/")<0 :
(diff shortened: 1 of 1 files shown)


schlafly:
I've made a few additional changes in this PR now.
* Delete the --new argument in favor of making --qastatus none the default.
* Print out a message when changing the status from something that isn't 'none' to something else.

The idea here is to make it harder for people to change unsure tiles without noticing it.  Previously we relied on users including the --new argument to avoid showing old unsure tiles.  Now I've removed that argument and made qastatus = none the default, essentially making this the default behavior.  One would need to explicitly run --qastatus unsure to see all unsure tiles.  The other change is that when we change the status of a tile whose status is not 'none', a message will be printed notifying the user of this.

schlafly:
Thanks Anand.  Yes, I imagine we've just been lucky here.  But it's kind of genuinely hard to do this right with the astropy table ecsv infrastructure, unless I've missed something.  So I worry about other ecsv files as well.

geordie666:
I think we (accidentally) circumvented this issue in the MTL ledgers by never overwriting data. We only ever append a new state to a file. Then the longest string in a column (in theory) never vanishes. That was a side effect, though, rather than an ideal solution.

akremin:
For what it's worth I ran into a similar issue with the exposure_table and processing_table's, which are basic csv's. My solution was pretty similar to yours in that I wrap my I/O in custom functions that ensure I have the correct data type and maximum string length for each column.
I'm not suggesting you look at that, mostly just saying I wasn't able to come up with anything more elegant either. Your solution is straight forward and does the job.

bin/desi_tile_vi
```diff
@@ -44,10 +44,10 @@ def parse(options=None):
...
     parser.add_argument('--viewer', type = str, default = "eog", required=False,
                         help = 'image viewer (default is eog)')
-    parser.add_argument('--new', action = 'store_true', required=False,
-                        help = 'only inspect new tiles (equivalent to --qastatus none)')
-    parser.add_argument('--qastatus', type=str, default=None, required=False,
-                        help = 'only inspect tiles with this QA status (e.g. "unsure")')
+    parser.add_argument('--qastatus', type=str, default='none', required=False,
+                        help = ('only inspect tiles with this QA status '
+                                '(e.g. "unsure").  Default of none shows only new '
+                                'tiles.'))
```
sbailey, line 50: It took me a minute to parse that "none" was an actual QA state, rather than a stand-in for the later test `if args.qastatus is not None` which will always be True since the default is the string 'none' (which is not None...).  Please add to this docstring all the standard states that could be selected.  It is good that you describe that the default "none" shows only new tiles, since that isn't obvious from "none" alone.

bin/desi_tile_vi
```diff
@@ -189,15 +186,13 @@ def main():
...
-        selection &= np.isin([x.lower() for x in tiles_table['FAPRGRM']],
-                             [x.lower() for x in args.program])
-    if args.qastatus is not None:
-        selection &= (tiles_table["QA"]==args.qastatus)
+        if args.survey is not None:
+            selection &= (tiles_table["SURVEY"]==args.survey)
+        if args.program is not None:
+            selection &= np.isin([x.lower() for x in tiles_table['FAPRGRM']],
+                                 [x.lower() for x in args.program])
+        if args.qastatus is not None:
```
sbailey, line 194: This test implies that `args.qastatus` could be None, which I think could never happen since the default value is not None (though it is "none", meaning new tiles without a QA status, which is confusingly similar to `None`).  Consider dropping the if test or at least adding a comment that this is a belt-and-suspenders check just in case the default later changes to `None` instead of `"none"`.

Bonus points: Since this is an opportunity for the user to supply an invalid value that would filter out all rows without it being obvious why; consider adding a safety check like
```python
keep = (tiles_table["QA"] == args.qastatus)
if np.any(keep):
    selection &= keep
else:
    print(f'ERROR: no tiles with QA={args.qastatus}; valid values are {set(tiles_table["QA"])}')
    sys.exit(1)
```

sbailey (commented):
Algorithmically this seems fine, though see comments inline about my confusion on `None` vs. `"none"` and optional suggested changes for maintainability.

schlafly:
Thanks.  I've made changes along the lines of your suggestions.  I added a new 'all' case for qastatus that duplicates the ~impossible qastatus = None behavior (though it's not clear to me that we would use that option in practice).  I am ready to merge this.

PR #2075, opened 2023-06-22
Loop over new exposures only once in desi_daily_proc_manager

marcelo-alvarez:
Adds the new command line option `--run-only-once` to  `desi_daily_proc_manager`.When `--run-only-once` is given,  the loop over new exposures happens only once, and `desi_daily_proc_manager` exits without processing the last science exposure.  Without the flag, the behavior is unchanged with respect to the main branch, except for the removal of a bug in which the last science exposure was written to the exposure table before the corresponding tilenight and ztile jobs had been submitted when `use_tilenight == True`.

I have tested this branch with the script
```
/global/cfs/cdirs/desi/users/malvarez/testscripts/desispec-2075.sh
```
that successfully submitted processing for a mock observation (by intermittently linking data to a test data directory every 3 seconds) for 20230612 that contained multiple tiles with multiple exposures each with the command
```
desi_daily_proc_manager --ignore-cori-node --raw-data-path /global/cfs/cdirs/desi/users/malvarez/mockdesiroot/spectro/data --data-cadence-time 1 --dry-run-level 1
```
(with a slightly modified version for testing with the night and end time hardcoded to 20230612 and 12pm today, respectively, and sleeps `data_cadence_time=1` when `dry_run==True`), demonstrating that the behavior without the new option is unchanged. This test prod [and log file] is in:
```
/global/cfs/cdirs/desi/users/malvarez/spectro/redux/mockdaily[/dailyproc.log]
```

I also tested this branch with the `--run-only-once` by running daily processing for 20230621 using the scrontab entry
```
*/5 * * * * /global/homes/d/desi/pdaily/bin/pdaily_dailyproc
```
such that `desi_daily_proc_manager` was run intermittently during the night with the `--run-only-once` argument before 9am, and without that argument after, with submission of processing jobs occurring as expected in an equivalent way to running persistently without the `--run-only-once` flag prior over a full night.

Because of this testing I am confident that merging this branch would not be disruptive to the daily spectroscopic pipeline and provides the capability to run intermittently over a night instead of persistently, through the `--run-only-once` option, making the pipeline more resilient to interruptions at NERSC.

@akremin @sbailey please review and merge when ready.

### DIFF
--- a/bin/desi_daily_proc_manager
+++ b/bin/desi_daily_proc_manager
@@ -69,8 +69,9 @@ def parse_args():#options=None):
                         help="Wait time between loops in checking queue statuses and resubmitting failures.")
     parser.add_argument("--override-night", type=str,default=None,
                         help="Specify the night to run on. Overrides the current day.")
-    parser.add_argument("--run-only-once", action="store_true",
-                        help="Run only once. Suitable for running repeateadly in a scron job.")
+    parser.add_argument("--loop-only-once", action="store_true",
+                        help="Loop over new exposures only once and does not process the tile corresponding to the last "+
+                             "science exposure before exiting. Suitable for running repeateadly as a scron job.")
     parser.add_argument("--ignore-expid-list", type=str,default=None,
                         help="Specify the expid's to ignore in a comma separated list given as a string.")
     parser.add_argument("--ignore-expid-file", type=str,default=None,
@@ -175,4 +176,4 @@ if __name__ == '__main__':
                              dont_check_job_outputs=args.dont_check_job_outputs,
                              dont_resubmit_partial_jobs=args.dont_resubmit_partial_jobs,
                              use_specter=args.use_specter,
-                             use_tilenight=args.use_tilenight,run_only_once=args.run_only_once)
+                             use_tilenight=args.use_tilenight,loop_only_once=args.loop_only_once)
--- a/bin/desi_daily_proc_manager
+++ b/bin/desi_daily_proc_manager
@@ -69,6 +69,8 @@ def parse_args():#options=None):
                         help="Wait time between loops in checking queue statuses and resubmitting failures.")
     parser.add_argument("--override-night", type=str,default=None,
                         help="Specify the night to run on. Overrides the current day.")
+    parser.add_argument("--run-only-once", action="store_true",
+                        help="Run only once. Suitable for running repeateadly in a scron job.")
     parser.add_argument("--ignore-expid-list", type=str,default=None,
                         help="Specify the expid's to ignore in a comma separated list given as a string.")
     parser.add_argument("--ignore-expid-file", type=str,default=None,
@@ -173,4 +175,4 @@ if __name__ == '__main__':
                              dont_check_job_outputs=args.dont_check_job_outputs,
                              dont_resubmit_partial_jobs=args.dont_resubmit_partial_jobs,
                              use_specter=args.use_specter,
-                             use_tilenight=args.use_tilenight)
+                             use_tilenight=args.use_tilenight,run_only_once=args.run_only_once)
--- a/py/desispec/scripts/daily_processing.py
+++ b/py/desispec/scripts/daily_processing.py
@@ -33,7 +33,7 @@ def daily_processing_manager(specprod=None, exp_table_path=None, proc_table_path
                              badamps=None, override_night=None, tab_filetype='csv', queue='realtime',
                              exps_to_ignore=None, data_cadence_time=300, queue_cadence_time=1800,
                              dry_run_level=0, dry_run=False, no_redshifts=False, continue_looping_debug=False, dont_check_job_outputs=False,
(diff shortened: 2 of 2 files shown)


py/desispec/scripts/daily_processing.py
```diff
@@ -355,6 +355,8 @@ def daily_processing_manager(specprod=None, exp_table_path=None, proc_table_path
                 # If done with science exposures for a tile and use_tilenight==True, use
                 # submit_tilenight_and_redshifts, otherwise use checkfor_and_submit_joint_job
                 if use_tilenight and lasttype == 'science' and len(sciences)>0:
+                    if dry_run_level < 3:
+                        write_tables([etable], tablenames=[exp_table_pathname])
```
akremin, line 359: This writes the latest exposures out to disk before trying to submit processing. In the event that the code crashes while trying to submit a job for processing, (e.g. when Slurm doesn't respond repeatedly), this will create the same type of issue we have previously seen. 

We can either move this code block to below the `submit_tilenight_and_redshifts()` call (just above the current `else` statement on line 370, or we can write the etable out when we write out the ptable on line 384. That will add an extra write for non-tilenight exposure tables, but it will be harmless.

py/desispec/scripts/daily_processing.py
```diff
@@ -443,6 +446,12 @@ def daily_processing_manager(specprod=None, exp_table_path=None, proc_table_path
                 write_table(ptable, tablename=proc_table_pathname)
             if override_night is None or continue_looping_debug:
                 sleep_and_report(10, message_suffix=f"after updating queue information", dry_run=dry_run)
+        if run_only_once:
+            print("Exiting after running only once")
```
akremin, line 450: Could we please make this a bit more explicit; "looping only once" or something even more explicit than that?

akremin (changes requested):
Thanks for adding this functionality and fixing a small bug (writing exposures to exposure tables before they are processed) while doing so. The additions look great, but I have a few minor requests that can be found inline.

bin/desi_daily_proc_manager
```diff
@@ -69,6 +69,8 @@ def parse_args():#options=None):
                         help="Wait time between loops in checking queue statuses and resubmitting failures.")
     parser.add_argument("--override-night", type=str,default=None,
                         help="Specify the night to run on. Overrides the current day.")
+    parser.add_argument("--run-only-once", action="store_true",
+                        help="Run only once. Suitable for running repeateadly in a scron job.")
```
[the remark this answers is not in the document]
marcelo-alvarez: Done.

py/desispec/scripts/daily_processing.py
```diff
@@ -33,7 +33,7 @@ def daily_processing_manager(specprod=None, exp_table_path=None, proc_table_path
                              badamps=None, override_night=None, tab_filetype='csv', queue='realtime',
                              exps_to_ignore=None, data_cadence_time=300, queue_cadence_time=1800,
                              dry_run_level=0, dry_run=False, no_redshifts=False, continue_looping_debug=False, dont_check_job_outputs=False,
-                             dont_resubmit_partial_jobs=False, verbose=False, use_specter=False, use_tilenight=False):
+                             dont_resubmit_partial_jobs=False, verbose=False, use_specter=False, use_tilenight=False, run_only_once=False):
```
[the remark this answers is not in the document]
marcelo-alvarez: Done.

py/desispec/scripts/daily_processing.py
marcelo-alvarez: Done.

py/desispec/scripts/daily_processing.py
marcelo-alvarez: @akremin just to clarify, this writes the exposure(s) in the tile about to be processed to disk, but not the latest exposure, since the row corresponding to the latest exposure is added to the exposure table below around line 391.

It's true that if something goes wrong in `submit_tilenight_and_redshifts()` such that the job is not submitted and the job crashes, then the exposures can end up in the exposure table, and thus would never be processed. I ended up placing the `write_table` call there because ztile jobs were missing the last exposure and thus incorrect (the tilenights were correct, however), and thus I assumed something about the ztile job creation was reading the exposure table. But I could have been mistaken that placing the `write_table` call before the `submit_tilenight_and_redshifts()` call actually fixed the missing ztile exposure error.

Do you think it's possible that somewhere in the ztile job creation that the exposure table is read in? I took a quick look and did not see it anywhere -- it seems to exclusively depend on the processing table passed in to `submit_tilenight_and_redshifts()` -- but it's complex enough that I could have just missed it.

In the meantime I will try moving the `write_table` call to just after `submit_tilenight_and_redshifts()` and see if that doesn't cause the ztile processing error to reappear in my mock observation tests.

marcelo-alvarez: Sounds good, thanks. I will have a look at the two options you've suggested.

sbailey:
superseded by pipeline refactor in #2187.  Closing without merging.

PR #2076, opened 2023-06-23
daily proc --exp-cadence-time option

sbailey:
This PR adds a `desi_daily_proc_manager --exp-cadence-time ...` option to adjust the amount of time that it sleeps between exposures.  The default is 2 seconds, the same as before (previously hardcoded).  This allows `desi_daily_proc_manager` tests that are purposefully spread out to mimic data arriving throughout the night.

I'm somewhat ambivalent about adding a new option that is intended to be used only for testing, which we might only need for a limit time period and then not use again (this was motivated by Perlmutter stability testing).  @akremin if you don't like this new option, I could also install the branch and run the tests out of the branch and never merge this PR into main.


akremin (approved):
The added option looks good and the option naming scheme is consistent previous variables, which is nice.

I am happy to merge this into main. It is relatively harmless, and this script will be refactored in the late summer or early fall anyway. If it isn't useful to us, we can drop it at that point.

Issue #2088, opened 2023-08-05
coadd_cameras and compute_coadd_scores 


araichoor:
unless I am mis-using the code as I m not familiar with it -- in which case apologies -- it looks to me that there s an inconsistency here in `desispec.coaddition.coadd_cameras()`:
https://github.com/desihub/desispec/blob/f1293d27ba9d23b65cbc7124c35bdb5b7b22196d/py/desispec/coaddition.py#L721-L740

because, when `compute_coadd_scores` is called, `res.scores=None`, and because of that `compute_coadd_scores` will fill `res.scores` with the dictionary it s using, not updating a Table:
https://github.com/desihub/desispec/blob/cfdbdab7444dbaaba6fd49ed813663ebbe401793/py/desispec/specscore.py#L84-L91

and having `scores` has a dictionary can break things downstream.

context:
my goal is to (1) take a coadd, (2) coadd the cameras, (3) select some rows.
the step (3) crashes in `Spectra._get_slices`, as it expects `scores` to be a Table.

see code snippet below.

```
from desispec.io import read_spectra
from desispec.coaddition import coadd_cameras

fn = "/global/cfs/cdirs/desi/spectro/redux/iron/healpix/main/dark/0/0/coadd-main-dark-0.fits"
s = read_spectra(fn)        # s.scores is a Table
s_brz = coadd_cameras(s)    # s_brz.scores is a dictionary
tids = s_brz.fibermap["TARGETID"][:5]
s_brz_cut = s_brz.select(targets=tids)
```
returns:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/desispec/main/py/desispec/spectra.py", line 375, in select
    sp = self._get_slice(keep_rows, bands=keep_bands)
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/desispec/main/py/desispec/spectra.py", line 318, in _get_slice
    for col in self.scores.dtype.names:
AttributeError: 'dict' object has no attribute 'dtype'

```

araichoor:
for what is worth, if my diagnosis above is correct, a simple workaround could be to replace this line:
https://github.com/desihub/desispec/blob/cfdbdab7444dbaaba6fd49ed813663ebbe401793/py/desispec/specscore.py#L90
by:
```
            # convert the dictionary to a table before feeding into coadd
[...]

Issue #2032, opened 2023-04-07
COADD_FIBERSTATUS bit propagation bug


sbailey:
If two input spectra have nonzero FIBERSTATUS but for different reasons, their results are not properly propagated into the output spectrum, resulting in an incorrectly optimistic `COADD_FIBERSTATUS==0`.  Example:

```python
import numpy as np
from desispec.io import read_spectra
import desispec.coaddition

sp = read_spectra('/global/cfs/cdirs/desi/spectro/redux/fuji/healpix/sv3/dark/281/28116/spectra-sv3-dark-28116.fits')
sp = sp[sp.fibermap['TARGETID'] == 39628362670410135]
print('Input spectra fibermap:')
print(sp.fibermap['TARGETID', 'FIBERSTATUS'])
print()

# make a copy since coaddition operates in-place
coadd = sp[:]
desispec.coaddition.coadd(coadd)
print('Coadded fibermap:')
print(coadd.fibermap['TARGETID', 'COADD_FIBERSTATUS'])
print()

print('Number of non-zero input flux bins:', np.sum(sp.flux['r'] != 0))
print('Number of non-zero coadded flux bins:', np.sum(coadd.flux['r'] != 0))
```
Results:
```
Input spectra fibermap:
     TARGETID     FIBERSTATUS
----------------- -----------
39628362670410135         256
39628362670410135           4

INFO:coaddition.py:350:coadd: Clipping cosmics with cosmics_nsig=4.0
Coadded fibermap:
     TARGETID     COADD_FIBERSTATUS
----------------- -----------------
39628362670410135                 0

Number of non-zero input flux bins: 4652
Number of non-zero coadded flux bins: 0
```

i.e. the coadded flux is zeroed out, but the non-zero FIBERSTATUS bits are propagated into a COADD_FIBERSTATUS=0.
Fortunately the zeroed flux triggers a bunch of Redrock ZWARN bits, but the COADD_FIBERSTATUS is misleading.

This impacts Fuji, Guadalupe, and Iron.

@ashleyjross @akremin heads up.

sbailey:
Fixed by PR #2093; closing ticket.

Issue #2092, opened 2023-08-10
record which EXP_FIBERMAP entries were used for coadded FIBERMAP


sbailey:
Context: coadd files have a coadded FIBERMAP (e.g. with MEAN_FIBER_RA) and an uncoadded EXP_FIBERMAP with the per-exposure information (e.g. FIBER_RA and EXPID).  We don't currently record which individual exposures actually contributed to the coadd for either the FIBERMAP or the FLUX, IVAR, etc. HDUs.

To do: Add column(s) to EXP_FIBERMAP to record which individual exposures were used for the coadd.  Due to FIBERSTATUS bits BADAMPB/R/Z this is a per-camera quantity, not a per-exposure quantity, so either use 3 boolean columns or one byte mask column (or something else that can describe per-amp and not just per-exp).

PR #2093, opened 2023-08-11
update COADD_FIBERSTATUS to bitwise OR when all inputs are bad

sbailey:
This PR fixes the COADD_FIBERSTATUS bit propagation bug reported in #2032  .

For good targets, COADD_FIBERSTATUS is unchanged = the bitwise AND of input FIBERSTATUS.

For targets where all input spectra are bad, this PR changes COADD_FIBERSTATUS to the bitwise OR of input FIBERSTATUS values.  This fixes the case in #2032 where the input spectra were all bad, but for different reasons, resulting in a bitwise OR of 0, incorrectly implying that the input FIBERSTATUS were ok.  This will also result in other bad targets getting additional bits set (due to OR instead of AND).

Note that the incorrect COADD_FIBERSTATUS=0 cases had all ivar=0 anyway, which leads to ZWARN!=0 so we weren't getting incorrectly "good" redshifts for these anyway, but some analyses/bookkeeping were treating COADD_FIBERSTATUS==0 as "good hardware" so this makes it a bit more accurate.

```
import numpy as np
from desispec.io import read_spectra
import desispec.coaddition

sp = read_spectra('/global/cfs/cdirs/desi/spectro/redux/fuji/healpix/sv3/dark/281/28116/spectra-sv3-dark-28116.fits')
sp = sp[sp.fibermap['TARGETID'] == 39628362670410135]
print('Input spectra fibermap:')
print(sp.fibermap['TARGETID', 'FIBERSTATUS'])
print()

# make a copy since coaddition operates in-place
coadd = sp[:]
desispec.coaddition.coadd(coadd)
print('Coadded fibermap:')
print(coadd.fibermap['TARGETID', 'COADD_FIBERSTATUS'])
```
Results in
```
Input spectra fibermap:
     TARGETID     FIBERSTATUS
----------------- -----------
39628362670410135         256
39628362670410135           4

Coadded fibermap:
     TARGETID     COADD_FIBERSTATUS
----------------- -----------------
39628362670410135               260
```
(previously that case was incorrectly 0).

@ashleyjross @stephjuneau @akremin heads up.  Comments welcome.

### DIFF
--- a/py/desispec/coaddition.py
+++ b/py/desispec/coaddition.py
@@ -288,9 +288,6 @@ def coadd_fibermap(fibermap, onetile=False):
     for i,tid in enumerate(targets) :
         jj = fibermap["TARGETID"]==tid
 
-        #- coadded FIBERSTATUS = bitwise AND of input FIBERSTATUS
-        tfmap['COADD_FIBERSTATUS'][i] = np.bitwise_and.reduce(fibermap[fiberstatus_key][jj])
-
         #- Only a subset of "good" FIBERSTATUS flags are included in the coadd
         fiberstatus_nonamp_bits = get_all_nonamp_fiberbitmask_val()
         fiberstatus_amp_bits = get_justamps_fiberbitmask()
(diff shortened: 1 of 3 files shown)


stephjuneau:
I haven't looked at the new code but from my recollection of the original code, the subtlety was that the fiberstatus needs to be determined on a per-amp basis in the case of the "badamp" bits and otherwise on a per spectrum basis. It might be worth to specifically find a case where only 1 or 2 amps are bad and the rest of the spectrum still useable for the unit tests.

sbailey:
Good question @stephjuneau .  I reviewed N>>1 FIBERSTATUS cases with @akremin for what we really want COADD_FIBERSTATUS to mean and expanded the unit tests, which caught a problem in the original version of this PR for the case where 1 input spectrum is flagged bad for a per-fiber reason (like BROKENFIBER) while another input spectrum for the same target is flagged bad for a per-camera reason (like BADAMPB).  Summarizing the final state:

```
If any input spectra are good:
    COADD_FIBERSTATUS = bitwise AND of the FIBERSTATUS of good input spectra, i.e.
                      = the AND of the FIBERSTATUSes actually used in the coadd
else (all are bad):
    COADD_FIBERSTATUS = bitwise OR of FIBERSTATUS input spectra
```
My first implementation added the else clause (bitwise OR), but was still taking the bitwise AND of all input spectra not just the ones actually used in the coadd.

Note: COADD_NUMEXP is the number of unique exposure IDs that contributed to the coadd, even if some cameras used fewer exposures than that due to BADAMPB/R/Z.  I considered also implementing #2092 in this PR to add a record of which exposures were used for each camera, but I'm going to save that for a separate PR to avoid feature creep here.  In any case, COADD_NUMEXP is only an approximately useful number since per-wavelength masking makes it a bin-by-bin value not just a camera-by-camera value, and the propagated IVAR is what you should really be using if the difference matters since coadding two high S/N spectra is very different from coadding one high S/N spectrum with one low S/N spectrum even if both are COADD_NUMEXP=2.

Along the way I also fixed #2088 reported by @araichoor where slicing spectra coadded in memory would crash due to coadd.scores being a dict instead of a Table.

Issue #1777, opened 2022-05-19
ZCAT_NSPEC type inconsistency


weaverba137:
`ztile-` and `zpix-` files always have the column `ZCAT_NSPEC` with type `int16`.  Somehow the `zall-` files change that to `int64`.  This is inconsistent with the data model, among other things.  Is there a specific reason for this?  It seems that for every other column, the `zall-` file preserves the type, so this is a real standout.

Although one could imagine many possible exposures of the same object, will that every *really* exceed the size of signed `int16`?  This motivates the question: why do `SV_NSPEC` or `MAIN_NSPEC` need to be `int32`, which is not consistent with `ZCAT_NSPEC` in the `ztile-` and `zpix-` files, *nor* consistent with `ZCAT_NSPEC` in the `zall-` files.

stephjuneau:
Hi @weaverba137 , I think we can set all the *_NSPEC columns to type `int16`. I added @Ragadeepika-Pucha as an assignee as she wrote the code for generating the zall- files.

weaverba137:
That's fine, thank you!

weaverba137:
@sbailey has this been fixed in the DR1 patching?

sbailey:
> @sbailey has this been fixed in the DR1 patching?

Not yet, but we should.  I added it to the DR1 Patching project. @Ragadeepika-Pucha are you available to fix this? It may just need adding a `dtype=np.int16` somewhere in `desispec.zcatalog.create_summary_catalog`.

stephjuneau:
@sbailey I sent you a message to close the loop on this.

sbailey:
It sounds like Raga is likely not available in the short term to work on this; if anyone else could help I would appreciate it.

LNapolitano:
Pull request: https://github.com/desihub/desispec/pull/2103

sbailey:
fixed by #2103, thanks!

PR #2103, opened 2023-08-24
Iss1777

LNapolitano:
Changed datatype of '_NSPEC' columns in zcatalog.py to int16 in order to be consistent with ztile- and zpix- files.


sbailey:
Looks good, thanks.  I explicitly verified by regenerating a fuji zall catalog and checked the output dtype.  fixes #1777 .

PR #2105, opened 2023-08-29
fix incorrect use of polyfit 

segasai:
Hi, 

while investigating trace_shifts I noticed that there is a couple places in desispec that use np.polyfit weights assuming those weigths are inverse variances, while they should be inverse errors:  https://github.com/desihub/desispec/blob/bf216b2e65b066367d9607315605e3f2f5863a5e/py/desispec/trace_shifts.py#L569
https://github.com/desihub/desispec/blob/bf216b2e65b066367d9607315605e3f2f5863a5e/py/desispec/qproc/qfiberflat.py#L83
cf. docs 
https://numpy.org/doc/stable/reference/generated/numpy.polyfit.html

The patch fixes that. (I am assuming that the patch should be pretty safe)


julienguy:
Wow, good catch. I should never use those built-in minimization (without reading the doc carefully!).

Issue #1827, opened 2022-08-18
Improve wavelength calibration adjustment on sky lines


julienguy:
The wavelength calibration adjustment on sky lines may be improved with a median filtering of the spectra prior to the cross-correlation of the measured spectra and the model sky spectrum. This improvement is striking for one exposure/camera which wavelength calibration failed as reported in https://github.com/desihub/desisurveyops/issues/45 . The trace shift is measured for each fiber independently in several intervals along the spectrum. And then averaged with outlier rejection. There are much fewer outliers if the spectra have been median-filtered for the test exposure/camera which suggests this could be an improvement for the wavelength calibration in general.

This is implement in desispec branch traceshift-median-filter .
This takes a bit more CPU time but more importantly it should be tested on a large data sample (with for instance radial velocities comparison).

segasai:
Hi Julien, do you have preference for the dates of the data taken ? I.e. fuji/guadalupe/daily ? 
The best exposures for my previous tests were usually fuji (pre-main survey) tiles with a large number of stars and many repeats.

julienguy:
Fuji data is fine. Please suggest a list of exposures that would be a good sample to validate this change. Ideally a fair number of stars with accurate external radial velocities.

segasai:
The best tile to compare with Gaia RV's is 80613, it has 1327 stars in common (counting individual exposures)
These are the EXPIDs 81831 81832 81833 81834 81835 68657 69042 68658 68659 69043 69044 69431 69045 69429 69430 69432 69433 69605 69606 69607 69608 69225 69226 69227 69228 69229 69230 69609 69610 69611

The stars with Gaia RVs will be very very bright 13-15 mag and RVs will be accurate to better than 1 km/s.
for APOGEE the best tile is  80738 with  ~ 1300 matches counting repeats.
These are the EXPID 83010 83011 83748 77579 77803 79308 79309 76663 79310 79311 80971 76661 77434 76662 83448 76314 77435 83166
These will be tad fainter.

julienguy:
Starting running in /global/cfs/cdirs/desi/spectro/redux/traceshift-median-filter/ .

segasai:
Thanks. I've looked there, but I can't see any spectra- files there only sframes. With spectra-*fits , I could easily run my RV code for the subset of stars we care about.

julienguy:
@segasai , the spectra from tile 80613 are in /global/cfs/cdirs/desi/spectro/redux/traceshift-median-filter/ .

[...]

PR #2106, opened 2023-08-30
Fix trace_shifts to deal with bright/backup data

segasai:
Hi, 

This is a tentative version of the patch that fixes the trace_shifts issue with backup/bright data. ( ( issue #1827 )  
Basically that uses the new sky template that has continuum subtracted. Also that applies a high-pass filter to the data before computing the shift.

Here's the result when applying it to a range of different expids 80967,81090,81091,81530,81531,83282,83283,101890,101891,102502,102503,102609,106453,106455,107543,108300,110864,111719,114604,114862,116444,118497,122057,123288,124238,124586,124726,124727,126581,130264,130555,131572,131940,131961,131966,132520,133256,134440,134923,139457,139464

left before, right new (using all 3 arms)
image

This is definitely better, but also not perfect.
Here's when I just use the blue arm to measure velocities
image

And finally for the z arm
image

In my mind the only test that is probably worth doing before committing this is a test of some tile that was repeated many times in SV, to test that the systematic floor in RV doesn't get worse.

I am still concerned that there are some systematics left even after the change, but I don't know if it's worth dealign with them one at a time. 

I.e. I haven't fully digested the 'internal calibration' part of the trace_shifts.py and whether that also needs to be fixed in a similar way to external calibration.

(maybe it is a separate point, but I also seem to understand now that the trace_shifts functionality also should allow us to do correct heliocentric corrections per fiber, as right now we use the heliocentric correction based on ra,dec of the center of the field).


julienguy:
Hi Sergey, 

- I think we should add more documentation on the provenance of the sky spectrum and in particular on the wavelength of the emission lines. I would add to the PR the script you used for the generation of this spectrum if it is at least partially based on data and references for the line calibration. 

- We should also test for the statistical precision of this shift and not include it if the uncertainty is larger than the know rms of the correction obtained in dark time (or add a prior).

- Yes we have not done the heliocentric correction per fiber and use instead for all fibers the correction derived at the center of the field of view. I suggest we keep it as a separate topic for now.

segasai:
Thanks Julien,

The current sky spectrum was a bit of quick hack, when I realized I couldn't use DESI actual sky spectrum, because a high res spectrum is used. 

For completeness this was the script I used. Basically this just detects prominent peaks in the current skyspec.dat and masks the rest


```python
import numpy as np
import scipy.signal


def doit(fname_in, fname_out, threshold=2):
    """ pick up prominent emission lines and filter the rest"""
    SKY0 = np.loadtxt(fname_in).T
    W, F = SKY0

    window = 17
    bg = scipy.ndimage.median_filter(
        F, footprint=np.abs(np.arange(window) - (window // 2)) > window // 4)
    Fbg = F - bg
    peaks = scipy.signal.find_peaks(Fbg, height=threshold)
    mask = np.zeros(len(F), dtype=bool)
    window2 = 5
    for p in peaks[0]:
        mask[p - window2:p + window2] = True

    ret = Fbg * 1
    ret[~mask] = 0
    ret = np.maximum(ret, 0)
    np.savetxt(fname_out, np.array([W, ret]).T)


if __name__ == '__main__':
    doit('/tmp/orig_skyspec.dat', '/tmp/skyspec.dat')
```


But now I can see that I should probably just use this instead:
https://ui.adsabs.harvard.edu/abs/2003yCat..34071157H/abstract
giving this if I add all the lines together.
image

I'll try to substitute the current spec-sky.dat with that and see if it is better.

* I agree on checking the precision of the shift.  I am not 100 % sure on using the prior (as that may prevent us from finding problematic cases). 

Actually your point about the prior motivated me to make the plot of 'MEANDY' for backup/bright/dark 
(here blue/green/red are B/R/Z)

here's dark
image

here's bright
image

here's backup
image

Assuming that the DY shift should not differ between backup/dark, that would actually suggest using say a gaussian 0.1 pix prior on the shift.

segasai:
Using just raw UVES based simulated spectrum for some reason produced worse results. 
I'll try to limit the line-list to only bright lines and see it works.


image

segasai:
I have added a Gaussian prior of 0.1A as a new patch. (it doesn't really do much given the updates from the PR which already gives much more reasonable dys comparing to before)
I also had run the UVES synthesizes sky and it can give similar results to the one I have here. 

A question -- how do I just run trace_shifts in the same way the desi_proc does it ? I.e. I'd like to be able to just debug one trace_shift --sky run without all other processing steps and without submitting them to the queue. 

For example I'm suspicious of this step https://github.com/desihub/desispec/blob/b98b567f2f82b445ff5e9154772c30454d3546f7/py/desispec/trace_shifts.py#L703 which probably doesn't work in the blue when reference spectrum has only emission lines.

segasai:
I figured that it takes too long to rerun the full reprocessing of expids (and it takes too much cpu/gpu time), so I'm going to focus on just looking at MEANDY values for a given petal. Here is a  plot that shows the MEANDY vs radial velocity  shift grouped by dark/bright/backup and b/r/z for  the whole iron looks like split by arm and program
image

I have now collected a list of 560 representative petal/arm combinations (for all conditions dark/bright/backup)  that I can run through trace_shifts in 5min nersc.

And here's the effect of the current patch. Left is current MEANDY vs median delta-radial velocity for all objects in a given petal. And on the right I show *new* MEANDY   and approximately *corrected* radial velocity using the difference between new MEANDY and old MEANDY.

image

This is already clearly much better than what we have now. and as far as I can see doesn't deteriorate anything for small meandy. But it clearly could be improved a bit, I'll try a bit more.

segasai:
Okay, now I have a version that I am happy with 
image

The main change on top of new template:
* High-pass filter applied to the data. The variance is assigned to the 1d spectrum based on MAD of this spectrum
* The rescaling of the spectrum is done through one constant, not function of wavelength
* The prior of 0.1 is applied to the shifts. 

I think possibly some other improvements can be done here, but IMO the current code is ready.
I'm rerunning the full pipeline now to do a final check

segasai:
Here's the full run and comparison of actually measured rvs to external rvs and compared to official iron for a sample of expids)
73771,73772,73920,73921,74324,74325,74821,74822,74823,74824,75123,75630,75631,76800,76801,76802,76803,76805,76806,76807,77219,77220,77221,77222,80967,81090,81091,81530,81531,83282,83283,101890,101891,102502,102503,102609,106453,106455,107543,108300,110864,111719,114604,114862,116444,118497,122057,123288,124238,124586,124726,124727,126581,130264,130555,131572,131940,131961,131966,132520,133256,134440,134923,136322,136323,139457,139464

image

It shows that for a 'typical' dark/bright nothing will change (although rare expids with large meandy should get better), but for backup there is a drastic improvement.

I think there are some systematics left there, but at much smaller level, and I'd prefer to deal with them after this is committed.


Also I looked at the repeated observations and fitted the pairwise differences with the model N(0,sqrt(2*floor^2+e_1^2+e_2^2) and the old reduction gave me 1.3 km/s floor, while new one gave 0.7 km/s ! systematic floor and that applies to bright, so it's not only backup that's got better.

julienguy:
The reduction in the systematic error floor is great. The prior fixes the issue for the backup tiles. I looked at the code and do not see any issue (I saw the more sophisticated estimation of the variance in the median spectrum after removal of the continuum). I also looked at the new reference sky line spectra compared to Hanuschik (2003) and do not find any issue, so I'm merging this.

sbailey:
@segasai thanks for this PR.  A post-merge followup request: please confirm what is the largest shift in this PR compared to iron, i.e. abs(DY_thisPR - DY_iron).  i.e. the core improvement looks great, but what is the upper limit on the outliers for what changes from iron?  Thanks.

segasai:
Thanks for your help @sbailey as well.
Regarding outliers in iron.
* I've re-processed (using trace_shifts only) 2.5k exposures spanning all the meandy range and all the permutations of survey/program.  (obviously that's still a small fraction).
For those that's how old dy vs new dy looks like 
image
I.e. the largest change in dy is of the order 4. But this is clearly because it was wrong before and because now there is a  N(0,0.1) prior on the shift. 

So I'm not sure the maximum shift is a particularly useful metric. I could run on more exposures in principle but i think my set of exposures is quite representative.
it's available in /desi/users/koposov/wavelength_fix/refit_exp_list.csv
FWIW The largest |new_dy| for cases where |old_dy|<0.1 is 0.22 for my set of exposures.

julienguy:
It reminds me we have to be a little bit careful here and investigate cases with a large shift. Some must be genuine because we did not use the correct PSF file. But these should be rare. We can think about applying the prior only when in poor observing conditions or short exposures or simply for the back up program. You sent me the list of outlier exposures, I have to look at them.

segasai:
But the prior is only a Gaussian penalty in chi-square, so I was assuming that if the data is informative enough that should still win, but it obviously depends on the numbers.

sbailey:
Credit to @akremin for pointing out off list that we didn't evaluate this as much for downstream impacts as we typically do for algorithmic changes.  I took a quick look at iron vs. desi/spectro/redux/koposov/wavelength_test_bulk_final/tiles/cumulative which I'm guessing is the prod related to this PR.  See desi/users/sjbailey/dev/pr2106_dy/README for some notes about assembling and cross matching the redshift catalogs.  A few notes:

* Out of 125806 science targets
  * 99 changed from ZWARN=0 to ZWARN != 0
  * 64 changed the other way from ZWARN != 0 to ZWARN=0.
  * 49 changed SPECTYPE
  * 13 had ZWARN=0 both before and after but changed their redshifts by >1000 km/s

These are tiny fractions, and might be due to other changes since iron that aren't directly due to this PR.  I'm not too concerned about these, but if others are we should re-run Sergey's test exposures with desispec hash b98b567, which is the state from where this PR branched so that we're sure we're testing just the changes of this PR before visually inspecting individual cases.

@julienguy I think we could test the case you are concerned about by running `desi_compute_trace_shifts` on the same science exposure with multiple different input PSFs with known wavelength differences to see if they consistently get shifted back to the correct solution, vs. whether the prior is too strong and preventing larger shifts.

julienguy:
Most of the exposures with |MEANDY|>1 are from NIGHT=20220106 CAMERA=b7 where the calibnight PSF was absent in the iron production (but not in daily?) and therefore the code used the default PSF which has some offsets.

```
desi_header -i /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b?-00116940.fits --hdu PSF -k IN_PSF MEANDY
# IN_PSF MEANDY filename
 /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/calibnight/20220106/psfnight-b0-20220106.fits 0.1181037849666189 /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b0-00116940.fits
 /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/calibnight/20220106/psfnight-b1-20220106.fits 0.07353502657263503 /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b1-00116940.fits
 /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/calibnight/20220106/psfnight-b2-20220106.fits 0.136159765841333 /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b2-00116940.fits
 /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/calibnight/20220106/psfnight-b3-20220106.fits 0.1335736384478097 /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b3-00116940.fits
 /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/calibnight/20220106/psfnight-b4-20220106.fits 0.1143213956497461 /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b4-00116940.fits
 /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/calibnight/20220106/psfnight-b5-20220106.fits 0.1285807352665853 /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b5-00116940.fits
 /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/calibnight/20220106/psfnight-b6-20220106.fits 0.1154187458003997 /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b6-00116940.fits
 /dvs_ro/cfs/cdirs/desi/spectro/desi_spectro_calib/0.4.0/spec/sm8/psf-sm8-b7-20210912.fits -1.336020744688121 /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b7-00116940.fits
 /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/calibnight/20220106/psfnight-b8-20220106.fits 0.1346197589384247 /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b8-00116940.fits
 /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/calibnight/20220106/psfnight-b9-20220106.fits 0.1014211902518728 /global/homes/j/jguy/redux/iron/exposures/20220106/00116940/psf-b9-00116940.fits
```
I checked that the offset of -1.3 pixel was correct by comparing the sky model between two cameras. The emission lines were well aligned (by eye).
I reran the trace shift fit with the new code and find a consistent shift (actually with a slightly larger amplitude). This confirms that the prior is weak enough that it does not bias the measured offset when this measurement is significant (for long dark time exposures).

```
desi_compute_trace_shifts -i /dvs_ro/cfs/cdirs/desi/spectro/redux/iron/preproc/20220106/00116940/preproc-b7-00116940.fits.gz --psf /dvs_ro/cfs/cdirs/desi/spectro/desi_spectro_calib/0.4.0/spec/sm8/psf-sm8-b7-20210912.fits --degxx 2 --degxy 0 --degyx 2 --degyy 0 --sky --outpsf psf-b7-00116940-new.fits
...
desi_header -i psf-b7-00116940-new.fits --hdu PSF -k MEANDY
# MEANDY filename
 -1.392191341423177 psf-b7-00116940-new.fits
```

I note however an error message (probably harmless) in the code and will open a separate issue about this.

PR #2109, opened 2023-08-31
desi_zcatalog --recoadd-fibermap option

### LINKED ISSUE #2087: add desi_zcatalog --recoadd-fibermap option
Add a new `desi_zcatalog --recoadd-fibermap` option to re-coadd the FIBERMAP from the input EXP_FIBERMAP per file while assembling the redshift catalogs, instead of directly propagating the already coadded FIBERMAPs.  This would be used to create updated redshift catalogs for DR1 after fixing the various issues in project https://github.com/orgs/desihub/projects/44/views/1 (DR1 Patching), without requiring that the original files themselves have to be patched first.

sbailey:
This PR fixes #2087 by adding a `desi_zcatalog --recoadd-fibermap` option to recoadd the spectra FIBERMAP -> FIBERMAP, EXP_FIBERMAP instead of propagating the previously coadded versions from the redrock files.  This enables us to rebuild the stacked redshift catalogs while bringing in the bug fixes from e.g. PRs #2100 (IN_COADD_B/R/Z), #2094 (target bit bitwise OR), and #2093 (COADD_FIBERSTATUS bug fixes).

A companion option `--ztile` flags whether this is a tile-based coadd (vs. a healpix based coadd).  I plan a separate PR to simplify the `desi_zcatalog` options and this might change again, but I've tried to focus this PR on the minimal changes needed to support adding re-coadding.

One non-minimal change for simplicity is dropping support for the "ZBEST" HDU (renamed "REDSHIFTS") and dropping support for missing TSNR2 and EXP_FIBERMAP; all of these cases were for internal releases that pre-date Fuji/EDR.

### DIFF
--- a/bin/desi_zcatalog
+++ b/bin/desi_zcatalog
@@ -27,7 +27,9 @@ from astropy.table import Table, hstack, vstack
 from desiutil.log import get_logger
 from desispec import io
 from desispec.zcatalog import find_primary_spectra
-from desispec.io.util import get_tempfilename
+from desispec.io.util import get_tempfilename, checkgzip, replace_prefix
+from desispec.io.table import read_table
+from desispec.coaddition import coadd_fibermap
 
 def match(table1,table2,key="TARGETID") :
     """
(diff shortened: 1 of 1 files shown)


sbailey:
I forgot to mention: example outputs are in /pscratch/sd/s/sjbailey/desi/dev/zcat_recoadd/:
  * ztile-sv2-bright-orig.fits: regenerating the iron ztile-sv2-bright-cumulative catalog without re-coadding
  * ztile-sv2-bright-recoadd.fits: including `--recoadd-fibermap --ztile`

the differences are due to the extra columns introduced in #2065 and #2100, and changes to COADD_FIBERSTATUS and MEAN/STD coadded columns due to bug fixes to fibermap coadds #2065, #2093, #2094, #2100.

akremin (approved):
Thanks, the changes are minimal and look good to me. The differences show that the new coadd fibermap has the correct values thanks to the updated code, while not showing any other differences.
The loss of backward compatibility is unfortunate, but after more than two years I think that is appropriate. We can use an older code tag if we need to work with earlier data.

PR #2110, opened 2023-09-01
fix HPXNEST header type

sbailey:
This PR fixes the datatype of the HPXNEST keyword in spectra files and stacked redshift catalogs.  This was coming in via `desi_group_spectra --header HPNEST=True ...` (i.e. not propagated from an input file), but that was incorrectly resulting in a string 'True    ' instead of the FITS boolean T.  Several layers of fix:
* `--header KEY=VAL` parsing explicitly handles booleans now, and is put into a new function `desispec.util.parse_keyval` since we were doing similar logic in both `desi_group_spectra` and `desi_zcatalog`.
* `desi_group_spectra --healpix N ...` automatically sets healpix-based keywords now, so it is no longer necessary to explicitly set `--header HPXNEST=True` anyway, but we still need to set other new header keywords so we might was well get it right for any future booleans that we might set.
* `desi_zcatalog` intercepts incorrect input HPXNEST='True' value and converts them to boolean, enabling us to rebuild stacked redshift catalogs with correct HPXNEST without having to first patch all upstream files.
* Tests are added that would have caught the problem in the first place

In a git branching mixup, this PR started off of the PR #2109 zcat_recoadd branch instead of main and thus also includes `--recoadd-fibermap` options too.  My attempt to rebase off of main left those changes in here and I don't want to muck it up more, so let's finish that PR and merge it, and then sort out any merge conflicts here.

Example outputs are in /pscratch/sd/s/sjbailey/desi/dev/zcat_hpxnest, generated with
```
desi_zcatalog -i /global/cfs/cdirs/desi/spectro/redux/iron/healpix/sv2/dark/ -o zpix-sv2-dark.fits -g healpix --header SURVEY=sv2 PROGRAM=dark

IRON=/global/cfs/cdirs/desi/spectro/redux/iron
desi_group_spectra --inframes \
    $IRON/exposures/20210519/00089278/cframe-?7-00089278.fits.gz \
    $IRON/exposures/20210519/00089278/cframe-?8-00089278.fits.gz \
    $IRON/exposures/20210519/00089280/cframe-?2-00089280.fits.gz \
    --outfile spectra-main-dark-10000.fits.gz \
    --coaddfile coadd-main-dark-10000.fits \
    --healpix 10000 \
    --header SURVEY=main PROGRAM=dark
```

For both output files `zpix-sv2-dark.fits ` and `spectra-main-dark-10000.fits.gz`, fitsheader now shows them to have HPXNEST = T (i.e. a boolean) instead of previously 'True    ' like the original files in `iron/zcatalog/zpix-sv2-dark.fits` and `iron/healpix/main/dark/100/10000/spectra-main-dark-10000.fits.gz`.

@weaverba137 please review.

### DIFF
--- a/bin/desi_zcatalog
+++ b/bin/desi_zcatalog
@@ -30,6 +30,7 @@ from desispec.zcatalog import find_primary_spectra
 from desispec.io.util import get_tempfilename, checkgzip, replace_prefix
 from desispec.io.table import read_table
 from desispec.coaddition import coadd_fibermap
+from desispec.util import parse_keyval
 
 def match(table1,table2,key="TARGETID") :
     """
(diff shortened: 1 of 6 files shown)


weaverba137 (approved):
This looks good to me.

Issue #2068, opened 2023-06-07
Propagate fibermap units through all steps of pipeline


weaverba137:
The DESI data model claims that columns in `redrock`, `zpix`, `ztile` and `zall` files have units, but in fact they do not.  There are no `TUNIT` keywords set in any of the files in the fuji zcatalog directory, and a spot check shows that they are not set in individual redrock files either.

This also makes it much more difficult to propagate units even further downstream to VACs derived from these files.

For the purposes of EDR, I think the response has to be documentation: "even though the data model *claims* these columns have units, in fact, the files do not contain unit metadata."

All along we've been strongly encouraging VAC authors to put units in data files, but the pipeline is not actually doing that for some of the most critical outputs, so I think for the future, we should really be more careful about propagating units.

weaverba137:
To speed up the discussion a little bit, I know that units are dropped internally for performance reasons, which explains why units are not present in the `redrock` files.  However, it seems like the assembly of `zpix` and `ztile` files would be IO-bound, not CPU-bound, so there would be no performance penalty to adding units to those files.

A better-phrased warning might be: "the units described in the data model for `DESI_SPECTRO_REDUX/SPECPROD` are the units of the values in those files, however, the files do not necessarily contain the unit metadata themselves (`TUNITxxx` header cards)."

sbailey:
> A better-phrased warning might be: "the units described in the data model for DESI_SPECTRO_REDUX/SPECPROD are the units of the values in those files, however, the files do not necessarily contain the unit metadata themselves (TUNITxxx header cards)."

That is how I have always interpreted the meaning of the datamodel — describing the actual units of the data, not the existence (or not) of TUNIT and BUNIT keywords.  I don't think that requires a specific caveat in the text of the data model.

We can separately address if there are any unintended side effects from adding TUNIT keywords to the data files in the future.  I agree that in general it is preferable for us to propagate those and it was an oversight (not purposeful) that they are not included in the final redshift catalogs.

weaverba137:
> That is how I have always interpreted the meaning of the datamodel.

But will the public interpret it that way? Especially since some files do have units and some down't

sbailey:
Adding units to the files is a post-EDR activity for future specprods / releases.

For the EDR Countdown dashboard I'm moving this into the optional category for adding caveats to the datamodel and/or the known issues section of desidatadocs.

weaverba137:
[...]

PR #2111, opened 2023-09-06
desi_zcatalog --add-units option for DR1 patching

sbailey:
This PR adds a `desi_zcatalog --add-units` option to reapply units when creating stacked redshift catalogs, using desidatamodel/py/desidatamodel/data/column_descriptions.csv .

Example outputs are in /pscratch/sd/s/sjbailey/desi/dev/zcat_units/zpix*.fits, compared to $CFS/desi/spectro/redux/iron/zcatalog/zpix*.fits which don't have TUNITnn header keywords.

This PR does *not* attempt to fix the underlying problem of why units aren't propagated in the fibermaps in the upstream files; see issue #2068 for context about that.  This is is just the minimal update for patching the final DR1 redshift catalogs.

This PR uses `desiutil.annotate.load_csv_units` to parse column_descriptions.csv, then it uses `desispec.io.util.write_bintable` to add the units while writing.  It does *not* use `desiutil.annotate.annotate_table` to add units to the table itself, because nanomaggy units get stripped back out by `astropy.io.fits.convenience.table_to_hdu`; see desihub/desiutil#191 starting at https://github.com/desihub/desiutil/issues/191#issuecomment-1708702084 for discussion about that.

### DIFF
--- a/py/desispec/io/util.py
+++ b/py/desispec/io/util.py
@@ -233,7 +233,7 @@ def write_bintable(filename, data, header=None, comments=None, units=None,
     #
     # Add comments and units to the *columns* of the table.
     #
-    for i in range(1, 999):
+    for i in range(1, 1000):
         key = 'TTYPE'+str(i)
         if key not in hdu.header:
             break
--- a/py/desispec/io/util.py
+++ b/py/desispec/io/util.py
@@ -187,6 +187,18 @@ def write_bintable(filename, data, header=None, comments=None, units=None,
     """Utility function to write a fits binary table complete with
     comments and units in the FITS header too.  DATA can either be
     dictionary, an Astropy Table, a numpy.recarray or a numpy.ndarray.
+
+    Args:
+        filename: full path to filename to write
+        data: dict or table-like data to write
+
+    Options:
+        header: dict-like header key/value pairs to propagate
+        comments (dict): comments[COLNAME] per column
+        units (dict): units[COLNAME] per column
+        extname (str): extension name for EXTNAME header
+        clobber (bool): if True, overwrite pre-existing file
+        primary_extname (str): EXTNAME to use for primary HDU=0
     """
     from astropy.table import Table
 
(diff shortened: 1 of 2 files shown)


py/desispec/io/util.py
```diff
@@ -187,6 +187,18 @@ def write_bintable(filename, data, header=None, comments=None, units=None,
...
+    Args:
+        filename: full path to filename to write
+        data: dict or table-like data to write
+
+    Options:
+        header: dict-like header key/value pairs to propagate
+        comments (dict): comments[COLNAME] per column
+        units (dict): units[COLNAME] per column
+        extname (str): extension name for EXTNAME header
+        clobber (bool): if True, overwrite pre-existing file
```
weaverba137, line 200: I thought we agreed a long time ago to follow the Astropy convention and use `overwrite` instead of `clobber`.

weaverba137 (approved):
I prefer `overwrite` to `clobber`, but otherwise this looks fine.  Can we generate an example file to look at before merging?

weaverba137:
One more gotcha:
```python
for i in range(1, 999):
```
this is *not* what you want.  You want:
```python
for i in range(1, 1000):
```
because `TTYPE999` is actually valid, but would never be accessed in the first case.

sbailey:
Example outputs are in /pscratch/sd/s/sjbailey/desi/dev/zcat_units/zpix*.fits, compared to $CFS/desi/spectro/redux/iron/zcatalog/zpix*.fits which don't have TUNITnn header keywords.

"clobber" pre-dates this PR; I was just adding a more complete docstring for the function `write_bintable` that already existed, without changing the interface of the existing function.  IIRC, that function actually dates back to ~9 years ago before even astropy was using "overwrite" instead of "clobber".

I'll fix the 999 vs. 1000 (which also predates this PR and we may never have that many columns, but you are right that was a conceptual bug).

weaverba137:
I'm willing to give `clobber` a pass for the purposes of this PR, but I think any new code should use `overwrite`.

The example files look good so far, but I didn't see any comments on the columns.  Did you intentionally not add those?

sbailey:
> I'm willing to give clobber a pass for the purposes of this PR, but I think any new code should use overwrite.

Indeed, for this PR I was prioritizing not breaking the API of a long-standing pre-existing function used elsewhere.

> The example files look good so far, but I didn't see any comments on the columns. Did you intentionally not add those?

I initially tried adding them, but ~1/3 of them were truncated, some of them confusingly so.  In the end I removed the comment-adding code from `desi_zcatalog` until we have a curated set of short comments appropriate for adding to FITS files.  That may be post DR1, which will continue to rely upon desidatamodel for the primary description of the columns.

PR #2117, opened 2023-09-13
desi_zcatalog refactor

sbailey:
This PR significantly updates desi_zcatalog to add parallelism and make it easier to use for productions.  Relative to current main, this PR changes the script options/usage but does not change the actual outputs other than their sort order.  The code changes are big enough that it might be easiest to evaluate the final state and outputs rather than the code diff.

## Usage ##

The updated script supports two modes:
  * `desi_zcatalog --indir ...` : recursively find redrock*.fits files under an arbitrary input directory and stack them.  This is useful for creating test redshift catalogs from custom runs of redrock that might be organized outside of the context of a production.
  * `desi_zcatalog --group healpix|cumulative|pernight|perexp --survey SURVEY --program PROGRAM ...` uses a standard specprod structure to group either by healpix or tiles.

Current main requires a pre-step to generate an ASCII list of tiles to include for each survey,program combination which is a hassle; this PR updates the script to use the tiles file from the production and filter by (survey, program) on-the-fly.

## Example outputs ##

See `/pscratch/sd/s/sjbailey/desi/dev/zcat_options/`

**`ztile-sv2-dark-cumulative-main.fits`**, produced with current main with:

```
desi_zcatalog --add-units --recoadd-fibermap --ztile \
    --group cumulative \
    --indir /global/cfs/cdirs/desi/spectro/redux/iron/tiles/cumulative/ \
    --tiles sv2-dark-tiles.txt \
    --outfile ztile-sv2-dark-cumulative-main.fits
```

**`ztile-sv2-dark-cumulative-pr.fits`**, produced with this branch with:
```
export SPECPROD=iron
desi_zcatalog --add-units --recoadd-fibermap --nproc 16 \
    --group cumulative \
    --survey sv2 --program dark \
    --outfile ztile-sv2-dark-cumulative-pr.fits
```

fitsdiff verifies that these files are the same except for the CHECKSUM/DATASUM comment timestamps, and the CHECKSUM value (due to the differing timestamps leading to differing CHECKSUMs).

Detail: main stacks the tiles in whatever order they appear in the `--tiles TILES` option; this PR stacks them in whatever order they appear in the `$DESI_SPECTRO_REDUX/$SPECPROD/tiles-$SPECPROD.fits` file.  In this case I made them match, but in general they could be in a different order.  The `compare.py` script in the same directory re-sorts before comparing, which also passes if the outputs are in a different order.

Other example outputs tests stacking healpix and arbitrary input directories using
```
desi_zcatalog --add-units --recoadd-fibermap --nproc 16 \
    --group healpix \
    --indir $CFS/desi/spectro/redux/iron/healpix/sv2/dark \
    --outfile zpix-sv2-dark-pr.fits

desi_zcatalog --add-units --nproc 16 \
    --indir $CFS/desi/users/desi/test/perlmutter/ptest-250 \
    --outfile ztile-ptest-250.fits
```

@weaverba137 I believe this is the last PR I need before making a tag and regenerating iron redshift catalogs for more detailed data review.  Please check.

### DIFF
--- a/bin/desi_zcatalog
+++ b/bin/desi_zcatalog
@@ -138,11 +138,25 @@ def load_sv1_ivar_w12(hpix, targetids):
     return targets
 
 def _wrap_read_redrock(optdict):
+    """read_redrock wrapper to expand dictionary of named args for multiprocessing"""
     return read_redrock(**optdict)
 
 def read_redrock(rrfile, group=None, recoadd_fibermap=False, minimal=False, pertile=False, counter=None):
     """
-    TODO: document
+    Read a Redrock file
+
+    Args:
+        rrfile (str): full path to redrock filename
+
+    Options:
+        group (str): add group-specific columns for cumulative, pernight, healpix
+        readcoadd_fibermap (bool): recoadd fibermap from spectra file in same dir
+        minimal (bool): only propagate minimal subet of columns
+        pertile (bool): input Redrock file is single tile (not healpix)
+        counter (tuple): (i,n) log loading ith file out of n
+
+    Returns (zcat, expfibermap) where zcat is a join of the redrock REDSHIFTS
+    catalog and the coadded FIBERMAP
     """
     log = get_logger()
     if counter is not None:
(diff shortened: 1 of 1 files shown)


weaverba137:
I will take a look and hopefully attempt a database test-load later this week.

weaverba137:
@sbailey, go ahead and merge when ready.  I took a look at the sample files and they seem fine.

Issue #2119, opened 2023-09-20
night_qa.py gives error if missing metadata


abhi0395:
Hi @araichoor 
night_qa.py is giving error if the metadata is missing. 
Context: While running the nightqa for 20230917, I got the following error: 

```
Traceback (most recent call last):
 File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/desispec/main/bin/desi_night_qa", line 208, in <module>
  main()
 File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/desispec/main/bin/desi_night_qa", line 140, in main
  expids, tileids, surveys = get_surveys_night_expids(args.night)
 File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/desispec/main/py/desispec/night_qa.py", line 108, in get_surveys_night_expids
  if hdr["OBSTYPE"] == "SCIENCE":
 File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/conda/lib/python3.10/site-packages/fitsio/header.py", line 354, in __getitem__
  raise KeyError("unknown record: %s" % item)
KeyError: 'unknown record: OBSTYPE'
```

When Anthony tried to debug, he found that it is crashing because of a bad exposure with a bunch of missing metadata:
 `/global/cfs/cdirs/desi/spectro/data/20230917/00196681`

I guess it should be relatively easy to check in get_surveys_night_expids()  if `"OBSTYPE" `is in` hdr`, then only proceed with the nightqa; otherwise, go to the next exposure or break. Could you please look into it.

PR #2120, opened 2023-09-20
Night_qa: get_surveys_night_expids() : protect against missing OBSTYPE keyword

araichoor:
This PR should address https://github.com/desihub/desispec/issues/2119.
In such cases, the new code throws a `log.warning()`, ignores the problematic exposure, and proceeds to the next one.
sounds good?


abhi0395:
looks good to me, thanks @araichoor

PR #2118, opened 2023-09-19
fix FIRSTNIGHT/LASTNIGHT when missing petals

sbailey:
This PR moves the FIRSTNIGHT,LASTNIGHT calculation from `desispec.coaddition.coadd_fibermap` to `desi_zcatalog` (the script the stacks N>>1 redrock files into the final redshift catalogs) to fix a problem on tiles that are missing different petals on different nights.  Although `coadd_fibermap` was correctly considering all nights in its input fibermap, the problem is that those per-petal input fibermaps can have a different set of nights per petal, resulting in different FIRSTNIGHT,LASTNIGHT values per-petal when they are supposed to be per-tile values.  This is especially important for LASTNIGHT which is used to trace back to the original spectra in tiles/cumulative/TILEID/LASTNIGHT/spectra-PETAL-*.fits.gz (same LASTNIGHT for all PETALs, even if some PETAL wasn't included on that LASTNIGHT)

For example, tile 80730 was missing petal 3 on the first night it was observed (20210218), and missing petal 6 on the last night it was observed (20210322), resulting in a mix of FIRSTNIGHT, LASTNIGHT values different per petal.

Example outputs are in $CFS/desi/users/sjbailey/dev/coadd_lastnight:
  * spectra*.fits.gz - original spectra from iron
  * coadd-*-iron.fits - original coadded spectra from iron
  * coadd-*-main.fits - coadded spectra from current main
  * coadd-*-pr.fits - coadded spectra from this PR
  * zcat-80730-iron.fits - stacked redshift catalog using iron-era code (release 23.1, desispec/0.56.5)
  * zcat-80730-main.fits - stacked redshift catalog from main
  * zcat-80730-pr.fits - stacked redshift catalog from this PR

Issues / differences to note:
  * coadd-*-iron.fits FIBERMAP does *not* have FIRSTNIGHT/LASTNIGHT
  * coadd-*-main.fits FIBERMAP has FIRSTNIGHT/LASTNIGHT, but they are different
    for each file
  * coadd-*-pr.fits FIBERMAP does *not* have FIRSTNIGHT/LASTNIGHT
    (like the original iron coadded FIBERMAPs)
  * zcat-80730-iron.fits no FIRSTNIGHT, has single LASTNIGHT=20210322
  * zcat-80730-main.fits multiple values of both FIRSTNIGHT and LASTNIGHT
  * zcat-80730-pr.fits FIRSTNIGHT=20210218 and LASTNIGHT=20210322, matching
[...]

PR #2121, opened 2023-09-22
zcat cleanup

sbailey:
This PR does some cleanup to summary redshift catalog creation:
  * Propagate `desi_zcatalog --survey ...` and `--program ...` options into SURVEY and PROGRAM header keywords, instead of requiring them to also be specified via `--header SURVEY=... PROGRAM=...`
  * Safer propagation of DEPNAMnn/DEPVERnn keywords, which previously could have been wrong if one of the inputs had a different set of dependency keywords that another (e.g. DEPNAM00='python' in one but DEPNAM01='python' in another)
  * Explicitly *don't* propagate CHECKSUM and DATASUM from inputs, but recalculate them at end
  * Silence nanomaggy units warnings (previously it was silencing nanomaggies warnings, but not nanomaggie or nanomaggy)
  * write zall catalogs via a tempfilename in case it is interrupted while writing, to not leave behind corrupted file with final name
  * When sorting final catalog include TARGETID or FIBER to ensure unique sort
  * Update `desispec.zcatalog.create_summary_catalog` interface to support creating catalogs from arbitrary input directories instead of assuming that they are in $DESI_SPECTRO_REDUX/$SPECPROD/zcatalog/.  This is useful for both testing, and future productions which will have them in zcatalog/v0, zcatalog/v1 etc.

Example outputs are in `$DESI_ROOT/users/sjbailey/dev/zcat_cleanup`, generated with

```
export SPECPROD=iron
for PROGRAM in dark bright backup; do
    desi_zcatalog --nproc 8 --add-units --group cumulative --survey sv2 --program $PROGRAM \
        -o ztile-sv2-$PROGRAM-cumulative.fits
done

# from python
from desispec.zcatalog import create_summary_catalog
create_summary_catalog(indir='.', specgroup='ztile')
```

Note that I included `--add-units` to test units propagation, but I didn't use `--recoadd-fibermap` so that exact columns will differ from the final DR1 patched redshift catalogs.

This PR plus #2118 are needed for generating the next round of patched DR1 redshift catalogs for vetting.


py/desispec/io/__init__.py
```diff
@@ -10,7 +10,8 @@
 # The line above will help with 2to3 support.
 
 import warnings
-warnings.filterwarnings('ignore', message="'.*nanomaggies.* did not parse as fits unit.*")
+warnings.filterwarnings('ignore', message="'.*nanomaggie.* did not parse as fits unit.*")
+warnings.filterwarnings('ignore', message="'.*nanomaggy.* did not parse as fits unit.*")
 warnings.filterwarnings('ignore', message=".*'10\*\*6 arcsec.* did not parse as fits unit.*")
```
weaverba137, line 15: This isn't a change that you made, but I'm puzzled by this: where does `10**6 arcsec` come from?

py/desispec/zcatalog.py
```diff
@@ -193,35 +196,49 @@ def create_summary_catalog(specprod, specgroup = 'zpix', \
...
     valid_specgroups = ('zpix', 'ztile')
     if specgroup not in valid_specgroups:
         errmsg = f'{specgroup=} not recognized, should be one of {valid_specgroups}'
         log.error(errmsg)
         raise ValueError(errmsg)
 
-    ######################################################################################
+    ## set indir if needed
+    if indir is None:
+        indir = specprod_root(specprod) + '/zcatalog'
```
weaverba137, line 210: You know, if you're going to join paths this way, you might as well do it with `PathLib`.

py/desispec/zcatalog.py
```diff
@@ -232,14 +249,33 @@ def create_summary_catalog(specprod, specgroup = 'zpix', \
...
     for filename in zcat:
         basefile = os.path.basename(filename)
 
         ## Load the ZCATALOG table, along with the meta data
-        log.debug(f'Reading {filename}')
+        log.info(f'Reading {filename}')
         t = Table.read(filename, hdu = 'ZCATALOG')
 
+        ## Merge DEPNAMnn and DEPVERnn, then remove from header
+        desiutil.depend.mergedep(t.meta, dependencies)
```
weaverba137, line 264: It looks like this code is exactly duplicated in the `collect_dependencies` function in `bin/desi_zcatalog`. Why does identical code need to be in two different places?

weaverba137 (changes requested):
Mostly looks fine, but I'm concerned about some possible code duplication.

In general it looks like this script is moving away from the idea of defining scripts in `py/desispec/scripts` and then having a minimal import of `main()` in the executable.  This makes it much more difficult to write tests.

py/desispec/io/__init__.py
sbailey: This was added as part of PR #717 to update to a new fibermap format, but I don't know what had/has that oddly scaled unit.

weaverba137: OK, noted. Fine to leave it for the purposes of this PR.

py/desispec/zcatalog.py
weaverba137: And you don't need to take the comment above particularly seriously; don't hold up the PR.

sbailey: It's admittedly a matter of taste/familiarity, but my complaint with pathlib.Path is that it overrode the division operator '/' to chain things together rather than '+' like strings have done since time immemorial.  My brain trips over whether the slash is incorrectly trying to divide strings, or whether it is a typo and was meant to be within a quotes like this case, or whether it actually is doing something fancy with overriding the / operator and relying upon left-to-right evaluation like `pathlib.Path('blat') / 'foo'`, which is especially confusing to me when it is `importlib_resources.files('blat') / 'foo'` where I have to independently know that `importlib_resources.files` is returning a `pathlib.Path` object with those properties.

i.e. I'm going to leave this as-is and declare it ok to assume that we will always be running DESI code on a posix-like filesystem with '/' as the directory hierarchy separator.

py/desispec/zcatalog.py
sbailey: good point; desihub/desiutil#202 will add functions to desiutil.depend to support removing dependencies to avoid the repeated code here.  I'll update this PR after that is merged.

sbailey:
> In general it looks like this script is moving away from the idea of defining scripts in py/desispec/scripts and then having a minimal import of main() in the executable. This makes it much more difficult to write tests

yeah...  This script used to be a lot simpler of just reading N files, stacking the results, and writing back out.  But every update also stayed within the script to make the functional diff more obvious, since that gets completely lost with a refactor into `desispec.scripts`.  There's even a note at the top of this script about that...

To avoid this problem in the future, after this PR converges on what the script should do, I'll refactor it into desispec.scripts to have it ready for the next time.

sbailey:
I moved bin/desi_zcatalog contents into desispec.scripts.zcatalog, leaving behind a lightweight wrapper.  This created merge conflicts due to PR #2118, rebasing + hand edits fixed that, but the final state diff of this PR is a bit of a mess, including changes due to PR #2118.  Lesson learned: I should have merged this PR as-is, and then separately done the bin -> scripts move.  I have verified that I can create and merge zcatalogs with the final state of this branch, so will merge now and proceed with tags and generating new draft DR1 zcatalogs.

PR #2124, opened 2023-10-06
Add script that summarizes info about an exposure in a specprod

akremin:
This is still a work in progress. I developed this tool to summarize info about an exposure from sources I thought were relevant. It includes showing the raw data location, the preproc and exposures directories, the pipeline exposure_table entry, the summary `exposures-SPECPROD.csv` entry, the surveyops `exposures.ecsv` file, and the entry in the workflow dashboard.

### DIFF
--- a/bin/desi_exposure_info
+++ b/bin/desi_exposure_info
@@ -20,20 +20,30 @@ from desiutil.log import get_logger
 from desispec.parallel import stdouterr_redirected
 import json
 
-p = argparse.ArgumentParser()
-p.add_argument('-e', '--expid', type=int, required=True,
-        help='Exposure ID to get information about')
-p.add_argument('-n', '--night', type=int, required=False,
-        help='night to get information about')
-#p.add_argument('-t', '--tileid', type=int, required=False,
-#        help='Tile ID to get information about')
-p.add_argument('-s', '--specprod', type=str, required=False,
-        help=('override $SPECPROD, or full path '
-              'to override $DESI_SPECTRO_REDUX/$SPECPROD'))
 
 
+## Create arg parser
+def get_parser():
+    p = argparse.ArgumentParser()
+    p.add_argument('-e', '--expid', type=int, required=True,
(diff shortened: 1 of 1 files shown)


sbailey:
Overall looks useful, though it looks like the --tileid option doesn't work.  I was hoping that it could be specified instead of --expid, in which case it would list all exposures for that tile on that night (or ideally without even having to specify --night), but the script still requires --expid to be set.  And it lets you specify an --expid that is inconsistent with a --tileid without complaining (it doesn't find the redshift data, but it says "Expected? True").

Examples:
```
desi_exposure_info -n 20231011 -e 200448
--> tileid 10057, which was also observed on expid 200447, ok

desi_exposure_info -n 20231011 -t 10057
--> says -e/--expid is required

desi_exposure_info -n 20231011 -e 200448 -t 1234
--> inconsistent tileid request, but no errors
```

akremin:
Right, I had originally wanted to write this to include use cases for looping over exposures for `--tileid` or even a `--night`, but those required more effort and if/else structure than I was willing to put in given my schedule at the moment and the fact that I didn't need it for my use case.

The intention of the script as it exists today is to get information about an exposure. Providing tile or night information reduces the amount of effort it takes to locate the exposure, which is the current purpose of those inputs. For now, if they are unknown or uncertain, the recommendation would be to not specify them. (In my first commit I removed the tileid as unnecessary, but that broke things so I've added it back in).

I have thought about the fact that an incorrect `--tileid` or `--night` would be accepted without checking, but for a first prototype, I didn't think it was malicious to assume the provided information was correct. Ideally, if it isn't then the user will get less information than they expected, and they would notice and correct it.

If you consider either of these things to be a dealbreaker, I can improve the relevant area before merging.

akremin:
Thank you for the review, Stephen. For now I've eliminated the `--tileid` option and have made the `--night` option more clear. I also improved the night handling to account for instances where the wrong night is specified. Now it will raise an error.

PR #2127, opened 2023-10-09
Minor PR: add option bin/desi_average_flux_calibration

julienguy:
This is really a tiny difference that does not affect the pipeline:
Add option to desi_average_flux_calibration to remove flatfield correction.

### DIFF
--- a/bin/desi_average_flux_calibration
+++ b/bin/desi_average_flux_calibration
@@ -42,10 +42,10 @@ def get_ffracflux_wave(seeing, ffracflux, wave, fac_wave_power):
     Returns:
         Fiber acceptance for seeing, normalized to ffracflux at 6500A
 
-    
+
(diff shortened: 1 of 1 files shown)


sbailey:
Looks good, but I made two small changes:
  * use `desispec.io.util.replace_prefix` instead of `filename.replace(...)`.  Simply using `replace` breaks if you have a specprod with a name like "test-fluxcalib" (we've encountered variants of this problem multiple times with other filename manipulation, thus the existence of replace_prefix to replace only the filename prefix and not any other places where it appears in the path or filename)
  * print -> log.info

Since I don't know example inputs to test this, please retry this branch with your work to make sure I didn't introduce a typo, then ok to merge.

julienguy:
I checked your change. Thanks.

PR #2129, opened 2023-10-09
Set psferr=0.04 for blue cameras

julienguy:
Set psferr=0.04 instead of 0.01 for the blue cameras to avoid having many spectral pixels with the ```specmask.BAD2DFIT``` bit set.
(in the call ```desi_extract_spectra --psferr xxx``` in ```desi_proc```).
This value has been tuned on exposure 197796 from night 20230923 and checked on exposure 192356 from night 20230824.

The psferr parameter increases the pixel variance by (psferr*model_flux)**2 in the chi2 of the 2D spectral extraction. It is used to account for modeling uncertainties. The value of 0.01 was not large enough in the blue cameras. It results in large chi2 on the brightest sky lines where psf errors dominate the noise.

The following plot shows the sum of masked spectral pixels per camera (among 500) as a function of wavelength for various psferr values. For psferr=0.04 most pixels are not masked. Larger values increase the risk to miss actual bad 2D fit from cosmic rays or CCD defects. One can see that indeed other pixels are unmasked (3 pixels for exposure 197796 and 1 for exposure 192356).

frame-b-00197796

frame-b-00192356

A few other pixels are also unmasked:

frame-b-00197796-bis

frame-b-00192356-zoom

This PR addresses the issue https://github.com/desihub/desisurveyops/issues/135 .

### DIFF
--- a/py/desispec/scripts/proc.py
+++ b/py/desispec/scripts/proc.py
@@ -725,7 +725,13 @@ def main(args=None, comm=None):
                 cmd += ' -i {}'.format(preprocfile)
                 cmd += ' -p {}'.format(psffile)
                 cmd += ' -o {}'.format(framefile)
-                cmd += ' --psferr 0.01'
+
+                #- Larger PSF model uncertainty for the blue cameras because a lower value
+                #- results in many pixels with specmask.BAD2DFIT on the 5578A sky line.
+                if camera.startswith('b'):
+                    cmd += ' --psferr 0.04'
+                else :
+                    cmd += ' --psferr 0.01'
 
                 if args.use_specter:
                     cmd += ' --use-specter'


julienguy:
The masked/unmasked spectral pixel at 4455A is indeed caused by the edge of a cosmic ray in b6 fiber 35 that is not masked by the cosmic finder at the point where it intersect the spectral trace. So there is some penalty in increasing this psferr threshold.
Screenshot from 2023-10-09 12-08-17

julienguy:
But this is a very small fraction of all the spectral pixels that are correctly masked by the cosmic ray finder. So I would merge this and maybe open a new ticket to improve further cosmic ray rejection.

sbailey:
I tested this PR vs. main on 99 tiles randomly selected over 88 main survey nights, while requiring that the tile was observed only on that night to avoid multi-night bookkeeping headaches, but this does include a mix of single-exposure and multi-exposure tiles.  Productions are in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/psferr-main and psferr-pr/, with combined redshift catalogs in /global/cfs/cdirs/desi/users/sjbailey/dev/psferr/redrock-main.fits and redrock-pr.fits.

Only 16 out of 494k targets change ZWARN.  Of the 374k targets where both prods have ZWARN=0, only 199 differ by more than 1 km/s (mostly QSOs), and only 1 is a >500 km/s difference with a different SPECTYPE.  That is TARGETID=2305843018277596717 on TILEID=40470 NIGHT=20220613 FIBER=4349 which has a major cosmetic problem and neither main nor this PR get it right (I'm not sure what the right answer is, but it isn't negative Lyman-alpha nor negative [OIII].)

i.e. I don't see any evidence that the reduced extraction masking is leading to unmasked cosmic rays in the final spectra leading to an increased number of bad spectra.  The original post provides the examples where the psferr change in this PR helps.  Looks good, merging now.

PR #2130, opened 2023-10-18
Update calibfinder.py

Waelthus:
This makes the error if new darks are not available a warning instead

### DIFF
--- a/py/desispec/calibfinder.py
+++ b/py/desispec/calibfinder.py
@@ -525,4 +525,4 @@ class CalibFinder() :
             self.data.update({"DARK": dark_filename,
                               "BIAS": bias_filename})
         else:
-            log.error(f"Didn't find matching {camera} calibration darks in $DESI_SPECTRO_DARK using default from $DESI_SPECTRO_CALIB instead")
+            log.warning(f"Didn't find matching {camera} calibration darks in $DESI_SPECTRO_DARK using default from $DESI_SPECTRO_CALIB instead")


sbailey:
On one hand, this is a cosmetic change to the logs and is fine.  But I want to make sure we're in agreement about calibfinder and $DESI_SPECTRO_DARK expected behavior, regardless of how we log it --

Normally an "error" is something that is not supposed to happen, but isn't fatal so the code can continue ("critical" is what is used for "this is so bad I have to stop").  A "warning" is something that isn't good, but is expected to happen occasionally, e.g. due to a statistical fluctuation in data quality.  I thought that not finding a dark model in $DESI_SPECTRO_DARK would be an "error" -- that's not supposed to happen, right?  i.e. it signals that something needs to be fixed, even if the code is temporarily falling back to default calibs.  Or otherwise what is the context/motivation behind renaming this "error" to a "warning"?

FYI https://docs.python.org/3/library/logging.html#logging-levels

Waelthus:
It can sometimes happen, e.g. when configs change and new darks have been made by Julien in DESI_SPECTRO_CALIB, or for e.g. pre-fuji data. I guess it should never happen anymore after the planned updates to how calibfinder works...

The reason was basically that the code still runs fine and still picks up the most reasonable dark frames to take and the change was due to https://desisurvey.slack.com/archives/C011JS5GW5U/p1697637710184689. That said I'm not yet fully sure if it should fall back to calib at that particular point mentioned in the slack discussion. So after all maybe keeping this an error is the more reasonable thing to do...

PR #2137, opened 2023-11-03
fix read_spectra skipping exp_fibermap

sbailey:
This PR is a followup to #2052, fixing a bug with `read_spectra(..., skip_hdus=['EXP_FIBERMAP', ...])`

The original code looping over HDUs had
```python
...
elif name == "EXP_FIBERMAP" and name not in skip_hdus:
    ...
elif ...
```
The problem was that if "EXP_FIBERMAP" was in skip_hdus, then it would continue down the elif chain and eventually get interpreted as a catchall user-provided "BAND_EXTRANAME" image HDU, which would then trip on reading a table as an image.

The fix is
```python
...
elif name == "EXP_FIBERMAP":
    if name not in skip_hdus:
        ...
elif ...
```
so that if it finds "EXP_FIBERMAP" but it is in `skip_hdus` then it will stop checking rather than continuing to try to parse the name as meaning other things.  I applied that same pattern to all other cases of known HDU names that might be in the skip_hdu list.

This includes a unit test that would have caught the original problem, but now passes.

@dirkscholte I'm pretty sure that I introduced this bug when updating your PR #2052, but if you could double-check my fix here I would appreciate it.  Thanks.

### DIFF
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -310,36 +310,40 @@ def read_spectra(
     for h in range(1, nhdu):
         name = hdus[h].read_header()["EXTNAME"]
         log.debug('Reading %s', name)
-        if name == "FIBERMAP" and name not in skip_hdus:
-            fmap = encode_table(
-                Table(
-                    hdus[h].read(rows=rows, columns=select_columns["FIBERMAP"]),
-                    copy=True,
-                ).as_array()
-            )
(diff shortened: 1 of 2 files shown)


dirkscholte (approved):
Hi @sbailey. This makes a lot of sense! In the earliest version I made of this (might not be the version that started this pull request) I purposefully broke out some functionality because I had a similar issue. Thanks for fixing this :)

PR #2139, opened 2023-11-03
support coadd_cameras without resolution

sbailey:
This PR adds support for running `coadd_cameras` on a Spectra object that purposefully doesn't have resolution data.  It adds a unit test that previously failed but now passes.


sbailey:
Self merging because I a) need this functionality and b) included additional unit tests testing the fix.  Post-facto review/comments are welcome if someone has the time.

PR #2141, opened 2023-11-15
Spectra to specutils

weaverba137:
This PR adds methods to convert `desispec.spectra.Spectra` to and from `specutils.SpectrumList`.

For testing purposes, `specutils` is only installed for the coverage test.

### DIFF
--- a/py/desispec/spectra.py
+++ b/py/desispec/spectra.py
@@ -139,9 +139,12 @@ class Spectra(object):
                 if resolution_data[b].shape[2] != wave[b].shape[0]:
                     raise RuntimeError("resolution array wavelength dimension for band {} does not match grid".format(b))
             if extra is not None:
-                for ex in extra[b].items():
-                    if ex[1].shape != flux[b].shape:
-                        raise RuntimeError("extra arrays must have the same shape as the flux array")
+                try:
+                    for ex in extra[b].items():
+                        if ex[1].shape != flux[b].shape:
+                            raise RuntimeError("extra arrays must have the same shape as the flux array")
+                except (KeyError, AttributeError):
+                    pass
 
         if fibermap is not None and extra_catalog is not None:
             if len(fibermap) != len(extra_catalog):
@@ -203,7 +206,6 @@ class Spectra(object):
                 self.extra[b] = {}
                 for ex in extra[b].items():
                     self.extra[b][ex[0]] = np.copy(ex[1].astype(self._ftype))
-
 
     @property
     def bands(self):
--- a/py/desispec/spectra.py
+++ b/py/desispec/spectra.py
@@ -19,11 +19,11 @@ import numbers
 
 import numpy as np
 from astropy.table import Table
+from astropy.units import Unit
 
 _specutils_imported = True
 try:
     from specutils import SpectrumList, Spectrum1D
-    from astropy.units import Unit
     from astropy.nddata import InverseVariance, StdDevUncertainty
 except ImportError:
     _specutils_imported = False
@@ -79,6 +79,9 @@ class Spectra(object):
         optional table of metadata, rowmatched to fibermap,
         e.g. a redshift catalog for these spectra
     """
+    wavelength_unit = Unit('Angstrom')
+    flux_density_unit = Unit('10-17 erg cm-2 s-1 AA-1')
+
     def __init__(self, bands=[], wave={}, flux={}, ivar={}, mask=None,
             resolution_data=None, fibermap=None, exp_fibermap=None,
             meta=None, extra=None,
@@ -682,13 +685,11 @@ class Spectra(object):
         if not _specutils_imported:
             raise NameError("specutils is not available in the environment.")
         sl = SpectrumList()
-        AA = Unit('Angstrom')
-        specunit = Unit('10-17 erg cm-2 s-1 AA-1')
         for i, band in enumerate(self.bands):
             meta = {'band': band}
-            spectral_axis = self.wave[band] * AA
-            flux = self.flux[band] * specunit
-            uncertainty = InverseVariance(self.ivar[band] * (specunit**-2))
+            spectral_axis = self.wave[band] * self.wavelength_unit
+            flux = self.flux[band] * self.flux_density_unit
+            uncertainty = InverseVariance(self.ivar[band] * (self.flux_density_unit**-2))
             mask = self.mask[band] != 0
             meta['int_mask'] = self.mask[band]
             meta['resolution_data'] = self.resolution_data[band]
@@ -774,16 +775,14 @@ class Spectra(object):
         mask = dict()
         resolution_data = None
         extra = None
-        AA = Unit('Angstrom')
-        specunit = Unit('10-17 erg cm-2 s-1 AA-1')
         for i, band in enumerate(bands):
-            wave[band] = sl[i].spectral_axis.to(AA).value
-            flux[band] = sl[i].flux.to(specunit).value
+            wave[band] = sl[i].spectral_axis.to(cls.wavelength_unit).value
+            flux[band] = sl[i].flux.to(cls.flux_density_unit).value
             if isinstance(sl[i].uncertainty, InverseVariance):
-                ivar[band] = sl[i].uncertainty.quantity.to(specunit**-2).value
+                ivar[band] = sl[i].uncertainty.quantity.to(cls.flux_density_unit**-2).value
             elif isinstance(sl[i].uncertainty, StdDevUncertainty):
                 # Future: may need np.isfinite() here?
-                ivar[band] = (sl[i].uncertainty.quantity.to(specunit).value)**-2
+                ivar[band] = (sl[i].uncertainty.quantity.to(cls.flux_density_unit).value)**-2
             else:
                 raise ValueError("Unknown uncertainty type!")
             try:
--- a/py/desispec/spectra.py
+++ b/py/desispec/spectra.py
@@ -698,20 +698,20 @@ class Spectra(object):
             meta['resolution_data'] = self.resolution_data[band].copy()
             try:
                 meta['extra'] = self.extra[band].copy()
-            except (KeyError, TypeError):
-                meta['extra'] = None
+            except (KeyError, TypeError, AttributeError):
+                pass
             if i == 0:
                 #
                 # Only add these to the first item in the list.
                 #
                 meta['bands'] = self.bands
-                meta['fibermap'] = self.fibermap.copy()
-                meta['exp_fibermap'] = self.exp_fibermap.copy()
-                meta['desi_meta'] = self.meta.copy()
                 meta['single'] = self._single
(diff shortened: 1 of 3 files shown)


py/desispec/spectra.py
```diff
@@ -658,6 +666,168 @@ def update(self, other):
...
+                meta['scores'] = self.scores
+                meta['scores_comments'] = self.scores_comments
+                meta['extra_catalog'] = self.extra_catalog
+            sl.append(Spectrum1D(flux=flux, spectral_axis=spectral_axis,
+                                 uncertainty=uncertainty, mask=mask, meta=meta))
+        return sl
+
+    @classmethod
+    def from_specutils(cls, spectra):
+        """Convert ``specutils`` objects to a :class:`~desiutil.spectra.Spectra` object.
```
akremin, line 717: `desiutil` -> `desispec`?

py/desispec/spectra.py
```diff
@@ -658,6 +666,168 @@ def update(self, other):
...
+
+        Parameters
+        ----------
+        spectra : specutils.Spectrum1D or specutils.SpectrumList
+            A ``specutils`` object.
+
+        Returns
+        -------
+        :class:`~desiutil.spectra.Spectra`
```
akremin, line 726: `desiutil` -> `desispec`?

py/desispec/spectra.py
```diff
@@ -658,6 +666,168 @@ def update(self, other):
...
+        """
+        if not _specutils_imported:
+            raise NameError("specutils is not available in the environment.")
+        if isinstance(spectra, SpectrumList):
+            sl = spectra
+            try:
+                bands = sl[0].meta['bands']
+            except KeyError:
+                #
+                # This is a big assumption; it doesn't capture ['b', 'z'] or ['r', 'z'].
```
akremin, line 744: Before jumping to this assumption, could you look over the elements of the list and check if their meta has `'band'` (singular)? If they did then you could build bands pretty easily, and if not then the information truly isn't available.

weaverba137:
PS, I'm still working on improving the test *coverage*, so this is not immediately ready to merge.

akremin (commented):
This looks great I left two comments on a docstring that may or may not be a typo and then one suggestion on something you acknowledge was an assumption.

And for the record I haven't tested the code, this is just a code review.

py/desispec/spectra.py
```diff
@@ -658,6 +666,168 @@ def update(self, other):
...
+        #
+        wave = dict()
+        flux = dict()
+        ivar = dict()
+        mask = dict()
+        resolution_data = None
+        extra = None
+        for i, band in enumerate(bands):
+            wave[band] = sl[i].spectral_axis.value
+            flux[band] = sl[i].flux.value
```
sbailey, line 800: as long as these are unit-ful items that you have to extract the non-unit values, consider explicitly converting them to `AA` and `specunit` before taking `.value`, just in case they are provided in some other units.

sbailey (commented):
Conceptually looks good, with the caveat that I haven't tried running it (leaving that to your tests for the moment).  I added comments inline.  I'm hoping that in the future Spectrum1D will have more natural support for custom columns, so I think we should consider this a temporary mapping of putting everything into the .meta header, but this looks like the best we can do given the current Spectrum1D.

weaverba137:
@sbailey, is it possible to manually align wavelength values for the test data in `test_spectra.py`? When I try to apply `coadd_cameras()` it complains that the wavelengths are not aligned.

py/desispec/spectra.py
weaverba137: Good catch, fixed.

py/desispec/spectra.py
weaverba137: Good catch, fixed.

py/desispec/spectra.py
```diff
@@ -658,6 +666,168 @@ def update(self, other):
...
+            raise NameError("specutils is not available in the environment.")
+        if isinstance(spectra, SpectrumList):
+            sl = spectra
+            try:
+                bands = sl[0].meta['bands']
+            except KeyError:
+                #
+                # This is a big assumption; it doesn't capture ['b', 'z'] or ['r', 'z'].
+                #
+                bands = ['b', 'r', 'z'][0:len(sl)]
```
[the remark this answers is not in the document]
weaverba137: @akremin's suggestion above is maybe to simply raise an exception if it is not possible to determine the bands. If I can quickly figure out a wavelength range solution, that's great, but if it takes too much time, I will fall back to an exception.

py/desispec/spectra.py
weaverba137: Moving discussion to @sbailey's comment below.

py/desispec/spectra.py
```diff
@@ -658,6 +666,168 @@ def update(self, other):
...
+            meta = {'band': band}
+            spectral_axis = self.wave[band] * AA
+            flux = self.flux[band] * specunit
+            uncertainty = InverseVariance(self.ivar[band] * (specunit**-2))
+            mask = self.mask[band] != 0
+            meta['int_mask'] = self.mask[band]
+            meta['resolution_data'] = self.resolution_data[band]
+            try:
+                meta['extra'] = self.extra[band]
+            except KeyError:
```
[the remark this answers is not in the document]
weaverba137: Acknowledged.

py/desispec/spectra.py
```diff
@@ -658,6 +666,168 @@ def update(self, other):
...
+                bands = ['brz']
+        else:
+            raise ValueError("Unknown type input to Spectra.from_specutils!")
+        #
+        # Load objects that are independent of band from the first item.
+        #
+        try:
+            fibermap = sl[0].meta['fibermap']
+        except KeyError:
+            fibermap = None
```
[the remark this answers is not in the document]
weaverba137: Acknowledged.

py/desispec/spectra.py
weaverba137: Good point, I think that's easy.

py/desispec/spectra.py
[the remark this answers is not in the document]
weaverba137: Done now.

py/desispec/spectra.py
weaverba137: Done now.

py/desispec/spectra.py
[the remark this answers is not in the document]
weaverba137: Done now.

weaverba137:
Please ignore the previous comment on spectral alignment, I was able to figure out how to do that myself.

py/desispec/spectra.py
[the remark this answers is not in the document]
weaverba137: Based on discussions today, @sbailey & I agree that when converting `specutils` to DESI internal, users really *should* supply band information, and raise an error if they do not.  It would be supplied via `my_specutils_object.meta['bands']`, so that's not hard.

weaverba137:
@sbailey, @akremin, this is ready for another look. In addition to the suggestions above, I tried to wherever possible use a `.copy()` method when converting between DESI Internal and `specutils`.  However, one of the suggestions above, *e.g.*:
```python
fibermap = sl[0].meta.get('fibermap', None)
```
Doesn't really fit into that model, because:
```python
fibermap = sl[0].meta.get('fibermap', None).copy()
```
might throw an exception anyway, since `None` would not have `.copy()`.  Any recommendations here?

weaverba137:
@sbailey, @akremin, I think the next step is for you to test this for yourselves, unless you'd like to merge this right away?

py/desispec/spectra.py
```diff
@@ -128,9 +139,12 @@ def __init__(self, bands=[], wave={}, flux={}, ivar={}, mask=None,
...
             if extra is not None:
-                for ex in extra[b].items():
-                    if ex[1].shape != flux[b].shape:
-                        raise RuntimeError("extra arrays must have the same shape as the flux array")
+                try:
+                    for ex in extra[b].items():
+                        if ex[1].shape != flux[b].shape:
+                            raise RuntimeError("extra arrays must have the same shape as the flux array")
+                except (KeyError, AttributeError):
+                    pass
```
sbailey, line 147: Please explain the case where this generates KeyError or AttributeError but is otherwise ok and should be silently passed.

py/desispec/spectra.py
```diff
@@ -189,10 +203,10 @@ def __init__(self, bands=[], wave={}, flux={}, ivar={}, mask=None,
...
                 self.R[b] = np.array( [ Resolution(r) for r in resolution_data[b] ] )
             if extra is not None:
-                self.extra[b] = {}
-                for ex in extra[b].items():
-                    self.extra[b][ex[0]] = np.copy(ex[1].astype(self._ftype))
-
+                if extra[b] is not None:
+                    self.extra[b] = {}
+                    for ex in extra[b].items():
+                        self.extra[b][ex[0]] = np.copy(ex[1].astype(self._ftype))
```
sbailey, line 209: Perhaps related to the previous comment: please explain the case that requires the additional `if extra[b] is not None` check.

sbailey (commented):
Thanks.  For the most part this is adding new functionality, but I put two comments inline requesting clarification on changes that might impact existing functionality, so I want to make sure those cases are clearly understood before merging.

py/desispec/spectra.py
weaverba137: The `except` is catching errors thrown by line 143 (which is not in the snippet above).

* If `extra` exists, but `extra[b]` does not exist, line 143 will throw `KeyError`.
* If `extra[b]` exists, but is `None`, line 143 will throw `AttributeError`.

py/desispec/spectra.py
weaverba137: One could be even more stringent.  Since `extra` is a dictionary of dictionaries, do we insist that not only do all bands exist, but that every band dictionary has the same keys?

Again, since I didn't see that level of stringency in the original code, I assumed that the definition of `extra` could be "loose".

weaverba137:
@sbailey, @akremin, this has been dangling for a while, could you please re-review?

sbailey:
The original docstring description of `extra`:
```
    extra : :class:`dict`, optional
        Optional dictionary of dictionaries containing extra
        floating point arrays.  The top-level is a dictionary over bands
        and each value is a dictionary containing string keys and values
        which are arrays of the same size as the flux array.
```
was intended to imply that it should include *all* of the bands ("dictionary over bands"; note that `bands` is a member variable of the Spectra object), and that "each value is a dictionary" was intended to mean that it should be a dictionary and not None of any other value.

That being said, we may have a future use case for being more flexible about `extra` so I left your try/except checks in place, and just added an inline comment about what cases they are catching.  I did *not* change the docstring to describe this additional flexibility; if we find we want to use that, let's add more rigorous tests to make sure it really is fully supported (e.g. when slicing, during I/O, etc.).

PR #2151, opened 2023-12-22
create_petalnz_pdf(): look for redrock*fits instead of zmtl*fits

araichoor:
With this PR, the night_qa`create_petalnz_pdf()` now parses the `redrock*fits` files to identify the processed petals, instead of the `zmtl*fits` file.

It addresses https://desisurvey.slack.com/archives/C01HNN87Y7J/p1703203812637849.
I.e. when we archived+mtl-updated a tile, then realized a petal was bad, so we re-process the tile, with creating a "dummy" `zmtl*fits` file (but no redrock or other files).
The `create_petalnz_pdf()` function was looking for `zmtl*fits` files to parse the existing petals.
Here it would find the "dummy" `zmtl*fits` file, but then crash as there s no `redrock*fits` file.

I ve successfully run on night here: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/nightqa_v25/20231209/.

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -27,6 +27,7 @@ from desitarget.geomask import match_to
 # AR desispec
 from desispec.fiberbitmasking import get_skysub_fiberbitmask_val
 from desispec.io import findfile
+from desispec.io.util import replace_prefix
 from desispec.calibfinder import CalibFinder
 from desispec.scripts import preproc
 from desispec.tile_qa_plot import get_tilecov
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks for the quick fix.  As a maintenance detail, I changed `fn.replace('redrock', 'zmtl')` to `desispec.io.util.replace_prefix(fn, "redrock", "zmtl")` which will replace only the prefix of the filename, but not other places where "redrock" might appear, e.g. in a SPECPROD="test_redrock" path.

PR #2155, opened 2024-01-02
remove deprecated test_suite() function from all unit tests that use it

### LINKED ISSUE #2142: test_suite functions are throwing pytest warnings
The many of the `test_*.py` files contain `unittest` boilerplate like this:
```python
def test_suite():
    """Allows testing of only this module with the command::

        python setup.py test -m <modulename>
    """
    return unittest.defaultTestLoader.loadTestsFromName(__name__)

#- This runs all test* functions in any TestCase class in this file
if __name__ == '__main__':
    unittest.main()
```
These are now starting to throw a lot of warnings when used with pytest. The warnings distract from other more serious warnings that should also be addressed. These should simply be removed, since we're not using `python setup.py test` any longer.

moustakas:
Should hopefully fix #2142.

### DIFF
--- a/py/desispec/test/test_binscripts.py
+++ b/py/desispec/test/test_binscripts.py
@@ -353,13 +353,6 @@ for legacy standards
             inputs=inputs, outputs=outputs, clobber=True)
         self.assertTrue(success)
 
-def test_suite():
-    """Allows testing of only this module with the command::
-
-        python setup.py test -m <modulename>
-    """
-    return unittest.defaultTestLoader.loadTestsFromName(__name__)
-
 
 #- This runs all test* functions in any TestCase class in this file
 if __name__ == '__main__':
--- a/py/desispec/test/test_binscripts.py
+++ b/py/desispec/test/test_binscripts.py
@@ -352,8 +352,3 @@ for legacy standards
         result, success = runcmd(desispec.scripts.sky.main, args=args,
             inputs=inputs, outputs=outputs, clobber=True)
         self.assertTrue(success)
-
-
-#- This runs all test* functions in any TestCase class in this file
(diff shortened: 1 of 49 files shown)


py/desispec/test/test_binscripts.py
```diff
@@ -353,13 +353,6 @@ def test_compute_sky(self):
...
-def test_suite():
-    """Allows testing of only this module with the command::
-
-        python setup.py test -m <modulename>
-    """
-    return unittest.defaultTestLoader.loadTestsFromName(__name__)
-
 
 #- This runs all test* functions in any TestCase class in this file
 if __name__ == '__main__':
```
weaverba137, line 358: These sections are also deprecated and should be removed.  The `pytest` equivalent is just:
```
pytest py/desispec/test/test_binscripts.py
```

weaverba137 (changes requested):
The `if __name__` sections should also be removed.

weaverba137:
Thanks to the reduced number of warnings, it is now clear that `numba.jit` warnings should be addressed next.

py/desispec/test/test_binscripts.py
moustakas: Done in https://github.com/desihub/desispec/pull/2155/commits/b4c874dfe82d61b91da9ec5f232d96fee39d4092 but please review that I didn't miss anything or over-deleted things.

moustakas:
Apologies for all the git-noise. I was trying to be helpful by cleaning up some of the other easy warnings but ended up reverting a couple of the changes I made. This PR is ready for final review and, hopefully, merging.

weaverba137 (approved):
This looks good, thank you for fixing the other warnings too.

PR #2156, opened 2024-01-03
fixes/updates to read_tile_spectra

sbailey:
This PR has several bug fixes and updates for `desispec.io.read_tile_spectra`:

**BUGFIX**: fixes the matching of a redshift table (with one row per TARGETID) to a uncoadded spectra fibermap
(with potentially multiple rows per TARGETID in a different order), when filtering by FIBER.  The previous logic
was simply wrong (mea culpa).  I introduced a new `desispec.util.argmatch` function to isolate and test this
matching logic, as well as introducing additional higher-level `read_tile_spectra` tests that would have caught the
original bug but now pass.  Usage: `ii = argmatch(a, b)` returns indices `ii` such that `a[ii] == b` is True for all entries.
  * @geordie666 and @araichoor heads up: `desispec.util.argmatch` is similar to `desitarget.geomask.match_to`, but it supports duplicate entries which `match_to` does not.  I *think* `argmatch` is a superset of the functionality of `match_to`, but it doesn't support the bi-directional matching of `desitarget.geomask.match` (which `match_to` uses under-the-hood), and I wasn't sure if there were other use cases for `match_to` that I was missing in my tests.  We could consider merging these in the future, but for now I introduced this new function in desispec instead of expanding the functionality of `desitarget.geomask.match_to` because I didn't want to risk breaking something there.

**BUGFIX**: fixes case where the spectra files are gzipped, but the redrock files are not.  It was previously assuming that the redrock files had the same level of gzipping or not; now it can be either way by using `checkgzip`.  Unit tests explicitly test this now.

**FEATURES** for `read_tile_spectra`:
  * `night` argument is now optional; will use latest night on disk if not specified.
  * `group` argument defaults to "cumulative" instead of None; the previous default produced paths structures that don't exist.
  * when filtering by fiber, skip reading any petals that aren't covered by those fibers.
    * full disclosure: this uses the filename to derive the petal to know whether to skip the file.  Normally it is bad practice to rely upon the filename to get metadata instead of using the file contents, but the whole point here is to avoid reading unnecessary files, and the function is already assuming that we are reading tiles within the standard datamodel directory structure and names, so I think this is ok.
  * adds functions in `desispec.test.util` for generating fake Spectra objects for testing.

@schlafly I think you were the original author (or advocate?) of read_tile_spectra, or at least a primary user of it, so could you review this PR?  Thanks.

### DIFF
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -35,6 +35,7 @@ from .fibermap import fibermap_comments
 
 from ..spectra import Spectra, stack
 from .meta import specprod_root
+from ..util import argmatch
 
 def write_spectra(outfile, spec, units=None):
     """
@@ -497,17 +498,17 @@ def read_frame_as_spectra(filename, night=None, expid=None, band=None, single=Fa
 
     return spec
 
-def read_tile_spectra(tileid, night, specprod=None, reduxdir=None, coadd=False,
+def read_tile_spectra(tileid, night=None, specprod=None, reduxdir=None, coadd=False,
                       single=False, targets=None, fibers=None, redrock=True,
-                      group=None):
+                      group='cumulative'):
     """
     Read and return combined spectra for a tile/night
 
     Args:
         tileid (int) : Tile ID
-        night (int or str) : YEARMMDD night or tile group, e.g. 'deep' or 'all'
 
     Options:
+        night (int or str) : YEARMMDD night
         specprod (str) : overrides $SPECPROD
         reduxdir (str) : overrides $DESI_SPECTRO_REDUX/$SPECPROD
         coadd (bool) : if True, read coadds instead of per-exp spectra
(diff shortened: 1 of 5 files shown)


schlafly (approved):

sbailey:
Thanks for the fast review.  I do need to specifically support the case of `argmatch(a,b)` having duplicate entries in `b`, because the use case here is matching TARGETIDs in a redshift table (one row per TARGETID) to an uncoadded fibermap which can have multiple rows for each TARGETID.  Supporting duplicates in `a` came along for the ride which seemed harmless, as did supporting `a` having extra entries not present in `b`.  But the opposite is a specific error, since if `b` has entries not in `a`, then no indices `ii` could make `a[ii] == b` .

For the record, `desitarget.geomask.match` (not `match_to`) doesn't support duplicates, but it does support bi-directional matching with extras in either list such that `ii, jj = match(a,b)` -> `a[ii] == b[jj]`.

araichoor:
[I see the PR has been merged while I was typing... I just send my message for completeness]

about `argmatch()`:
as @schlafly, I may be a bit concerned about the handling of duplicates; i.e. it can be risky if the code returns something, but that s not what the user would expect; for ease, you may want to throw an error if duplicates are present.

btw, @geordie666: even though it s written in the docstr that the arguments shouldn t contain duplicates, I always felt we should add such a test of uniqueness in `desitarget.geomask.match()`, and return an error if arguments have duplicates; because, as far as I remember, the current code does not crash, but the returned indexes do not make sense.

moustakas:
> btw, @geordie666: even though it s written in the docstr that the arguments shouldn t contain duplicates, I always felt we should add such a test of uniqueness in desitarget.geomask.match(), and return an error if arguments have duplicates; because, as far as I remember, the current code does not crash, but the returned indexes do not make sense.

Yes! This bit me shortly after I started using `desitarget.geomask.match()` in `FastSpecFit` and it was a little painful to (eventually) track down the issue.

My only comment---too late now that this has been merged---is that a function like this should perhaps have gone into `desiutil` over `desispec`.

schlafly:
My objection is to allowing duplicates in both a & b; it just feels like that demands "undefined behavior" surrounding how you pick which indices map to which indices there.  As long as there are duplicates only in one of the two, things are unambiguous.

geordie666:
@araichoor: Feel free to open an issue in `desitarget` and I'll update the `match` function to include a test for duplicates.

PR #2159, opened 2024-01-05
CCDTEMP checking for DARK

Waelthus:
based on #2158, this adds additional checks for CCDTEMP. 

After merging the creating of dark models based on sets with heterogeneous temperatures will be forbidden explicitely.

When we look for calib files, a check is performed regarding difference of dark model temperature and to-be-reduced frames temperature.

A 1K tolerance is applied in both cases.


sbailey:
Thanks.  This appears to be a superset of branch dark_changes_and_calib_ledger (PR #2158) so I'm going to merge this PR and close that PR without merging.  If I misunderstood the git history network graph and PR 2158 has unique changes not here, please let me know and I'll sort it out.

PR #2165, opened 2024-01-25
add desi_link_calibnight script

sbailey:
New `desi_link_calibnight` script to simplify linking calibnight files from one night to another:
```
usage: desi_link_calibnight [-h] --refnight REFNIGHT --newnight NEWNIGHT [-c CAMERAS]
                            [--include INCLUDE] [--exclude EXCLUDE] [--dryrun]

options:
  --refnight REFNIGHT   Reference night with calibs
  --newnight NEWNIGHT   New night without calibs, to link to refnight
  -c CAMERAS, --cameras CAMERAS
                        Camword of cameras to link, e.g. a0123b6r7z8
  --include INCLUDE     prefixes of types of calibnight files to create links
  --exclude EXCLUDE     prefixes of types of calibnight files to exclude from links
  --dryrun              dry run; don't actually create links
```

By default this links badcolumns,biasnight,fiberflatnight,psfnight files, with a placeholder for ctecorr (to be updated after PR #2163), but the `--include` and `--exclude` options can adjust that.

It *won't* overwrite pre-existing files, unless those file happen to already be links to the same destination anyway.

Includes a full suite of unit tests.  @akremin please review.  Does this match what you need for pipeline calibnight linking?

### DIFF
--- a/bin/desi_link_calibnight
+++ b/bin/desi_link_calibnight
@@ -1,0 +1,7 @@
+#!/usr/bin/env python
+
+import sys
+from desispec.scripts import link_calibnight
+
+sys.exit(link_calibnight.main())
+
--- a/py/desispec/io/meta.py
+++ b/py/desispec/io/meta.py
@@ -173,6 +173,8 @@ def findfile(filetype, night=None, expid=None, camera=None,
         biasnight = '{specprod_dir}/calibnight/{night}/biasnight-{camera}-{night}.fits.gz',
         badfibers =  '{specprod_dir}/calibnight/{night}/badfibers-{night}.csv',
         badcolumns = '{specprod_dir}/calibnight/{night}/badcolumns-{camera}-{night}.csv',
+        ctecorrnight = '{specprod_dir}/calibnight/{night}/ctecorr-{night}.csv',
+        ctecorr      = '{specprod_dir}/calibnight/{night}/ctecorr-{night}.csv', #- alias, same file
         #
         # spectra- healpix based
         #
--- a/py/desispec/scripts/link_calibnight.py
+++ b/py/desispec/scripts/link_calibnight.py
@@ -1,0 +1,138 @@
+
+import os, sys
+import argparse
+
+from desiutil.log import get_logger
+
+from desispec.io import findfile
+from desispec.io.util import decode_camword, relsymlink
+
+def parse(options=None):
+    """parse options from sys.argv or input list of options"""
+    p = argparse.ArgumentParser()
+    p.add_argument('--refnight', type=int, required=True,
+                   help='Reference night with calibs')
+    p.add_argument('--newnight', type=int, required=True,
+                   help='New night without calibs, to link to refnight')
+    p.add_argument('-c', '--cameras', type=str, default='a0123456789',
+                   help='Camword of cameras to link, e.g. a0123b6r7z8')
+    p.add_argument('--include', type=str, 
+                   default='badcolumns,biasnight,fiberflatnight,psfnight,ctecorr',
+                   help='prefixes of types of calibnight files to create links')
+    p.add_argument('--exclude', type=str, 
+                   help='prefixes of types of calibnight files to exclude from links')
+    p.add_argument('--dryrun', action='store_true',
+                   help="dry run; don't actually create links")
+
+    args = p.parse_args(options)
+
+    if args.exclude is None:
+        args.exclude = set()
+    else:
+        args.exclude = set(args.exclude.split(','))
+
+    args.include = set(args.include.split(',')) - args.exclude
+
+    args.cameras = decode_camword(args.cameras)
+
+    return args
+
+def check_link(newfile, reffile):
+    """Check if link from newfile -> reffile is ok
+
+    Args:
+        newfile (str): path to new link location
+        reffile (str): apth to reference file to link to
+
+    Returns True if everything is ok.
+
+    Returns False if refile is missing, or newfile already exists
+    and isn't a link.
+
+    Logs a warning if newfile exists as a link, but different than reffile.
+    """
+    log = get_logger()
+    if not os.path.exists(reffile):
+        log.error(f'Missing {reffile}')
+        return False
+
+    #- notes on islink vs. exists:
+    #-   a bad link passes islink but not exists;
+    #-   a good link passes both;
+    #-   a non-link file passes exists but not islink
+    if os.path.islink(newfile):
+        #- if link, check if it is a different destination
+        orig_link = os.readlink(newfile)
+        new_link = os.path.relpath(reffile, os.path.dirname(newfile))
+        if orig_link != new_link:
+            log.error(f'Pre-existing link {newfile} -> {orig_link} != {new_link}')
+            return False
+
+    elif os.path.exists(newfile):
+        #- pre-existing but not a link is bad
+        log.error(f"Pre-existing non-link {newfile}; won't override")
+        return False
+
+    return True
+
+
+def main(args=None):
+
+    log = get_logger()
+
+    if not isinstance(args, argparse.Namespace):
+        args = parse(args)
+
+    if len(args.include) == 0:
+        log.critical(f'No prefixes to include? check --include and --exclude options')
+        sys.exit(1)
+
+    #- check for any problems before creating any links
+    num_errors = 0
+    reffiles = list()
+    newfiles = list()
+    for prefix in args.include:
+        if prefix == 'ctecorr':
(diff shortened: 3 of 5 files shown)


sbailey:
Tests failed while installing astropy 5.0 (mis-match of checksums).  Will retry in morning in case it is transient download and/or packaging glitch.

py/desispec/scripts/link_calibnight.py
```diff
@@ -0,0 +1,138 @@
...
+def parse(options=None):
+    """parse options from sys.argv or input list of options"""
+    p = argparse.ArgumentParser()
+    p.add_argument('--refnight', type=int, required=True,
+                   help='Reference night with calibs')
+    p.add_argument('--newnight', type=int, required=True,
+                   help='New night without calibs, to link to refnight')
+    p.add_argument('-c', '--cameras', type=str, default='a0123456789',
+                   help='Camword of cameras to link, e.g. a0123b6r7z8')
+    p.add_argument('--include', type=str,
```
akremin, line 19: Can you update this and exclude to mention that the code expects a comma-separated list without spaces?

py/desispec/scripts/link_calibnight.py
```diff
@@ -0,0 +1,138 @@
...
+                   help="dry run; don't actually create links")
+
+    args = p.parse_args(options)
+
+    if args.exclude is None:
+        args.exclude = set()
+    else:
+        args.exclude = set(args.exclude.split(','))
+
+    args.include = set(args.include.split(',')) - args.exclude
```
akremin, line 34: I'd personally prefer if the default of include was None, and here you added logic:
1) If `include` is defined by the user, make sure that the set is independent of the `exclude` set. Then tell the user that only one or the other is used and you will proceed with the `include` list.
2) `If include is None`, then define it as the `set('badcolumns,biasnight,fiberflatnight,psfnight,ctecorr'.split(','))` - set().

py/desispec/test/test_link_calibnight.py
```diff
@@ -0,0 +1,153 @@
...
+        for name in cls.cache_env:
+            if cls.cache_env[name] is None:
+                del os.environ[name]
+            else:
+                os.environ[name] = cls.cache_env[name]
+
+    def test_basic_links(self):
+        options = f'--refnight {self.refnight} --newnight {self.newnight}'.split()
+        link_calibnight.main(options)
+        for prefix in self.prefixes:
```
akremin, line 60: Should you bring in the link_calibnight parser directly and then define `self.prefixes` as `args.include`? Or do you only want to test the subset you define above?

This won't automatically test new filetypes like `cte` unless added explicitly, and would break if we renamed or remove one of the above filenames. Not a major issue but something to consider.

akremin (commented):
Thank you for the useful script that does exactly what we need, and kudos for the unit tests. This is very well covered. I have made some comments throughout. Some of these I have already discussed with you, while others are new. None of these are critical blocking factors, but there are choices that I'd like more insights on, and a few comments that you can choose to consider or ignore.

bin/desi_link_calibnight
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
+
+import sys
+from desispec.scripts import link_calibnight
+
+sys.exit(link_calibnight.main())
```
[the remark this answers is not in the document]
sbailey: Context: I originally implemented this as returning an integer and having bin/desi_link_calibnight wrap that with a sys.exit(main()).  But then I realized that could be dangerous if desispec.scripts.link_calibnight.main() was called directly as part of a pipeline that didn't know to treat non-zero integers as bad, so I wanted something that would fail more dramatically so switched to sys.exit(errcode).

Prompted by this question, I checked how we've been doing it elsewhere, and we haven't been consistent.  e.g. scripts.preproc also calls sys.exit, while scripts.extract raises exceptions, while scripts.zproc does some of both.  In the end I think either way is fine, and I checked that sys.exit(sys.exit(7)) results in a unix error code of 7, so I'm going to leave this as is.  If we find a compelling reason why sys.exit(sys.exit(n)) is bad, we'll need to cleanup more than just this script.

py/desispec/scripts/link_calibnight.py
sbailey: done: help string says comma separated, and splitting code also strips spaces now.

py/desispec/scripts/link_calibnight.py
sbailey: Done, with modifications: I made --include and --exclude mutually exclusive rather than warning and proceeding with using only include.  I also check that --exclude is a strict subset of the default include so that the user doesn't accidentally not exclude something due to a typo.

py/desispec/scripts/link_calibnight.py
```diff
@@ -0,0 +1,138 @@
...
+
+    args.cameras = decode_camword(args.cameras)
```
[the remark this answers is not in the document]
sbailey: This is now decode_cameras(parse_cameras(args.cameras)) which I think supports all forms of lists, comma separated strings of cameras, and camwords.

py/desispec/scripts/link_calibnight.py
```diff
@@ -0,0 +1,138 @@
...
+            if check_link(newfile, reffile):
+                reffiles.append(reffile)
+                newfiles.append(newfile)
+            else:
+                num_errors += 1
+
+            #- ctecorr is per-night, not per-camera per-night so we only
+            #- have to check/link it once. Break out of camera loop and
+            #- continuewith other prefixes.
+            if prefix == 'ctecorr':
```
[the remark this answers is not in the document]
sbailey: I refactored this a bit to make links to ctecorr files if they exist, but don't treat it as a fatal error if they are missing from the reference night.  This allows us to use desi_link_calibnight for older nights and during the transition after the ctecorr PR is merged (PR #2163).  This could be updated in the future to require ctecorr to exist like the other calibnight files.

py/desispec/scripts/link_calibnight.py
```diff
@@ -0,0 +1,138 @@
...
+        newdir = os.path.dirname(newfiles[0])
+        os.makedirs(newdir, exist_ok=True)
+
+    for reffile, newfile in zip(reffiles, newfiles):
+        relpath = os.path.relpath(reffile, os.path.dirname(newfile))
+        if args.dryrun:
+            log.info(f'Dry run: would create link {newfile} -> {relpath}')
+        else:
+            #- pre-flight checks confirmed this is ok
+            if os.path.islink(newfile) or os.path.exists(newfile):
```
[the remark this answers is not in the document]
sbailey: I kept the `islink` check but removed the `or exists` check.

py/desispec/scripts/link_calibnight.py
```diff
@@ -0,0 +1,138 @@
...
+                os.remove(newfile)
+
+            log.debug(f'Linking {newfile} -> {relpath}')
```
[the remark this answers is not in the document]
sbailey: Done

py/desispec/test/test_link_calibnight.py
sbailey: As part of argparse updates, I factored out the --include prefix defaults into a `link_calibnight.calibnight_prefixes` module-level variable used by both the argparser and the unit tests (though I still cache that in self.prefixes, mostly as the minimal update to the existing tests).  I think that addresses the concern here, albeit in a different implementation.

sbailey:
@akremin thanks for the review comments.  I think I have addressed all of them; please re-review.

py/desispec/io/meta.py
```diff
@@ -173,6 +173,8 @@ def findfile(filetype, night=None, expid=None, camera=None,
         biasnight = '{specprod_dir}/calibnight/{night}/biasnight-{camera}-{night}.fits.gz',
         badfibers =  '{specprod_dir}/calibnight/{night}/badfibers-{night}.csv',
         badcolumns = '{specprod_dir}/calibnight/{night}/badcolumns-{camera}-{night}.csv',
+        ctecorrnight = '{specprod_dir}/calibnight/{night}/ctecorr-{night}.csv',
+        ctecorr      = '{specprod_dir}/calibnight/{night}/ctecorr-{night}.csv', #- alias, same file
```
akremin, line 177: I apologize for not following the cte developments closely, but are these meant to be identical in path and file name?

akremin (approved):
Thank you for the responses and for the revisions. I originally had a question but have since understood the answer based on the code comments. I'm happy to approve this and merge it whenever you are ready.

PR #2169, opened 2024-01-29
Read spectra parallel

sbailey:
This PR that adds `read_spectra_parallel(targets_table, ...)` to simplify the process of reading N>>1 targets from M>>1 different files, taking subsets as needed and then stacking them back together into a single Spectra result.  It supports multiprocessing or MPI for parallelism, and auto-derives from the columns of the targets_table whether to read tiles/cumulative spectra or healpix spectra.

The intended use case would be to read a zcatalog/ or equivalent redshift catalog, filter down to the targets you want, then call `read_spectra_parallel` to efficiently read them.  Beware: you'll probably blow memory if you try to read 1M quasars in a single call, but this is still useful to read in batches by nside=16 healpix or groups of 10 tiles, etc.

This includes a suite of unit tests for tiles and healpix spectra, and multiprocessing and serial reading.

@dylanagreen @moustakas, this PR admittedly would have been more useful a few weeks ago, but could one of you take a look at this and kick the tires and let me know if it would make your life easier for the next time you need to read N>>1 targets, and/or broken pieces or feature requests (within reason; I reserve the right to punt future requests to other PRs if too complicated to implement).

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -8,9 +8,11 @@ desispec Change Log
 * Better handling of copyprod links (PR `#2160`_).
 * Dark missing from DESI_SPECTRO_DARK is now a fatal error unless
   preproc with --fallback-on-dark-not-found (PR `#2162`_).
+* Add read_spectra_parallel (PR `#2169`_).
 
 .. _`#2160`: https://github.com/desihub/desispec/pull/2160
 .. _`#2162`: https://github.com/desihub/desispec/pull/2162
+.. _`#2169`: https://github.com/desihub/desispec/pull/2169
 
 0.61.0 (2024-01-15)
 -------------------
--- a/py/desispec/io/__init__.py
+++ b/py/desispec/io/__init__.py
@@ -22,7 +22,7 @@ from .fluxcalibration import (read_stdstar_templates, write_stdstar_models,
                               read_stdstar_models, read_flux_calibration,
                               write_flux_calibration, read_average_flux_calibration)
 from .spectra import (read_spectra, write_spectra, read_frame_as_spectra,
-                      read_tile_spectra)
+                      read_tile_spectra, read_spectra_parallel)
 from .frame import read_meta_frame, read_frame, write_frame
 from .xytraceset import read_xytraceset, write_xytraceset
 from .image import read_image, write_image
--- a/py/desispec/io/meta.py
+++ b/py/desispec/io/meta.py
@@ -73,7 +73,7 @@ def findfile(filetype, night=None, expid=None, camera=None,
         healpix=None, nside=64,
         band=None, spectrograph=None,
         survey=None, faprogram=None,
-        rawdata_dir=None, specprod_dir=None,
+        rawdata_dir=None, specprod_dir=None, specprod=None,
         download=False, outdir=None, qaprod_dir=None,
         return_exists=False,
         readonly=False, logfile=False):
@@ -98,6 +98,7 @@ def findfile(filetype, night=None, expid=None, camera=None,
     Options:
         rawdata_dir : overrides $DESI_SPECTRO_DATA
         specprod_dir : overrides $DESI_SPECTRO_REDUX/$SPECPROD/
+        specprod : production name, or full path to production
         qaprod_dir : defaults to $DESI_SPECTRO_REDUX/$SPECPROD/QA/ if not provided
         download : if not found locally, try to fetch remotely
         outdir : use this directory for output instead of canonical location
@@ -316,8 +317,8 @@ def findfile(filetype, night=None, expid=None, camera=None,
         log.debug("rawdata_dir = '%s'", rawdata_dir)
 
     if specprod_dir is None and 'specprod_dir' in required_inputs and outdir is None :
-        specprod_dir = specprod_root()
-        log.debug("specprod_dir = '%s'", specprod_dir)
+        specprod_dir = specprod_root(specprod)
+        log.debug("specprod_dir = '%s', specprod = '%s'", specprod_dir, specprod)
     elif outdir is not None :
         # if outdir is set, we will replace specprod_dir anyway
         # but we may need the variable to be set in the meantime
@@ -326,11 +327,9 @@ def findfile(filetype, night=None, expid=None, camera=None,
     if qaprod_dir is None and 'qaprod_dir' in required_inputs:
         qaprod_dir = qaprod_root(specprod_dir=specprod_dir)
 
-    if 'specprod' in required_inputs and outdir is None :
-        #- Replace / with _ in $SPECPROD so we can use it in a filename
-        specprod = os.environ['SPECPROD'].replace('/', '_')
-    else:
-        specprod = None
+    if 'specprod' in required_inputs and specprod is None and outdir is None :
+        specprod = os.path.basename(specprod_dir)
+        log.debug("Setting specprod = '%s'", specprod)
 
     if camera is not None:
         camera = camera.lower()
@@ -686,15 +685,23 @@ def specprod_root(specprod=None):
     ``$DESI_SPECTRO_REDUX/$SPECPROD``.
 
     Options:
-        specprod (str): overrides $SPECPROD
+        specprod (str): production name or full path to prodution
 
     Raises:
         KeyError: if these environment variables aren't set.
+
+    If specprod contains '/', treat as path and return that.
+    If specprod is None, return $DESI_SPECTRO_REDUX/$SPECPROD.
+    Otherwise, treat specprod as production name to override $SPECPROD
+    and return $DESI_SPECTRO_REDUX/$SPECPROD
     """
     if specprod is None:
         specprod = os.environ['SPECPROD']
 
-    return os.path.join(os.environ['DESI_SPECTRO_REDUX'], specprod)
+    if '/' in specprod:
+        return specprod
+    else:
+        return os.path.join(os.environ['DESI_SPECTRO_REDUX'], specprod)
 
 
 def qaprod_root(specprod_dir=None):
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -912,10 +912,16 @@ def read_spectra_parallel(targets, nproc=None, prefix='coadd',
         spectra = stack_spectra(spectra)
 
         #- reorder spectra to match input target table if needed
-        if match_order and np.any(spectra.fibermap[keycols] != targets[keycols]):
-            ii = argmatch(spectra.fibermap[keycols], targets[keycols])
-            spectra = spectra[ii]
-            assert np.all(spectra.fibermap[keycols] == targets[keycols])
+        if match_order:
+            need_to_reorder = False
+            for col in keycols:
+                if np.any(spectra.fibermap[col] != targets[col]):
+                    need_to_reorder = True
+                    break
+
+            if need_to_reorder:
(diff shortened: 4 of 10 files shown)


moustakas:
Thanks, this is a terrific contribution! I am happy to test this as part of my template work. My one quick thought about the top-level wrapper is that it's not obvious how to specify the specprod, e.g., `iron` vs `fuji`, etc. I *think* it's derived in `findfile` via an environment variable, but that's a little annoying. :) Could we add a `specprod` optional input at the top level?

sbailey:
@moustakas good idea about more flexible specprod options.  I added a `read_spectra_parallel(..., specprod=SPECPROD)` option which can take either the name of a specprod (e.g. "fuji") or the full path to a specprod (e.g. /global/cfs/cdirs/desi/spectro/redux/fuji), and those will override $DESI_SPECTRO_REDUX/$SPECPROD.

Under the hood, I implemented this in a more general way than just `read_spectra_parallel`, so now `desispec.io.findfile` also has a `specprod` option (instead of previously only supporting the full path though `specprod_dir`).  Two non-backwards compatible implications:
* $SPECPROD can no longer have a '/' in the name (previously supported, but in a hacky way where sometimes the slash was replaced with an underscore because sometimes it was a path and sometimes it was part of a filename).
* `findfile` enforces that the production is kept in a directory that matches what it considers "specprod" to be (e.g. specprod "fuji" must be in a directory /path/to/somewhere/fuji).  Previously it allowed you to mix-and-match in an inconsistent way.

moustakas (commented):
Thanks @sbailey, that's a really useful change to `findfile`. The next issue I ran into is that the input order isn't respected. I think the default behavior should be to sort the spectra at the end. However, while trying to create a MWE I ran into a couple different problems:
```
import fitsio
from desispec.io.spectra import read_spectra_parallel

zz = Table(fitsio.read('/global/cfs/cdirs/desi/spectro/redux/iron/zcatalog/zpix-sv1-bright.fits', rows=[0, 300, 900])
zz['SURVEY'] = 'sv1'
zz['PROGRAM'] = 'bright'
ss = read_spectra_parallel(zz['TARGETID', 'HEALPIX', 'SURVEY', 'PROGRAM'])
TypeError                                 Traceback (most recent call last)
File /global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/conda/lib/python3.10/site-packages/numpy/lib/shape_base.py:772, in array_split(ary, indices_or_sections, axis)
    770 try:
    771     # handle array case.
--> 772     Nsections = len(indices_or_sections) + 1
[snip]
--> 776     Nsections = int(indices_or_sections)
    777     if Nsections <= 0:
    778         raise ValueError('number sections must be larger than 0.') from None

TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

The problem seems to be the `nproc=None` default, which we should capture. Setting this to be unity gives a different crash:
```
ss = read_spectra_parallel(zz['TARGETID', 'HEALPIX', 'SURVEY', 'PROGRAM'], nproc=1)
FileNotFoundError                         Traceback (most recent call last)
Cell In[21], line 1
----> 1 ss = read_spectra_parallel(zz['TARGETID', 'HEALPIX', 'SURVEY', 'PROGRAM'], nproc=1)

File ~/code/desihub/desispec/py/desispec/io/spectra.py:874, in read_spectra_parallel(targets, nproc, prefix, rdspec_kwargs, specprod, comm)
[snip]
File ~/code/desihub/desispec/py/desispec/io/util.py:41, in checkgzip(filename)
     39     return altfilename
     40 else:
---> 41     raise FileNotFoundError(f'Neither {filename} nor {altfilename}')

FileNotFoundError: Neither /dvs_ro/cfs/cdirs/desi/spectro/redux/ioannis/healpix/sv1/bright/103/10378/coadd-sv1-bright-10378.fits nor /dvs_ro/cfs/cdirs/desi/spectro/redux/ioannis/healpix/sv1/bright/103/10378/coadd-sv1-bright-10378.fits.gz
```

The issue, of course, is that the code falls back to the $SPECPROD environment variable (which I didn't think I had set but it must default to one's username):
```
echo $SPECPROD
ioannis
```
Setting the specprod explicitly finally gives the requested output:
```
ss = read_spectra_parallel(zz['TARGETID', 'HEALPIX', 'SURVEY', 'PROGRAM'], nproc=1, specprod='iron')
INFO:spectra.py:396:read_spectra: iotime 0.037 sec to read coadd-sv1-bright-10378.fits at 2024-01-31T06:16:09.246099
INFO:spectra.py:396:read_spectra: iotime 0.123 sec to read coadd-sv1-bright-10400.fits at 2024-01-31T06:16:09.378789
INFO:spectra.py:396:read_spectra: iotime 0.041 sec to read coadd-sv1-bright-17670.fits at 2024-01-31T06:16:09.426543
```

Hmm, it looks like with all the simple examples I'm coming up with the input and output order *does* match. However, I definitely ran into a sorting issue with a separate piece of code, plus I'm not seeing any explicit sorting in the code, so I do think this is a failure mode we should fix.

BTW, as we discussed briefly off-list, it would be great to be able to temporarily suppress these informational messages without having to set the log-level environment variable. E.g., when reading N>>1 spectra I have to do
```
os.environ['DESI_LOGLEVEL'] = 'warning'
ss = read_spectra_parallel(zz['TARGETID', 'HEALPIX', 'SURVEY', 'PROGRAM'], nproc=1, specprod='iron')
os.environ['DESI_LOGLEVEL'] = 'info'
```
But it would be *nice* to be able to do:
```
with "temporarily-suppress-info-messages":
    ss = read_spectra_parallel(zz['TARGETID', 'HEALPIX', 'SURVEY', 'PROGRAM'], nproc=1, specprod='iron')
```
But that's a different ticket.

sbailey:
Thanks for the testing and the crash cases!  I really thought I had the tests covering all the cases, alas.  Will fix.

For sorting, the intended behavior is that the output spectra match the sort order of the input targets list, even if that is different from the order that the targets appear in the files on disk.  I think matching input order is better than doing something like sorting on TARGETID.  The unit tests cover some cases of input order = output order != disk order, but if you find a reproducer case where that isn't true, please post that as well.

sbailey:
I've fixed the nproc=None (default) case, with some embarrassment that none of my unit tests had actually checked the default behavior (they do now)

Regarding your nproc=1, specprod=None (default) case:
> The issue, of course, is that the code falls back to the $SPECPROD environment variable

I think the code is doing the right thing, sort of — if you don't specify specprod, it should use the $SPECPROD environment variable because the `zz` Table you provided has no knowledge in memory that it came from the iron production.  Note that the `Table(fitsio(..., rows=...))` wrapper stripped out the header info.

Although it wouldn't have helped in this case, I could see defining that if the input `targets` table is as astropy Table and has either `SPECPROD` or `DEPNAMnn='SPECPROD'`, `DEPVERnn=something` combo in `targets.meta`, then that should should be used instead of $SPECPROD if the `specprod` input isn't set.

If we're going into checking for `targets.meta`, I might also support allowing `targets.meta['SURVEY']` and `targets.meta['PROGRAM']` to override missing SURVEY/PROGRAM columns.

Feature creep!  While we're at it, we could augment `desispec.io.read_table` (that already uses `Table(fitsio.read(...))` while re-adding the header to `result.meta`) to also support other fitsio-optimizations like `columns` and `rows` and then you could have your cake and eat it too as long as you like the `desispec.io.read_table` flavor of cake.

sbailey:
@moustakas this now works, auto-deriving the SPECPROD, PROGRAM, and SURVEY from the catalog header:
```
from desispec.io import read_table, read_spectra_parallel
zz = read_table('/global/cfs/cdirs/desi/spectro/redux/iron/zcatalog/zpix-sv1-bright.fits', rows=[0, 300, 900])
ss = read_spectra_parallel(zz)
```

moustakas:
Thanks @sbailey. Here's an example where the sorting isn't preserved (inspired by something that's actually happening to me in a different piece of code).
```
from desispec.io import read_table, read_spectra_parallel
zz = read_table('/global/cfs/cdirs/desi/spectro/redux/iron/zcatalog/zpix-sv1-bright.fits', rows=[0, 300, 900])
ss = read_spectra_parallel(zz[[2, 0, 1]])
ss.target_ids() == zz[[2, 0, 1]]['TARGETID']
  array([False, False, False])
```

sbailey:
@moustakas got it, thanks for example.  Underlying problem fixed, with tests to confirm.  New arg `match_order` (default True) will ensure that the output order matches the input order.  Setting `match_order=False` will reduce a temporary memory spike, but the output order may or may not match the input in that case.

moustakas:
Thanks @sbailey. I apologize for not mentioning this earlier, but, I don't think it's sufficient to sort on `targetid`. I think for healpix coadds we need to sort on a "key" of `survey-program-healpix-targetid` and for cumulative coadds we need `tileid-targetid`. (I don't think we should support other types of coadds to avoid mission creep.)

Here's an example failure in the case that someone is trying to coadd spectra across surveys or programs (which is not a crazy use case)--

```
import fitsio, pdb
from desispec.io import read_spectra_parallel

zall = Table(fitsio.read('/global/cfs/cdirs/desi/spectro/redux/iron/zcatalog/zall-pix-iron.fits',
                       columns=['TARGETID', 'SURVEY', 'PROGRAM', 'HEALPIX', 'MAIN_NSPEC', 'MAIN_PRIMARY', 'TSNR2_LRG']))
# this is the first example where MAIN_PRIMARY == True and MAIN_NSPEC>1
zz = zall[zall['TARGETID'] == 616089273883427662]
zz = zz[[1, 0]]

ss = read_spectra_parallel(zz, specprod='iron')
print(ss.scores['TSNR2_LRG'] == zz['TSNR2_LRG'])
  [False False]
```
Also, oddly, `target_ids()` returns a scalar. Is this expected behavior?
```
ss.fibermap
<Table length=2>
     TARGETID      COADD_FIBERSTATUS     TARGET_RA          TARGET_DEC       PMRA   PMDEC  REF_EPOCH ... STD_FIBER_RA   MEAN_FIBER_DEC   STD_FIBER_DEC MEAN_PSF_TO_FIBER_SPECFLUX HEALPIX SURVEY PROGRAM
      int64              int32            float64            float64       float32 float32  float32  ...   float32         float64          float32             float32            int64   str4    str6
------------------ ----------------- ------------------ ------------------ ------- ------- --------- ... ------------ ------------------ ------------- -------------------------- ------- ------ -------
616089273883427662                 0 134.65160350096173 32.053057584507805     0.0     0.0       0.0 ...  0.010053255 32.053078586707976  0.0028417162                 0.79246867    5059   main  backup
616089273883427662               512 134.65160350096173 32.053057584507805     0.0     0.0       0.0 ...          0.0  32.05351351608743           0.0                   1.484397    5059   main  bright

ss.target_ids()
<Column name='TARGETID' dtype='int64' length=1>
616089273883427662
```

sbailey:
@moustakas good point about TARGETIDs that span SURVEYs and PROGRAMs.  Fixed, with tests.  Anything else that I've missed?

> Also, oddly, target_ids() returns a scalar. Is this expected behavior?

`spectra.target_ids()` returns the *unique* TARGETIDs that appear in the object, not the `spectra.fibermap['TARGETID']` column which could have repeats.  In your above example, it isn't returning a scalar, but it is returning an array/column of length 1 since there is only one unique TARGETID in your `ss` object.  Also note that `ss.num_spectra() == 2` and `ss.num_targets() == 1`

moustakas (approved):
Looks awesome, thank you!

PR #2183, opened 2024-02-22
simplify installation requirements

sbailey:
This PR moves some imports around to simplify the dependencies needed to do basic operations with DESI spectra:
  * desispec.io.read_spectra
  * filter spectra by desitarget masks
  * desispec.coaddition.coadd and .coadd_cameras
  * desispec.spectra.stack
  * desispec.io.write_spectra

These actions can be done with a minimal environment of:
```
conda create -n desilite --yes python=3.10 astropy scipy numba pytest ipython
conda activate desilite
pip install fitsio

python -m pip install git+https://github.com/desihub/desiutil.git
python -m pip install git+https://github.com/desihub/desitarget.git
python -m pip install git+https://github.com/desihub/desispec.git@desilite
```
(after merging that `@desilite` branch specification won't be needed).

e.g. you no longer need specter, speclite, or desimodel to do those basic tasks.

py/desispec/test/test_lite.py is a minimal test of those spectra tasks, and a new github actions workflow installs just the minimal environment and runs just that test.

numba is the remaining unfortunate extra requirement.  This is because module-level
`@numba.jit` decorators can't be hidden by moving numba imports inside the functions that
need numba, even if those functions are never called.  We could work around this by creating
a fake numba alternative with `fakenumba.jit = contextlib.nullcontext` etc., but I'd like to think
a bit more about the design of that, and consider adding it to desiutil instead of desispec.
So for now, numba is still required.

@stephjuneau @moustakas @akremin et al are there any other "basic end-user functions" that you
think we should support with a minimal environment?

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -112,7 +112,7 @@ jobs:
             - name: Install Python dependencies
               run: |
                 python -m pip install --upgrade pip setuptools wheel
-                python -m pip install astropy scipy numba
+                python -m pip install astropy scipy numba "numpy<2.0"
                 python -m pip install pytest
                 python -m pip cache remove fitsio
                 python -m pip install --no-deps --force-reinstall --ignore-installed fitsio
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -92,6 +92,35 @@ jobs:
                 GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               run: coveralls
 
+    desilite:
+        name: Test minimal env
+        runs-on: ${{ matrix.os }}
+        strategy:
+            fail-fast: true
+            matrix:
+                os: [ubuntu-latest]
+                python-version: ['3.10']
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v2
+              with:
+                fetch-depth: 0
+            - name: Set up Python ${{ matrix.python-version }}
+              uses: actions/setup-python@v2
+              with:
+                python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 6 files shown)


weaverba137:
A few suggestions:

* I recommend selecting specific versions for desiutil and desitarget.
* Numpy 2.0 is due out soon, so these instructions should probably select more specific versions of the conda packages as well. Unfortunately, I don't think we're ready for the major changes in Numpy 2.0 yet.
* Does this work in both NERSC and non-NERSC environments?

weaverba137:
What about specutils? specutils 1.13.0 now supports reading DESI spectra and coadd files, and there are several notebooks that use it.

weaverba137:
@sbailey, I agree with you mostly above, but what actually happens when you do
```
python -m pip install git+https://github.com/desihub/desiutil.git
```
I don't actually know what version results from that.  I've never *not* specified a version.

If it is the latest tag, then that's fine.  If it is effectively `main`, then it becomes harder for external users to communicate the version of these packages when reporting problems.

For context, astropy strongly encourages users to report not only the version of astropy, but of all dependencies, numpy, scipy, etc.

sbailey:
pip install git without specifying a branch/tag installs the latest commit of the default branch, i.e. main.  I don't know of a way to specify "the latest tag" without hardcoding what that tag is and having to keep that in sync for docs, though that "latest tag on github" would be handy if anyone knows syntax for that.

At the same time, I think that is a documentation issue for what we tell people to install and what information to include when reporting issues.  For the purposes of this PR and testing, I think installing latest main is still the right thing to do.

weaverba137:
Good to know. OK to go with that for now.

PR #2192, opened 2024-03-08
Fix CTE direction flip.

schlafly:
This fixes an issue with the CTE correction on amplifiers which read out in the "flipped" direction.  The correction was being done "correctly" but needed to be flipped back before being applied to the image.  This now occurs.

Figure shows 20240304/00228660/preproc-r7-00228660.fits.gz, left: daily preproc, middle: CTE corrected without this PR, right: CTE-corrected with this PR.
image

### DIFF
--- a/py/desispec/correct_cte.py
+++ b/py/desispec/correct_cte.py
@@ -839,8 +839,8 @@ def correct_image_via_model(image, niter=5, cte_params_filename=None):
 
             cteimage[ampreg] = apply_multiple_cte_effects(
                 imamp[:, ::sign], locations=ctelocs,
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks.  To avoid other future surprises, could you also check other amps for up/down left/right flip/not correctness?  I think if you invent a fake CTECOLSx range for something that doesn't actually need a correction, you should get back approximately the same image, but if there is a sign flip you'll get something dramatically bad like your middle image.

schlafly:
Yes, I did a run where I added corrections to all four amps and it looks fine.

PR #2194, opened 2024-03-14
ctecorr flexibility

sbailey:
This PR makes the CTE correction code more flexible for cases when not correction is needed:
* when generating the CTE corrections, desi_fit_cte_night first checks if a CTE correction is even needed for a camera before complaining about things like BADCAMWORD being set.  It also checks the science exposures and doesn't try to generate a CTE correction for a camera that is flagged as bad all night anyway.
  * @akremin this still only checks CAMWORD and BADCAMWORD; I could use some help for how to also check BADAMPS
* when using the CTE corrections, desi_preproc first checks if the correction is needed for a given camera before checking for the existence of $DESI_SPECTRO_REDUX/$SPECPROD/calibnight/NIGHT/ctecorr-NIGHT.csv .  This restores the ability to preproc non-CTE-problematic cameras outside of the context of a production without having to specify --no-cte-corr.  If DESI_SPECTRO_CALIB says that this camera on this night needs a CTE correction, then $DESI_SPECTRO_REDUX/$SPECPROD/calibnight/NIGHT/ctecorr-NIGHT.csv is still required unless the user opts-out with --no-cte-corr.  @julienguy this should make you happier.

### DIFF
--- a/py/desispec/correct_cte.py
+++ b/py/desispec/correct_cte.py
@@ -10,11 +10,6 @@ import os
 import numpy as np
 import fitsio
 from desispec.calibfinder import CalibFinder
-import desispec.io
-import desispec.io.util
-from desispec.io.util import decode_camword
-import desispec.io.xytraceset
-import desispec.io.fiberflat
 from desispec.trace_shifts import compute_dx_from_cross_dispersion_profiles
 import desispec.preproc
 from desiutil.log import get_logger
@@ -232,7 +227,7 @@ def get_amp_regions_to_cte_correct(header):
             nvals = len(vals)
             if nvals != 2 :
                 mess = "cannot decode {}={}".format(key, value)
-                log.error(mess)
+                log.critical(mess)
                 raise KeyError(mess)
 
             start, stop = int(vals[0]), int(vals[1])
@@ -282,6 +277,8 @@ def get_cte_params(header, cte_params_filename=None):
 
     Returns pair of empty dictionaries if there are no amps needing CTE corrections.
     """
+    import desispec.io   # inside func to avoid circular import
+
     log = get_logger()
 
     amp_regions, cte_regions = get_amp_regions_to_cte_correct(header)
@@ -572,10 +569,21 @@ def get_cte_images(night, camera, expids=None):
     -------
     List of preprocessed Images without CTE correction applied
     """
+    # inside func to avoid circular import
+    import desispec.io
+    import desispec.io.util
 
     log = get_logger()
 
-    if expids is None:
+    if expids is not None:
+        filename = desispec.io.findfile('raw', night, expids[0])
+        header   = fitsio.read_header(filename, camera.upper())
+        if not needs_ctecorr(header):
+            log.info(f"No CTE correction needed for {night} {camera}")
+            return None
+
+    else:
+        #- Look up exposures in production exposure table
         exptablefn = os.path.join(os.environ['DESI_SPECTRO_REDUX'],
                                   os.environ['SPECPROD'],
                                   'exposure_tables', str(night // 100),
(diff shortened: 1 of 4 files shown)


py/desispec/scripts/fit_cte_night.py
```diff
@@ -62,7 +64,27 @@ def main(args=None, comm=None):
...
+                          os.environ['SPECPROD'],
+                          'exposure_tables', str(args.night // 100),
+                          f'exposure_table_{args.night}.csv')
+        etable = load_table(etablefile, tabletype='exptable')
+        keep = etable['OBSTYPE'] == 'science'
+        sci_etable = etable[keep]
+        if len(sci_etable) == 0:
+            log.warning(f'No science exposures on {args.night}, but calculating CTE corrections anyway...')
+        else:
+            camword = parse_cameras(args.cameras)
```
akremin, line 82: This works when `args.cameras` is a superset of CAMWORD, but will give incorrect results if CAMWORD is restricted but cameras is e.g. all cameras=a0123456789.

I think what you want is to find the GOODCAMWORD=CAMWORD-BADCAMWORD for each exposure, find the union of those, then take the intersection with `input_camword=parse_cameras(args.cameras)`. You can do the logic for GOODCAMWORD yourself, but there is a function that will take an erow and provide you with that. There is also a function for taking the intersection.

Useful functions for this:
`desispec.io.util.erow_to_goodcamword(erow, suppress_logging=True)`
`desispec.io.util.camword_intersection(camwords)`

py/desispec/correct_cte.py
```diff
@@ -542,71 +569,86 @@ def get_cte_images(night, camera, expids=None):
...
+            return None
+
+        #- CTE correction is needed, so also filter by CAMWORD and BADCAMWORD
         keep = np.ones(len(exptable), dtype=bool)
         for i in range(len(keep)):
-            if camera not in decode_camword(str(exptable['CAMWORD'][i])):
+            if camera not in desispec.io.util.decode_camword(str(exptable['CAMWORD'][i])):
                 keep[i] = False
-            elif camera in decode_camword(str(exptable['BADCAMWORD'][i])):
+            elif camera in desispec.io.util.decode_camword(str(exptable['BADCAMWORD'][i])):
```
akremin, line 621: In the future, after `pipelinerefactor` is merged, `desispec.io.util.all_impacted_cameras(badcamword, badamps)` returns the set of cameras reflected in either badcamword or badamps, where each is a single string. You could change this to:
`if camera in desispec.io.util.all_impacted_cameras(str(exptable['BADCAMWORD'][i]), str(exptable['BADAMPS'][i])):`

But that obviously won't work for the interim. For now you could leverage the fact that BADAMPS is a comma-seperated list with cameras in the identifiers, so e.g `'b4' in 'z1A,b4B,z7C'` is `True`, as desired.

So you can do the following:
`if camera in desispec.io.util.decode_camword(str(exptable['BADCAMWORD'][i])) or camera in str(exptable['BADAMPS'][i]):`

py/desispec/correct_cte.py
```diff
@@ -542,71 +569,86 @@ def get_cte_images(night, camera, expids=None):
...
     log.info(f"Will use exposures {expids} for {night} {camera} CTE corrections")
 
     images = list()
     for expid in expids:
         preproc_filename = desispec.io.findfile('preproc_for_cte', night, expid, camera)
         if not os.path.isfile(preproc_filename) :
             log.info(f"Computing {preproc_filename}")
-            infile = desispec.io.findfile('raw',night,expid)
-            image  = desispec.io.read_raw(infile, camera, no_cte_corr = True)
+            infile = desispec.io.findfile('raw', night, expid)
```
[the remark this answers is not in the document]
sbailey: Done

sbailey:
@akremin I believe I have updated the camword logic to cover the cases you raised, including BADAMPS via erow_to_goodcamword.  Please re-doublecheck my logic.

akremin (approved):
All of my requests were addressed in the latest commits. The implementation at line 604 in `get_cte_images()` is better than my original suggestion because it quickly skips nights where the camera isn't used, while also capturing the case I was concerned about.

PR #2199, opened 2024-03-19
Fix the desi_group_spectra with more than 1000 files

### LINKED ISSUE #2198: desi_group_spectra >1000 files
Hi, 

While running desi_group_spectra on some tertiary program data I see this error: 
ValueError: The header keyword 'INFIL1000' with its value is too long
which is clearly caused by >1000 input files. I'm stacking data in small HPX'es, but in the center of the tile with more than 34 exposures you get >1000 cframe files. 

I understand that this is somewhat 'special' case, but it is not that unreasonable situation.

My proposal is to put a warning here: 
https://github.com/desihub/desispec/blob/c2222d2b019634f1d9c125cf9ba47bd12aa3863e/py/desispec/scripts/group_spectra.py#L157
and stop writing INFIL after 1000 files. 

Comments  ?

segasai:
Fix #2198 

I also write a INFILNUM keyword to the header. I wasn't sure if it's better to always write it, or only if the number is >1000.
I did the latter.

### DIFF
--- a/py/desispec/scripts/group_spectra.py
+++ b/py/desispec/scripts/group_spectra.py
@@ -159,6 +159,7 @@ def main(args=None):
     for i, filename in enumerate(foundframefiles[:max_header_names]):
         spectra.meta[f'INFIL{i:03d}'] = shorten_filename(filename)
     if nframefiles > max_header_names:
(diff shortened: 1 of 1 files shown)


segasai:
Thank you @sbailey !

Issue #2204, opened 2024-03-28
NightQA script error


rongpu:
The nightqa script failed with the following error
```
############################################
# tail of nightqa log nightqa-daily-20240327.log:#
############################################
INFO:night_qa.py:142:get_surveys_night_expids: for NIGHT=20240327 found 34 exposures from 32 tiles for SURVEY=main and 2 exposures from 2 tiles for SURVEY=special
INFO:night_qa.py:168:get_dark_night_expid: proctable_fn = /global/cfs/cdirs/desi/spectro/redux/daily/processing_tables/processing_table_daily-20240327.csv
Traceback (most recent call last):
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/desispec/main/bin/desi_night_qa", line 208, in <module>
    main()
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/desispec/main/bin/desi_night_qa", line 144, in main
    dark_expid = get_dark_night_expid(args.night, args.prod)
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/desispec/main/py/desispec/night_qa.py", line 185, in get_dark_night_expid
    expid = int(str(d["EXPID"][sel][0]).strip("|"))
ValueError: invalid literal for int() with base 10: '232598|232668|232665'
```

akremin:
Yes that is expected due to changes in the pipeline. The cte correction is also bundled into the ccdcalib and we now correctly list the flat exposures included in those fits. The dark should be the first expid in the list, if there is more than one and the OBSTYPE is dark. So the logic looks good except we now need to parse the line for the first entry. I'll take care of this.

araichoor:
I see, thanks.

akremin:
I fixed the original problem in PR #2206 , but now the code is crashing at a later step:
```
Traceback (most recent call last):
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/desispec/main/bin/desi_night_qa", line 208, in <module>
    main()
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/desispec/main/bin/desi_night_qa", line 156, in main
    create_dark_pdf(outfns["dark"], args.night, args.prod, dark_expid, args.nproc, bkgsub_science_cameras=args.dark_bkgsub_science_cameras)
[...]

PR #2206, opened 2024-03-28
Make nightqa dark sel robust to new ccdcalib

akremin:
Quick fix to issue #2204 . ccdcalib now lists flats used for cte corrections, if performed, which broke the qa code looking for the dark exposure. This fixes it by taking the first exposure if multiple are present, which is the dark exposure based on the pipeline logic. We could do additional sanity checks to make sure it is a dark expid, but this is programmatically how the pipeline currently works. If it breaks in the future then I suggest we do a more rigorous and in-depth patch.

@anand please check to see if you're happy with this intellectually. Then we're safe to merge and rerun nightqa for last night.

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -182,6 +182,8 @@ def get_dark_night_expid(night, prod):
                 )
             )
         else:
+            ## ccdcalib jobs can have more than one exposure if also doing cte
+            ## corrections. The first is expected to be the dark, so take that one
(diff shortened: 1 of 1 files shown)


araichoor:
looks good to me, thanks!
minor comment: would it be worth it to add a comment describing that logic just before the "if"?

akremin:
Thanks for the check and suggestion. I've added a comment to explain what the string manipulation is doing. I'll self merge

Issue #2189, opened 2024-03-01
Pipeline call autocalib_fiberflat with --solve-gradient option for selected nights


sbailey:
Add hooks and override files to the pipeline to call `desi_autocalib_fiberflat` with the `--solve-gradient` option for a list of nights with known gradients (likely due to calib screen misalignment with the telescope).  From @rongpu and @sybenzvi in #2172 these nights are:

20230925, 20231001, 20231002, 20231003, 20231004, 20231005, 20231006, 20231007, 20231008, 20231009, 20231010, 20231011, 20231012, 20231013, 20231014, 20231015, 20231203, 20231204, 20231205, 20231207, 20231214, 20231216, 20231217, 20231218, 20231219, 20231220, 20231221, 20231228, 20231229, 20231230, 20231231, 20240101, 20240102

Do this after PR #2187 is merged.

sbailey:
Further details -- I think this would look like an override_NIGHT.yaml entry like
```
## 20230925 offset between telescope and dome screen so correct for fiberflat gradient
calibration:
    autocalib_fiberflat:
        solve_gradient: True 
```

which would propagate into `desi_proc_joint_fit` calling `desispec.scripts.autocalib_fiberflat.main` with an additional option `--solve-gradient`.  Note that `desispec.scripts.autocalib_fiberflat.main` takes a list of strings as if it was constructing a command line for subprocess, and thus it needs a string `'--solve-gradient'`; it is not a function with key/value args that would need `main(..., solve_gradient=True)`.  We do *not* need to specify a reference night, nor does any reference night need to be processed before the night with the `--solve-gradient` override since the reference fiberflats are in $DESI_SPECTRO_CALIB itself.

Currently the override files are handled in PR #2187 desi_proc_night at the pipelining level.  I think this is the first case where we need to be passing an option down into the algorithmic level.

akremin:
As noted in [Issue #2174](https://github.com/desihub/desispec/issues/2174#issuecomment-2057611435_), these 33 nights are now covered by override files specifying that we should set `--solve-gradient` in `autocalib_fiberflat`.

```
A    202401/override_20240101.yaml
A    202401/override_20240102.yaml
A    202309/override_20230925.yaml
A    202310/override_20231001.yaml
A    202310/override_20231011.yaml
A    202310/override_20231003.yaml
A    202310/override_20231013.yaml
A    202310/override_20231005.yaml
U    202310/override_20231015.yaml
A    202310/override_20231007.yaml
[...]

PR #2208, opened 2024-03-28
Add hooks for autocalib_fiberflat flag into override file and pipeline

akremin:
This is the code changes to resolve issue #2189 by introducing new allowed override file keywords and implementing the code to interpret those and pass them to `autocalib_fiberflat` inside `desi_proc_joint_fit`. This is done by introducing a new flag to `desi_proc_joint_fit` that when set, runs `autocalib_fiberflat` with the `--solve-gradient` flag.

The new override file format differs from the recommendation in the issue. The format is:

```
calibration:
    nightlyflat:
        flags: ['--autocal-ff-solve-grad'] 
```

This follows the logic from the calibration linking, where jobs that are only performed once per night can be uniquely identified by their `JOBDESC` (job description) in the override files. This approach breaks down for per-exposure or per-tile specifications but the simplicity here outweighed that for me. When we introduce per-exp or per-tile overrides, we will have to determine how to approach that and allow the override to have multiple meanings for second (and third and forth) levels. That is left for a future day.

I have also added a unit test which ensures that the nightlyflat script being produced includes the `--solve-gradient` flag when `--autocal-ff-solve-grad` is specified as a flag in the override file.

### DIFF
--- a/py/desispec/scripts/proc_joint_fit.py
+++ b/py/desispec/scripts/proc_joint_fit.py
@@ -356,6 +356,8 @@ def main(args=None, comm=None):
             cmd.append(f"-i")
             for flat in flats_for_arm[camera_arm]:
                 cmd.append(f"{flat}")
+            if args.autocal_ff_solve_grad:
+                cmd.append("--solve-gradient")
             num_cmd += 1
             cmdargs = cmd[1:]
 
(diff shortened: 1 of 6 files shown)


sbailey (approved):
Looks good.  In addition to the unit tests (thanks!) I also tried generating and submitting a job with desi_proc_night and an override file and confirmed that the `desi_proc_joint_fit --autocal-ff-solve-grad` option in the script correctly propagated through to calling `desi_autocalib_fiberflat --solve-gradient` when run (it did).

My only suggestion is cosmetic bike shedding: should we change the override "flags" entry to "options" (or "extra_opts" or "extra_options")?  I think of "flags" as meaning boolean items, which is the case for `--autocal-ff-solve-grad`, but structurally this works for any additional options like `--option value` too.  But if you prefer "flags" over "options", that's fine too.

Issue #2197, opened 2024-03-18
cctecorr code doesn't support links to a ctecorr file from another night


akremin:
`ctecorr-{NIGHT}.csv` includes a NIGHT column that appears to be checked by the code and raises a critical error if it differs from the night of the data the code is trying to process. This means that cctecorr doesn't currently support the linking of calibrations from a different night.

sbailey:
Ugh!  The motivation for including the NIGHT column was so that we could stack N>>1 different ctecorr-NIGHT.csv files and get a table to track corrections vs. time.

1. I think the solution here is to relax the check on NIGHT to be fairly flexible, like +/- 2 weeks to the current night in the hopes that we'd never have to span an outage longer than that.

Added to Jura To Do list.

sbailey:
@akremin for consideration, it might be a better solution for linkcal to *copy* the contents of the reference ctecorr file, replace any NIGHT entries with the new night, and write it back out for the new night.  This would allow code using the file to be agnostic about how the file got there and just use it, without requiring parsing which night in the file is closest to the desired night.

Ideally we would include a comment that this night had been replace, but unfortunately vanilla csv doesn't formally support comments.  They could be read with `Table.read(filename, comment='#')` but that's a potential user-interface landmine, and writing it with a comment would require re-opening to add the comment.  Given that we have ~50 ctecorr files in daily already, I don't think we should change the format to ecsv or yaml or add a COMMENTS column though.  Or we could e.g. define a new COMMENTS column and retroactively add them to all daily ctecorr files.  Or just live without having a comment about what night this came from.

akremin:
I'm not sure that it's better to alter the information in the file rather than just adding a few lines of code to interpret whether the night where the correction was taken is compatible with the current night of processing. Especially given that there is no metadata propagation, as you point out, I think it would be better to keep the night column as the true night the values were estimated from. This is in the same line as not changing the NIGHT header keyword in the other linked cals, so that we know where they actually came from (in addition to the links which also indicate where they're coming from).

akremin:
Resolved with PR #2213

PR #2213, opened 2024-04-02
Ctenightcheck

akremin:
This resolved issue #2197 in which the ctecorr code crashes if all table entries are from a different night than the night being processed. However, we want to support this case since there are nights with no valid calibration data where we need to link the information from another night.

This introduces a new function `desispec.util.difference_nights` that uses `datetime.date` to handle the difference in days between two nights. The ctecorr code is updated to accept any calibration data within 14 days of the current night being processed (either before or after the current night).

I've tested this on the test night that originally identified the problem, and it now runs in this branch. I also tested on a new night I set up to link the ctecorrnight from a different night, and it worked for jobs on that test night as well.

### DIFF
--- a/py/desispec/util.py
+++ b/py/desispec/util.py
@@ -11,8 +11,10 @@ import os
 import sys
 import errno
 import time
+import datetime
 import collections
 import numbers
+import datetime
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


py/desispec/util.py
```diff
@@ -398,18 +400,36 @@ def night2ymd(night):
...
+
+def difference_nights(firstnight, secondnight):
+    """
+    parse two YEARMMDD nights (ints or strings) and determine the number of
+    days between them
+    """
+    dt1 = night2dateobj(str(firstnight))
+    dt2 = night2dateobj(str(secondnight))
+    difference = dt1 - dt2
+    return np.abs(difference.days)
```
sbailey, line 431: Although the current usage of this function only needs abs(difference), I suggest that this generic utility function return a signed value of days(secondnight - firstnight).  Note that is different in sign from your `difference` variable, but matches the sign convention of `numpy.subtract(a,b) == b-a`.

I admit that I don't have a use case in mind for when we'd need a signed value, but the Principle of Least Surprise would imply to me that a difference function should return a signed value.

akremin: Thank you for the suggestion. I was referring to difference it in the colloquial sense rather than the mathematical sense, but I've updated the code to returned a signed quantity and the cte takes the absolute value. I also updated the unit tests.

sbailey (approved):

PR #2214, opened 2024-04-15
Add script to create override file

akremin:
This adds a script called desi_create_override_file that will prompt the user for relevant information that can be included in an override file. Some of the information can be provided via the command line for easier scripting. This was used to create the override files needed to set the `autocalib_fiberflat` `--solve_gradient` flag for about a month's worth of nights in 2023 and 2024 (issue #2174). The files are in the proper format, which I tested by running `desi_proc_night` on a test night, which correctly propagated the `--autocal-ff-solve-grad` to `desi_proc_joint_fit`.

Note, because of the way the `pyyaml` module dumps information, the lists will be printed in long form rather than abbreviated bracket ([]) notation if `linkcal` is specified. This minor inconvenience was outweighed by the convenience of being able to script the generation of these files.

Important note: This is not intended to edit existing override files. If an existing override file is detected, it is moved out of the way by inserting a timestamp. The contents of the old override file are NOT used in the generation of the new override. A warning is printed about this, but it is important for the user to either include all relevant override information again to generate the new override via the script (preferred), or copy over the relevant entries from the old file manually after generating the new entries.

### DIFF
--- a/bin/desi_create_override_file
+++ b/bin/desi_create_override_file
@@ -1,0 +1,27 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+import argparse
+from desispec.scripts.createoverride import create_override_file
+
+def get_parser():
+    """
+    Creates an arguments parser for the desi_create_override_file script
+    """
+    parser = argparse.ArgumentParser(usage="{prog} [options]")
+    parser.add_argument("-n", "--night", type=int, default=None,
+                        required=False,
+                        help="Night that the override file should be create for.")
+    parser.add_argument("--linkcal", type=str, default=None, required=False,
+                        help="If True then caliblink will be added to the override file with "
+                             + " prompts obtaining relevant information.")
+    parser.add_argument("--ff-solve-grad", type=str, default=None,
+                        required=False,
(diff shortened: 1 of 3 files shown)


sbailey:
For `--linkcal` and `--ff-solve-grad` let's use
```
parser.add_argument('--linkcal', action=argparse.BooleanOptionalAction, help=...)
parser.add_argument('--ff-solve-grad', action=argparse.BooleanOptionalAction, help=...)
```
which will result in the args being `None` if not specified, but allow the user to specify `--linkcal` / `--no-linkcal` to specifically override in either direction and avoid the interactive prompts.

Suggested rewording: "NOTE: Remember to set the reference night in desi_spectro_calib" -> "NOTE: check that reference fiberflats in desi_spectro_calib are valid for gradient correction on YEARMMDD"

Other minor items:
* in prompts and/or error message, say valid options are y/n or yes/no instead of just "Format of response x was incorrect, please try again"
* at the end, print what file was written

akremin:
Thank you for the useful suggestions. I have implemented all of them in the most recent commit.

PR #2232, opened 2024-05-01
fix QuasarNET afterburner parsing of Redrock templates

### LINKED ISSUE #2231: Fix QuasarNet afterburner usage of Redrock QSO templates
The new redrock-templates organization includes historical templates along with the current ones, and also standardized upon UPPERCASE SPECTYPE instead of lowercase.  This breaks this line from `desispec.scripts.qsoqn`:
```python
    templates_default_qso = glob.glob(os.path.join(os.path.dirname(find_templates()[0]),
                                                   'rrtemplate-qso*.fits'))
```
as-is it will find zero templates because they are now named rrtemplate-QSO*.fits instead of rrtemplate-qso*.fits, and even if that capitalization is changed, it would find 4 templates instead of 2:
```
In [7]: glob.glob(os.path.join(os.path.dirname(find_templates()[0]),
   ...:                                                    'rrtemplate-QSO*.fits'))
DEBUG: Reading templates from /global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/redrock-templates/main
Out[7]: 
['/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/redrock-templates/main/rrtemplate-QSO-None-v0.1.fits',
 '/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/redrock-templates/main/rrtemplate-QSO-LOZ-v1.0.fits',
 '/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/redrock-templates/main/rrtemplate-QSO-HIZ-v1.1.fits',
 '/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/code/redrock-templates/main/rrtemplate-QSO-HIZ-v1.0.fits']
```

I think the right solution is to use `find_templates()` to return the names of the current default templates and then parse those filenames to find the QSO templates instead of re-grepping the directory.

Check whether there is similar code in zmtl too.

sbailey:
This PR fixes #2231 to update the qsoqn (quasarnet) afterburner to correctly parse the new structure of redrock-templates.  It will raise an exception if it doesn't find any templates, and it will log an error but not crash if the number found isn't 2 (HIZ and LOZ expected, but potentially could be different in the future).

This is required for correct qsoqn afterburner runs in Jura (current main silently finds 0 templates and never does the "rerun Redrock with a prior around the QN redshift" step).

I'll leave this open for an hour or so if there are rapid comments, but we need to merge it for testing tonight after Perlmutter is back.  Heads up @dylanagreen @echaussidon @akremin @ashleyjross @geordie666 .

I tested this with
```
desispec.scripts.qsoqn.main([
  '--coadd', '/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/badampcals/tiles/cumulative/2824/20240408/coadd-0-2824-thru20240408.fits',
  '--redrock', '/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/badampcals/tiles/cumulative/2824/20240408/redrock-0-2824-thru20240408.fits',
  '--output', '/global/cfs/cdirs/desi/users/sjbailey/spectro/redux/badampcals/tiles/cumulative/2824/20240408/qso_qn-0-2824-thru20240408.fits',
  '--target_selection', 'all', '--save_target', 'all'])
```
and verified that previously it wasn't running rrdesi at all, and now it re-runs it for LOZ and HIZ templates as expected.

### DIFF
--- a/py/desispec/scripts/qsoqn.py
+++ b/py/desispec/scripts/qsoqn.py
@@ -31,12 +31,26 @@ from redrock.external.desi import rrdesi
 from quasarnp.io import load_model, load_desi_coadd
 from quasarnp.utils import process_preds
 
+def get_default_qso_templates():
+    """Return list of default Redrock QSO template filenames to use"""
+    log = get_logger()
+    all_templates = find_templates()
+    qso_templates = list()
+    for filename in all_templates:
+        if os.path.basename(filename).startswith('rrtemplate-QSO-'):
+            qso_templates.append(filename)
+
+    #- Expect HIZ and LOZ but not more than that
+    #- otherwise log error but don't actually crash
+    if len(qso_templates) != 2:
+        log.error(f'Unexpected number of QSO templates found: {qso_templates}')
+    else:
+        log.debug('Using Redrock templates %s', qso_templates)
+
+    return qso_templates
+
 
 def parse(options=None):
-    # LOAD ONLY QSO templates to RE-RUN redrock
-    templates_default_qso = glob.glob(os.path.join(os.path.dirname(find_templates()[0]),
-                                                   'rrtemplate-qso*.fits'))
-
     parser = argparse.ArgumentParser(description="Run QN and rerun RR (only for SPECTYPE != QSO or for SPECTYPE == QSO && |z_QN - z_RR| > 0.05) on a coadd file to find true quasars with correct redshift")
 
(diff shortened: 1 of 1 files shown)


echaussidon:
Sounds good to me. 

I even not realized that the template names were modified.

PR #2235, opened 2024-05-03
add reservation parsing for cpu vs. gpu

sbailey:
This PR adds support for specifying separate reservation names for CPU and GPU reservations, with either of them optionally being None, e.g. if we had a CPU reservation but not a GPU reservation.

Examples in $CFS/desi/users/sjbailey/spectro/redux/j1
```
desi_proc_night -n 20240409 --dry-run-level 3 --reservation blat &> submit-20240409-singlereservation.log
desi_proc_night -n 20240409 --dry-run-level 3 --reservation blat_cpu,foo_gpu &> submit-20240409-dualreservations.log
desi_proc_night -n 20240409 --dry-run-level 3 --reservation blat_cpu,None &> submit-20240409-cpuonlyreservation.log
```

grep for "sbatch" in each of those log files:
  * singlereservation: all jobs have `--reservation=blat`
  * dualreservation: jobs have `--reservation=blat_cpu` or `--reservation=foo_gpu` depending upon their job type
  * cpuonlyreservation: CPU jobs have `--reservation=blat_cpu`, but GPU jobs don't have any --reservation specified.

This also works for desi_resubmit_queue_failures since that uses the same `submit_batch_script` function.  Yay!  I tested this with the ongoing j1 production:
```
desi_resubmit_queue_failures -n 20240403 --dry-run --reservation blat_cpu,foo_gpu > $SCRATCH/resubmit-test.log
grep sbatch /pscratch/sd/d/desi/resubmit-test.log
```
--> The 4 ztile jobs to resubmit had `--reservation=foo_gpu`

Implementation note: I opted to not enforce _cpu/_gpu endings to reservation names, which gives us more flexibility for reservation names in the future and makes it easier to support having a reservation for one but not the other.

@akremin please check

### DIFF
--- a/py/desispec/workflow/batch.py
+++ b/py/desispec/workflow/batch.py
@@ -11,6 +11,7 @@ import yaml
 
 from desiutil.log import get_logger
 
+_config_cache = dict()
 def get_config(name):
     """
     Return configuration dictionary for system `name`
@@ -30,12 +31,19 @@ def get_config(name):
     if name is None:
         name = default_system()
 
+    global _config_cache
+    if name in _config_cache:
+        return _config_cache[name]
+
     configfile = resources.files('desispec').joinpath('data/batch_config.yaml')
     with open(configfile) as fx:
         config = yaml.safe_load(fx)
 
     #- Add the name for reference, in case it was default selected
     config['name'] = name
+
+    #- Add to cache so that we don't have to re-read batch_config.yaml every time
+    _config_cache[name] = config[name]
 
     return config[name]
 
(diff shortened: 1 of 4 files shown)


py/desispec/workflow/batch.py
```diff
@@ -76,6 +84,44 @@ def default_system(jobdesc=None, no_gpu=False):
...
+    else:
+        raise ValueError(f'Unable to parse {reservation} as rescpu,resgpu')
+
+    if reservation_cpu.lower() == 'none':
+        reservation_cpu = None
+
+    if reservation_gpu.lower() == 'none':
+        reservation_gpu = None
+
+    system_name = default_system(jobdesc)
```
akremin, line 118: Given that the cpu/gpu seperation is somewhat specific to how Perlmutter queues are partitioned, it's not obvious to me that we need to generalized this to checking the configuration for GPU's. We could just use the system name, which is either perlmutter_cpu or perlmutter_gpu.

For NERSC10 the partitions may be divided differently in a way that requires a refactor anyway.

I don't feel very strongly about this, so if you do then it is fine as-is. My proposal simplifies the need for getting and caching the config file, but otherwise is slightly less generic.

akremin (commented):
Thanks, the code looks good. I put one comment inline. We could simplify this if we're willing to go all-in on Perlmutter and update for the next machine (which we will likely have to do anyway since it may not split partitions based on cpu's and gpu's).

py/desispec/workflow/batch.py
sbailey: I agree that this is a step more than we need for Perlmutter itself, but I think it gives us slightly more flexibility for the future, though you are right that NERSC-10 might throw us a monkey wrench either way.  Note that if we specify just a single reservation name it is the only one used so we at least have that fallback.  I suggest leaving this as-is for now.

sbailey:
@akremin please also update to move nightlyflat jobs to GPU nodes, and then merge when ready.  Let's test this on j1 while we still have the jura1_cpu and jura1_gpu reservations.

PR #2245, opened 2024-05-13
Update desispec.module

Waelthus:
This patch updates the DESI_SPECTRO_DARK directory to a new one for upcoming observations. Currently, the new directory only contains symlinks to previous files, but new calibs for post-Y3 observations will go to the new directory only.

### DIFF
--- a/etc/desispec.module
+++ b/etc/desispec.module
@@ -77,7 +77,7 @@ setenv [string toupper $product] $PRODUCT_DIR
 # Add any non-standard Module code below this point.
 #
 setenv DESI_SPECTRO_CALIB $env(DESI_ROOT)/spectro/desi_spectro_calib/trunk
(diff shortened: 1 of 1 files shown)


Waelthus:
added a symlink pointing to the most up-to-date directory, using that by default

sbailey:
Good idea to have a "latest" link so that we can update darks for desispec/main without having to re-install the desispec/main module file.

The jura production environment specifically enforces v2209 so it is safe to update this here.  I will merge and update the desispec/main environment at NERSC.

sbailey:
Having just merged this, I'm no longer sure this is what we want to do, because this will also impact future tags, not just main.  I don't think it is viable to assume that whenever we make a desispec tag will will remember to temporarily change this back to a tagged version.  Rather than immediately undo this, I'm going to continue the conversation on desihub/desimodules#49 to see if we can converge on what the best practices should be.

PR #2252, opened 2024-05-14
Refactor read_gfa_data function

### LINKED ISSUE #2250: desi_tsnr_afterburner can't find more recent GFA files
In `desi_tsnr_afterburner`, `read_gfa_data()` will not find recent GFA files, because the 'SV1', 'SV2', 'SV3' labels are hard-coded into the function. We should also verify which directory the function is using to find the files.

weaverba137:
This PR closes #2250.

* `read_gfa_data()` returns the GFA summary file with the most recent date, independently of any label (`SV1`, `SV2`, `SV3`, `main`, `foobar`, etc.).
* `read_gfa_data()` is moved into `desispec.tilecompleteness`, which makes it easier to write a unit test.
* Cleaned up some unused import statements.

### DIFF
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -36,7 +36,7 @@ from   desiutil.log import get_logger
 from   desispec.tsnr import calc_tsnr2,tsnr2_to_efftime
 from   astropy.table import Table, vstack
 from   desiutil.depend import getdep
-from   desispec.tilecompleteness import compute_tile_completeness_table,merge_tile_completeness_table
+from   desispec.tilecompleteness import read_gfa_data, compute_tile_completeness_table, merge_tile_completeness_table
 from   desispec.skymag import compute_skymag
 from   desispec.efftime import compute_efftime
 from   desispec.parallel import  stdouterr_redirected, use_mpi
@@ -161,35 +161,6 @@ def update_targ_info(entry, targ_in):
         entry['GOALTYPE'] = faflavor2program(entry['FAFLAVOR'])
 
     return entry
-
-
-def read_gfa_data(gfa_proc_dir) :
-    """
-    Read the directory with the offline GFA data reduction (like /global/cfs/cdirs/desi/survey/GFA/),
-    find the latest version of the table files for the various surveys, return the merged table.
-    See documentation here https://desi.lbl.gov/trac/wiki/SurveyValidation/SV1/conditions/summary_files
-    Args:
-      gfa_proc_dir: str, directory path
-    returns astropy.table.Table
-    """
-    log = get_logger()
-    tables=[]
-    for survey in ["SV1","SV2","SV3"] :
-        filenames=sorted(glob.glob("{}/offline_matched_coadd_ccds_{}-thru_????????.fits".format(gfa_proc_dir,survey)))
-        if len(filenames)==0 : continue
-        filename=filenames[-1]
-        log.info(f"Reading {filename}")
-        table=read_table(filename,2)# HDU2 is average over frames during spectro exposure and median across CCDs
-        tables.append(table)
-    if len(tables)==0 :
-        log=get_logger()
-        mess="did not find any file offline_matched_coadd_ccds_*-thru_????????.fits in {}".format(gfa_proc_dir)
-        log.critical(mess)
-        raise RuntimeError(mess)
-    table=vstack(tables)
-    log.info(f'{len(table)} GFA table entries')
-    return table
-
 
 
 def compute_tsnr_values(cframe_filename,cframe_hdulist,night,expid,camera,specprod_dir, alpha_only=False) :
--- a/py/desispec/test/test_tilecompleteness.py
+++ b/py/desispec/test/test_tilecompleteness.py
@@ -12,24 +12,31 @@ class TestTileCompleteness(unittest.TestCase):
     """Test desispec.tilecompleteness.
     """
 
+    @patch('desispec.tilecompleteness.vstack')
     @patch('desispec.tilecompleteness.read_table')
     @patch('desispec.tilecompleteness.glob')
     @patch('desispec.tilecompleteness.get_logger')
-    def test_read_gfa_data(self, mock_log, mock_glob, mock_table):
+    def test_read_gfa_data(self, mock_log, mock_glob, mock_table, mock_vstack):
         """Test identification of the most recent GFA file.
         """
         gfa_proc_dir = '/global/cfs/cdirs/desi/survey/GFA'
-        mock_glob.glob.return_value = [os.path.join(gfa_proc_dir, 'offline_matched_coadd_ccds_sv1-thru_20201231.fits'),
-                                       os.path.join(gfa_proc_dir, 'offline_matched_coadd_ccds_sv2-thru_20210101.fits'),
-                                       os.path.join(gfa_proc_dir, 'offline_matched_coadd_ccds_sv3-thru_20210102.fits'),
+        mock_glob.glob.side_effect = [[os.path.join(gfa_proc_dir, 'offline_matched_coadd_ccds_SV1-thru_20201231.fits'),],
+                                      [os.path.join(gfa_proc_dir, 'offline_matched_coadd_ccds_SV1-thru_20201231.fits'),
+                                       os.path.join(gfa_proc_dir, 'offline_matched_coadd_ccds_SV2-thru_20210101.fits'),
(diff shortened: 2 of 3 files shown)


weaverba137:
@sbailey, please take a look at `desispec.tilecompleteness.number_of_good_redrock`. I don't think that function ever would have worked, in part due to an import error, but also it is not used by anything at all in desispec.

araichoor:
sorry I m late here; I don t have time right now to dive in the details of the PR and the related issues, neither to dig into the data.
if my comment is relevant, apologies in advance, just discard it!

my remark, with just reading the PR comment:
- the latest GFA file will be some `offline_matched_coadd_ccds_SV3-thru_{NIGHT}.fits` file:  those files contain exposures starting from SV3:
```
>>> d = Table.read("/global/cfs/cdirs/desi/survey/GFA/offline_matched_coadd_ccds_SV3-thru_20240409.fits", 3)
>>> d["NIGHT"].min()
20210405
```
- SV1 and SV2 exposures should be in some `offline_matched_coadd_ccds_{SV1,SV2}-thru_{NIGHT}.fits` files.

for the daily operations, it shouldn t be a problem, as earlier (including SV1, SV2) GFA info already is in the `exposures-daily.csv` file, so the code will just look for GFA info for the latest exposures, which will be in the `offline_matched_coadd_ccds_SV3-thru_{NIGHT}.fits`.

but for a production, e.g. `jura`, which will reprocess all exposures since SV1,  it sounds to me that the code won t find the relevant GFA info files for pre-SV3 exposures.

I see those two "latest" files:
```
ls -l /global/cfs/cdirs/desi/survey/GFA_orig/offline_matched_coadd_ccds_SV1-thru_* | tail -n 1
-r--r--r-- 1 ameisner desi 53864640 Sep 29  2021 /global/cfs/cdirs/desi/survey/GFA_orig/offline_matched_coadd_ccds_SV1-thru_20210928.fits

ls -l /global/cfs/cdirs/desi/survey/GFA_orig/offline_matched_coadd_ccds_SV2-thru_* | tail -n 1
-r--r--r-- 1 ameisner desi 29888640 Sep 29  2021 /global/cfs/cdirs/desi/survey/GFA_orig/offline_matched_coadd_ccds_SV2-thru_20210928.fits
```

looks like the SV1 file contains all exposures from the SV2 file, ie we may just need `/global/cfs/cdirs/desi/survey/GFA_orig/offline_matched_coadd_ccds_SV1-thru_20210928.fits` for pre-SV3 exposures.
as this exposure list static, I guess it s fine to hard-code this path (unless we do at some point a whole reprocessing of the GFA data).

weaverba137:
Do we know specifically how this was done for `fuji`, `guadalupe` and `iron`? In other words, did these read from `/global/cfs/cdirs/desi/survey/GFA_orig` or `/global/cfs/cdirs/desi/survey/GFA`?

sbailey:
Iron used `/global/cfs/cdirs/desi/survey/GFA`on Feb 8 2023.  That path is currently a symlink to `GFA.NERSC`, created on Feb 16 2024, i.e. created a year after Iron was run.  `GFA_orig` is a symlink, also created on Feb 16 2024, to `../users/ameisner/GFA/conditions`.  I was expecting `GFA_orig` to point to what we used for Iron, but that `ameisner/GFA/conditions` directory only goes though `offline_matched_coadd_ccds_SV3-thru_20211013.fits`, which is different from what we cached in `/global/cfs/cdirs/desi/public/dr1/survey/GFA/offline_matched_coadd_ccds_SV3-thru_20220613.fits`.

i.e. I'm not sure what we actually used for iron and prior.

weaverba137:
A possible way to rewrite the function `read_gfa_data()` would be:

1. Read the SV1 file. There will only ever be one of them at this point.
2. Concatenate the most recent GFA summary file, ignoring any label, considering only the last night.

sbailey:
Thanks @weaverba137 for digging into the files and the set theory.  My understanding of the situation now:
* not including `offline_matched_coadd_ccds_SV1-thru_20210928.fits` in EDR, draft DR1, and current `/global/cfs/cdirs/desi/survey/GFA` appears to be an oversight.  At minimum we should add it to EDR and DR1.
* If we also add it to `/global/cfs/cdirs/desi/survey/GFA`, the current main code would work because it grabs the latest of each of `offline_matched_coadd_ccds_{SV1,SV2,SV3}-thru_????????.fits` and concatenates them together (albeit resulting in duplicate rows due to date overlaps in the files).  In some sense, the problem isn't in the code but in our misunderstandings about what files needed to be preserved when transitioning from KPNO -> NERSC processing; we should not have dropped `offline_matched_coadd_ccds_SV1-thru_20210928.fits`.
* None-the-less, this PR has advantages of adding unit tests and forcing us to actually understand the situation.  But as is, it would not work because it doesn't pickup the SV1 data even if it was there as @araichoor noticed (thanks!).
* For the purposes of going back on-sky next week, nothing is urgent because daily ops only needs the latest file that had the latest night appended to it and it doesn't matter that SV1 is missing.
* For the purposes of Jura, re-adding `offline_matched_coadd_ccds_SV1-thru_20210928.fits` is the minimum needed, though we may want/need to cleanup how all of this works anyway.

Options:
  * Re-add `offline_matched_coadd_ccds_SV1-thru_20210928.fits` immediately, then in this PR revert `read_gfa_data()` to the "stack the latest SV1,SV2,SV3-thru_????????.fits file" method, while keeping the additional unit tests of this PR (updated to match the logic if needed).
  * Do some file cleanup to make this conceptually cleaner so that we are less likely to misunderstand it.  Part of the conceptual mess is that the naming scheme "{SURVEY}-thru_{NIGHT}" doesn't work because surveys overlapped in time so our start and end markers aren't the same thing.  Options could include
    *  go back to "the latest file actually has everything starting with SV1" rather than needing to mix-and-match anything.
    *  non-overlapping files grouped by SURVEY but only containing the exposures for that survey; nights with multiple SURVEYs would have their exposures split across multiple files.  I'm not sure we really have a use case for wanting to read the GFA data for a single survey, so this smells like a hassle.
    * change to FIRSTNIGHT-thru-LASTNIGHT regardless of what SURVEYs were represented in those nights.
    * Make one file per night rather than continuously concatenating and writing a new file.  The tsnr afterburner would then read N>>1 small files when rebuilding the catalog from scratch, but I suspect that wouldn't actually be too bad.  In particular the tsnr afterburner works night-by-night anyway so it might already be reading the latest big files N>>1 times.

weaverba137:
> Re-add offline_matched_coadd_ccds_SV1-thru_20210928.fits immediately, then in this PR revert read_gfa_data() to the "stack the latest SV1,SV2,SV3-thru_????????.fits file" method, while keeping the additional unit tests of this PR (updated to match the logic if needed).

I think my proposal above is a simplified version of that.  Would that be acceptable?

sbailey:
Yes, I think that would be fine, with the inclusion of actually adding the SV1 file back to the GFA directory as well.  This PR already has the code for sorting by night regardless of what future SURVEY prefixes may exist, so the only change is to also read the latest SV1 file as a special case and concatenate the two.  Good.

weaverba137:
OK, I'll probably have to update that on Monday at this point, and I'll take care of adding in the SV1 file.

weaverba137:
I can also add some of your explanations above to the documentation.

weaverba137:
For the SV1 file: `offline_matched_coadd_ccds_SV1-thru_20210405.fits` would have the minimal overlap with any current file. The last EXPID is 83543, and the first expid in SV3 and later is 83522. However, given the intermingling of SV1 & SV3, perhaps we really need to take the last SV1 file, and just accept the overlap. Thoughts?

weaverba137:
`offline_matched_coadd_ccds_SV1-thru_20210928.fits` is now visible in EDR, DR1 and the GFA directory for daily processing.

weaverba137:
I think this is ready for testing and review now.

weaverba137:
@sbailey, just a reminder to test this once we start getting new GFA data.

sbailey:
Summarizing a chat with @weaverba137 
* this PR is needed for daily ops with new data, hopefully starting tonight
* at the same time, this was not used for Jura, so if we need any additional final tags for Jura, they should *not* be based on main after merging this

To resolve this, I have made a new branch "jura" which we can use for any last minute Jura updates + tags, thus opening up "main" for daily ops again, including this PR.  Until we are completely done with Jura, we will need to carefully consider each PR for whether it should be to main or jura.  The basic workflow rule is that we can merge PR -> jura -> main, or PR -> main, but not PR -> main -> jura because that could bring in other changes on main that we don't want in jura tags.

PR #2265, opened 2024-05-24
optimizations for creating specprod tiles table

sbailey:
This PR adds several I/O optimizations and a placeholder script for creating a tiles-SPECPROD.fits file outside of the context of the tsnr afterburner.  This script is what is used for the tiles file currently in jura (as of 2024-05-24) but will need to be updated before the final Jura release.

  * new script bin/desi_create_tiletable
  * zcatalog uses /dvs_ro instead of /global, and glob instead of iterfiles
  * tsnr caches the tsnr_ensemble instead of re-reading it from disk every time `tsnr2_to_efftime` is called

### DIFF
--- a/bin/desi_create_tiletable
+++ b/bin/desi_create_tiletable
@@ -1,0 +1,150 @@
+#!/usr/bin/env python
+
+"""
+Create the tiles table for a specprod
+"""
+
+import os, sys, glob
+import multiprocessing
+import argparse
+import numpy as np
+from astropy.table import Table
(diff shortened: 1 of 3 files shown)


akremin (approved):
Thanks, the changes look good. Interesting use of `itertools.chain.from_iterable`, I've never seen that used before.

I have some thoughts on `desi_create_tiletable` including possible name changes, but since that is still a work in progress we can revisit it in a future PR.

PR #2280, opened 2024-06-17
revert QN_MODEL_FILE to old weights

sbailey:
This PR reverts the desispec modules file to use the old QuasarNET weights file, $DESI_ROOT/target/catalogs/lya/qn_models/qn_train_coadd_indtrain_0_0_boss10.h5

We'll use this one for daily until we've further debugged the new weights file + afterburners.  The new weights file qn_train_desi_guadalupe_linear.h5 was used in Jura and for data processed 20240603 - 20240616.

I've discussed with @akremin and plan to self merge to get this update in before tonight's processing (20240617).


akremin (approved):
Thanks for the change and for documenting why we are changing back for the time being

PR #2291, opened 2024-07-12
optionally parallelize group_spectra

sbailey:
This PR adds optional MPI parallelization to desi_group_spectra.  I have not yet integrated this into desi_zproc because that will require more plumbing on the zproc side, but at minimum this provides a standalone way to generate large healpix if needed so I'm opening this as an independent PR and will do the zproc integration in a followup PR.

e.g. one of our largest most problematic healpix in jura was special other 27258 with over 4000 input frame files.  This branch can group those spectra in 11 minutes with
```
HPIX=27258
time srun -n 64 -c 4 desi_group_spectra --mpi --healpix $HPIX \
  --expfile $DESI_SPECTRO_REDUX/$SPECPROD/healpix/special/other/272/$HPIX/hpixexp-special-other-$HPIX.csv \
  -o $SCRATCH/temp/spectra-special-other-$HPIX.fits -c $SCRATCH/temp/coadd-special-other-$HPIX.fits
```
The equivalent took hours in jura with the serial code in main.
FYI `srun -n 128 -c 2 ...` also works without blowing memory, but isn't actually any faster (likely saturating node I/O).

On smaller healpix (e.g. sv1 dark 7015) I also verified that this still works in non-MPI mode and that the outputs are data-identical in both cases (header timestamps and dependency versions differ, ok).

Most of the changes are for MPI boilerplate "if rank == 0" sort of stuff, plus factoring out the frame file reading + filtering into a separate function so that it can be parallelized using
```python
with MPICommExecutor(comm, root=0) as pool:
    frames = list(pool.starmap(_read_framefile, read_args))
```
This pattern could also be used to parallelize with multiprocessing if needed, but I didn't add that option.

### DIFF
--- a/py/desispec/scripts/group_spectra.py
+++ b/py/desispec/scripts/group_spectra.py
@@ -50,6 +50,8 @@ def parse(options=None):
             help="output coadded spectra filename")
     parser.add_argument("--onetile", action="store_true",
             help="input spectra are from a single tile")
+    parser.add_argument("--mpi", action="store_true",
+            help="use MPI for parallelism")
 
     if options is None:
         args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


akremin (approved):
Thanks the changes all look good and I will take your tests as sufficient. We will run tests before Kibo which should catch any corner cases I don't see reading the code.

The use of `log0` for only printing info once is clever in reducing the number of `if root == 0` lines and indents.

Issue #2264, opened 2024-05-24
cross-night ztile -> tilenight dependency tracking


sbailey:
Followup to #2263:

ztile jobs require the cframe files from all nights/expids to exist to make the combined spectra files, and also the exposure-qa files from all nights/expids to make the tile-qa file.  We do not track cross-night dependencies for ztile -> tilenight jobs, but instead just let them run and crash if they don't find the cframes they need, and then resubmit them.  However, this procedure doesn't work if the cframes exist but the exposure-qa does not.  An example is the Jura processing of tile 23551 was observed on 20220403 and 20220404:
* tilenight-20220403-23551 jobid 25371075 ran on May 8 and produced cframes, but timed out before completing exposure-qa.
* tilenight-20220404-23551 job 25394465 and ztile-23551-thru20220404 job 25394469 ran on May 9.
* ztile-23551-thru20220404 did *not* immediately crash because it found all of the cframes it needed.  When it got to tile-qa, it attempted to re-generate the missing exposure-qa file but due to a `night` vs. `exposure_night` but, it failed to re-generate the exposure-qa and treated that exposure as QA_EFFTIME=0, causing the tile to not pass QA.
* On May 13, a resubmission cleanup pass noticed that tilenight-20220403-23551 had failed and resbumitted it, which successfully generated the previously missing exposure-qa-00128581.fits file.

Action items:
* fix the `night` vs. `exposure_night` bug identified in #2263; and/or
* ztile should exit with an error an input exposure-qa and/or it can't re-generate it, instead of falling back to treating it was QA_EFFTIME=0 and declaring success.

For the purposes of Jura, I'm going to remove the tile-qa files and rerun tile-qa for the impacted tiles so that they will pick up the previously-missing-but-now-existing exposure-qa files.

Adding this to the Kilimanjaro dashboard.

sbailey:
I meant to post this here instead of #2263; reposting here to keep the comment with the ticket that will be open until we fix the underlying problem.

Belt-and-suspenders-and-duct-tape: zproc knows that tile-qa needs the coadd and redrock files as input:

INFO:util.py:128:runcmd: RUNNING: desispec.scripts.tileqa.main(['-g', 'cumulative', '-n', '20220404', '-t', '23551'])
  Inputs
    /global/cfs/cdirs/desi/spectro/redux/jura/tiles/cumulative/23551/20220404/coadd-0-23551-thru20220404.fits
[...]

Issue #2263, opened 2024-05-24
Jura tile 82265 much lower qa_efftime than Iron


sbailey:
@araichoor can you debug why tile 82265 (special-bright) in Jura has much lower QA_EFFTIME than it did in Iron:

https://data.desi.lbl.gov/desi/spectro/redux/iron/tiles/cumulative/82265/20211006/tile-qa-82265-thru20211006.png
https://data.desi.lbl.gov/desi/spectro/redux/jura/tiles/cumulative/82265/20211006/tile-qa-82265-thru20211006.png

Night 20211003 expid 102826 is the only exposure with a decent EFFTIME_SPEC=191 but the QA_EFFTIME when from 195 to 0.  Why?

Thanks to @abrodze for identifying this in Jura QA.

sbailey:
Other similar cases also noted by @abrodze 

Tile 23551 (main-bright) exposure 20220403/128581 went from QA_EFFTIME 76 -> 0
https://data.desi.lbl.gov/desi/spectro/redux/iron/tiles/cumulative/23551/20220404/tile-qa-23551-thru20220404.png
https://data.desi.lbl.gov/desi/spectro/redux/jura/tiles/cumulative/23551/20220404/tile-qa-23551-thru20220404.png

tile 42621 (main-backup) exposure 20211218/114628 went from QA_EFFTIME 41 -> 0
https://data.desi.lbl.gov/desi/spectro/redux/iron/tiles/cumulative/42621/20220114/tile-qa-42621-thru20220114.png
https://data.desi.lbl.gov/desi/spectro/redux/jura/tiles/cumulative/42621/20220114/tile-qa-42621-thru20220114.png

araichoor:
from a quick debug on 23551:

I think it s a combination of two things, which ends up setting efftime=0 for expid=128581 (instead of ~76s):

(1) when the tile-qa has been run, the `exposure-qa-00128581.fits` (from 20220403) was missing:
```
-rw-r----- 1 desi desi 630K May  9 12:16 /global/cfs/cdirs/desi/spectro/redux/jura/tiles/cumulative/23551/20220404/tile-qa-23551-thru20220404.fits
-rw-r----- 1 desi desi 445K May 13 11:09 /global/cfs/cdirs/desi/spectro/redux/jura/exposures/20220403/00128581/exposure-qa-00128581.fits
```

so the code tries to get the efftime for that exposure:
[...]

PR #2306, opened 2024-07-23
Have zproc require input exposure-qa files for tileqa 

akremin:
### Overview
This more fully solves the issue presented in issue #2263, before beginning the large implementation of cross-night dependencies needed for issue #2264.

First, in `zproc.py` it now requires the input exposure-qa's to exist before proceeding in producing the tileqa file. Previously it would run even if they didn't exist.

Second, this fixes a typo in `tile_qa.py` such that the exposure-qa file will be created for each night if the code is run and the exposure-qa file is missing. Of course, the first change in this PR makes it so that this doesn't occur when running from `zproc`, but this is still a useful fix when running outside of the pipeline where old data may not have the exposure-qa files pre-generated.

### Test
I ran a test on tile 9203  observed on nights 20240408 and 20240409 with exposures 235010 and 235200 

I first sym-linked the files in the preproc and exposures directories from Jura. I then removed the exposure-qa link from 23510 on night 20240408.

I then ran:
`srun -N 1 -n 64 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores desi_zproc -t 9203 -g cumulative -n 20240408 20240409 -e 235010 235200 -c a0123456789 --mpi --run-zmtl --max-gpuprocs 4 --starttime $(date +%s) |& tee jobout_desi_zproc_9203_cumulative_jura`

This failed at the tileqa step because of the missing exposure-qa file. This is the new correct behavior.

I then ran `desi_tile_qa -g cumulative -n 20240409 -t 9203` which identified the missing exposure-qa file and successfully created it. This fails in main because of a typo that is fixed in this branch.

I then reran `srun -N 1 -n 64 -c 2 --gpu-bind=map_gpu:3,2,1,0 --cpu-bind=cores desi_zproc -t 9203 -g cumulative -n 20240408 20240409 -e 235010 235200 -c a0123456789 --mpi --run-zmtl --max-gpuprocs 4 --starttime $(date +%s) |& tee jobout_desi_zproc_9203_cumulative_jura`,

which was successful now that the exposure-qa file had been generated.

This successfully tested both changes to the code in this PR.

### DIFF
--- a/py/desispec/scripts/zproc.py
+++ b/py/desispec/scripts/zproc.py
@@ -612,9 +612,7 @@ def main(args=None, comm=None):
         ## requires all coadd and redrock outputs in addition to exposureqa
         infiles = []
         for expid, night in zip(expids, nights):
-            findfileopts['expid'] = expid
-            findfileopts['night'] = night
-            infiles.append(findfile('exposureqa', **findfileopts))
+            infiles.append(findfile('exposureqa', expid=expid, night=night, readonly=True))
         for spectro in all_subgroups:
             findfileopts['spectrograph'] = spectro
             infiles.append(findfile('coadd', **findfileopts))
--- a/py/desispec/scripts/zproc.py
+++ b/py/desispec/scripts/zproc.py
@@ -608,7 +608,9 @@ def main(args=None, comm=None):
         qafile = findfile('tileqa', **findfileopts)
         qapng = findfile('tileqapng', **findfileopts)
         qalog = findfile('tileqa', logfile=True, **findfileopts)
+        ## requires all coadd and redrock outputs in addition to exposureqa
         infiles = []
+        infiles.append(findfile('exposureqa', **findfileopts))
         for spectro in all_subgroups:
             findfileopts['spectrograph'] = spectro
             infiles.append(findfile('coadd', **findfileopts))
--- a/py/desispec/scripts/zproc.py
+++ b/py/desispec/scripts/zproc.py
@@ -203,8 +203,9 @@ def main(args=None, comm=None):
             log.error(msg)
             raise ValueError(msg)
         else:
-            msg = f"Only using exposures specified with --expids {args.expids}"
-            log.info(msg)
+            if rank == 0:
+                msg = f"Only using exposures specified with --expids {args.expids}"
+                log.info(msg)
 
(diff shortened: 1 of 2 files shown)


py/desispec/scripts/zproc.py
```diff
@@ -608,7 +609,12 @@ def main(args=None, comm=None):
         qafile = findfile('tileqa', **findfileopts)
         qapng = findfile('tileqapng', **findfileopts)
         qalog = findfile('tileqa', logfile=True, **findfileopts)
+        ## requires all coadd and redrock outputs in addition to exposureqa
         infiles = []
+        for expid, night in zip(expids, nights):
+            findfileopts['expid'] = expid
+            findfileopts['night'] = night
+            infiles.append(findfile('exposureqa', **findfileopts))
```
sbailey, line 617: This loop has the side effect of leaving `findfileopts['night']` to be the last entry in the `nights` list (similar for expid).  That happens to be the same as the original `findfileopts['night'] = thrunight` value if `groupname='cumulative'`, but that depends upon how `nights` and `thrunight` were constructed hundreds of lines earlier.  That could change in the future without realizing the impact here.

This is admittedly not the only place that modifies `findfileopts` on the fly (and those other places likely trace back to me), but it still feels especially fragile here since it is touching nights that aren't the thrunight of the zproc.  Since `findfile('exposureqa', ...)` only needs `night` and `expid` and doesn't need the other variables that are currently set in `findfileopts`, I suggest bypassing that with something like
```python
for expid, night in zip(expids, nights):
    infiles.append(findfile('exposureqa', night=night, expid=expid, readonly=True))
```
or otherwise making a copy of `findfileopts` into a different dictionary that is modified on the fly without changing the original `findfileopts` and risking leaving it with a different night and expid than before.

sbailey (changes requested):
Thanks for fixing this; it will help Kibo run just a bit smoother.  I put one inline comment about the maintainability of modifying `findfileopts` on the fly in a way that might not leave it in the same state as before.  It currently does leave it in an equivalent state, but it's a long term maintenance issue that I'm concerned about.  Please consider that, or document any pushback ideas.  Otherwise looks good.  Thanks.

akremin:
Thanks @sbailey, I've made the requested change. Please review and approve+merge if you're happy.

sbailey (approved):

PR #2311, opened 2024-07-26
cleanup redirected stdout crash before re-redirecting

### LINKED ISSUE #2269: stdouterr_redirected recovery from mid-stream crash
`desispec.parallel.stdouterr_redirected` uses `desispec.io.util.backup_filename` to make a backup instead of clobbering the original, e.g. redrock-blah.log -> redrock-blah.log.1.

However, if the code crashes due to an MPI abort in the middle of a redirect, it leaves behind a bunch of per-rank files like redrock-blah.log_0, redrock-blah.log_1, redrock-blah.log_2 etc. i.e. it never gets a change to run the per-rank merging -> final logfile in the try/finally block ([L465](https://github.com/desihub/desispec/blob/main/py/desispec/parallel.py#L465)).  Then if the code is  re-run, those files get overwritten, losing the log info from the crash.

I think what we want is for `stdouterr_redirected` to auto-detect that per-rank files exist, merge those into a single log, and back that up before proceeding with writing new per-rank files.

sbailey:
This PR fixes #2269 by merging any per-rank log files leftover from a previous crash before proceeding with a new stdout/stderr log redirect per-rank.  Previously these logfiles would get overwritten, losing the record of the crash.

I tested this with $CFS/desi/users/sjbailey/dev/stdout_redirect/mpi_test_redirect which purposefully crashes a requested rank.  e.g.
```
# run without crashing any ranks; generates a blat.log file
$> srun -n 4 python mpi_test_redirect
Starting up; will crash rank -1
INFO:parallel.py:465:stdouterr_redirected: Begin log redirection to blat.log at Thu Jul 25 17:46:51 2024
INFO:parallel.py:524:stdouterr_redirected: End log redirection to blat.log at Thu Jul 25 17:46:51 2024
All done
$> ls
blat.log  mpi_test_redirect

# run while crashing rank 2, leaves behind backup of previous run and per-rank files from this run
$> srun -n 4 python mpi_test_redirect 2
Starting up; will crash rank 2
INFO:parallel.py:465:stdouterr_redirected: Begin log redirection to blat.log at Thu Jul 25 17:48:12 2024
srun: error: nid200053: task 2: Segmentation fault
srun: Terminating StepId=28590762.10
slurmstepd: error: *** STEP 28590762.10 ON nid200053 CANCELLED AT 2024-07-26T00:48:15 ***
srun: error: nid200053: tasks 0-1,3: Terminated
srun: Force Terminated StepId=28590762.10
$> ls
blat.log.0  blat.log-rank0  blat.log-rank1  blat.log-rank2  blat.log-rank3  mpi_test_redirect

# run again without crashing; will cleanup blat.log-rank* files, back them up to blat.log.1
# then generate blat.log
$> srun -n 4 python mpi_test_redirect
Starting up; will crash rank -1
INFO:parallel.py:465:stdouterr_redirected: Begin log redirection to blat.log at Thu Jul 25 17:50:24 2024
INFO:parallel.py:524:stdouterr_redirected: End log redirection to blat.log at Thu Jul 25 17:50:24 2024
All done
$> ls
blat.log  blat.log.0  blat.log.1  mpi_test_redirect
```

Detail: previously the per-rank files had names like `output.log_0` which was confusingly similar to the backup names like `output.log.0`.  I changed this to `output.log-rank0` instead to make it clearer which files are leftover per-rank files and which are the backups.

Note: if a crash happens cleanly with an exception, `stdouterr_redirected` recovers and is able to merge the per-rank files.  The leftover per-rank logfiles case comes from hard crashes like OOM or in this test case, from a purposeful segfault.

### DIFF
--- a/py/desispec/parallel.py
+++ b/py/desispec/parallel.py
@@ -10,6 +10,7 @@ from __future__ import print_function, absolute_import, division
 import os
 import sys
 import time
+import glob
 import io
 from contextlib import contextmanager
 import logging
@@ -380,6 +381,13 @@ def stdouterr_redirected(to=None, comm=None, overwrite=False):
     # The DESI loggers.
     desi_loggers = desiutil.log._desiutil_log_root
 
+    def _rank_filename(base, rank):
+        """
+        Return standard name for output file of individual rank.
+        Rank can be a wildcard to generate a glob string.
+        """
+        return f"{base}-rank{rank}"
+
     def _redirect(out_to, err_to):
 
         # Flush the C-level buffers
@@ -424,6 +432,29 @@ def stdouterr_redirected(to=None, comm=None, overwrite=False):
             ch.setFormatter(formatter)
             logger.addHandler(ch)
 
+    def _combine_individual_outputs(to, nproc=None):
+        """
+        Combine individual {to}_{n} files into a single {to} file
+
+        If nproc is specified, assume there are exactly that many per-rank
+        files; otherwise glob to find out what is there.
+        """
+        if nproc is None:
+            individual_files = sorted(glob.glob(_rank_filename(to, '*')))
+        else:
+            individual_files = [_rank_filename(to, p) for p in range(nproc)]
+
+        with open(to, "w") as outfile:
+            for p, fname in enumerate(individual_files):
+                outfile.write("================ Start of Process {} ================\n".format(p))
+                with open(fname) as infile:
+                    outfile.write(infile.read())
+                outfile.write("================= End of Process {} =================\n\n".format(p))
+
+        # only remove input files after successfully finishing merging
+        for fname in individual_files:
+            os.remove(fname)
+
     # redirect both stdout and stderr to the same file
 
     if to is None:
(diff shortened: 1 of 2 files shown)


py/desispec/parallel.py
```diff
@@ -424,6 +432,29 @@ def _redirect(out_to, err_to):
...
+        """
+        if nproc is None:
+            individual_files = sorted(glob.glob(_rank_filename(to, '*')))
+        else:
+            individual_files = [_rank_filename(to, p) for p in range(nproc)]
+
+        with open(to, "w") as outfile:
+            for p, fname in enumerate(individual_files):
+                outfile.write("================ Start of Process {} ================\n".format(p))
+                with open(fname) as infile:
```
akremin, line 450: This already existed in the original implementation, but when `nproc` is specified should we check for the existence of the files before trying to open them? They _should_ all be there, so maybe crashing on a missing rank's log file would help point to a processing issue. I don't have strong feelings either way, but I noticed it so I wanted to flag it for discussion.

py/desispec/parallel.py
```diff
@@ -380,6 +381,13 @@ def stdouterr_redirected(to=None, comm=None, overwrite=False):
     # The DESI loggers.
     desi_loggers = desiutil.log._desiutil_log_root
 
+    def _rank_filename(base, rank):
```
akremin, line 384: A very minor thing, but could this be called basename or something more descriptive? When I read base and rank I immediately expected two integers (though I didn't know what base referred to), but then base turned out to be a filename. It is obvious from the code what it is, but still.

akremin (commented):
Thanks for this useful update to make the pipeline logging more robust. I have two very minor comments. Both should take two seconds to implement or they can both be ignored, as I don't feel strongly about either of them.

py/desispec/parallel.py
sbailey: Good suggestion.  I updated it to "basename" and expanded the docstring.

py/desispec/parallel.py
sbailey: Good question. A missing per-rank file would indicate a core logic bug or a basic I/O failure so I think crashing here would be the right thing to do, not just merging whatever it can find and moving on.  Pre-checking for the file and printing an error message would be more code for approximately the same result as just letting Python raise a FileNotFound error with the missing filename, so I'm going to leave this as-is.

Issue #2249, opened 2024-05-13
Deal with varying readnoise in SM6 b3.


julienguy:
SM6 b3 has had a peculiar noise since the beginning of the survey with the noise level jumping from 3.5 to 9 electrons for some rows.
We haven't done anything about it so far, but it would be an improvement to measure and record this increased readnoise for the affected CCD rows.

![Screenshot from 2024-05-10 14-04-03](https://github.com/desihub/desispec/assets/5192160/25689c6b-e7a9-4cf7-a02e-4f7c9dabbf36)

akremin:
Closed with PR #2314

PR #2314, opened 2024-08-09
Read noise per CCD row based on keyword READNOISEPERROW in config

julienguy:
This PR addresses issue #2249.

If the keyword READNOISEPERROW is present in the calibration configuration, and the value contain the amp being preprocessed, the value of the read noise per CCD row is used instead of the average if this value is a positive 4 sigma outlier.

Here are some plots illustrating what happens. The first figure is a profile of the read noise per CCD row of amps A and C for one zero of SM7 b6 (it's the same situation for most images of b6 before we replaced the FEE in 2024/05/23).  One can see the problematic amp C with excess read noise in some rows.

Figure_1

Now looking only at amp C. The new code records for the read noise the profile in orange (if the config contains 'READNOISEPERROW: C') instead of an average over all rows.

Figure_2a

Zoom

Figure_2b

And this is what it looks like as images. Left is the data (of a zero exposure), right is the 'READNOISE' HDU (portions of all four amps can be seen).

Screenshot from 2024-08-09 09-39-38

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -19,6 +19,7 @@ desispec Change Log
 * New FIBERSTATUS NEARCHARGETRAP and VARIABLETHRU set in
   desispec.io.fibermap.assemble_fibermap based on content
   of DESI_SPECTRO_CALIB yaml files (PR `#2313`_).
+* Use read noise estimated per CCD row for some amplifiers (PR `#2314`_).
 
 .. _`#2290`: https://github.com/desihub/desispec/pull/2290
 .. _`#2294`: https://github.com/desihub/desispec/pull/2294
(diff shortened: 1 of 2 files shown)


julienguy:
The calibration config file ```sm6-b.yaml``` has been modified and committed in DESI_SPECTRO_CALIB. It now contains the entry ```READNOISEPERROW: C``` for all configurations except for the most recent one when the FEE has been replaced.

akremin:
The code looks good here. Speaking with Julien offline I requested a new header keyword to document which method was used for the readnoise for a given amplifier. He has agreed to implement something like "RMETH" to parallel the overscan method "OMETH". Once done I think this will be ready to merge.

julienguy:
Added RMETHX  with X in A,B,C,D header keyword. For `preproc-b3-00228269.fits` it now looks like:
```
OMETHA  = 'AVERAGE '           / use average overscan                           
RMETHA  = 'AVERAGE '           / use average readnoise                          
OMETHB  = 'AVERAGE '           / use average overscan                           
RMETHB  = 'AVERAGE '           / use average readnoise                          
OMETHC  = 'AVERAGE '           / use average overscan                           
RMETHC  = 'PER_ROW '           / use average readnoise per row for some rows    
OMETHD  = 'AVERAGE '           / use average overscan                           
RMETHD  = 'AVERAGE '           / use average readnoise        
```

akremin:
Thank you, Julien the update looks good.

PR #2313, opened 2024-08-08
More FIBERSTATUS bits

julienguy:
This PR addresses issue #2304

- Add desispec.maskbits.fibermask  bits NEARCHARGETRAP ("Fiber trace near charge trap in one of the CCDs") and VARIABLETHRU ("Fiber has throughput variations we cannot model well")
- Add NEARCHARGETRAPFIBERS and VARIABLETHRUFIBERS to the list of keywords of calibfinder.badfibers
- Add code in desispec.io.fibermap.assemble_fibermap to set NEARCHARGETRAP and VARIABLETHRU fibermask bits by parsing the DESI_SPECTRO_CALIB configs of all cameras using the calibfinder.badfibers function and the raw fits file headers.

Example:
```
from desispec.io.fibermap import assemble_fibermap
from desispec.maskbits import fibermask

hdulist = assemble_fibermap(night=20220102, expid=116442)
# INFO:fibermap.py:1147:assemble_fibermap: Bad fibers from DESI_SPECTRO_CALIB: NEARCHARGETRAPFIBERS:[551 552 553]

fmap = hdulist["FIBERMAP"].data
print(fmap["FIBER"][(fmap["FIBERSTATUS"] & fibermask.NEARCHARGETRAP) > 0])
# [551 552 553]
```

To run this, you need an updated version of ```DESI_SPECTRO_CALIB``` svn and you need the environment variable ```FIBER_ASSIGN_DIR=/global/cfs/cdirs/desi/target/fiberassign/tiles/trunk```. It is set by the 'main' version of the DESI environment ```source /global/cfs/cdirs/desi/software/desi_environment.sh main``` but not the default one.
 
So far, DESI_SPECTRO_CALIB has been updated only for the problematic fibers 551,552,553 of SM10 r1. I will work on flagging more bad fibers in the yaml files (but we don't need to wait for those changes to the calibrations to merge this PR).

PR #2319, opened 2024-08-14
Fix unit tests

### LINKED ISSUE #2030: Many bootcalib tests are skipped because of missing data
While working on #2029, I discovered that `test_bootcalib.py` tries to download two data files:
```python
        # Grab the data
        url_arc = 'https://portal.nersc.gov/project/desi/data/spectest/pix-sub_b0-00000000.fits.gz'
        url_flat = 'https://portal.nersc.gov/project/desi/data/spectest/pix-sub_b0-00000001.fits.gz'
```
While `https://portal.nersc.gov/project/desi` still works, those files have been missing for some time, possibly years.

I scanned the "metadata" reports for those file names but could not find them, so we should assume they need to be regenerated.  Alternatively, we could use real arc and flat files.

### LINKED ISSUE #2315: desispec calib unit tests failing
desispec unit tests starting failing at NERSC last Friday night August 9 for reasons related to $DESI_SPECTRO_DARK and $DESI_SPECTRO_CALIB, which might be due to PRs #2313 (more FIBERSTATUS bits) or PR #2314 (readnoise per CCD row) which were merged around that time.  py/desispec/test/test_io_fibermap.py fails both when run individually and when run as part of the full test suite, albeit with different failures, which is usually a symptom of one test changing an environment variable but not resetting it, which then impacts a different test.

When run individually:
```
pytest py/desispec/test/test_io_fibermap.py
...
====================================================== FAILURES =======================================================
________________________________________ TestIOFibermap.test_assemble_fibermap ________________________________________

self = <desispec.test.test_io_fibermap.TestIOFibermap testMethod=test_assemble_fibermap>

    @unittest.skipUnless(standard_nersc_environment, "not at NERSC")
    def test_assemble_fibermap(self):
        """Test creation of fibermaps from raw inputs"""
        for night, expid in [
            (20200219, 51039),  #- old SPS header
            (20200315, 55611),  #- new SPEC header
            ]:
            print(f'Creating fibermap for {night}/{expid}')
>           fm = assemble_fibermap(night, expid)['FIBERMAP'].data

py/desispec/test/test_io_fibermap.py:260: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
py/desispec/io/fibermap.py:1121: in assemble_fibermap
    cfinder=CalibFinder([rawheader,camheader])
py/desispec/calibfinder.py:361: in __init__
    self.find_darks_in_desi_spectro_dark(header)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <desispec.calibfinder.CalibFinder object at 0x7f62b5525030>
header = {'ACQCAM': 'GUIDE0,GUIDE2,GUIDE3,GUIDE5,GUIDE7,GUIDE8', 'ACQTIME': 15.0, 'ADC1HOME': False, 'ADC1NREV': -2.0, ...}

    def find_darks_in_desi_spectro_dark(self, header):
...
        if found:
            self.data.update({"DARK": dark_filename,
                              "BIAS": bias_filename})
        else:
            if not self.fallback_on_dark_not_found:
                log.critical(f"Didn't find matching {camera} calibration darks in $DESI_SPECTRO_DARK, quitting")
>               raise IOError(f"Didn't find matching {camera} calibration darks in $DESI_SPECTRO_DARK, quitting")
E               OSError: Didn't find matching b0 calibration darks in $DESI_SPECTRO_DARK, quitting

py/desispec/calibfinder.py:560: OSError
```

When run as part of the full test suite
```
[login23 desispec] pytest py/desispec/test/
============================= test session starts ==============================
platform linux -- Python 3.10.14, pytest-8.1.1, pluggy-1.4.0
rootdir: /global/common/software/desi/users/sjbailey/desispec
plugins: remotedata-0.4.1, anyio-4.3.0, cov-5.0.0, astropy-0.11.0, arraydiff-0.6.1, astropy-header-0.2.2, doctestplus-1.2.1, asdf-3.2.0, mock-3.14.0, hypothesis-6.100.1, filter-subpackage-0.2.0
collected 305 items                                                            

py/desispec/test/test_binscripts.py ...FF                                [  1%]    [Editorial detail: this was due GPU memory on a shared login node]
py/desispec/test/test_bitmask.py .                                       [  1%]
py/desispec/test/test_bootcalib.py x.......                              [  4%]
py/desispec/test/test_calibfinder.py .                                   [  4%]
py/desispec/test/test_ccdcalib.py .                                      [  5%]
py/desispec/test/test_coadd.py ...................                       [ 11%]
py/desispec/test/test_cosmics.py ....                                    [ 12%]
py/desispec/test/test_database.py .....                                  [ 14%]
py/desispec/test/test_emlinefit.py s.                                    [ 15%]
py/desispec/test/test_extract.py .....                                   [ 16%]
py/desispec/test/test_fiberbitmask.py .                                  [ 17%]
py/desispec/test/test_fibercrosstalk.py .                                [ 17%]
py/desispec/test/test_fiberflat.py ..........                            [ 20%]
py/desispec/test/test_flux_calibration.py ........                       [ 23%]
py/desispec/test/test_frame.py ...                                       [ 24%]
py/desispec/test/test_gpu.py ...                                         [ 25%]
py/desispec/test/test_image.py ......                                    [ 27%]
py/desispec/test/test_io.py ............................................ [ 41%]
..                                                                       [ 42%]
py/desispec/test/test_io_fibermap.py .FF....FF.                          [ 45%]
...
____________________ TestIOFibermap.test_assemble_fibermap _____________________

self = <desispec.test.test_io_fibermap.TestIOFibermap testMethod=test_assemble_fibermap>

    @unittest.skipUnless(standard_nersc_environment, "not at NERSC")
    def test_assemble_fibermap(self):
        """Test creation of fibermaps from raw inputs"""
        for night, expid in [
            (20200219, 51039),  #- old SPS header
            (20200315, 55611),  #- new SPEC header
            ]:
            print(f'Creating fibermap for {night}/{expid}')
>           fm = assemble_fibermap(night, expid)['FIBERMAP'].data

py/desispec/test/test_io_fibermap.py:260: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
py/desispec/io/fibermap.py:1121: in assemble_fibermap
    cfinder=CalibFinder([rawheader,camheader])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <desispec.calibfinder.CalibFinder object at 0x7f92f3bc33a0>
headers = [XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / 8-b...um updated 2020-02-20T08:26:56       
DATASUM = '3006616571'         / data unit checksum updated 2020-02-20T08:26:56 ]
yaml_file = None, fallback_on_dark_not_found = False

    def __init__(self,headers,yaml_file=None, fallback_on_dark_not_found=False) :
...
            self.directory = os.environ["DESI_SPECTRO_CALIB"]
...
        if not os.path.isdir(self.directory):
>           raise IOError("Calibration directory {} not found".format(self.directory))
E           OSError: Calibration directory /global/homes/s/sjbailey/preproc_unit_test not found

py/desispec/calibfinder.py:271: OSError
```

Note that it appears that something set $DESI_SPECTRO_CALIB=$HOME/preproc_unit_test which is then mucking up the calibfinder as part of the fibermap tests.

This is likely a problem in the tests triggered by the PR updates, not the underlying PR algorithms, but we should clean it up so that tests pass again.

sbailey:
This PR
* fixes #2315 (calib-related unit tests failing)
* fixes #2030 (skipped bootcalib tests)
* does general cleanup of tests to use tempdirs instead of the current directory

For the newly failing unit tests (#2315), the problem was triggered by assemble_fibermap in PR #2313 needing CalibFinder to look up which fibers need FIBERSTATUS bits set, but
1. the unit tests were checking corner cases in very old data that didn't have a dark model in $DESI_SPECTRO_DARK, causing the CalibFinder creation to crash even though assemble_fibermap itself doesn't need a dark; this was fixed using `fallback_on_dark_not_found=True`.  I considered updating CalibFinder itself to only crash if a missing dark was requested (instead of crashing at initialization time whether or not the dark was needed), but that was going to be a bigger refactor and `fallback_on_dark_not_found=True` within assemble_fibermap worked fine.
2. test_calibfinder was changing $DESI_SPECTRO_CALIB but not resetting it when done, causing test_io_fibermap to fail due to a leftover bad $DESI_SPECTRO_CALIB value.  Now fixed.

For bootcalib #2030, the unit tests now download from https://data.desi.lbl.gov/public/epo/example_files/spectest instead of https://portal.nersc.gov/project/desi/data/spectest which no longer exists.  Full disclosure: previously the tests were run out of the desispec clone directory and the tests files were left behind in the top-level dir, and re-used if they already existed from a previous test run.  Now the test is using a temp directory every time, but that means it also has to download the files every time.  If the download fails it just skips the test (which is what is has been doing for over a year since the old URL disappeared anyway).  When running at NERSC, the download is a fraction of a second, but we should keep an eye on this if it becomes problematically slow from elsewhere.

In general, I cleaned up the other tests so that they don't write test files into the desispec git clone directory.  In addition to being good practice, this now enables tests to be run from a GPU compute node which mounts /global/common/software readonly.  To minimize changes to the tests themselves, the basic model I followed was:
  * setUpClass, run before any tests, caches `origdir=os.getcwd()`, creates a temporary directory `testdir=tempfile.mkdtemp()`, and move there with `os.chdir(testdir)`.
  * setUp, run at the start of every test, re-runs `os.chdir(self.testdir)` in case any previous test had moved out of that dir
  * tearDownClass, run at the end of all tests, removes any temporary files and goes back to the original `os.chdir(origdir)` so that the next test suite starts from a known location.

Some tests already had some usage of tempfile, e.g. to specify the full path of every file instead of `os.chdir` + writing files without a path.  I left those as-is.

Summarizing, this now works to run tests from a GPU compute node that has mounted a git clone directory in read-only mode:
```
salloc -N 1 -C gpu -A desi_g --gpus-per-node=4 -t 04:00:00 -q interactive
cd /global/common/software/desi/users/sjbailey/desispec
pytest
```

### DIFF
--- a/py/desispec/test/test_bootcalib.py
+++ b/py/desispec/test/test_bootcalib.py
@@ -5,6 +5,7 @@ tests bootcalib code
 import unittest
 from uuid import uuid1
 import tempfile
+import shutil
 import os
 import numpy as np
 import glob
@@ -51,18 +52,9 @@ class TestBoot(unittest.TestCase):
 
     @classmethod
     def tearDownClass(cls):
-        """We deliberately don't clean up the testarc and testflat files,
-        since they are useful for offline testing.
-        """
-        os.chdir(cls.testdir)
-        # if os.path.exists(cls.testarc):
-        #     os.unlink(cls.testarc)
-        # if os.path.exists(cls.testflat):
-        #     os.unlink(cls.testflat)
-        if os.path.exists(cls.testout):
-            os.unlink(cls.testout)
-        if os.path.isfile(cls.qafile):
-            os.unlink(cls.qafile)
+        #- Remove testdir only if it was created by tempfile.mkdtemp
+        if cls.testdir.startswith(tempfile.gettempdir()) and os.path.exists(cls.testdir):
+            shutil.rmtree(cls.testdir)
 
         os.chdir(cls.origdir)
 
--- a/py/desispec/test/test_bootcalib.py
+++ b/py/desispec/test/test_bootcalib.py
@@ -4,6 +4,7 @@ tests bootcalib code
 
 import unittest
 from uuid import uuid1
+import tempfile
 import os
 import numpy as np
 import glob
@@ -21,6 +22,9 @@ class TestBoot(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
+        cls.origdir = os.getcwd()
+        cls.testdir = tempfile.mkdtemp()
+        os.chdir(cls.testdir)
         cls.testarc = 'test_arc.fits.gz'
         cls.testflat = 'test_flat.fits.gz'
         cls.testout = 'test_bootcalib_{}.fits'.format(uuid1())
@@ -28,8 +32,8 @@ class TestBoot(unittest.TestCase):
         cls.data_unavailable = False
 
         # Grab the data
-        url_arc = 'https://portal.nersc.gov/project/desi/data/spectest/pix-sub_b0-00000000.fits.gz'
-        url_flat = 'https://portal.nersc.gov/project/desi/data/spectest/pix-sub_b0-00000001.fits.gz'
+        url_arc = 'https://data.desi.lbl.gov/public/epo/example_files/spectest/test_arc.fits.gz'
+        url_flat = 'https://data.desi.lbl.gov/public/epo/example_files/spectest/test_flat.fits.gz'
         for url, outfile in [(url_arc, cls.testarc), (url_flat, cls.testflat)]:
             if not os.path.exists(outfile):
                 try:
@@ -50,6 +54,7 @@ class TestBoot(unittest.TestCase):
         """We deliberately don't clean up the testarc and testflat files,
         since they are useful for offline testing.
         """
+        os.chdir(cls.testdir)
         # if os.path.exists(cls.testarc):
         #     os.unlink(cls.testarc)
         # if os.path.exists(cls.testflat):
@@ -58,6 +63,11 @@ class TestBoot(unittest.TestCase):
             os.unlink(cls.testout)
         if os.path.isfile(cls.qafile):
             os.unlink(cls.qafile)
+
+        os.chdir(cls.origdir)
+
+    def setUp(self):
+        os.chdir(self.testdir)
 
     def test_fiber_peaks(self):
         if self.data_unavailable:
--- a/py/desispec/test/test_calibfinder.py
+++ b/py/desispec/test/test_calibfinder.py
@@ -7,7 +7,7 @@ import unittest
 import os
 import shutil
 from importlib import resources
-
+import tempfile
 
 
 from desispec.calibfinder import CalibFinder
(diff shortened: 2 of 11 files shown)


py/desispec/test/test_bootcalib.py
```diff
@@ -21,15 +22,18 @@ class TestBoot(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
+        cls.origdir = os.getcwd()
+        cls.testdir = tempfile.mkdtemp()
```
weaverba137, line 26: I think we should also do `shutil.rmtree(cls.testdir)` in `tearDownClass`. `mkdtemp` doesn't clean up after itself, and it does not guarantee that the temporary directory is written to storage that disappears *e.g.* on reboot. In principle, enough tests could fill up a storage device.

py/desispec/test/test_calibfinder.py
```diff
@@ -7,40 +7,74 @@
...
+        os.makedirs(specdir)
         for c in "brz" :
             shutil.copy(str(resources.files('desispec').joinpath(f'test/data/ql/{c}0.yaml')), os.path.join(specdir,f"{c}0.yaml"))
-        #- Set calibration environment variable    
-        os.environ["DESI_SPECTRO_CALIB"] = self.calibdir
-        
+
+    @classmethod
+    def tearDownClass(cls):
+        #- remove temporary calibration directory
```
weaverba137, line 37: Here's a perfect example of running `shutil.rmtree()`. Just a reference to another comment.

weaverba137 (changes requested):
Besides the comment about cleaning up temporary directories, this looks fine.

py/desispec/test/test_bootcalib.py
sbailey: Thanks.  I agree with that in principle, but in practice I'm a bit nervous that someone in the future might do something like change the `testdir` to some place like $SCRATCH, without realizing that `tearDownClass` will recursively remove the entire directory at the end.  It would be safer to only remove files that are expected to have been generated by the tests, and then `os.rmdir(cls.testdir)` which will fail if the directory isn't empty.  But that could be a maintenance headache maintaining the list of files that each test could generate.  Some of our tests do a brute force `shutil.rmtree` of a tempdir, while others individually remove files (usually because they thought they were otherwise leaving them in the git clone).

I'm going to sit on this for the evening to think about safety vs. maintainability.  Followup thoughts/suggestion welcome.

weaverba137: That's a valid point. However you decide to do the clean-up is fine with me. It's more important to clean up than to clean up in a specific way.

weaverba137: PS, but it would be nice to have a _consistent_ clean-up method for all tests.

sbailey: I think the best solution is for `tearDownClass` to check that `cls.testdir` starts with `tempfile.gettempdir()` (the base directory used for creating temp subdirectories) before doing a `shutil.rmtree` on it.  That guarantees that the test dir actually is a temporary directory and not something like $SCRATCH.

sbailey:
I have updated the tests to remove their temporary directories using the pattern
```python
if cls.testdir.startswith(tempfile.gettempdir()) and os.path.exists(cls.testdir):
    shutil.rmtree(cls.testdir)
```
I think this is sufficient to ensure that we only remove directories that were created by `tempfile.mkdtemp` and won't accidentally remove $SCRATCH or $HOME etc.

@weaverba137 please re-review.

weaverba137 (approved):
Looks good to merge!

PR #2322, opened 2024-08-15
Modernize submit_prod

akremin:
## Overview
This PR modernizes the `submit_prod.py` script and renames the desi_run_prod to `desi_submit_prod` to make the names consistent. The code accepts a limited number of parameters via a yaml file. An example is provided in the tests section of this PR.

## Implementation Details
Most things should be specified via the yaml file, including the nights to process (either as a first and last night or as a list of nights), the name of the SPECPROD, the queue and reservation to submit to, surveys to submit, and what types of redshift jobs to submit. Other arguments to `desi_proc_night` are not currently exposed but could be in the future with small changes to the code to expect those parameters in the yaml file.

If given a first and last night the code loads all available exposure_tables and identifies all nights between the two dates (inclusive on both ends) that have valid science observations. The "THRUNIGHT" for tile completeness checking is defined as the last night unless specified in the yaml file. The first night can be left blank and defaults to 20201214, the first night of Fuji and Iron. Redshifts jobs default to submitting cumulative jobs. The code doesn't currently allow the user to request cumulatives on every night of observations (like we do in daily), it only allows cumulatives to be run on the last night of observation (as is done in productions). This could be allowed in the future with minimal changes to `submit_prod.py`.

Each night is submitted in chronological order. If a processing table already exists, that night is assumed to be complete and is skipped. If more than `queue_threshold` jobs are in the user's queue then the script stops submitting nights and exits. Because it skips over nights with processing tables this allows for repeated running of the script to submit all nights even with a queue limit. The default `queue_threshold` is 4800.

The logs for each night are redirected to $DESI_SPECTRO_REDUX/$SPECPROD/run/logs/<NIGHT>.log.

If a "sentinel" file exists in `$DESI_SPECTRO_REDUX/$SPECPROD/run/sentinel.txt`, the code doesn't check for processing tables and exists assuming the production is done. This is only written by the script if all nights have been submitted.

The selection of nights was factored out into a standalone function that can take the yaml file or a path to the yaml file. This is useful for other scripts to be able to derive the list of nights from the yaml file.

## Tests
Two test runs were performed, one of a few specified nights in:
`/global/cfs/cdirs/desi/users/kremin/PRs/submit_prod/`

The yaml file looks like:

`>cat example.yaml`
```
# EXAMPLE Test Prod


SPECPROD: 'submit_prod'
NIGHTS: [20231002, 20231203]
#FIRST_NIGHT: 20201214
#LAST_NIGHT: 20240409
THRU_NIGHT: 20240409
#Z_SUBMIT_TYPES: ['cumulative', 'pernight']
#SURVEYS: ['sv3', 'main']
QUEUE: 'regular'
#RESERVATION: 'K1res'
```
And a rerun of all of jura (in dry_run_level=2):
`/global/cfs/cdirs/desi/users/kremin/PRs/submit_jura/`

The yaml file looks like:

`> cat jura.yaml`
```
# Test Reproducibility of jura


SPECPROD: 'submit_jura'
FIRST_NIGHT: 20201214
LAST_NIGHT: 20240409
THRU_NIGHT: 20240409
Z_SUBMIT_TYPES: cumulative
QUEUE: 'regular'
```
note that THRU_NIGHT, Z_SUBMIT_TYPES, and QUEUE are not strictly necessary as they are set to defaults, but are given to be explicit.

Neither produced errors and the outputs look sensible. The tables in `submit_prod` are the same length as in jura. I will do a more in-depth check of these output tables and provide an update in this PR.


bin/desi_submit_prod
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+# coding: utf-8
```
sbailey, line 2: detail: I think the proper syntax for this is
```
# -*- coding: utf-8 -*-
```
but I'm also pretty sure that that is a leftover from python 2 times, and utf-8 is the default for python 3 so this isn't needed anymore.

py/desispec/workflow/queue.py
```diff
@@ -452,3 +453,132 @@ def any_jobs_failed(statuses, failed_states=None):
...
+
+    Returns
+    -------
+    Table
+        Table with the columns JOBID, PARTITION, NAME, USER, ST, TIME, NODES,
+        NODELIST(REASON) for the specified user.
+    """
+    log = get_logger()
+
+    cmd = f'squeue -u {user} -o "%i,%P,%j,%u,%t,%M,%D,%R"'
```
sbailey, line 482: let's intercept the default `user==None` case and set `user=os.environ['USER']`.

bin/desi_submit_prod
```diff
@@ -0,0 +1,38 @@
...
+
+def parse_args():  # options=None):
+    """
+    Creates an arguments parser for the desi run production
+    """
+    parser = argparse.ArgumentParser(description="Submit a full production run of the DESI data pipeline for processing.")
+
+    parser.add_argument("-p", "--production-yaml", type=str, required=True,
+                        help="Relative or absolute pathname to the yaml file summarizing the production.")
+    parser.add_argument("-q", "--queue-threshold", type=int, default=4800,
```
sbailey, line 17: Let's make this 4500 to give a little bit more overhead before the 5000 hard limit.

py/desispec/scripts/submit_prod.py
```diff
@@ -9,109 +9,302 @@
...
+                                                env_name='SPECPROD')
+    user = os.environ['USER']
+
+    ## Look for sentinal
+    sentinel_file = os.path.join(os.environ['DESI_SPECTRO_REDUX'],
+                                 os.environ['SPECPROD'], 'run', 'sentinel.txt')
+    if os.path.exists(sentinel_file):
+        log.info(f"Sentinel file {sentinel_file} exists, therefore all "
+                 + f"nights already submitted.")
+        return 0
```
sbailey, line 169: One use of this sentintel file is for submit_prod to tell itself that it is already done so that future scron invocations don't keep trying.  Is another potential use to temporarily turn off job submissions without having to edit the scrontab, and then it could resume after the file is removed?

Minor: let's give this a more descriptive name like "done.txt"

py/desispec/scripts/submit_prod.py
```diff
@@ -9,109 +9,302 @@
...
+    with open(production_yaml, 'rb') as yamlfile:
+        conf = yaml.safe_load(yamlfile)
+
+    ## Make sure the specprod matches, if not set it to that in the file
+    if 'SPECPROD' not in conf:
+        raise ValueError(f"SPECPROD required in yaml file {production_yaml}")
+    specprod = str(conf['SPECPROD']).lower()
+    specprod = verify_variable_with_environment(var=specprod, var_name='specprod',
+                                                env_name='SPECPROD')
+    user = os.environ['USER']
```
sbailey, line 161: I see here that you get `user=os.environ['USER']` to later pass to `check_queue_count(user=user, ...)`, but I think `check_queue_count` itself should also have that as the default if `user=None`.

py/desispec/scripts/submit_prod.py
```diff
@@ -9,109 +9,302 @@
...
-        raise IOError(f"Prod Yaml file doesn't exist: {production_yaml} not found. Exiting.")
-    conf = yaml.safe_load(open(production_yaml, 'rb'))
-    specprod = str(conf['name']).lower()
-    specprod = verify_variable_with_environment(var=specprod, var_name='specprod', env_name='SPECPROD')
-    if 'reservation' in conf:
-        reservation = str(conf['reservation'])
+        raise IOError(f"Prod yaml file doesn't exist: {production_yaml} not found.")
+    with open(production_yaml, 'rb') as yamlfile:
+        conf = yaml.safe_load(yamlfile)
+
```
sbailey, line 154: Somewhere near the beginning while getting setup, let's remove SLURM_MEM_PER_CPU and SLURM_OPEN_MODE from `os.environ` if they are set, since we know that those need to be unset for a scronjob to submit other jobs.  I think that would enable this script to be directly called by scron instead of requiring a wrapper bash script just to unset those.

sbailey (commented):
Overall looks good; I am trusting your real-world testing on this.  I had forgotten about the existence of desi_run_prod (now desi_submit_prod).  We didn't use that for Jura but let's use it for K1 and Kibo.

Minor comments inline; none are blocking factors.

bin/desi_submit_prod
akremin: This was added automatically by my IDE. It's present in several other scripts we use in the pipeline and doesn't seem to be harming anything, but I've replaced it with the syntax you recommend.

py/desispec/workflow/queue.py
akremin: Done

py/desispec/scripts/submit_prod.py
akremin: I had considered that use case, but I agree it could be used for that and it is a clever idea to avoid editing the scrontab. It's similar to the idea behind the yaml file also protecting us from having to change command line arguments mid-prod. 

I've renamed it `prod_submission_complete.txt`, which is very explicit.

py/desispec/scripts/submit_prod.py
akremin: Done

py/desispec/scripts/submit_prod.py
akremin: I've created a new function `desispec.workflow.utils.remove_slurm_environment_variables()` that removes these if present. I call that function in `submit_prod.py`. The purpose of a new function is that we can use it in other pipeline scripts in the future if we wish.

akremin:
I've rerun my very simple "example.yaml" 2-night prod in `dry-run-level=2` and it succeeded and wrote the new `prod_submission_complete.txt`. This is ready for re-review, but please don't merge until after the crossnight redshift branch as there is one change that needs to be made.

sbailey:
I addressed the TODO items in submit_prod.py for the structural changes from PR #2321, and re-tested using a version of this branch rebased with main.  The doc test failures are due to remaining structural issues in the non-rebased branch of this PR, but should be fine once merged with main.  After other tests pass I will merge this and then re-retest main.

PR #2341, opened 2024-08-23
Fix TSNR value when cframe mask or ivar=0 

julienguy:
This PR fixes 2 issues in TSNR calculation:
 - avoid situations where we apply a correction based on a ratio of 2 small numbers for fibers that are off target (and have a low TSNR anyway).
 - use the cframe.mask==0 and cframe.ivar>0 instead of frame.mask==0 and frame.ivar>0 in the computation of TSNR because some spectal values are masked during calibration.

### DIFF
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -210,12 +210,13 @@ def compute_tsnr_values(cframe_filename,cframe_hdulist,night,expid,camera,specpr
     sky = cframe_filename.replace('cframe', 'sky')
     psf = cframe_filename.replace('cframe', 'psf')
 
+    cframe=read_frame(cframe_filename, skip_resolution=True)
     frame=read_frame(iin, skip_resolution=True)
     fiberflat=read_fiberflat(flat)
     fluxcalib=read_flux_calibration(calib)
     skymodel=read_sky(sky)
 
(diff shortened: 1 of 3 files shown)


julienguy:
Validation example:
```
desi_process_exposure --infile /global/cfs/cdirs/desi/spectro/redux/jura/exposures/20231206/00208376/frame-r2-00208376.fits.gz --fiberflat /global/cfs/cdirs/desi/spectro/redux/jura/exposures/20231206/00208376/fiberflatexp-r2-00208376.fits.gz --sky /global/cfs/cdirs/desi/spectro/redux/jura/exposures/20231206/00208376/sky-r2-00208376.fits.gz --calib /global/cfs/cdirs/desi/spectro/redux/jura/exposures/20231206/00208376/fluxcalib-r2-00208376.fits.gz --outfile cframe-r2-00208376-new.fits.gz --cosmics-nsig 6
```
Two fibers that had large TSNR now have TSNR values=0 (because masked), the others are unchanged.

sbailey:
Thanks.  I double checked this case for main vs. this PR and confirm that 2 fibers that are very offset and completely masked now have TSNR2=0 instead of a crazy large value, and 4 other fibers have a very small decrease in TSNR2 which we interpret as being because cframes have slightly more masked pixels than frames (e.g. due to fiberflats or fluxcalib having masked pixels).

Issue #2342, opened 2024-08-25
bug in tileqa dependency tracking


sbailey:
There is a bug in the ztile job dependency tracking for the tileqa step, causing many ztile jobs in Kibo to exit early.  Example from kibo/tiles/cumulative/80616/20201223/logs/tile-qa-80616-thru20201223.log:
```
ERROR:util.py:90:runcmd: missing input /dvs_ro/cfs/cdirs/desi/spectro/redux/kibo/exposures/20201220/00068794/exposure-qa-00068794.fits
ERROR:util.py:90:runcmd: missing input /dvs_ro/cfs/cdirs/desi/spectro/redux/kibo/exposures/20201222/00068795/exposure-qa-00068795.fits
ERROR:util.py:90:runcmd: missing input /dvs_ro/cfs/cdirs/desi/spectro/redux/kibo/exposures/20201223/00068796/exposure-qa-00068796.fits
CRITICAL:util.py:101:runcmd: FAILED missing required inputs: desispec.scripts.tileqa.main(['-g', 'cumulative', '-n', '20201223', '-t', '80616'])
```

Those exposure IDs are from tile 80616, but are from night 20201219, not 20201220, 20201222, and 20201223.

For reference, this tileqa dependency checking was introduced in PR #2306, which included cross-night testing, and has been used for a month of daily operations with cross night tiles so it isn't immediately obvious what is wrong.  After fixing we'll need to cut a new tag and resubmit a bunch of ztile jobs.  In the meantime I'm letting Kibo proceed

Thanks for @dmargala for watching the kibo production closely and noticing the large number of ztile crashes.

akremin:
There is a flawed assumption in that PR that nights and exposure ids are index-matched lists. It turns out that they are not, so when we  do:
```
        for expid, night in zip(expids, nights):
            infiles.append(findfile('exposureqa', expid=expid, night=night, readonly=True))
```
on lines 614 - 615 in zproc.py, this produces the erroneous filepaths.

There are two ways to tackle this:

1. Update the code to open the exposure tables for just the nights specified and create the desired lists.
2. Drop this input check since we now have cross-night redshift job dependencies and we should therefore never be running zproc if the files haven't been generated by the tilenight job.

(1) is the most complete but requires a little bit of coding. (2) is simplest. I think we want belt-and-suspenders here, so (1) is the best option, but if it turns into too much surgery then (2) might be the pragmatic choice at that point.

PR #2344, opened 2024-08-26
Fix indexing bug in zproc expqa check

akremin:
#### Overview

This solves the issue identified in Issue #2342 that was introduced in PR #2306 .

zproc allows many modes to process healpix based redshifts and tile-based redshifts in per-exposure, per-night, cumulative, and custom varieties. Because of this the input structure and handling of information is complex and sometimes dependent on the type of job being requested. 

Originally the code looped over expids and night to get the list of expid-night combinations, but these are unique lists of the requested variables rather than index-matched lists. However, for pernight, cumulative, and custom tile-based jobs an exposure_table is populated from the set of nights and then subselected to the requested set of exposure ID's. So this table has the index-matched information we want to use. 

This PR updates the code to use the exposure_table for identifying exposure_qa files that are expected to exist. The exposure_table variable should exist for pernight and cumulative jobs, which are the only types that perform the tileqa step that we're updating here.

#### Test 
I ran a test of the failing job (tile=80616 thrunight=20201223) mentioned in the issue. I sym-linked the `calibnight`, `exposure_tables`, `processing_tables`, and `exposures` directories from kibo; copied the relevant script from kibo; and ran it using this branch. The job succeeded.

Location: `/global/cfs/cdirs/desi/spectro/redux/test_kibo`
zproc log:`/global/cfs/cdirs/desi/spectro/redux/test_kibo/ztile-80616-thru20201223-29816844.log`
tileqa log: `/global/cfs/cdirs/desi/spectro/redux/test_kibo/tiles/cumulative/80616/20201223/logs/tile-qa-80616-thru20201223.log`

### DIFF
--- a/py/desispec/scripts/zproc.py
+++ b/py/desispec/scripts/zproc.py
@@ -626,9 +626,12 @@ def main(args=None, comm=None):
         qapng = findfile('tileqapng', **findfileopts)
         qalog = findfile('tileqa', logfile=True, **findfileopts)
         ## requires all coadd and redrock outputs in addition to exposureqa
+        ## note that exposure_table is populated for both pernight and cumulative
+        ## so we can use it to loop over exposures
         infiles = []
-        for expid, night in zip(expids, nights):
-            infiles.append(findfile('exposureqa', expid=expid, night=night, readonly=True))
+        for erow in exposure_table:
+            infiles.append(findfile('exposureqa', expid=erow['EXPID'],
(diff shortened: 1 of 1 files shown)


sbailey (approved):
Nice! That was a much cleaner fix than I was expecting. Great that you found a way to update this without having to re-define what the `nights` variable means and thus potentially break some other piece of the code.

Issue #2347, opened 2024-08-27
desi_submit_prod rechecks running jobs hundreds of times


sbailey:
desi_submit_prod starts with log messages like:
```
INFO:queue.py:533:get_jobs_in_queue: Querying Slurm with the following: squeue -u desi -o "%i,%P,%j,%u,%t,%M,%D,%R"
INFO:submit_prod.py:257:submit_production: night=20201214 already has a proctable, skipping.
INFO:queue.py:533:get_jobs_in_queue: Querying Slurm with the following: squeue -u desi -o "%i,%P,%j,%u,%t,%M,%D,%R"
INFO:submit_prod.py:257:submit_production: night=20201215 already has a proctable, skipping.
INFO:queue.py:533:get_jobs_in_queue: Querying Slurm with the following: squeue -u desi -o "%i,%P,%j,%u,%t,%M,%D,%R"
INFO:submit_prod.py:257:submit_production: night=20201216 already has a proctable, skipping.
INFO:queue.py:533:get_jobs_in_queue: Querying Slurm with the following: squeue -u desi -o "%i,%P,%j,%u,%t,%M,%D,%R"
INFO:submit_prod.py:257:submit_production: night=20201217 already has a proctable, skipping.
...
```
I.e. it appears to be invoking squeue to check the currently running jobs for every night that it is skipping while trying to find the next time to submit.  This adds up to a lot of slurm load and delayed start of actual submission while trying to find a night that needs to be submitted.

Check if this squeue check could be run once, skip over N nights already submitted, and then proceed.

sbailey:
fixed by PR #2348; closing.

PR #2348, opened 2024-08-27
Minor Kibo workflow fixes

akremin:
This PR implements three small changes to the workflow.

First I improved the logging of the processing table caching, which was printing a line for every file it read it. It now prints one line with a list of files it will load in the most concise way possible. This just improves the logging but doesn't impact the downstream jobs at all.

Second I changed the order of operations in desi_submit_prod so that it now checks for the existence of a processing table first before checking the number of jobs in the queue. This allows us to skip hundreds of completed nights without needing to query the queue each time. This resolves Issue #2347 .

Third, I fixed a bud in the calibration submission that arose when cleaning up nights that crashed due to Slurm issues. When some calibrations are in the processing table the code correctly skipped over them but didn't include those jobs in the joint fitting of e.g. the `psfnight`. This is wrong, as we want all 5 arcs to be included in the `psfnight` job if they are good. The same situation was happening with flats. This now fixes that by using the existing rows in the processing table and passing the complete set of jobs to the `psfnight` and `nightlyflat` job submitter.

I tested this in `/global/cfs/cdirs/desi/spectro/redux/test_kibo` by sym-linking the `calibnight`, `exposures`, `preproc`, and `exposure_tables` directors; copying the incomplete processing table from 20211220 currently in Kibo; and running `desi_proc_night` out of my branch. It identifies the processing table entries and includes them in the joint fit (this currently fails in `main`). The output of the processing table caching is far fewer lines and works as expected.

Finally, I copied over the config file from Kibo and ran a test of `desi_submit_prod`, which now checks for processing table existence before checking the queue, and is therefore much faster in getting to nights that actually need to be submitted.

### DIFF
--- a/py/desispec/scripts/proc_night.py
+++ b/py/desispec/scripts/proc_night.py
@@ -756,6 +756,10 @@ def submit_calibrations(cal_etable, ptable, cal_override, calibjobs, int_id,
         arc_prows = []
         for arc_erow in arcs:
             if arc_erow['EXPID'] in processed_cal_expids:
+                matches = np.where(ptable['EXPID'] == arc_erow['EXPID'])[0]
+                if len(matches) == 1:
+                    prow = ptable[matches[0]]
+                    arc_prows.append(prow)
                 continue
             prow, int_id = make_exposure_prow(arc_erow, int_id, calibjobs)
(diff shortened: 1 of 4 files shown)


sbailey (approved):
Looks good, thanks

Issue #2358, opened 2024-09-06
Nightqa crashes because of "special" survey tile


akremin:
Nightqa crashes because of "special" survey tile in Kibo processing.

```
ERROR:tile_qa_plot.py:1114:get_tilecov: surveys=special not in allowed_surveys=sv1,sv2,sv3,main,catchall
Traceback (most recent call last):
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/bin/desi_night_qa", line 210, in <module>
    main()
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/bin/desi_night_qa", line 192, in main
    create_petalnz_pdf(outfns["petalnz"], args.night, args.prod, unq_tileids, unq_surveys, dchi2_threshold=25,
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/lib/python3.10/site-packages/desispec/night_qa.py", line 1796, in create_petalnz_pdf
    _, pix_ntilecovs, _, _, _ = get_tilecov(tileid, surveys=survey, programs=faprgrm.upper(), lastnight=night)
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/lib/python3.10/site-packages/desispec/tile_qa_plot.py", line 1115, in get_tilecov
    raise ValueError(msg)
ValueError: surveys=special not in allowed_surveys=sv1,sv2,sv3,main,catchall
```

Log:
`/global/cfs/cdirs/desi/spectro/redux/kibo/run/scripts/tsnr_nightqa/logs/nightqa-kibo-20210917.log`

Script:
`/global/cfs/cdirs/desi/spectro/redux/kibo/run/scripts/tsnr_nightqa/scripts/nightqa-kibo-20210917.slurm`

Command:
`desi_night_qa --nproc=64 --recompute -p kibo -n 20210917 -o /global/cfs/cdirs/desi/spectro/redux/kibo/nightqa/20210917`

araichoor:
ok, this one is a bug / typo in the code, the `create_petalnz_pdf()` function:

here:
https://github.com/desihub/desispec/blob/33d11ac38df5c7e9b18cefe3d28b1cd1278b6dac/py/desispec/night_qa.py#L1670-L1678

[...]

Issue #2357, opened 2024-09-06
nightqa crashes because it can't find a tile in surveyops tiles-*.ecsv file


akremin:
Nightqa is crashing because it assumes a tile should be in a `tiles*.ecsv` file but isn't. We should investigate whether the code or ecsv file is correct here. In any case, it may also help to have this be a non-fatal error (especially for an 80000 series tile from sv2).

```
ERROR:tile_qa_plot.py:1146:get_tilecov: no TILEID=81099 found in /global/cfs/cdirs/desi/survey/ops/surveyops/tags/2.0/ops/tiles-sv2.ecsv
Traceback (most recent call last):
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/bin/desi_night_qa", line 210, in <module>
    main()
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/bin/desi_night_qa", line 192, in main
    create_petalnz_pdf(outfns["petalnz"], args.night, args.prod, unq_tileids, unq_surveys, dchi2_threshold=25,
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/lib/python3.10/site-packages/desispec/night_qa.py", line 1796, in create_petalnz_pdf
    _, pix_ntilecovs, _, _, _ = get_tilecov(tileid, surveys=survey, programs=faprgrm.upper(), lastnight=night)
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/lib/python3.10/site-packages/desispec/tile_qa_plot.py", line 1147, in get_tilecov
    raise ValueError(msg)
ValueError: no TILEID=81099 found in /global/cfs/cdirs/desi/survey/ops/surveyops/tags/2.0/ops/tiles-sv2.ecsv
```

Log:
`/global/cfs/cdirs/desi/spectro/redux/kibo/run/scripts/tsnr_nightqa/logs/nightqa-kibo-20210511.log`

Script:
`/global/cfs/cdirs/desi/spectro/redux/kibo/run/scripts/tsnr_nightqa/scripts/nightqa-kibo-20210511.slurm`

Command:
`desi_night_qa --nproc=64 --recompute -p kibo -n 20210511 -o /global/cfs/cdirs/desi/spectro/redux/kibo/nightqa/20210511`

araichoor:
[...]

Issue #2356, opened 2024-09-06
nightqa fails in some early data when TILEID isn't defined in data header


akremin:
When running nightqa in Kibo we've hit an error when running on data from 2020 and early 2021.

Example:
`/global/cfs/cdirs/desi/spectro/redux/kibo/run/scripts/tsnr_nightqa/logs/nightqa-kibo-20201214.log`

```
Traceback (most recent call last):
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/bin/desi_night_qa", line 210, in <module>
    main()
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/bin/desi_night_qa", line 141, in main
    expids, tileids, surveys = get_surveys_night_expids(args.night)
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/0.66.2/lib/python3.10/site-packages/desispec/night_qa.py", line 136, in get_surveys_night_expids
    tileids.append(hdr["TILEID"])
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/conda/lib/python3.10/site-packages/fitsio/header.py", line 354, in __getitem__
    raise KeyError("unknown record: %s" % item)
KeyError: 'unknown record: TILEID'
```

akremin:
While this is being fixed there is also a warning that appears since it can't find the SURVEY either:
`WARNING:night_qa.py:133:get_surveys_night_expids: SURVEY could not be identified for /dvs_ro/cfs/cdirs/desi/spectro/data/20201214/00067797/desi-00067797.fits.fz; setting to 'unknown'`

The KeyError and warning happened on 7 nights in Kibo:
```
/global/cfs/cdirs/desi/spectro/redux/kibo/run/scripts/tsnr_nightqa/logs/nightqa-kibo-20201214.log
/global/cfs/cdirs/desi/spectro/redux/kibo/run/scripts/tsnr_nightqa/logs/nightqa-kibo-20201215.log
/global/cfs/cdirs/desi/spectro/redux/kibo/run/scripts/tsnr_nightqa/logs/nightqa-kibo-20210114.log
/global/cfs/cdirs/desi/spectro/redux/kibo/run/scripts/tsnr_nightqa/logs/nightqa-kibo-20210218.log
[...]

PR #2361, opened 2024-09-07
Nightqa: kibo fixes

araichoor:
This PR should address:
- https://github.com/desihub/desispec/issues/2356: ignore `SCIENCE` exposure with no associated fiberassign file (kind of a bugfix, actually...);
- https://github.com/desihub/desispec/issues/2357: ignore tileids for a given survey if it s not in `tiles-{survey}.ecsv`;
- https://github.com/desihub/desispec/issues/2358: bugfix (in `create_petal_nz()`) to restrict to `sv1`, `sv2`, `sv3`, `main` in the case where all the tileids should be ignored.

Outputs of the `desi_night_qa` runs for those three cases / nights are here: https://data.desi.lbl.gov/desi/users/raichoor/nightqa_dev/nightqa_v28/

For 20201214 and 20210917, the `petalnz-{night}.pdf` file is empty, as no tiles are kept for the analysis.

I didn t run `desi_night_qa` for other nights; hopefully the changes are simple enough that they don t require additional testing... though, I m happy to do so, if you d like me to.

ps: @akremin: sorry for all those issues in night_qa!

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -129,6 +129,9 @@ def get_surveys_night_expids(
                     survey = fahdr["SURVEY"]
                 else:
                     survey = fahdr["FA_SURV"]
+            else:
+                log.warning("no associated fiberassign file for {}; ignoring".format(fns[i]))
+                continue
             if survey == "unknown":
                 log.warning("SURVEY could not be identified for {}; setting to 'unknown'".format(fns[i]))
             # AR append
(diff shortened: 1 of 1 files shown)


akremin (approved):
Thank you, Anand. No need for an apology. These issues are all on very old data that your code was never designed to run on. I'm glad the fixes to make it robust were straightforward and not too invasive.

I have tested this branch on kibo for nights 20210223, 20210511, and 20210917; which previously failed but now succeed in this branch.

PR #2365, opened 2024-09-10
fix a logical, but likely harmless indexing bug with backup efftime

segasai:
While looking at the backup efftime calculations, I spotted this place where there is a check for len(indices) where indices is a boolean mask (seems to be an IDL'ism). 
I think the current code is not what the author meant, but probably still harmless. It's probably still better to fix.


sbailey:
Thanks.  I agree that this is a bug fix.  I think it has no impact on Kibo, but just in case let's wait until we've finished the final Kibo catalogs and tags so that this is merged post-Kibo.

Issue #2363, opened 2024-09-09
Kibo dropped TSNR2_LYA and TSNR2_QSO columns


sbailey:
Compared to Iron, kibo/zcatalog/v1/zpix-all-kibo.fits has dropped the columns TSNR2_LYA and TSNR2_QSO, though it retains the per-camera values TSNR2_LYA_[BRZ] and TSNR2_QSO_[BRZ].  Track down if this is a bug or a purposeful choice that we forgot about.

Thanks to @stephjuneau for reporting.

sbailey:
Update: Those columns are present in the ztile*.fits and zpix*.fits files, but are being dropped by `desispec.zcatalog.update_table_columns` when it is trying to standardize the order of the columns across input files.  It is incorrectly assuming that TSNR2_LRG is the last of the TSNR2 columns, which for some reason changed in Kibo compared to Iron.

sbailey:
Fixed by PR #2368

PR #2368, opened 2024-09-11
fix missing zall TSNR2 columns

sbailey:
This PR fixes the missing Kibo zall TSNR2_LYA and TSNR2_QSO columns reported in #2363 .  This is required for making the final kibo/zcatalogs/v1/zall*.fits files.

The underlying issue is that `desispec.zcatalog.update_table_columns` makes some assumptions about column order of the input files while trying to standardize the output column order.  The input order changed in Kibo (for reasons I haven't investigated) which broke those assumptions.

This is a minimal fix to get things to work for Kibo while still being backwards compatible with Iron etc.  This update is only slightly less fragile than the original code, but while wrapping up Kibo I'm also trying to make the minimal working change and save deeper fixes for later.

The change: instead of assuming that TNSR2_LRG is the last TSNR2 column, check for for the last column that starts with TSNR2.

Test outputs using this branch are in $CFS/desi/spectro/redux/kibo/zcatalog/v1/zall-*-v1a.fits .  These have the missing columns `TSNR2_QSO` and `TSNR2_LYA` at the intended location (after the other `TSNR2*` columns, and before the `*_NSPEC` and `*_PRIMARY` columns).  Due to the size of the files and Perlmutter being out today, I have not checked that every row of every other column is unchanged, but I did check the first million rows and they exactly match as expected.

### DIFF
--- a/py/desispec/zcatalog.py
+++ b/py/desispec/zcatalog.py
@@ -456,10 +456,10 @@ def update_table_columns(table, specgroup = 'zpix', all_columns = True, columns_
         ## Last column in TSNR2_LRG in all the redshift catalogs
         ## We will add the PRIMARY columns in the end
 
-        ## The indices of NUMOBS_INIT, PLATE_RA, and ZCAT_PRIMARY columns
+        ## The indices of NUMOBS_INIT, PLATE_RA, and last TSNR2_* columns
         nobs = np.where(np.array(tab.colnames) == 'NUMOBS_INIT')[0][0]
         pra = np.where(np.array(tab.colnames) == 'PLATE_RA')[0][0]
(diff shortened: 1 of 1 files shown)


sbailey:
I agree that this PR still has fragile assumptions, especially about the location of the `NUMOBS_INIT` and `PLATE_RA` columns relative to the `*_TARGET` and `TSNR2_*` columns.  I don't think it is worth further generalizing the TSNR2 columns without also addressing those issues.  I opened #2369 to track this after this PR is merged.

akremin:
Thanks. We will address further robustness improvements in that PR. Merging this as good enough for now.

Issue #2335, opened 2024-08-22
Recommended operation of `desi_tsnr_afterburner`


weaverba137:
`desi_tsnr_afterburner` creates the top-level `exposures-${SPECPROD}.(fits|csv)` files. In addition, it *can* create the top-level `tiles-${SPECPROD}.(fits|csv)` files, however, `desi_tiles_completeness` can *also* do that.

Going from `fuji`/`guadalupe` to `iron` to `jura`, this afterburner process has been run in different ways each time, and each time, effectively "by hand".  This should be standardized.

More detail on `jura`:

1. `desi_tsnr_afterburner` was run on each night in `jura` to create an exposure summary table for each night. This was done with `jura/run/scripts/afterburner/code/generate_slurm_files.sh` which used `tsnr_template_script.slurm` (same directory).
2. The per-night exposure tables were stacked with `combine_tables.py` (executed inside `create_merged_tables.sh`).
3. `create_merged_tables.sh` ran `desi_tiles_completeness` to create the tiles summary table.
4. The summary files were *created* in the run directory and then *copied* to the top-level specprod directory.

Options for `desi_tsnr_afterburner` in `jura`:
```
desi_tsnr_afterburner -o ${DESI_SPECTRO_REDUX}/${SPECPROD}/run/scripts/afterburner/outfiles/exposures-${SPECPROD}-{NIGHT}.fits --prod ${SPECPROD} \
                      --aux /global/cfs/cdirs/desi/survey/observations/SV1/sv1-tiles.fits \
                      --gfa-proc-dir /global/cfs/cdirs/desi/survey/GFA/ \
                      --compute-skymags --update --nproc 64 \
                      --nights {NIGHT} &>> ${DESI_SPECTRO_REDUX}/${SPECPROD}/run/scripts/afterburner/expburner-${SPECPROD}-{NIGHT}.log
```

Options for `desi_tiles_completeness` in `jura`:
```
desi_tiles_completeness -i /global/cfs/cdirs/desi/spectro/redux/jura/run/scripts/afterburner/outfiles/exposures-jura.fits -o /global/cfs/cdirs/desi/spectro/redux/jura/run/scripts/afterburner/outfiles/tiles-jura.fits --prod jura --aux /global/cfs/cdirs/desi/survey/observations/SV1/sv1-tiles.fits
[...]

PR #2373, opened 2024-09-17
Add UPDATED timestamp column

### LINKED ISSUE #2367: Filter or remove NaN warnings when reading GFA summary files
This issue replaces desihub/gfa_reduce#24.

When processing daily data, `desi_tsnr_afterburner` reads and replaces NaN in *all* exposures in *all* GFA summary files, with the result that rare NaNs trigger warnings for every night, even though the exposures with NaN are not associated with the night. For example:
```
tsnr-daily-20240824.log:WARNING:desi_tsnr_afterburner:1119:main: gfa_table column 'TRANSPARENCY' contains NaN or other non-finite values. Invalid values will be replaced with zero (0).
```
This has the potential to distract from more serious warnings.

This may require a refactor of `desi_tsnr_afterburner` along the lines of #1724. However it may be possible to fix the warnings without a full rewrite. Further investigation is needed.

weaverba137:
This PR:

* Updates `desispec.tilecompleteness.merge_tile_completeness_table()` to add an `UPDATED` column for database use.
* Closes #2367. Warnings now will only be printed if the corresponding exposure table entry is not already set to zero.
* When changes to GFA entries are detected, the night is recorded, which modifies the input to `compute_tile_completeness_table`. In one scenario, delayed GFA processing can result in updates to the tiles file, even if a night was not included in `--nights $NIGHT`.
* Adds some cosmetic and documentation changes to `desi_tsnr_afterburner`.

Open questions:

* There was an open question about how to return the output of `one_night()` to `main()` when running in MPI mode.  Is `desi_tsnr_afterburner` ever run in MPI mode?
* Why are `args.expids` and `args.nights` parsed more than once?

Some operational testing should be performed before merging this PR.

### DIFF
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -1273,14 +1273,12 @@ def main():
             if os.path.isfile(args.tile_completeness) :
                 previous_table = Table.read(args.tile_completeness)
                 new_tile_table = merge_tile_completeness_table(previous_table, new_tile_table)
-            new_tile_table.write(args.tile_completeness,overwrite=True)
-            log.info("wrote {}".format(args.tile_completeness))
 
-            if args.tile_completeness.endswith('.fits'):
-                head, ext = os.path.splitext(args.tile_completeness)
-                csvtiles = head + '.csv'
-                new_tile_table.write(csvtiles, overwrite=True)
-                log.info("wrote {}".format(csvtiles))
+            head, foo = os.path.splitext(args.tile_completeness)
+            write_formats = {'.fits': 'fits', '.csv': 'ascii.csv'}
+            for ext in write_formats:
+                new_tile_table.write(head + ext, format=write_formats[ext], overwrite=True)
+                log.info("Wrote %s.", head + ext)
 
             # compute_tile_completeness_table may have found new GOALTIME info
             # from auxiliary_table_filename, so update exposures if needed
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -2,7 +2,17 @@
 '''
 Calculate stand alone tsnr tables for a given desispec prod.
 
-Use MPI:
+Standard invocation for ``daily`` specprod on ``perlmutter``::
+
+    desi_tsnr_afterburner -o ${DESI_SPECTRO_REDUX}/${SPECPROD}/exposures-${SPECPROD}.fits --prod ${SPECPROD} \
+                          --tile-completeness ${DESI_SPECTRO_REDUX}/${SPECPROD}/tiles-${SPECPROD}.csv \
+                          --aux /global/cfs/cdirs/desi/survey/observations/SV1/sv1-tiles.fits \
+                          --gfa-proc-dir /global/cfs/cdirs/desi/survey/GFA/ \
+                          --compute-skymags --add-badexp --update --nproc 32 \
+                          --nights $NIGHT &>> ${LOGLOCATION}/tsnr-${SPECPROD}-${NIGHT}.log
+
+Use MPI::
+
     salloc -N 2 -C haswell -q interactive -t 00:30:00
 
     # 2 nodes, 2 mpi ranks reducing 2 nights (1 per rank) with 32 processes.
@@ -34,7 +44,7 @@ from   desispec.io.fluxcalibration import read_flux_calibration
 from   desispec.io.util import get_tempfilename, decode_camword, difference_camwords
 from   desiutil.log import get_logger
 from   desispec.tsnr import calc_tsnr2,tsnr2_to_efftime
-from   astropy.table import Table, vstack
+# from   astropy.table import Table, vstack
 from   desiutil.depend import getdep
 from   desispec.tilecompleteness import read_gfa_data, compute_tile_completeness_table, merge_tile_completeness_table
 from   desispec.skymag import compute_skymag
@@ -269,7 +279,7 @@ def compute_summary_tables(summary_rows, preexisting_tsnr2_expid_table,
     """ Compute summary tables.
 
     Args:
-      summary_rows: list of dictionnaries
+      summary_rows: list of dictionaries
       preexisting_tsnr2_expid_table: None or astropy.table.Table, update this table if not None
       preexisting_tsnr2_frame_table: None or astropy.table.Table, update this table if not None
       specprod_dir: str, production directory
@@ -614,7 +624,7 @@ def func(prod,night,expid,camera,recompute,alpha_only,details_dir) :
         alpha_only: bool, only recompute alpha
         details_dir: str or None, save details per frame in this directory if not None
 
-    Returns a dictionnary with NIGHT,EXPID,TILEID,EXPTIME,CAMERA, and the TSNR2 values for this camera
+    Returns a dictionary with NIGHT,EXPID,TILEID,EXPTIME,CAMERA, and the TSNR2 values for this camera
     """
     log = get_logger()
 
@@ -824,7 +834,7 @@ def main():
 
     if not args.outfile.endswith(".fits") :
         log.critical("Output filename '{}' is incorrect. It has to end with '.fits'.".format(args.outfile))
-        sys.exit(1)
+        return 1
 
     if args.mpi and use_mpi():
         from mpi4py import  MPI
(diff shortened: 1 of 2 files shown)


sbailey:
> There was an open question about how to return the output of one_night() to main() when running in MPI mode. Is desi_tsnr_afterburner ever run in MPI mode?

We have used desi_tsnr_afterburner in MPI mode in the past, but that didn't scale up to current production sizes due to limitations of how the MPI path was implemented (needing to recalculate things that aren't recalculated in the multiprocessing path).  For Kibo we used multiprocessing with one job per night, then combined them afterward.  Daily mode also uses multiprocessing, not MPI.  i.e. I don't think we are actively using the MPI option now (but as part of a future refactor, I would like to re-enable an MPI option with better scaling).

> Why are args.expids and args.nights parsed more than once?

Smells like a leftover from iterative code development.  i.e. I don't see a specific reason for that (e.g. a modification of the originally parsed list of nights/expids, requiring an re-parsing of the original args).

bin/desi_tsnr_afterburner
```diff
@@ -908,14 +942,30 @@ def main():
...
 
-
-
     # starting computing
     # one night at a time
 
     summary_rows  = list()
 
     def one_night(count, night, rows, tmpfilename):
+        """Prepare multiprocessing steps for analyzing one night.
```
sbailey, line 951: If I understand correctly, this function is more than just preparing steps, it is actually running the TSNR calculations for this night.  If that is correct, let's update this docstring description.

sbailey (commented):
Conceptually this seems fine, though I haven't done any operational testing of it myself.  @akremin could you help with that to make sure it works as expected for daily ops?

@weaverba137 does this require the patched exposures and tiles files to be in place with the pre-added UPDATED column before this can be used, or will it add that UPDATED column to the current files if we merge this before signing off on the patched files?  i.e. what is the order of operations for deployment (which might also be relevant for @akremin testing)?

bin/desi_tsnr_afterburner
weaverba137: OK, will do.

weaverba137:
> does this require the patched exposures and tiles files to be in place with the pre-added UPDATED column?

That would be my preference. It would increase code complexity if we had to assume `UPDATED` may or may not be present in both the previous data and the new tiles.

akremin (commented):
Thanks Ben, the changes look good. I agree with the above comments that we'll want to coordinate things such that we put the updated tables in place before we try to run this with the merged code.

One question I have is whether we also want to have this column in the exposures file? Maybe I misread the code, but it seems like the `UPDATED` is only added for the tiles completeness that feeds into the tiles file.

weaverba137:
When the `daily` database is updated, that update is tile-based, not exposure based, so we actually need `UPDATED` in `tiles-daily.csv`, not in `exposures-daily.(fits|csv)`.

That brings up another open question. As I mention in #2335, `desi_tsnr_afterburner` and its relatives should write both fits and csv. Is there a specific, operational reason why we don't write a `tiles-daily.fits` file?

akremin:
Okay, if we don't need it, that is fine. We can always add a column in the future if we decided we like having it. 

There is no major reason why we don't write a `tiles-daily.fits`. Some (myself included) prefer csv's because they are easily parsable with e.g. `grep` and `awk`. `exposures-daily.fits` was necessary because of the additional HDU's that are included in the fits version of the file that can't be included naturally in a csv. Since `tiles-daily.*` only contains a single table there was no reason why we needed an additional `fits` version. However, for the productions we do produce a fits version for symmetry with the exposures file, and I would be open to a similar argument in daily.

On a more practical side, the tiles code only writes one format at a time, so we have to run the code twice to produce the files for e.g. Kibo. However, I'm sure that would be an easy modification to the code to make it write both simultaneously.

bin/desi_tsnr_afterburner
weaverba137: On further review, technically all `one_night` does is prepare the inputs to `func`, which performs the actual calculation.

weaverba137:
> On a more practical side, the tiles code only writes one format at a time, so we have to run the code twice to produce the files for e.g. Kibo.

Right, I mentioned that in #2335, and yes it would be easy to fix

But since you mention it, there's actually a "feature" in `desi_tsnr_afterburner` such that if you pass the path to the tiles file with extension .csv, it will write the csv. However, if you pass it with extension .fits, it will write both.  I don't think that should be a "feature" and these should just always write both.

akremin:
I am fine with having the code always write both. I agree that it should either always write both or it should only write what is requested. The behavior or writing both under one circumstance but only one under another circumstance should be removed.

weaverba137:
I've modified the code so that both files are always written out.

weaverba137:
@sbailey @akremin, in preparation for (hopefully) merging this soon, I think we agreed to punt these open questions to a future refactor of `desi_tsnr_afterburner`:

> There was an open question about how to return the output of one_night() to main() when running in MPI mode. Is desi_tsnr_afterburner ever run in MPI mode?
> Why are args.expids and args.nights parsed more than once?

So I think the only thing left is to run operational tests after patching the daily tiles and exposures files.

akremin:
With the latest push the code now works and I see updates for the night I tested.

Test:
I copied the latest files in daily to a local directory, checked out this branch, and ran the following:
`desi_tsnr_afterburner -o exposures-daily_2.fits --prod daily --tile-completeness ./tiles-daily.csv --aux /global/cfs/cdirs/desi/survey/observations/SV1/sv1-tiles.fits --gfa-proc-dir /global/cfs/cdirs/desi/survey/GFA/ --compute-skymags --add-badexp --update --nproc 32 --nights 20240928` 

The updated table had the same "UPDATED" date for all nights except those on the night requested, which were updated to the time I ran it. 

Two entries from 20240928 in the new file and the pre-updated file. They are different time stamps, which is good:
```
kremin@perlmutter:login04:/global/cfs/cdirs/desi/users/kremin/workspace/add-updated-column> grep 20240928 tiles-daily.csv /global/cfs/cdirs/desi/spectro/redux/daily/tiles-daily.csv 
tiles-daily.csv:22258,main,bright,bright,mainbright,1,773.0,40.505,-9.963,29.2,35.5,0.0,180.0,obsstart,32.7,39.9,35.5,16.1,37.808,0.85,20240928,2024-10-07T14:35:55-0700
tiles-daily.csv:11839,main,dark,dark,maindark,2,2385.7,48.809,-13.555,1006.5,1053.4,0.0,1000.0,obsend,1053.4,1042.0,1081.7,1065.0,485.86,0.85,20240928,2024-10-07T14:35:55-0700
/global/cfs/cdirs/desi/spectro/redux/daily/tiles-daily.csv:9332,main,dark,dark,maindark,2,1976.3,344.928,-10.544,1006.5,1068.2,1093.5,1000.0,obsend,1068.2,1077.8,1099.4,1191.6,dark,0.85,20240928,2024-10-07T12:03:11-0700
/global/cfs/cdirs/desi/spectro/redux/daily/tiles-daily.csv:9345,main,dark,dark,maindark,2,2240.7,339.44,-12.966,1008.9,1058.9,1160.4,1000.0,obsend,1058.9,1071.8,1086.8,1098.6,dark,0.85,20240928,2024-10-07T12:03:11-0700
```

Two entries from 20240927 in the new file and the pre-updated file. They are the same, which is good:
```
tiles-daily.csv:27154,main,bright,bright,mainbright,2,2169.2,331.757,-13.178,181.9,245.3,284.7,180.0,obsend,228.3,219.4,245.3,149.8,bright,0.85,20240927,2024-10-07T12:03:11-0700
tiles-daily.csv:6733,main,dark,dark,maindark,1,1464.0,28.484,18.944,1008.3,1042.7,1181.1,1000.0,obsend,1042.7,1046.1,1043.0,1527.8,dark,0.85,20240927,2024-10-07T12:03:11-0700
/global/cfs/cdirs/desi/spectro/redux/daily/tiles-daily.csv:27154,main,bright,bright,mainbright,2,2169.2,331.757,-13.178,181.9,245.3,284.7,180.0,obsend,228.3,219.4,245.3,149.8,bright,0.85,20240927,2024-10-07T12:03:11-0700
/global/cfs/cdirs/desi/spectro/redux/daily/tiles-daily.csv:1565,main,dark,dark,maindark,2,2625.2,32.936,19.265,1002.5,1112.4,1239.9,1000.0,obsend,1112.4,1160.5,1095.3,969.7,dark,0.85,20240927,2024-10-07T12:03:11-0700
```

PR #2384, opened 2024-10-08
specstatus compatibility fix

weaverba137:
This PR removes the `UPDATED` column from a `tiles` table so it remains compatible with the `specstatus` table.


weaverba137:
In addition @sbailey found a bug in `desi_tsnr_afterburner` that was placing `EFFTIME_SPEC` values in the `GOALTYPE` column. That should be fixed now.

sbailey:
This fixes a bug in desi_tsnr_afterburner, and fixes a crash-inducing incompatibility with the new UPDATED column in the tiles-daily.csv file.  We're still checking the impact of the patched changes in tiles-daily.csv and how they propagate to tiles-specstatus.ecsv, but these changes at minimum are necessary so I will merge this now.

Issue #2113, opened 2023-09-11
Internal wavelength calibration errors


segasai:
After tackling the sky corrections, I wanted to take a look at the internal wavelength corrections done by trace_shifts.
Specifically so far I just reran the traceshifts on the representative sample of exposures with different conditions (that i used for previous patch)
For each run I saved the dwave corrections in different wavelength bins for each fiber that are computed by compute_dy_using_boxcar_extraction. 

I then saved the wavelength offsets for each wavelength bin and each fiber
Here's an good example of b9 with expid=126067.
![image](https://github.com/desihub/desispec/assets/109513/9bd227fa-a002-4d2f-a11a-054b940fbccb)
Each row of panel show different wavelength bin. The right panels show the ylim of -3,3 the left panels show the zoom in on -.2, .2

One problem that i see is shown here for example. The uncertainties assigned to shifts seem kind'a wrong most of the time
![image](https://github.com/desihub/desispec/assets/109513/bc1d19f4-4d62-40cd-a236-3133020c281f) because they are clearly too small.

There are some cases where the uncertainties seem right like here 

![image](https://github.com/desihub/desispec/assets/109513/bf2e0d8f-936d-48bb-9637-99ccf0d76fd3)

I  do see that the polynomial model that fits these offsets has some way of dealing with outliers, but I don't quite know if it behaves well if all the errors are vastly underestimated. Also I havent' checked if there are cases where the offsets for the whole frame are completely wrong (as they all are just uniformly scattered between -3 and 3 )

Just in case the images for all the frames I fitted are available here /desi/users/koposov/wavelength_fix/outpsf/

julienguy:
I must have added this to enhance the weight of sky lines, but I agree it's questionable and it can be revisited. Formally we should not add this term if the model spectrum we are comparing to is representative of the data but this is not the case here: the flux of the objects can be very different from the reference flux.
Changes to the weights could improve the measurements but have to be tested in various conditions (inc. the backup survey!).

segasai:
Thanks @julienguy  
I have previously thought that the internal wavelength calibration is harmless for backup because one can still use the continuum solar spectra for this cross-correlation, but I haven't realised that the internal calibration step is done with spectra that have objects in them as well.

Altogether, I think the blue corrections even in the dark time are highly problematic (other than in two wavelength bins with bright emission lines.

[...]

PR #2386, opened 2024-10-09
Trace fix improvements

### LINKED ISSUE #2380: Fixing 1 km/s radial velocity systematic
Hi, 

One of the long-standing issues we notice in MWS when looking at RVs wrt external data is the offset of ~ 1km/s that has been present for long time. (see i.e. figure
image
from 
MWS EDR VAC paper K+2024).

We have already done some improvements in RV accuracy with trace_shifts https://github.com/desihub/desispec/pull/2106 (but mostly for backup data), but I thought I'd look at it again. And I've looked at the MEANDY from trace_shifts across the whole kibo, and this is how it looks like 

image

There is a clear bias of 0.03 pixels. This corresponds to about 1km/s. 
The sign is also in the right direction for +1km/s offst see figures in the #2106  for the plots of meandy vs velocity offset.

From what I can see the most likely cause is the sky spectrum sky-spec.dat, rather than anything algorithmic (that's my hypothesis though, not the fact). So my proposal/plan is to try to  either 
1) substitute sky-spec.dat by similar synthesized spectrum bazed on Hanuschik+2003 atlas
2) update the algorithm compute_dy_from_spectral_cross_correlation() to something a bit more sophisticated  where a sky spectrum is not compared to a fixed template, but for each offset we on the fly solve the line amplitudes (given the line-list from Hanuschik and LSF).

In #2106  I briefly looked at Hanuschik, but then decided to deal with that later as it was tangential to the issue with backup data and large MEANDY offsets.

I don't currently have a patch, but have written a code somewhere for 2nd method.  I'll aim to try it on some subset of data to see if that gets rid of the offset and/or also improves the meandy accuracy.

This is just an issue to keep track of this problem, avoid forgetting it, as well as solicit any help/feedback.

segasai:
Hi, 

This is a PR WIP trying to address #2380 (and maybe also improve systematic rv floor) 

ATM I don't think it fixes anything much but currently does a couple of things I was hoping could help.
1) Use multiple PSFs at different wavelengths (rather than PSF at central wavelength)  when cross-correlating with external spectra to avoid possible biases vs wavelength
2)  in the same external spectrum x-correlation routine use PSFs from multiple fibers as well for the same reason
3) when doing 'internal wavelength' calibration, subtract the continuum in the same way we do for external calibration, to avoid issues if we have a bright stars and weak lines

I did also try to restructure the code a little bit/reduce duplication to make it easier to change. 

I would ideally like to merge this type of patch even if it cannot fully fix #2380 (when it is ready and assuming nothing gets worse)


py/desispec/trace_shifts.py
```diff
@@ -363,13 +363,18 @@ def compute_dy_from_spectral_cross_correlations_of_frame(flux, ivar, wave , xcoe
...
                 continue
+            block_wave = np.sum(ivar[fiber,ok]*flux[fiber,ok]*(flux[fiber,ok]>0)*wave[ok])/sw
 
-            dwave,err = compute_dy_from_spectral_cross_correlation(flux[fiber,ok],wave[ok],reference_flux[ok],ivar=ivar[fiber,ok]*reference_flux[ok],hw=3., calibrate=True)
+            dwave,err = compute_dy_from_spectral_cross_correlation(flux[fiber,ok], wave[ok],
+                                                                   reference_flux[ok],
+                                                                   ivar=ivar[fiber,ok],
+                                                                   hw=3., calibrate=True)
+            if fiber %10==0 :
+                log.info("Wavelength offset %f +/- %f for fiber #%03d at wave %f "%(dwave, err, fiber, block_wave))
```
sbailey, line 373: Minor style comment, but for the record so that it doesn't propagate to more code:

When formatting strings for logging, it is better to use the structure
```python
log.info("Wavelength offset %f +/- %f for fiber #%03d at wave %f", dwave, err, fiber, block_wave)
```
so that the string formatting is only evaluated by the logger if the log level would actually print it.  In practice we almost always have INFO-level on so this particular case doesn't make a performance difference, but the readability is nearly the same as old-syle %-formatting and we have had cases in the past of string evaluation of unprinted debug-level logging taking a significant amount of time, so this style is something to keep in mind when adding new log messages.

Personal opinion: I'm also fine with pre-evaluated new-style format strings for INFO-level and above if the author feels it helps with readability, e.g.
```python
log.info(f"Wavelength offset {dwave} +/- {err} for fiber #{fiber:03d} at wave {block_wave}")
```
My basic motivation is that readability trumps performance given that we use INFO-level for production running anyway so there isn't a performance impact.  This should *not* be used for DEBUG-level logging though, due to performance issues especially if it is deep in loops.

[actual review of algorithmic changes takes more thought/time...]

segasai:
I think to avoid feature creep and very large patch, I have decided to test what we have here. 
I ran full processing of several tiles/nights (same set I used for previous version of trace_shifts patch)
/global/homes/k/koposov/desi_koposov/wavelength_fix/bulk.sh

To my surprise this already lead to significant improvements. I looked at the xmatch of velocities ( from redrock-) to APOGEE.

image

And here, the MAD wrt to APOGEE changes from 1.3 km/s to 1.1 km/s.
(I checked for all other surveys there are still improvements, but less visible due to their bigger errors)
Also I checked and the MEANDY's scatter also improves (MAD goes from 0.018 to 0.015)

So I think that's good evidence of things improving and I think it would be good to commit this.
Summarizing the changes here
1) background subtraction when doing internal wavelength offsets (in the same way it's done when doing external wavelength corrections)
2) using correct variance when doing internal wavelength offsets ( see https://github.com/desihub/desispec/issues/2113 )
3) refactor by putting the background subtraction into separate function
4) When doing external wavelength correction use PSFs sampled across 20 points along the spectrum, and 20 fibers across the petal as opposed to one single wavelength point for one fiber
5) using a consistent and more reasonable weighting scheme for determined the 'mean' wavelength of the wavelength bin considered.
None of these changes fixes the #2380 

One thing that I feel makes these changes harder to analyse is that these internal/external wavelength offsets for each fiber and wavelength bins are not saved anywhere (other than printed in the log and saved on average in MEANDY). I think for future improvements, it'd be beneficial to save these kinds of offsets (i.e. either in psf- file or some other kind of output file). It'd be ~ 2500 numbers per single frame, so it's not that much. But maybe that should be dealt with separately.

segasai:
I've verified on the the test subset of data processed /global/cfs/cdirs/desi/spectro/redux/koposov/wavelength_test_bulk_new/
that repeated observations also show reduced scatter in velocities See figure for z-band arm measurements
image
Improvement is MAD change from .7 km/s to .6 km/s. 
I've done a separate analysis when using blue only arm there. MAD improves from 1 km/s to .9 km/s 
Also looking at the large offsets, the number of those is significantly reduced with new code (this is B arm)
image

julienguy (changes requested):
* I agree weighting with ivar x flux^2 x (flux>0) is better than ivar x flux x (flux>0)  to get the effective wavelength in a bin.

* Subtracting the continuum is a good idea to improve the precision on the cross-match (similarily to what was done to the match of the sky)

* I see the effort to sample the PSF across wavelength and fibers to convolve the external reference spectrum (original code was using only wavelength and fiber for the  PSF)

* I ran the code on a random dark time exposure and 3 cameras and found tiny shifts <0.01A as expected

* I have only one change request: you wrote a comment "The reason why we oversample is unclear to me".
It is there because the wavelength arrays with the boxcar extraction are not aligned and for this reason I thought oversampling before stacking would help. Given the careful study you have done, maybe you have seen that this does not help,  in which case please change the code, otherwise, you may remove the comment.

I think we can merge after that. Thank you for all the work.

segasai:
Thanks for looking over the patch @julienguy . On the resampling, My comment was just a thought (it wasn't obvious to me there would be benefit from oversampling), but I didn't verify that. So I've removed that comment. 
I think the only other comment in the patch that's worth dealing with, concerns this line of code 
https://github.com/desihub/desispec/blob/fe88bbdcaa67093c31a77459eb9dbb767adacb2a/py/desispec/trace_shifts.py#L725
(it's now line 688 in my patch). I did not fully follow the reason behind the y[-1]-y[0] in this original code

so my code is just a more straightforward 2d grid in x,y, but maybe I missed the reason for that.

julienguy:
I checked the new changes and they are fine with me. I also do not really understand my intent with the term (y[-1]-y[0])/(2*hw+1) which must always be very close to one. So I agree with your simplification. I am merging this. Thanks for all the work on this.

julienguy (approved):

julienguy:
This is a problem with the the documentation with the error:
``` /home/runner/work/desispec/desispec/doc/changes.rst:16: ERROR: Unknown target name: "#2409"```
Can @weaverba137 help fix that so that we can merge?

sbailey:
The documentation error was mine, while making an earlier tag, and it has been corrected on main.  We can merge this PR without fixing it in this branch.

PR #2391, opened 2024-10-11
enable zproc.main to be called from other scripts

sbailey:
This PR provides two changes to `desispec.scripts.zproc.main` to enable it to be called from other scripts:
* Return an integer error code instead of calling `sys.exit` directly.  Previously even upon success, it would call `sys.exit(0)` instead of returning to the calling function.  The command line script `bin/desi_zproc` already includes `sys.exit(zproc.main())` so that did not need to be changed.
* Generate the command line from input `args` or `sys.argv`.  This is needed for the `--batch` option, which previously assumed the command line was coming only from `sys.argv` which isn't the case if a wrapper script is calling `zproc.main` with a list of strings.
  * Caveat: if the wrapper script calls `zproc.main` with an `args.Namespace` object, this will still incorrectly generate the command line from `sys.argv`.  That's the same as what the current code does, i.e. this PR doesn't solve all cases, but it at least provides a workaround via supporting passing a list of strings to main.

This was motivated by Loa testing wanting to batch submit an individual tile/nights without having to call the command line script and re-import packages etc every time (slow).  An alternative would have been to use `desi_tile_redshifts` and `desi_healpix_redshifts` instead, but I was mostly done with this update before realizing that from talking with @akremin, and I think these are useful functionality updates anyway.

### DIFF
--- a/py/desispec/scripts/zproc.py
+++ b/py/desispec/scripts/zproc.py
@@ -885,10 +885,7 @@ def main(args=None, comm=None):
         else:
             log.info(goodbye)
 
-    if error_count > 0:
-        return int(error_count)
-    else:
-        return 0
+    return int(error_count)
 
 
 def distribute_ranks_to_blocks(nblocks, rank=None, size=None, comm=None,
--- a/py/desispec/scripts/zproc.py
+++ b/py/desispec/scripts/zproc.py
@@ -135,6 +135,22 @@ def parse(options=None):
 
 
 def main(args=None, comm=None):
+    """
+    Run spectral grouping, coaddition, redshifts, and afterburners
+
+    Options:
+        args (argparse.Namespace or list of str): options to use instead of sys.argv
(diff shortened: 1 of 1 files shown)


py/desispec/scripts/zproc.py
```diff
@@ -870,7 +886,7 @@ def main(args=None, comm=None):
             log.info(goodbye)
 
     if error_count > 0:
```
akremin, line 888: since we're returning in both instances, we can remove the if statement and just `return int(error_count)`

akremin (commented):
Thanks, as stated there are other existing ways to create N>1 `desi_zproc` batch scripts, but it's useful to be able to do it directly from `desi_zproc --batch`. 

I have one comment inline to simplify an exit.

akremin:
Thanks for the quick turnaround. Once the checks are finished, I'll merge.

Issue #2379, opened 2024-10-01
improve import time by postponing optional imports


sbailey:
@dmargala identified that our occasional PMI init failures are related to large variations in the startup time of different MPI ranks before they connect to MPI.  These were traced to variations in module-level import times, which further identified several surprisingly expensive steps.  Some of these could be deferred into importing optional libraries only if needed by specific functions.

To explore further:

desispec.io.spectra -> desispec.spectra -> specutils dominates the import time, but is only needed for Spectra.to_specutils and Spectra.from_specutils.  Part of the specutils expense is importing astropy.nddata which imports dask stuff, so the astropy.nddata imports would also have to be postponed.

desispec.pixgroup -> import healpy is expensive and appears to be unneeded.

desispec.tsnr -> astropy.convolution is expensive again because it touches astropy.nddata -> dask.  This import could be deferred and/or replaced with scipy.signal.fftconvolve.

etc.

`python -Ximporttime -c "import desispec.spectra"` gives a dizzying amount of info.  @dmargala has visualization wrappers on that to simplify the output.  Sharing that and/or a hackfest with him to refactor some of these imports could be useful.

moustakas:
For the record, I've noted the desispec import overhead when benchmarking fastspecfit (which has `desispec` as one of its core dependencies) as well, so I appreciate the type of cleanup associated with this ticket.

The other place where a large number of (potentially unnecessary) imports occur is in, e.g., 
https://github.com/desihub/desispec/blob/main/py/desispec/io/__init__.py, and potentially other `__init__.py` files. 

E.g., if all I want is `desispec.io.spectra.read_spectra`, a bunch of stuff related to data reduction are imported along for the ride.

weaverba137:
I would not object to deferring the import of `specutils` unless and until the mentioned methods are used.

sbailey:
Use with caution, but...

While chasing a different issue, @segasai noticed https://peps.python.org/pep-0562/ that we can define module-level `__getattr__` to implement lazy (just-in-time) loading of submodules.  This might be useful for `desispec.io` where for convenience we import submodule functions like `read_frame`, `read_spectra`, `read_image`, `read_raw` into the top-level `desispec.io` namespace, but that currently has the side effect of loading all of them and their sub-dependencies even if you only need one of them.  If implemented, include lots of comments about what is really going on, since @segasai discovered this while we were scratching our heads about odd side effects from scipy doing this.

sbailey:
[...]

PR #2394, opened 2024-10-15
improve top-level import speed with deferred imports

sbailey:
This PR addresses several import speed issues identified in #2379, especially for `desispec.io`.

`desispec.io`: deferring imports of specutils, speclite.filters, and desispec.preproc improves "import desispec.io" from 4.1 seconds to 1.3 seconds (best of 5, pwd on scratch).  The remaining time is dominated by importing astropy.table (1.0 sec), which would be messy to defer everywhere.

`desispec.tsnr`: deferring import astropy.convolution improves "import desispec.tsnr" from 2.9 to 2.3 sec. Remaining big imports are astropy.table and astropy.coordinates via desiutil.dust.

`desispec.pixgroup`: removing unused import healpix improves "import desispec.pixgroup" from 3.7 to 3.15 sec.  Remaining big imports are desispec.tsnr (could be deferred, though this is a bit messy since TSNR is re-calculated on-the-fly if columns are missing from older data)

### DIFF
--- a/py/desispec/io/filters.py
+++ b/py/desispec/io/filters.py
@@ -4,7 +4,9 @@ desispec.io.filters
 
 """
 import numpy as np
-import speclite.filters
+
+#- defer import of speclite.filters for faster desispec.io import
+### import speclite.filters
 
 def load_filter(given_filter):
     """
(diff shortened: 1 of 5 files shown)


akremin (approved):
Thanks, all these changes look good to me and I checked that there are no missing imports in any functions.

sbailey:
For the record: I also ran a mini-prod with this branch in /global/cfs/cdirs/desi/users/sjbailey/spectro/redux/mini-pr2394 with jobs ccdcalib, arc, psfnight, flat, nightlyflat, tilenight, cumulative ztile, zpix, plus desi_tsnr_afterburner and desi_zcatalog.  I think that covers everything touched by the deferred imports in this PR, e.g. to make sure that I didn't defer too much.

PR #2395, opened 2024-10-22
get rid of the ebv factor in the backup efftime calculation in one more code location

segasai:
Address the last point in the #2370 
by removing the E(B-V) factor for backup survey in efftime.py


akremin:
Thank you for the quick fix. I'll merge this one-line change so that we can get the three EFFTIME's back in sync.

PR #2396, opened 2024-10-23
v2 zcatalog

rongpu:
This PR revamps the zcatalogs to 1) add columns needed for selecting confident redshifts and add the redshift confidence flags and 2) reduce the file sizes by splitting a single catalog into row-by-row matched catalogs. This PR also updates validredshifts.py which is now required by zcatalog.

Summary:
- The core catalog (e.g., zpix-main-dark.fits) now only contains the frequently used columns. The less frequently used columns are stored in two row-by-row matched FITS catalogs, *-imaging.fits that stores the imaging-related columns and *-extra.fits that stores the rest of the columns (e.g., zpix-main-dark-imaging.fits zpix-main-dark-extra.fits).
- Add the GOOD_SPEC column to indicate good hardware status. It is defined as: ((COADD_FIBERSTATUS==0) OR (COADD_FIBERSTATUS==2**3)) AND (OBJTYPE=='TGT'). The "TGT" requirement means that sky fibers will always have GOOD_SPEC==False.
- Add the Z_QSO column (and ZERR_QSO), which is the recommended redshift for quasars. If an object meets both the criteria for QuasarNET afterburner rerun (i.e., rerunning redrock using on QSO templates and QuasarNET prior and returns "Z_NEW") and the criteria for Z_NEW adoption, Z_QSO takes the Z_NEW value, otherwise it takes the original redrock Z value. Also add the boolean GOOD_Z_QSO flag that indicates the redshift confidence of Z_QSO, and it is True if 1) it is a QSO target AND 2) it passes the QSO redshift quality cut AND 3) GOOD_SPEC==True.
- Add the boolean flags GOOD_Z_BGS, GOOD_Z_LRG and GOOD_Z_ELG that indicate the redshift confidence of Z for the three tracers. GOOD_Z_BGS is True if 1) it is a BGS target AND 2) it passes the BGS redshift quality cut AND 3) GOOD_SPEC==True. Similarly for the LRG and ELG flags.
- Add a merged redshift confidence flag Z_CONF for Z. It is the recommended redshift quality flag to use for non-quasar objects. It is an integer that can take one of three values:
2: high confidence; definition: is a primary galaxy (BGS, LRG or ELG) target, AND passes the corresponding redshift quality cut, AND has GOOD_SPEC==True.
1: low confidence; definition: fails the criteria for Z_CONF=2, AND has GOOD_SPEC==True AND ZWARN==0
0: no confidence (fails the criteria for both 1 and 2).

The relevant email thread is [desi-data 6673].

### DIFF
--- a/py/desispec/scripts/zcatalog.py
+++ b/py/desispec/scripts/zcatalog.py
@@ -451,12 +451,11 @@ def main(args=None):
     good_spec = validredshifts.get_good_fiberstatus(zcat)
     good_spec &= zcat['OBJTYPE']=='TGT'
     zqual['GOOD_SPEC'] = good_spec.copy()  # GOOD_SPEC: true if it is a science spectrum with good hardware status
-    zqual['GOOD_Z'] = np.int16(-1)  # GOOD_Z: -1 if redshift quality is unknown
+    zqual['Z_CONF'] = np.uint8(0)  # Z_CONF: 0 if no confidence
     for col in ['GOOD_Z_BGS', 'GOOD_Z_LRG', 'GOOD_Z_ELG', 'GOOD_Z_QSO']:
         zqual[col] = zqual[col] & zqual['GOOD_SPEC']  # require good hardware quality for GOOD_Z_TRACER
-        zqual[col] = zqual[col].astype(np.int16)  # convert boolean to integer
 
-    # Add GOOD_Z columns
+    # evaluate Z_CONF
     if survey in ['main', 'sv1', 'sv2', 'sv3']:
         if survey=='main':
             desi_target_col = 'DESI_TARGET'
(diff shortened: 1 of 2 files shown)


akremin:
I don't have a solution (or even a strong opinion) but I thought I should write this down as an item for discussion before I forgot the offline discussion we had:

Here you're adding 'GOOD_Z_QSO', 'Z_QSO', and 'ZERR_QSO'; but your definition of Z_CONF has no "level 2" criteria for QSO. I think many people will be confused why something called 'GOOD_Z_QSO' is not 'Z_CONF'==2. The answer, I believe, is that you're distinguishing between 'Z' and 'Z_QSO'. The question is whether we should be doing that or not. 

Offline we had discussed moving 'Z' to 'Z_RR' and creating a derived redshift column that merges 'Z' and 'Z_QSO'. If we do that, which we probably should, then we should also set 'GOOD_Z_QSO'==True as 'Z_CONF'==2.

Whether we want to do this or leave quasars as a separate set of columns is up for discussion, so let's discuss.

rongpu:
@akremin I can think of several reasons why we cannot and should not merge Z and Z_QSO, at least for the v2 zcatalog (in no particular order):
1. We use both Z (for galaxies) and Z_QSO (for QSOs) for the clustering measurements; we never merge them in the LSS catalog.
2. Someone needs to do a serious study on how to actual merge them, and they would first need to fix the ~150 km/s offset in Z_QSO-Z. I don't see that happening any time soon, and we should not wait until then to release this zcatalog.
3. SUBTYPE, DELTACHI2 and ZWARN are all missing from the QN afterburner file. This means that for objects that adopt the Z_QSO value, these three columns will have missing/wrong values. Fixing that would require fixing and rerunning the QN afterburner to produce these necessary columns. It's also unclear how to interpret the QN DELTACHI2 value because of its narrow redshift prior.

The proposed scheme would tell users to use Z for galaxies and Z_QSO for QSOs. It's certainly not perfect but I think acceptable and will have to suffice until we have a permanent solution to the dual redshifts.

rongpu:
A note for future reference: running the code for Y3 uses a maximum of ~50% of Perlmutter's 500 GB memory. So it will probably exceed the memory limit when our data volume doubles. There is probably be some easy tricks we can do to free up memory in the final steps of catalog stacking, splitting and numpy array conversion.

rongpu:
@sbailey Here is the latest zcatalog for Kibo:
`/global/cfs/cdirs/desicollab/users/rongpu/tmp/zcatalog/v0.3`
and the Loa version is here:
`/global/cfs/cdirs/desicollab/users/rongpu/tmp/zcatalog/v0.3_loa`
Note that the `ztile-main-bright-cumulative*` files are currently missing from the Loa zcatalogs because some emline files are missing (see https://github.com/desihub/desispec/issues/2434).

rongpu:
In `v0.3_loa/zpix-main-dark.fits`, there are 176512 objects with good redshift quality in both redrock Z and Z_QSO (`Z_CONF==2 & GOOD_Z_QSO==True`). See comparison below. Out of those, 3696 objects with redshift discrepancy larger than 1000 km/s (orange points below), and 3608 objects with >5000 km/s discrepancy.

image

weaverba137:
I'm sorry if this has already been discussed. I agree that row-by-row matching is good in this case, but is there enough information in each file to reconstruct the unique ID for each object?  For healpix-based redshift, that means every file has (targetid, survey, program) and for tile-based redshift, every file has (targetid, spgrp, spgrpval, tileid), where spgrp is effectively always 'cumulative' and spgrpval should be the same as lastnight.

rongpu:
@weaverba137 All tile-based catalog files include `TARGETID` and `TILEID`. All healpix-based catalog files include `TARGETID`. That should enable the construction of a unique ID in all cases, except for the healpix zall catalogs where there are duplicates in `TARGETID`. I'm not sure why you would need to construct unique ID's for the row-matched catalogs, but I can add `SURVEY` and `PROGRAM` to all healpix catalogs that would guarantee uniqueness if you think it's necessary.

weaverba137:
The reason is that even with row matching, in principle I need to be able to construct a unique ID to load a database. Row ordering doesn't matter there, only being able to find the unique ID.

weaverba137:
> All tile-based catalog files include TARGETID and TILEID. 

But not lastnight or spgrpval?

rongpu:
I'm pretty sure `TARGETID` and `TILEID` specifies a unique object+observation.

I have never used `spgrpval` (in the current scheme it will be in the -extra.fits file).

rongpu:
> The reason is that even with row matching, in principle I need to be able to construct a unique ID to load a database. Row ordering doesn't matter there, only being able to find the unique ID.

Dumb question: is it possible to just use the row number as the unique ID in the database?

weaverba137:
> I think everything you need for the unique DB ID is in the primary catalog files, but since the DB will nominally be loading all columns, this means reading all 3 flavors of files anyway and thus the full information is available.

Based on the email sent to desi-data today, no that is not the case at all. In fact, I think the situation is worse than I thought. We should maybe have a 1:1 Zoom call to talk about this.

stephjuneau:
For the record, QSO-maker cuts start[ here in the code](https://github.com/desihub/LSS/blob/1280e41cd1f02571799b48d5fe9928b321531e84/py/LSS/qso_cat_utils.py#L370).

Namely, here are the QN threshold definitions (to decide if it's confident): 
```
# QN afterburner is run with a threshold 0.5. With VI, we choose 0.95 as final threshold. Note the new version of QN from Jura imporve the older one. Increase the confidence level to 0.99.
    log.info('Increase the QN threshold selection from 0.5 to 0.99.')
    qn['IS_QSO_QN_099'] = np.max(np.array([qn[name] for name in ['C_LYA', 'C_CIV', 'C_CIII', 'C_MgII', 'C_Hbeta', 'C_Halpha']]), axis=0) > 0.99
    qn['IS_QSO_QN_06'] = np.max(np.array([qn[name] for name in ['C_LYA', 'C_CIV', 'C_CIII', 'C_MgII', 'C_Hbeta', 'C_Halpha']]), axis=0) > 0.6
```
Criteria for BGS (applied to BGS targets that are *not* also ELG or QSO):
```
    is_OK_for_BGS = (QSO_cat['SPECTYPE'] == 'QSO') & (QSO_cat['IS_QSO_QN_06'] | QSO_cat['IS_QSO_MGII'])
```
Criteria for ELG (applied to ELG targets that are *not* also a QSO target):
```
    # selection for ELG targets:
    is_OK_for_ELG = (QSO_cat['SPECTYPE'] == 'QSO') & QSO_cat['IS_QSO_QN_06']
```
Same criteria for QSO targets and for WISE_VAR_QSO targets (irrespective of whether they're targeted as anything else):
```
    # selection for WISE_VAR_QSO targets (secondary target) --> same as QSO
    is_OK_for_VAR_WISE_QSO = (QSO_cat['SPECTYPE'] == 'QSO') | QSO_cat['IS_QSO_MGII'] | QSO_cat['IS_QSO_QN_099']
```
And there are not criteria for LRG, even though I think with the latest improved QN and QSO templates, there is now better hope than in the early days when @reac2  & I were studying the QN thresholds per target type.

ashleyjross:
Please don't split the tile-based "extra" catalog any further (we would much of the current extra catalog in the LSS data collection).

PR #2400, opened 2024-10-27
fix few obvious issues with undefined variables reported by static checker

segasai:
While running pyright checker on desispec, I've seen a few obvious undefined variable cases. 
This fixes most of them where it was clear to me what was code's intention.

### DIFF
--- a/py/desispec/scripts/autocalib_fiberflat.py
+++ b/py/desispec/scripts/autocalib_fiberflat.py
@@ -4,7 +4,6 @@ desispec.scripts.autocalib_fiberflat
 
 """
 from __future__ import absolute_import, division
-import sys
 import time
 
 import numpy as np
@@ -38,7 +37,7 @@ def main(args=None) :
     log=get_logger()
     if ( args.night is None or args.arm is None ) and args.prefix is None :
         log.error("ERROR in arguments, need night and arm or prefix for output file names")
-        sys.exit(1)
+        return 1
 
     log=get_logger()
     log.info("starting at {}".format(time.asctime()))
--- a/py/desispec/scripts/autocalib_fiberflat.py
+++ b/py/desispec/scripts/autocalib_fiberflat.py
@@ -4,6 +4,7 @@ desispec.scripts.autocalib_fiberflat
 
 """
 from __future__ import absolute_import, division
+import sys
 import time
 
 import numpy as np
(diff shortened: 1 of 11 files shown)


py/desispec/trace_shifts.py
```diff
@@ -491,9 +491,6 @@ def compute_dx_from_cross_dispersion_profiles(xcoef,ycoef,wavemin,wavemax, image
 
     t0=time.time()
 
-    if fibers is None :
-        fibers = np.arange(psf.nspec)
-
```
sbailey, line 496: I think the correct change here is to set the default `fibers=None` in the function definition, not remove this logic.  `fibers` is an input to `compute_dx_from_cross_dispersion_profiles`, and if the user provides `fibers=None` then it is supposed to default to all fibers.

py/desispec/scripts/autocalib_fiberflat.py
```diff
@@ -4,6 +4,7 @@
 
 """
 from __future__ import absolute_import, division
+import sys
```
sbailey, line 7: for the record: this `sys` was missed because it was used for a `sys.exit` upon a fatal error condition that we don't normally encounter.  The better solution is for `main` to `return 1` instead of calling `sys.exit(1)`, i.e. let the `bin/desi_autocalib_fiberflat` wrapper call `sys.exit(main(...))`.  It probably doesn't matter in this case, but that allows other functions to call `desispec.scripts.autocalib_fiberflat.main` without getting completely booted out by a `sys.exit`.  This matters more for scripts that have both a command line wrapper and are called by the pipeline, so just noting this as a design pattern.

sbailey (approved):
Wow, that was an embarrassing set of misses. Thanks for the update. I made two comments inline; one was substantive and one was stylistic. I'll take care of making both changes, then leave this open a bit longer before merging.

py/desispec/trace_shifts.py
segasai: The problem is that PSF is not defined. I couldn't quickly figure out where to get nfibers from if not psf

sbailey: Whoops! Indeed. I corrected this to be
```python
    if fibers is None :
        fibers = np.arange(xcoef.shape[0])
```
since `xcoef` is a 2D array of shape `(nspec, ncoef)`.

Issue #2398, opened 2024-10-25
Quasarnet afterburner doesn't record QN version


sbailey:
The output files from desi_qso_qn_afterburner do not seem to record the QN model file used, nor does the log output seem to have that either.  I suggest that the script should explicitly log the value of $QN_MODEL_FILE that will be read, and also add that to a DEPNAMnn/DEPVERnn header using `desiutil.depend`.

Assigning to @abrodze who had expressed interest in cleaning up multiple aspects of the QSO QN afterburner.

sbailey:
Fixed by PR #2402; closing.

Issue #2375, opened 2024-09-18
Missing columns in QuasarNET afterburner catalog


rongpu:
Currently only the best-fit redshift, redshift error and template coefficients are saved in the QuasarNET afterburner catalogs `qso_qn-*.fits`, and some of the important columns, such as `SUBTYPE` and `ZWARN`, are missing. Since we are using these QN redshifts for science analyses, I think I would be useful to at least save the redrock output for the best-fit solution.

The relevant QN code: https://github.com/desihub/desispec/blob/97b174838a2e951a32aa719f19da07e184585c39/py/desispec/scripts/qsoqn.py#L262

akremin:
This was resolved in https://github.com/desihub/desispec/pull/2402

Issue #2299, opened 2024-07-19
QuasarNET afterburner only uses HIZ template results in overlap region


dylanagreen:
When rerunning redrock the qos-qn afterburner runs the HIZ and LOZ templates separately due to the way RR accepts priors. The intention in the code is to then compare the two results in the overlap region, and keep the one with the lower chi^2 (this is normally done internally in redrock). 

When running the two templates separately the afterburner initializes global variables:
https://github.com/desihub/desispec/blob/e9ac85427046f6e475a3b11ee8a468fbecf9c773/py/desispec/scripts/qsoqn.py#L198

When aggregating the output the results are compared to the global variables, and if the fit has a better chi^2 than the globally stored chi^2 variable the fit is accepted and updated in the global variables:

https://github.com/desihub/desispec/blob/e9ac85427046f6e475a3b11ee8a468fbecf9c773/py/desispec/scripts/qsoqn.py#L251-L256

However, the loop neglects to update the global chi2 variable, so line 253 above is always comparing against `np.inf * np.ones(targetid.size)` as initialized in line 198. The effect of this is that since the HIZ templates are run _after_ the LOZ templates the HIZ results will always overwrite the LOZ results in the overlap region, regardless of whether the fit is better or at a different redshift than the LOZ templates. 

This is probably a one line fix, adding

```python
chi2[best_chi2] = chi2_tmp[best_chi2_tmp]
```

to update the global chi2 variable. after L256 should solve this issue, although validating this might take a little bit of effort. Reminder that the overlap region is 1.4 < z < 1.6.

sbailey:
Good catch.  An arguably better long term solution is to update `rrdesi --templates` to accept more than one template.  Then the afterburner can call Redrock once, let Redrock sort out which HIZ/LOZ is best by its standard algorithm, and then the after burner just uses that.  It feels like a lot of the complication / fragility here is that rrdesi currently does not allow you to specify more than one template unless you give it a full directory worth of templates (which then brings in templates that the afterburner doesn't want to use).

abrodze:
Stephen's suggestion should already be possible since Redrock can take a text file for the --templates argument. 

https://github.com/desihub/redrock/blob/main/py/redrock/templates.py#L298-307

I'll play around with the qnqso.py script to get it up to speed with this "new" redrock feature.

sbailey:
Fixed by PR #2402; closing

Issue #2281, opened 2024-06-18
Quasarnet afterburner tries to write to input directory


akremin:
While trying to run some reproducibility tests of QuasarNP, I attempted to use `daily` `redrock` and `coadd` files as inputs but save the outputs to an alternate location for comparison with the outputs in `daily`.  This is failing with the following traceback. The solution is to either write temporary files to the current working directory, or more ideally to the output directory rather than assuming the user will have write permission on the input directory.

Command:
`desi_qso_qn_afterburner --coadd /global/cfs/cdirs/desi/spectro/redux/daily/tiles/cumulative/9203/20240409/coadd-0-9203-thru20240409.fits --redrock /global/cfs/cdirs/desi/spectro/redux/daily/tiles/cumulative/9203/20240409/redrock-0-9203-thru20240409.fits --output ./test-qso_qn-0-9203-thru20240409.fits --target_selection all --save_target all`

Output:

```
DEBUG: Reading templates from /global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/redrock-templates/main
INFO:qsoqn.py:502:main: SANITY CHECK: The indices of REDROCK HDU and FIBERMAP HDU match.
INFO:qsoqn.py:537:main: Nbr objetcs for QN: 500
Model grid not found in file, defaulting to logarithmic
INFO:qsoqn.py:318:selection_targets_with_QN: We select QSO from QN with c_thresh=0.5 and n_thresh=1 --> 41 objects are QSO for QN
INFO:qsoqn.py:326:selection_targets_with_QN: Remove 35 objetcs with SPECTYPE==QSO and |z_RR - z_QN| < 0.05 (since even with the prior, RR will give the same result)
INFO:qsoqn.py:331:selection_targets_with_QN: RUN RR on 6
Traceback (most recent call last):
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/main/bin/desi_qso_qn_afterburner", line 10, in <module>
    sys.exit(qsoqn.main(args))
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/main/py/desispec/scripts/qsoqn.py", line 538, in main
    QSO_from_QN = selection_targets_with_QN(redrock, fibermap, sel_to_QN, DESI_TARGET,
[...]

PR #2402, opened 2024-10-30
QuasarNet afterburner updates part I

sbailey:
This PR requires desihub/redrock#320 and fixes:
  * #2299 (now it runs both HIZ and LOZ templates in a single call to Redrock, thus treating the overlap region self-consistently)
  * #2281 (now it writes temporary files in the output directory instead of the input directory)
  * #2398 (now it includes header keywords to record code versions, the location of $RR_TEMPLATE_DIR, $QN_MODEL_FILE, and exactly which Redrock template files were used)

I have not addressed #2375 in this PR because that will change the data model of the QN output in more substantive ways so I want it isolated in its own PR.

Heads up @abrodze @dylanagreen @akremin

### DIFF
--- a/py/desispec/scripts/qsoqn.py
+++ b/py/desispec/scripts/qsoqn.py
@@ -251,20 +251,13 @@ def collect_redshift_with_new_RR_run(spectra_name, targetid, z_qn, z_prior, para
         log.info('Done running redrock')
 
         # Extract information from the new run of RR
-        redshift_tmp, err_redshift_tmp, chi2_tmp, coeffs_tmp = extract_redshift_info_from_RR(filename_redrock_rerun_RR, targetid[sel])
+        redshift, err_redshift, chi2, coeffs = extract_redshift_info_from_RR(filename_redrock_rerun_RR, targetid[sel])
 
         if param_RR['delete_RR_output'] == 'True':
             log.debug("Remove output from the new run of RR")
             os.remove(filename_priors)
             os.remove(filename_output_rerun_RR)
             os.remove(filename_redrock_rerun_RR)
-
-        # aggregate the result:
-        best_chi2 = np.zeros(targetid.size, dtype='bool')
-        best_chi2_tmp = chi2[sel] > chi2_tmp
-        best_chi2[sel] = best_chi2_tmp
-
-        redshift[best_chi2], err_redshift[best_chi2], coeffs[best_chi2] = redshift_tmp[best_chi2_tmp], err_redshift_tmp[best_chi2_tmp], coeffs_tmp[best_chi2_tmp]
 
     return redshift, err_redshift, coeffs
 
@@ -548,7 +541,7 @@ def main(args=None, comm=None):
 
             # from everest REDSHIFTS hdu and FIBERMAP hdu have the same order (the indices match)
             if np.sum(redrock['TARGETID'] == fibermap['TARGETID']) == redrock['TARGETID'].size:
-                log.info("SANITY CHECK: The indices of REDROCK HDU and FIBERMAP HDU match.")
+                log.info("SANITY CHECK: The indices of REDSHIFTS HDU and FIBERMAP HDU match.")
             else:
                 log.error("**** The indices of REDROCK HDU AND FIBERMAP DHU do not match. This is not expected ! ****")
                 return 1
@@ -600,7 +593,7 @@ def main(args=None, comm=None):
                 log.warning(f"No objects selected to save; blank file {os.path.splitext(args.output)[0]+'.notargets.txt'} is written")
 
     else:  # file for the consider Tile / Night / petal does not exist
-        log.error(f"**** There is problem with files: {args.coadd} or {args.redrock} ****")
+        log.error(f"**** Missing {args.coadd} or {args.redrock} ****")
         return 1
 
     log.info(f"EXECUTION TIME: {time.time() - start:3.2f} s.")
--- a/py/desispec/scripts/qsoqn.py
+++ b/py/desispec/scripts/qsoqn.py
@@ -22,6 +22,7 @@ from desitarget.sv1.sv1_targetmask import desi_mask as sv1_mask
 from desitarget.cmx.cmx_targetmask import cmx_mask
 
 from desiutil.log import get_logger
+import desiutil.depend
 
 from desispec.io.util import get_tempfilename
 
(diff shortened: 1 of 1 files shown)


py/desispec/scripts/qsoqn.py
```diff
@@ -209,55 +210,61 @@ def extract_redshift_info_from_RR(filename_redrock, targetid):
...
-            rrdesi(argument_for_rerun_RR, comm=comm)
+        argument_for_rerun_RR  = ['--infiles', spectra_name]
+        argument_for_rerun_RR += ['--templates',] + param_RR['templates_filename']
+        argument_for_rerun_RR += ['--targetids', ' '+",".join(reversed(targetid[sel].astype(str))),
+                                                 # see long comment above about need for preceeding space
+                                 '--priors', filename_priors,
+                                 '--details', filename_output_rerun_RR,
+                                 '--outfile', filename_redrock_rerun_RR]
+        # NEW RUN RR
+        log.info(f'Running redrock with priors on selected targets with {template_filename}')
```
akremin, line 246: Now that `param_RR['templates_filename']` is a list of files, I think this should maybe be `param_RR['templates_filename']`? Otherwise you are only listing the last file, and this could even raise an error if `param_RR['templates_filename']` is an empty list such that `template_filename` is never defined.

py/desispec/scripts/qsoqn.py
```diff
@@ -209,55 +210,61 @@ def extract_redshift_info_from_RR(filename_redrock, targetid):
...
 
-        # find redshift range of the template
+    filename_priors = param_RR['filename_priors']
+    filename_output_rerun_RR = param_RR['filename_output_rerun_RR']
+    filename_redrock_rerun_RR = param_RR['filename_redrock_rerun_RR']
+
+    # find redshift range spanned by templates
+    zmin = 100
+    zmax = -100
+    for template_filename in param_RR['templates_filename']:
```
akremin, line 221: If the ship has sailed, that's okay, but it's a little odd to me that this has a singular name but can be a list of multiple files.

akremin (commented):
I'll leave the more detailed review to Allyson, since she has more in-depth knowledge of this code and how it interplays with redrock and the templates. I have two comments, one can be ignored if too much work, and one log message that I think needs to be updated.

abrodze:
I talked with Stephen in person about this, but I am writing it here for better tracking. 
Right now, it looks like a decent percentage of the spectra being rerun through redrock are from sky fibers. I am unsure if there is a legitimate reason to keep the implementation as such, but this could easily be fixed by requiring `fibermap['OBJTYPE'] == 'TGT' ` in the selection mask `sel_index_with_QN `.  Though the only benefit to making such a change would be a (maybe negligible) boost in runtime, as these will still be filtered out in catalog construction.

py/desispec/scripts/qsoqn.py
sbailey: good point; log message updated

py/desispec/scripts/qsoqn.py
sbailey: That naming scheme was leftover from the original code, but given that this is only used internally to the afterburner (not a public API), I updated "templates_filename" to "template_filenames" throughout.

abrodze (commented):
A few comments related to the bestchi2 section we have discussed in person. Fine to push this to future PR

py/desispec/scripts/qsoqn.py
```diff
@@ -201,63 +202,69 @@ def extract_redshift_info_from_RR(filename_redrock, targetid):
...
+            os.remove(filename_redrock_rerun_RR)
 
-            # aggregate the result:
-            best_chi2 = np.zeros(targetid.size, dtype='bool')
-            best_chi2_tmp = chi2[sel] > chi2_tmp
-            best_chi2[sel] = best_chi2_tmp
+        # aggregate the result:
+        best_chi2 = np.zeros(targetid.size, dtype='bool')
+        best_chi2_tmp = chi2[sel] > chi2_tmp
+        best_chi2[sel] = best_chi2_tmp
```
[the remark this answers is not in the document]
sbailey: Agreed; this was leftover from the previous code.  I have removed this vestigial logic from this branch (including the other `redshift_tmp` et al. comment below).

py/desispec/scripts/qsoqn.py
```diff
@@ -201,63 +202,69 @@ def extract_redshift_info_from_RR(filename_redrock, targetid):
...
 
-            redshift[best_chi2], err_redshift[best_chi2], coeffs[best_chi2] = redshift_tmp[best_chi2_tmp], err_redshift_tmp[best_chi2_tmp], coeffs_tmp[best_chi2_tmp]
+        redshift[best_chi2], err_redshift[best_chi2], coeffs[best_chi2] = redshift_tmp[best_chi2_tmp], err_redshift_tmp[best_chi2_tmp], coeffs_tmp[best_chi2_tmp]
```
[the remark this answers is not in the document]
sbailey: agreed; updated.

py/desispec/scripts/qsoqn.py
```diff
@@ -201,63 +202,69 @@ def extract_redshift_info_from_RR(filename_redrock, targetid):
...
+
+    sel = (z_qn >= zmin) & (z_qn <= zmax)
 
-        # In case where all the objects have priors outside the redshift template range
-        # Skip the template to avoid any undesired errors
-        if sel.sum() != 0:
-            write_prior_for_RR(targetid[sel], z_qn[sel], z_prior[sel], filename_priors)
+    # In case where all the objects have priors outside the redshift template range
+    # Skip the template to avoid any undesired errors
+    if sel.sum() != 0:
```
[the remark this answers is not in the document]
sbailey: Agreed, but I already have it implemented in another branch for another PR, so I'm not going to update it here yet since that would cause merge conflicts.  i.e. punting to another forthcoming PR.

sbailey:
@abrodze raised the additional issue of whether the simultaneous run of LOZ and HIZ templates will result in more ZWARN=4 (SMALL_DELTA_CHI2) flags in the 1.4 < z < 1.6 overlap region (close redshifts, but still |dv|>1000).  Previously these would not occur because LOZ and HIZ were run separately so they couldn't trigger warnings on each other.  I will try to quantify this before merging.

sbailey:
Rerun redshifts have a high ZWARN>0 rate, but they don't appear to be any higher in the 1.4 < z < 1.6 LOZ:HIZ overlap region:
image

This was based on spot checking the 10 main-dark healpix with the highest number of rerun QSOs in the overlap region.

I'm going to proceed with final cross checks on the outputs and then merge and deploy.

abrodze:
Unless you explicitly removed them, I would guess the ~50% flags is from 50% of reruns that are sky fibers

sbailey:
Good point.  At the time of this PR I had not yet removed sky fibers from the reruns, so indeed many of the ZWARNs are for sky fibers.  Additionally, low S/N spectra are more likely to have a Redrock / QuasarNet disagreement and also more likely to have a LOW_DELTA_CHI2 warning.

PR #2404, opened 2024-11-05
use better error message

### LINKED ISSUE #2397: desi_make_dark_tables crash with pre-existing tables
Running desi_make_dark_tables to update a pre-existing $DESI_SPECTRO_DARK/dark_table.csv crashes:
```
$> cd /global/cfs/cdirs/desi/spectro/desi_spectro_dark/latest
$> desi_make_dark_tables -i . -o .
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/main/bin/desi_make_dark_tables", line 62, in <module>
    outarr[:,:-1], names=["FILENAME", "DETECTOR", "CCDCFG", "CCDTMING"],
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```

It does work to send the output to a different blank directory and then copy the results back.

@Waelthus please inspect.  It is possible that I'm running it in a different mode than you do, but I think this case is supposed to work and at minimum shouldn't crash with this dimensionality error.

Waelthus:
should fix #2397

### DIFF
--- a/bin/desi_make_dark_tables
+++ b/bin/desi_make_dark_tables
@@ -47,6 +47,8 @@ for files,outfile in zip([dark_files,bias_files],
         files_new = [f for f in files if f.replace(f"{args.indir}/","") not in outtab['FILENAME']]
     else:
         files_new = files
+    if len(files_new)==0:
+        raise ValueError(f"No new files found in {args.indir} to add to {outfile}")
     for fi in files_new:
         with fits.open(fi, "readonly") as f:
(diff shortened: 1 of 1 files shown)


sbailey:
I'm not entirely sure that this will fix the symptom that I was seeing in #2397, but I also don't see any harm in this update and I'm on pipeline duty again this week including updating calib files, so I'll go ahead and merge this and use it and report again if there are hiccups.

PR #2405, opened 2024-11-05
allow fallback to darks/biases in DESI_SPECTRO_CALIB if running ccdcalib

### LINKED ISSUE #2399: desi_compute_bias requires unsetting $DESI_SPECTRO_DARK
To create a new bias with desi_compute_bias, you first have to `unset DESI_SPECTRO_DARK`, or otherwise it fails due to not finding the bias file that you are trying to create in the first place.

I did not find this procedure documented, but I have a vague memory of @julienguy mentioning needing to do this.  We should fix this so that "missing" biases and darks are only fatal if the processing is actually needs that bias/dark.

Waelthus:
should fix #2399


Waelthus:
not sure if the fix should be applied when computing nightlybias or if this should actually fail...
Also still needs to be tested.

sbailey (approved):

Issue #2360, opened 2024-09-06
Streamline prod zcatalog creation


sbailey:
For Kibo, the redshift catalogs were generated by hand by cutting and pasting commands like
```
salloc -N 1 -C cpu -A desi -t 04:00:00 -q interactive
cd $DESI_SPECTRO_REDUX/$SPECPROD

SURVEY=cmx
PROGRAM=other
desi_zcatalog -i /dvs_ro/cfs/cdirs/desi/spectro/redux/$SPECPROD/healpix/$SURVEY/$PROGRAM/ \
              -o zcatalog/v1/zpix-$SURVEY-$PROGRAM.fits -g healpix --nproc 64 |& tee zcatalog/v1/logs/zpix-$SURVEY-$PROGRAM.log

SURVEY=sv1
PROGRAM=other
desi_zcatalog -i /dvs_ro/cfs/cdirs/desi/spectro/redux/$SPECPROD/healpix/$SURVEY/$PROGRAM/ \
              -o zcatalog/v1/zpix-$SURVEY-$PROGRAM.fits -g healpix --nproc 64 |& tee zcatalog/v1/logs/zpix-$SURVEY-$PROGRAM.log

for SURVEY in sv1 sv2 sv3 main special; do
  for PROGRAM in dark bright backup; do
    echo ----- $SURVEY $PROGRAM -----
    nohup desi_zcatalog -i /dvs_ro/cfs/cdirs/desi/spectro/redux/$SPECPROD/healpix/$SURVEY/$PROGRAM/ \
                        -o zcatalog/v1/zpix-$SURVEY-$PROGRAM.fits -g healpix --nproc 64 &>> zcatalog/v1/logs/zpix-$SURVEY-$PROGRAM.log
  done
done
```

The particular combination of SURVEY,PROGRAM was derived from what exists in the tiles-kibo.csv file and then typed into bash by hand.  This could be standardized and streamlined into a single job that reads tiles-kibo.csv to know what to do and generate any missing zcatalog files.

sbailey:
Additionally, Kibo required post-facto adding of units and column descriptions with `annotate_fits -C $DESIDATAMODEL/py/desidatamodel/data/column_descriptions.csv ...` .  Make sure that is included in the wrapper, or otherwise automatically part of `desi_zcatalog`.

sbailey:
[...]

Issue #2364, opened 2024-09-09
Kibo zcatalog lacking units and descriptions


sbailey:
Kibo zcatalog creation forgot to include `--add-units` causing the output files to not have TUNITnn entries for the table columns.  This is not the default because it requires a non-standard dependency on desidatamodel to lookup the units, but that is arguably the wrong choice because it leads to errors like this.

At minimum, regenerate the Kibo zcatalog files to include units.

They are also missing column descriptions.  `desispec.io.fibermap.annotate_fibermap` and `desiutil.annotate` have some of the functionality for adding comments using `astropy.io.fits`, but that might not scale to the level needed for the giant kibo/zcatalog/v1 files, which use fitsio for efficiency and still take a long time to write.  Add column descriptions if reasonably possible, but this might be a bit ask and better saved as part of a zcatalog/v2/ update with more streamlined formats.

Thanks to @stephjuneau for reporting.

sbailey:
Using desidatamodel tag 24.9 with the updated short column descriptions from @stephjuneau, I ran `annotate_fits --overwrite $FILENAME ...` on all of the kibo/zcatalog/v1/*.fits files to add units and column descriptions.

PR #2410, opened 2024-11-08
Add a new wrapper to create all of the zcatalogs with a single command

akremin:
### Overview
This addresses https://github.com/desihub/desispec/issues/2360.

This creates a new script, currently called `desi_zcatalog_wrapper`, that will use the tiles-SPECPROD.fits file in a production to identify which SURVEY+PROGRAM pairs exist and produce zcatalogs for each of them. If all are created successfully, then the zall-* catalog is created with all of the individual zcatalogs.

Required arguments are:
` --group `to specify whether you're running healpix redshifts or cumulative redshifts. It should also work for other tile-based flavors but I have not tested that.
`--cat-version` OR `--outdir` to specify where to save the zcatalogs. If `--cat-version` is specified then it is saved under `$DESI_SPECTRO_REDUX/$SPECPROD/zcatalog/<cat_version>/`.

The user can specify `--indir`, but it is optional and defaults to the typical places for a production.

All of the boolean flags accepted by desi_zcatalog are also accepted by this script, which propagates them to desi_zcatalog. Though I did not thoroughly vet my implementation of these.

This PR also includes updates to `desispec.io.meta.findfile` to include the `ztile`, `zpix`, and `zall` files. That required adding a new function argument `version` that is only used for these redshift catalogs at the moment but could be used for other things in the future. Because there is ongoing work on the zcatalogs v2, I did not edit the `desi_zcatalog` code itself to use `findfile`, but that should be updated to use `findfile` at some point during the v2 refactor.

#### Recommended usage, which uses only --cat-version and --group:
`desi_zcatalog_wrapper -g healpix --cat-version=v1 --nproc=2 -v`
`desi_zcatalog_wrapper -g cumulative --cat-version=v1 --nproc=2 -v`

#### Alternate method that uses --outdir
`desi_zcatalog_wrapper -g healpix --outdir=/global/cfs/cdirs/desi/users/kremin/PRs/zcat_maker/test --nproc=2`
`desi_zcatalog_wrapper -g cumulative --outdir=/global/cfs/cdirs/desi/users/kremin/PRs/zcat_maker/test --nproc=2`

#### Alternate method that uses cat-version for output but specify input
`desi_zcatalog_wrapper -g cumulative --cat-version=v2 --nproc=2 -i /global/cfs/cdirs/desi/spectro/redux/loa/tiles/cumulative`
`desi_zcatalog_wrapper -g healpix --cat-version=v2 --nproc=2 -i /global/cfs/cdirs/desi/spectro/redux/loa/healpix`

#### Alternate method that uses --indir and --outdir
`desi_zcatalog_wrapper -g healpix -i /global/cfs/cdirs/desi/spectro/redux/loa/healpix -outdir=/global/cfs/cdirs/desi/users/kremin/PRs/zcat_maker/test2 --nproc=2`
`desi_zcatalog_wrapper -g cumulative -i /global/cfs/cdirs/desi/spectro/redux/loa/tiles/cumulative --outdir=/global/cfs/cdirs/desi/users/kremin/PRs/zcat_maker/test2 --nproc=2`

These all work. I also linked the zcatalogs from loa and ran again, and the zall generation appears to work as expected.


akremin:
The doc generation passes on my checkout at NERSC and there is nothing informative I have been able to glean from the GitHub run logs.

py/desispec/zcatalog.py
```diff
@@ -157,6 +157,13 @@ def _get_survey_program_from_filename(filename):
     program = arr[2]
     return survey, program
 
+def create_summary_catalog_wrapper(kwargs):
+    """
+    Trivial wrapper around create_summary_catalog that takes a dict
+    and passes the key-value pairs to create_summary_catalog as keyword arguments
+    """
+    return create_summary_catalog(**kwargs)
```
sbailey, line 165: It appears that this function exists only so that `desispec.scripts.zcatalog_wrap` can call it with `runcmd(create_summary_catalog_wrapper, args=kwargs, ...)` because `runcmd` only supports a list of `args` and not a dict of `kwargs`.  Ideally `runcmd` would also support `kwargs`, but that is admittedly kind of messy given everything else that it is doing (function vs. script, MPI vs. not...).  Lacking that, I suggest moving this function into `desispec/scripts/zcatalog_wrap.py` as a `_hidden` function for this one use-case and not making it part of the `desispec.zcatalog` namespace.

py/desispec/scripts/zcatalog_wrap.py
```diff
@@ -0,0 +1,226 @@
...
+        args = parse(options=args)
+
+    if args.verbose:
+        log=get_logger(DEBUG)
+    else:
+        log=get_logger()
+
+    ## Ensure we know where to save the files
+    if args.outdir is None and args.cat_version is None:
+        log.critical(f"Either --outdir or --cat-version must be specifiied. Exiting")
```
sbailey, line 72: minor typo "specifiied" -> "specified"

py/desispec/scripts/zcatalog_wrap.py
```diff
@@ -0,0 +1,226 @@
...
+        args.outdir = os.path.dirname(findfile(ftype,
+                                               version=args.cat_version,
+                                               groupname=args.group,
+                                               survey='dummy', faprogram='dummy'))
+
+    ## Define catalog version (currently not used except with findfile,
+    ## but since we are only using findfile for the filename and not the
+    ## directory tree, the version isn't actually relevant
+    if args.cat_version is None:
+        args.cat_version = os.path.basename(args.outdir)
```
sbailey, line 106: This default `--cat-version` logic is a big subtle, turning the basename into the catalog version which then propagates into the output filenames.  Consider
* making `--cat-version` a required parameter (a slight hassle but safer, and potentially mitigated by making a `-V` shortcut, or `-v` and making `--verbose` be the long thing without a `-v` shortcut because it is afterall, verbose, like this comment...); or
* logging a warning message about how `cat_version` is being derived given that it wasn't specified on the command line.

sbailey:
Looking at /global/cfs/cdirs/desi/users/kremin/PRs/zcat_maker/test/zcatalog/v1/zpix-cmx-other.fits, it appears that this didn't add units.  I made the same mistake when running Loa by hand (thanks to @weaverba137 for noticing; I'll post-facto fix that), but let's get this wrapper updated so that it automatically includes units too.

akremin:
I had assumed this was fixed downstream since https://github.com/desihub/desispec/issues/2364 was closed. The wrapper includes the `--add-units` option, just as `desi_zcatalog` does. 

Given that the issue isn't resolved, I am happy to fix it in this PR. I can think of two options:

1) Remove the `--add-unit` option for both the wrapper and `desi_zcatalog` and always add units. 
2) Flip the option to be opt-out, i.e. `--do-not-add-units` for both the wrapper and `desi_zcatalog`

Also happy to hear other ideas

weaverba137:
@akremin, I know this is somewhat deeply buried, but in order for `--add-unit` to do anything at all, `module load desidatamodel` must be run before the script starts.

py/desispec/scripts/zcatalog_wrap.py
akremin: I've added a -V shortcut and made it required.

py/desispec/scripts/zcatalog_wrap.py
akremin: Done

py/desispec/zcatalog.py
akremin: I couldn't agree more about your `runcmd` comments. That was what I initially wanted to do before realizing it didn't accept kwargs. That change would be subtle and feature creep on this PR, so I've taken the second suggestion of making it a _hidden function in `zcatalog_wrap.py`.

akremin:
@sbailey I believe all changes have been addressed, including making the addition of units opt-out rather than opt-in.

sbailey (approved):

PR #2415, opened 2024-11-20
Fix zcatalog columns when reading from many epochs of redshift catalog types

weaverba137:
This PR updates the `read_redrock` function to handle many different styles of per-tile redshift catalogs and return consistent results.

Note that some columns are created with dummy values. For example, when reading from zbest files, there is no obvious way to re-generate the TSNR2 columns (unlike regenerating the fibermap and per-exposure fibermap columns).

If there _is_ an easy way to re-generate TSNR2 columns, we should do that.

I've generated `ztile-*-*-cumulative.fits` files for all survey-programs in `daily` and the columns were consistent across all files, and also basically the same as `loa`.

### DIFF
--- a/py/desispec/scripts/zcatalog.py
+++ b/py/desispec/scripts/zcatalog.py
@@ -132,15 +132,6 @@ def read_redrock(rrfile, group=None, recoadd_fibermap=False, minimal=False, pert
             redshifts = fx['ZBEST'].read()
             zbest_file = True
 
-        #
-        # These old columns show up in zbest files. They have been replaced with
-        # COADD_NUMEXP, COADD_NUMTILE, which are obtained from coadd_fibermapp()
-        # for zbest files.
-        #
-        for drop_column in ('NUMEXP', 'NUMTILE'):
-            if drop_column in redshifts.dtype.names:
-                redshifts = drop_fields(redshifts, drop_column, usemask=False, asrecarray=False)
-
         if recoadd_fibermap or zbest_file:
             if zbest_file:
                 spectra_filename = checkgzip(replace_prefix(rrfile, 'zbest', 'spectra'))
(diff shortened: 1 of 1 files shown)


weaverba137:
@sbailey, could you take a look at this when you get a chance?

sbailey:
Overall this looks fine to me; I will trust your real-world testing as sufficient for catching the corner cases.

> Note that some columns are created with dummy values. For example, when reading from zbest files, there is no obvious way to re-generate the TSNR2 columns (unlike regenerating the fibermap and per-exposure fibermap columns).
> If there is an easy way to re-generate TSNR2 columns, we should do that.


In the PR description you mentioned the TSNR2 columns, but in the PR changes I don't actually see anything specific to the TSNR2 columns, e.g. filling in with dummy values like other missing columns, so I'm confused how those are handled here if they aren't in the redrock/zbest files.

I spot checked an old daily TILEID=100 LASTNIGHT=20210505 which has zbest files and the old single FIBERMAP (not FIBERMAP + EXP_FIBERMAP) and the TSNR2 columns are in the coadd files SCORES table.

weaverba137:
Thanx, I'll see if I can reasonably quickly extract what I need from `SCORES`.

weaverba137:
@sbailey, it's been so long now that I'm starting to forget what I already did. If you look at `desispec.scripts.zcatalog`, line 155++, I've read the SCORES table in the `spectra` file and then called `compute_coadd_tsnr_scores()`. This is being called in the latest round of zcatalog files I generated for daily. However, the `TSNR2_GPB*` columns are still missing in many cases, and *these* are getting filled with dummy values. Is there another step that generates these columns, or are these simply missing entirely from older files?

In particularly old reductions (cmx, sv1), the `SCORES` table is missing from the `spectra` or otherwise doesn't contain additional columns, so we still have to fill the majority of the `TSNR2` columns with dummy values.

So basically, unless we really want to put in the work to regenerate a `SCORES` table from scratch rather than reading it from the `spectra` file, we ought to merge this PR, which technically doesn't change anything about `TSNR2`, it is just about making the columns returned by `read_redrock()` uniform.

sbailey:
Thanks for itemizing the cases where TSNR2 columns are filled with dummy values.  Those sound fine to me, and I'm not aware of any simple alternatives to filling them in when they are missing from old daily prod files.  Will merge now.

sbailey (approved):

Issue #2420, opened 2024-11-30
Bug in the flux-calibration variance calculation ? 


segasai:
Hi, 

While preparing for #2419, 
I've looked into fluxcalibration code and I believe the the current code that compute the ivar's of flux-calibrated frames is not quite correct:

https://github.com/desihub/desispec/blob/255647f0d2b7038c4747d9abb86af49bcceea9d6/py/desispec/fluxcalibration.py#L1478

Specifically the current code uses this eqn:

$$Newivar =  1. / ( 1. /(iVar(F) C^2) + F^2 /(iVar(C) C^4))$$
where Ivar(C) is the inverse variance of fluxcalib vector, and  ivar(F) is the inverse variance of spectra.
The problem here, is that it applies this eqn to already calibrated spectra (i..e original flux divided by C) 
see https://github.com/desihub/desispec/blob/255647f0d2b7038c4747d9abb86af49bcceea9d6/py/desispec/fluxcalibration.py#L1473

While the equation only applies to original undivided flux. 

Derivation:

$$F1 = F/C$$

$$Var(F1) = 1/C^2 Var(F) + F^2  Var(1/C) = 1/C^2 Var(F) + F^2 /C^4  Var(C)$$

$$iVar(F1) = 1./(1/C^2 Var(F) + F^2 /C^4 Var(C)) = 1./(1/(C^2 iVar(F)) + F^2 /(C^4  iVar(C)) )  $$
This is exactly what the current code uses, but instead of F in it, it uses F1.
Therefore this leads to incorrect C^2 factor in the second term.
Essentially the current code does: 
$$iVar(F1) = 1./(1/C^2 Var(F) + F^2 /C^4 Var(C)) = 1./(1/(C^2 iVar(F)) + F^2 /(C^6  IVar(C)) )  $$
A quick check shows the typical value of C is ~20-70, so that will lower the second term significantly, reducing the effect from flux-calibration error.

I am expecting that this will affect variances, where flux calibration uncertainty starts to become comparable to the spectrum uncertainty.
I don't know a typical S/N of flux calibration vectors a look at a few random files they can be pretty low, i.e. 20.

Am I missing something here ? 
        S

segasai:
Assuming that I don't do something stupid here, the effect from this bug is massive I believe.
I.e. I compared the reduction with the #2422 fix and loa, and here's a S/N new vs S/N original
[...]

PR #2422, opened 2024-12-02
Fluxcalib variance fix

segasai:
Proposed  fix to #2420 
I've added an extra test is for test_fluxcalibration which tests  a case of infinite S/N spectra, but finite S/N for calibration vector. In that case the correct S/N of the flux-calibrated spectrum should be the same as of the calibration vector 
Previously the tests only considered the case of infinite S/N fluxcalib vector.


julienguy (approved):
I approve the changes.

julienguy:
I checked the code is running and results is smaller inverse variance for bright objects.

segasai:
thank you for merging quickly!

Issue #2419, opened 2024-11-29
Flux-calibration and resolution matrices


segasai:
Hi, 

It is my understanding that while at extraction time it is true that the extracted spectrum S probably can be written as 
$S = M * T$ where T is the "unconvolved" spectrum, and M is the resolution matrix. The unconvolved spectrum however is the uncalibrated spectrum.
Therefore, strictly speaking, after dividing  your uncalibrated spectra by the flux calibration vector, the resolution matrix must be updated to still apply. 

To first order I think if your the calibration vector is  C you need a new resolution matrix which is  
$C_d M C_d^{-1}$
where $C_d$ is the diagonal matrix from the calibration vector. 
Obviously if calibration vector doesn't change with wavelength ( or is very smooth) , it's a non-issue, but if does have sharp changes (which it does for DESI), then the effect will be noticeable.
From my quick look at the code, I don't think this correction is applied in desispec (please correct me if I am wrong).

A quick test on a single z-band exposure shows that ~ 1% of pixels the resolution matrix values change by more than 0.05 (i.e 10% of the peak value). 

From what I can see the change is kind'a trivial -- just update this :

https://github.com/desihub/desispec/blob/255647f0d2b7038c4747d9abb86af49bcceea9d6/py/desispec/fluxcalibration.py#L1437

by changing the resolution matrix after calibration. 

Thoughts ? 

Thanks, 
      S

PS I haven't tested this obviously, but if that is a correct thing to do I'd expect this to lead to improvement in chi-squares.

segasai:
With the #2422 merged, I intend to proceed to make a patch for this issue, as I think this is the right approach.
The only tricky thing I can see is what to do if the calibration vector is invalid somewhere (==0), hence you can't invert the matrix that I specified. 
Here is my thinking for the correct procedure. 

If we have and extracted spectrum S and some kind of true model spectrum M and resolution matrix R, then
$E[S] = R M$

If we now want to use the calibrated spectrum $S_c = C_d^{-1} S$ where ($C_d$ is the diagonal matrix made from calibration vector), and calibrated model spectrum $M_c = C_d^{-1} M$, 
then for $E[S_c] = {\tilde R} M_c$ we need ${\tilde R} = C_d^{-1} R C_d$

[...]

PR #2423, opened 2024-12-04
Correct the resolution matrix when doing flux calibration

segasai:
This addresses #2419
and is WIP as one of the newly written tests fails ATM.

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -1502,10 +1502,7 @@ def _calibrate_resolution_matrix(res_mat, calib, good_calib):
         res_mat_fc: corrected resolution matrix
     """
     # We use calculation from #2419
-    # res_mat is the banded representation of the resolution matrix
-    # calib is the calibration vector
-    # good_calib is the subset of pixel where calib is valid
-    # the caclulation is C^{-1} R C
+    # the calculation is C^{-1} R C
     # the only complication if C has zeros.
     # Then for the left we take the pseudo inverse
     # and for the right we interpolate over bad regions
@@ -1518,8 +1515,9 @@ def _calibrate_resolution_matrix(res_mat, calib, good_calib):
         rcalib = _interpolate_bad_regions(calib, ~good_calib)
     # now we align the diagonals to the banded storage
     # of the resolution matrix
-    M1 = [np.roll(icalib, _) for _ in np.arange(-5, 6)[::-1]]
-    M2 = [1./rcalib for _ in np.arange(-5, 6)]
+    width = res_mat.shape[0]
+    M1 = [np.roll(icalib, _) for _ in np.arange(-width//2,(width//2)+1)[::-1]]
+    M2 = [1./rcalib for _ in np.arange(width)]
     res_mat_out = res_mat * M1 * M2
     # This has been tested with X0 and X1 being indentical in this calculation
     # M1 = [np.roll(D1, _) for _ in np.arange(-5, 6)[::-1]]
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -1476,15 +1476,67 @@ def apply_flux_calibration(frame, fluxcalib):
                                  (C[i, ok]**2 * fluxcalib.ivar[i, ok] +
                                   frame.flux[i, ok]**2 * frame.ivar[i, ok]
                                   ))
+            frame.resolution_data[i] = _calibrate_resolution_matrix(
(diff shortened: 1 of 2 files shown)


segasai:
I think this patch is ready.
The new tests check the following.
1) If the calibration factor is constant then the resolution matrix should not change
2) if the calibration is not, then if before calibration the flux was subject to the condition
$F= R * F_0$
then after the calibration 
$F_c = R_c * F_0,c$
where $F_{0,c} = C^{-1} F_0$

I also test the masked calibration case that it still works. For that test  I use a constant calibration vector.

dylanagreen:
I did some investigation using a smoothed version of the calibration vector instead of the full version. Here are the updated redrock results for this tile: 
```
For 5000 spectra for ZWARN == 0 for both fits...
Changed Spectype: 0.0
All redshift changes:
	 Changed z by more than dv=10: 0.0012
	 Changed z by more than dv=50: 0.0004
	 Changed z by more than dv=100: 0.0
	 Changed z by more than dv=500: 0.0
	 Changed z by more than dv=1000: 0.0
Changed ZWARN !=0 to ZWARN == 0: 0.0004
Changed ZWARN ==0 to ZWARN != 0: 0.0008
``` 
This is essentially a negligible change in redrock results while still maintaining the "spirit" of the PR. With the smoothed calib vector, the median number of pixels masked in each spectrum still rises slightly, from 41 to 53 (although significantly smaller still than the full vector...) The histogram of resolution sums shows that it seems relatively similar to the main production run.

res_sum(4)
spectype(4)
The model is now essentially the same between the two fits now (here they lie on top of each other).

 
spectype(5)
This is the largest redshift difference in all the results and the difference is essentially inconceivable.

The $\chi^2$ values are also now extremely comparable:
chi2_calc(3)

The $\Delta \chi^2$ plot is essentially the same. 

@segasai What are your thoughts on smoothing the calibration vector before applying it to the resolution matrix? I think we are weary of making a downstream change to redrock to couple with the change here, especially when that cut in redrock has other benefits (namely, masking unmasked cosmic rays and the edge of each CCD wavelength range, where the resolution matrix is ill defined). Stephen and I are uncertain though to what extent that change would have an effect on your code. To put it another way, we are not certain whether your suggested change to the resolution matrix here is motivated _just_ because it is the correct math, or if you had observed some downstream effects in your stellar fitting code that would be resolved with this change to the resolution matrix? 

In my opinion smoothing the calibration vector is actually the "closest" we can get to the correct thing to do here. The math outlined in #2419 is only strictly correct if there is no uncertainty in the calibration vector as the covariance matrix generated by the resolution matrices does not account for variance in the calibration vector (recalling this from memory because I left my notes at my other desk...so hopefully I have this right). Nominally though our calibration vector _is_ noisy, but to handle that 100% correctly is computationally expensive, whereas smoothing the calibration vector can be seen as a "good enough" proxy for removing the noise and uncertainty.

segasai:
Thanks for looking at this further @dylanagreen 

1) Regarding the impact of this on the stellar fits. I do not have a specific case where the current pipeline causes problems, i.e. 'correct maths' was the main motivation for the change as I needed to understand 100% precisely how to correctly go from infinite resolution model spectra to DESI extracted spectra. The flux calibration as it is currently done breaks the "maths"
If I had to guess, the redshifts may not be affected too much, but things like line-fluxes, especially if they are in the z-arm where the throughput is varying will be affected
2) Regarding the smoothing of the calibration vector, personally I do not think that it is a good idea, because if the noisy calibration vector is the best approximation to the truth we have, we should just apply it. And the exact equations in #2419 are applicable in expectation sense. If we think the smooth calibration vector is better approximation to the truth it should be applied for both the calculation of the spectra and resolution matrix, if not there is no reason to use different vectors for spectra calibration and resolution mat calibration. (but this is my handwaving)
3) I obviously understand historical reasons for relying on sum_i(resol_{i,j})!=1 as a flag of something  being wrong with extraction, but I think it is probably not a good idea IMO (probably it would be too much work to change, but I thought I'd add my 2c)

dylanagreen:
@segasai, you may have seen my other tag, but Stephen and I have been working on this the last week and believe that while adjusting the resolution matrix for the calibration is the right thing to do, the implementation in this PR is not correct. The mathematics of our new methodology is included in the other PR, at #2642. The implementation deviated far enough from this work that it was tantamount to writing it from scratch anyway, and it was cleaner to create a new branch off main to implement the new method. The gist of it is that instead of using the per spectrum calibration on the right hand side, we use the original estimated global deconvolved calibration for that exposure.

PR #2426, opened 2024-12-11
Update to new output from fit_trace_shifts.

sybenzvi:
The function `fit_trace_shifts` now outputs three data structures instead of a single `XYTraceSet` instance when `shift_psf` is set to `True`. This fix will unbreak qproc for processing of science and calibration exposures ([see nightwatch#444](https://github.com/desihub/nightwatch/issues/444)).


sbailey:
Apologies that we broke that in PR #2411!  Merging now and I'll update desispec at NERSC for tonight.

Issue #2427, opened 2024-12-11
qproc issues processing arc exposures


sybenzvi:
PR #2426 fixed the processing of science exposures and flats with qproc, but not arcs, which are failing with the exception

```
desispec/scripts/trace_shifts.py: line 210, in fit_trace_shifts

x_for_dy,y_for_dy,dy,ey,fiber_for_dy,wave_for_dy,dwave,dwave_err = compute_dy_using_boxcar_extraction(tset, image=image, fibers=fibers, width=args.width, continuum_subtract=continuum_subtract)

UnboundLocalError: local variable 'continuum_subtract' referenced before assignment
```

Spec tests of obstype `OTHER` are also impacted by the same exception.

julienguy:
Hi Segev, the fix looks trivial but do you have an example nightwatch script 
command that I can use to verify this?

sybenzvi:
@julienguy , yes, at NERSC in a DESI environment (with updated desispec) you can run

```
module load nightwatch/main
nightwatch run -n YYYYMMDD -e 00NNNNNN --cameras b0,r0,z0 -o output_dir
```

PR #2428, opened 2024-12-13
Fix the traceshift undefined variable bug

segasai:
Fixes the #2427 
Sorry for introducing it. 
(can't test it myself, but I think it should do it).


julienguy:
I tested the fix in the context of a nightwatch run of an arc exposure (which crashes with main and works with this branch):
```
nightwatch run -n 20241211 -e 00267785 --cameras b0,r0,z0 -o .
```
Merging.

PR #2435, opened 2025-01-24
Add option to not resub failed exposures in proc_night

akremin:
This adds a new feature in `desi_proc_night` that is turned off by default. When explicitly enabled it prevents the code from trying to resubmit old jobs in the processing table that were found to have failed, timed out, etc.

The reason for this feature is that I'm working in the daily prod on old nights where multiple problems exist. I want to be able to solve the problems one by one, but the code crashes when trying first to resubmit old jobs that can no longer be run without special intervention.

I've tested that this solves my current problem and the code works fine with this minimal change.


sbailey (approved):
Looks good, thanks.  I will trust your real-world testing.

sbailey:
tests are failing for unrelated reasons about installing svn; merging anyway.

Issue #2433, opened 2025-01-22
Hot column contaminating the overscan region


rongpu:
As mentioned in [this slack message](https://desisurvey.slack.com/archives/C02KKL0PBA7/p1737562066967539), since a readout parameter change on 20250114 there is a degradation of a hot column in z1, and it significantly affects the overscan region of amp C. This led to a number of fibers in petal 1 to be flagged as "LOWEFFTIME" (https://github.com/desihub/desisurveyops/issues/274).

To illustrate, below are morning darks before and after the readout change. The overscan region is on the right side of the red line. We will try a software fix by discarding the contaminated rows in the overscan image. The right panel shows another hot column on the same CCD which may be insufficiently masked at the moment.

![Image](https://github.com/user-attachments/assets/746c3ffa-6f3c-40ac-af65-fc1c6934052f)
![Image](https://github.com/user-attachments/assets/f4c714b4-3e8c-4f23-bfa6-391389d44251)

PR #2438, opened 2025-02-02
Add option to select "good" overscan region in preproc

rongpu:
This PR adds the option to only use a "good" section of the overscan region for preproc. This is done by reading the GOODBIASSEC{amp} parameter from the yaml file and it overrides BIASSEC{amp} in the header.

This is option is needed to fix the overscan contamination issue https://github.com/desihub/desispec/issues/2433 in z1 (amps A and C). The updated yaml files (with the GOODBIASSEC parameters added in sm10-z.yaml) are already committed to svn, but they won't do anything until this PR is merged.

### DIFF
--- a/bin/desi_compute_pixmask
+++ b/bin/desi_compute_pixmask
@@ -18,6 +18,7 @@ from desispec.preproc import parse_sec_keyword, get_readout_mode, get_amp_ids
 from desispec.maskbits import ccdmask
 from desispec.workflow.tableio import load_table, load_tables
 from desispec.calibfinder import CalibFinder
+from desispec.io.raw import read_raw_primary_header
 
 #Parser to take arguments from command line
 parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter,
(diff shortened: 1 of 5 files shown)


julienguy (approved):
Verified the code. Tested on a few exposures including , including desi_compute_pixmask.
We will rerun a full night of data after fixing the master bias and master dark that are also affected.

julienguy:
We have unit test errors: Failed to build fitsio? @weaverba137 can you look at it?

julienguy:
Merged main following Ben's advice.

araichoor:
thanks @rongpu!

quick sanity check: I look here at `EFFTIME_NORM=f(FIBER)` for all the petal=1 main bright/dark reduxs in `/global/cfs/cdirs/desi/spectro/redux/jguy/tiles/cumulative/*/20250119` and for a comparison night in the daily (20250105).
for each tile I normalize the `EFFTIME` to the median value over the petal.

the pattern look similar for the reference (daily/20250105) and the fixed reduxs (jguy/20250119), great!

tmp-main
tmp-new

Issue #2387, opened 2024-10-09
Allow desi_proc_night to submit only redshift jobs or desi_zproc to add entries to processing table


akremin:
For daily operations, it is sometimes the case that we remove data from an old night and need to rerun cumulative redshifts on a later night.

As of now there is now "easy" way to resubmit the later cumulative night without human editing of tables or removing the spectral processing on that later night (even though the exposure-level data on the later night is independent).

See whether we can easily add logic in `desi_proc_night` to handle this case, or alternatively add `processing_table` logic to `desi_tile_redshifts` or `desi_zproc`

PR #2441, opened 2025-02-11
Allow desi_proc_night to submit new redshift jobs even if tilenight exists

akremin:
#### Overview

This solves issue #2387. There are circumstances where the tilenight job has already been submitted but we wish to submit additional redshift jobs that depend on those spectra. A common example is deleting spectra and redshifts from an earlier night in daily operations. Any future cumulative redshifts must also be deleted since they used the deleted data. This allows us to resubmit those cumulative jobs within the structure of the pipeline with proper dependencies and processing_table tracking.

#### Testing

I did a number of tests to make sure the new functionality works:

1. That on a night with all requested redshifts the code submits nothing and exists.
2. On a night that hasn't been submitted, it submits everything with the appropriate dependencies.
3. On nights with missing redshifts, the code submits the missing redshifts with appropriate dependencies and nothing more.
4. On nights with everything where we request new perexp redshifts that it submits those but nothing else.
5. I've ensured that it works on old processing tables that have per-exposure "poststdstar" jobs instead of tilenight jobs.

### DIFF
--- a/py/desispec/scripts/proc_night.py
+++ b/py/desispec/scripts/proc_night.py
@@ -35,7 +35,8 @@ from desispec.workflow.processing import define_and_assign_dependency, \
     night_to_starting_iid, make_joint_prow, \
     set_calibrator_flag, make_exposure_prow, \
     all_calibs_submitted, \
-    update_and_recursively_submit, update_accounted_for_with_linking
+    update_and_recursively_submit, update_accounted_for_with_linking, \
+    submit_redshifts
 from desispec.workflow.queue import update_from_queue, any_jobs_need_resubmission, \
     get_resubmission_states
 from desispec.io.util import decode_camword, difference_camwords, \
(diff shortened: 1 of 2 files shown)


akremin:
Self-merging, but I would like to hear comments from @sbailey if he has any post-merge.

py/desispec/scripts/proc_night.py
```diff
@@ -540,7 +540,43 @@ def create_submit_add_and_save(prow, proctable, check_outputs=check_for_outputs,
...
+            ## old proctables have poststdstar, check for that or tilenight
+            if 'tilenight' in tile_prows['JOBDESC']:
+                tnight = tile_prows[tile_prows['JOBDESC']=='tilenight'][0]
+            elif 'poststdstar' in tile_prows['JOBDESC']:
+                poststdstars = tile_prows[tile_prows['JOBDESC']=='poststdstar']
+                tnight = tile_prows[tile_prows['JOBDESC']=='poststdstar'][-1]
+                ## Try to gather all the expids, but if that fails just move on
+                ## with the EXPID's from the last entry. This only happens in daily
+                ## for old proctabs and doesn't matter for cumulative redshifts
+                if len(tnight) > 1:
```
sbailey, line 564: I'm not following the logic here.  If `tile_prows` is a Table, then `tnight = tile_prows[...][-1]` is a Row object for the last row of the Table, and `len(tnight)` is counting the number of columns which will always be more than 1.  It looks like this logic was thinking that `tnight` is a Table and it is checking how many rows it has, but I don't see how that could happen.  I'm also not understanding the `np.concatenate` in the following since `poststdstars['EXPID']` appears to be a Column, not an array of arrays?  What's the case that would trigger "Tried and failed to populate full EXPIDs..." but we would still want to pass?

akremin: Good catch, I added this in my last commit and meant to write `len(poststdstars)`. If there was more than one row in the table slice, then we want to concatenate all of the EXPID's.

`poststdstars['EXPID']` is a column, but each element is a `numpy` array with an arbitrary number of EXPIDs. In practice that number is always 1 for `poststdstar` jobs, but the code is still the easiest for combining an arbitrary number of arrays into a single sorted array.

As I note in the comment, whether the EXPID column is correct or not isn't critical since `desi_zproc` re-derives the correct exposure id's for cumulative redshifts, which are the only ones we run for `daily`.  Since `poststdstar` jobs are deprecated and only exist in `daily`, we're therefore safe. However, it's always better to be accurate when we can be.

I'll make the change in a future PR, but since this is an inconsequential bug I won't do it today.

Issue #2408, opened 2024-11-07
dashboard shows unknown for job that should be known


akremin:
On Nov 5th 2024 in daily a tile was observed on two exposures. The first was good, but the second had low exposure time (aborted) and was only run through sky subtractions. The column coloring is all correct but the status shows as "unknown" and is colored orange for the skysub exposure. Below is a screenshot. We should fix the dashboard to do the correct thing in this case.

<img width="1590" alt="image" src="https://github.com/user-attachments/assets/72fd2460-6a81-4596-a02f-49b0ca5ec0cd">

akremin:
The issue was that for this tile and 2 other examples I found, the skysub exposure was not in the processing_table for the tile but was rather in the unprocessed_table. 

My hypothesis is that `desi_proc_night` was run without `--daily` mode, such that the script would not submit skysub jobs. However, that doesn't explain why the outputs exist on disk, as I would have expected them to have been removed if the tile was purged.

Because I don't have an explanation for the root cause, I will leave this ticket open, even if the issue is not in the dashboard.

sbailey:
Other recent examples of this issue are nights 20241205 (expid 266913), 20241206 (expid 267061), and 20241209 (expid 267511), all of which were LASTSTEP=skysub.  We're starting to build up several false positive orange alerts per month, so let's revisit this to try to identify and fix the underlying cause.

akremin:
Luckily there is a simple explanation and an equally simple fix. In the desi_proc_night development, the science exposure selection included a cut on EXPTIME of 60s. So no short exposures were being propagated to the processing table. However, since the tilenight jobs independently identify what to process for a given tile on a given night based on the LASTSTEP's, they were still correctly processing these short exposures through sky subtraction.

Because the exposures were processed but the exposure ID's weren't present in the processing table, they were correctly listed as "unknown". I.e. this is a bug in the pipeline that the dashboard caught, rather than a bug in the dashboard.

I will make a PR momentarily to fix this from happening in the future. For old processing tables in daily, I will hand-edit them to add the `skysub` exposure and the dashboard should then be happy.

akremin:
Tables in daily have been updated, so once the code is committed this issue will be solved for future nights in daily and all future prods.

akremin:
PR #2442 is now committed. Historical data is already fixed, so this issue is resolved.

PR #2442, opened 2025-02-12
Resolve unknown statuses in the dashboard and fix typo

akremin:
This resolves issue #2408 and fixes a typo identified after merging PR #2441. 

The solution to the issue is to remove a redundant cut on exposure time. All science exposures less than 60s are already labeled as `LASTSTEP=skysub` automatically by the pipeline. If the user doesn't want such low signal-to-noise data, they can specify `science_laststeps="all"`.


akremin:
Checks are failing but it appears to be an issue with coveralls and not with the PR.

sbailey (approved):

PR #2446, opened 2025-02-28
Update RTD configuration

### LINKED ISSUE #2445: desispec.readthedocs.io search function doesn't work?
The search function of desispec.readthedocs.io doesn't appear to work.  I just get a "Searching..." message that never seems to resolve into an answer.  It is unclear to me whether this is a bug in readthedocs itself, or a bug in our config that prevents it from working.

Example search for "Spectra".

@weaverba137 please investigate.  If it is a config problem on our side, please also check desiutil, desitarget, redrock, desimodel for similar issues.  For some odd reason github isn't letting me assign the ticket to you.

I'm flagging this for DR1 in the hopes of getting our code docs updated and searchable for DR1.

weaverba137:
This PR fixes #2445.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -143,7 +143,7 @@ jobs:
               with:
                 python-version: ${{ matrix.python-version }}
             - name: Install Python dependencies
-              run: python -m pip install --upgrade pip setuptools wheel Sphinx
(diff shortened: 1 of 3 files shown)


weaverba137:
Self-merging at @sbailey's request.

PR #2447, opened 2025-03-04
Fix tsnr afterburner bug causing crash on nights with no "good" data

akremin:
It appears that PR #1950 didn't fully resolve issues related to nights with no valid science data, or another bug was reintroduced. This PR fixes that.

The code was crashing because it was trying to use a list of computed sky magnitudes to get the list of columns in each row of that list. However, on nights with no valid exposures, no rows are computed, and that caused a crash.

The fix was a simple one-line change with an extra line for logging. I also took the opportunity to update the afterburner to use `findfile` to identify the pipeline `exposure_table`(s).

I tested this in `/global/cfs/cdirs/desi/users/kremin/PRs/tsnr_skymagcrashfix` and the code now runs and produces the correct results on the test night that is crashing in `daily`.

### DIFF
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -407,42 +407,42 @@ def compute_summary_tables(summary_rows, preexisting_tsnr2_expid_table,
         if nights is None:
             nights = np.unique(exp_summary["NIGHT"])
         for night in nights:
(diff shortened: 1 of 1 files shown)


akremin:
I've fixed a bug in the logging and upgraded the info message to a warning message if no sky magnitudes were computed when the user requested them.

I've tested this on the night that crashes (20250122) and a night that didn't crash (20250123) and this branch runs in both cases and produces plausible outputs.

sbailey (approved):

PR #2448, opened 2025-03-04
default DESI_SPECTRO_REDUX

sbailey:
This PR makes `$DESI_SPECTRO_REDUX` optional for the purposes of `desispec.io.findfile`.  If it is set, it will be used just like it currently is, but if not set it will default to `$DESI_ROOT/spectro/redux`.  I updated the tests to confirm this behavior. The motivation for this is to make the required environment setup for end-users as simple as possible, in particular for the tutorials (`$DESI_ROOT` will be the only required environment variable for end-user analysis, while still allowing `$DESI_SPECTRO_REDUX` override).  Ditto for `$DESI_SPECTRO_DATA` defaulting to `$DESI_ROOT/spectro/data` if not set.

I did *not* scrub `$DESI_SPECTRO_REDUX` from absolutely everywhere in desispec, so I suggest that we continue to set it and use it like before.  I did update two custom path creations with their equivalent `findfile` invocations.

### DIFF
--- a/py/desispec/io/meta.py
+++ b/py/desispec/io/meta.py
@@ -782,7 +782,10 @@ def rawdata_root():
     Raises:
         KeyError: if these environment variables aren't set.
     """
-    return os.environ['DESI_SPECTRO_DATA']
+    if 'DESI_SPECTRO_DATA' in os.environ:
+        return os.environ['DESI_SPECTRO_DATA']
+    else:
+        return os.path.expandvars('$DESI_ROOT/spectro/data')
 
 
 def specprod_root(specprod=None, readonly=False):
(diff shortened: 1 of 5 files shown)


py/desispec/io/meta.py
```diff
@@ -782,7 +782,10 @@ def rawdata_root():
     Raises:
         KeyError: if these environment variables aren't set.
     """
-    return os.environ['DESI_SPECTRO_DATA']
+    if 'DESI_SPECTRO_DATA' in os.environ:
+        return os.environ['DESI_SPECTRO_DATA']
+    else:
+        return os.path.expandvars('$DESI_ROOT/spectro/data')
```
akremin, line 788: Could we make this `os.path.join(os.environ['DESI_ROOT'], 'spectro', 'data')` to be more similar to the other code below?

akremin (commented):
Thanks, the updates look good and bonus points for the changes.rst update and unit tests. I have one quick request inline that should be trivial to change.

PR #2454, opened 2025-03-13
Fix fiber aperture correction: normalization and sky fibers

rongpu:
This PR include two changes to fluxcalib:
1. Excluding sky fibers from the `point_source_correction` normalization calculation; sky fibers are now always assigned `point_source_correction=1`;
2. Changing the normalization calculation from mean to median so that it is robust against outliers.

Before this fix, fibers with very large fiber positioning offsets (mostly used as sky fibers) biases the normalization calculation, causing the `FLAT_TO_PSF_FLUX` value (`point_source_correction` in the code) to be significantly lower than 1 for the "good" fibers (see below the per-petal median `FLAT_TO_PSF_FLUX` value for 100 random tiles). Fortunately, downstream code cancels out this "correction" and the final flux calibration is unbiased for the good fibers. In addition to bringing the `FLAT_TO_PSF_FLUX` value back to a small correction around 1 as originally intended, with this PR we also fix the flux calibration of sky fibers which effectively had arbitrary normalization due to this "aperture correction".

image

I have tested the new code on one blue camera of a dark-time tile, and the changes to fluxcalib is very small (<1%) for all fibers except for sky fibers which now have the reasonable fluxcalib.

### DIFF
--- a/py/desispec/fiberfluxcorr.py
+++ b/py/desispec/fiberfluxcorr.py
@@ -70,11 +70,16 @@ def flat_to_psf_flux_correction(fibermap,exposure_seeing_fwhm=1.1) :
     #  fiber flat correction is larger
     #  have to divide by isotropic_platescale^2
     ok = (fiber_frac>0.01)
+    skyfibers = fibermap["OBJTYPE"]=="SKY"
+    ok &= (~skyfibers)  # also exclude sky fibers from the point_source_correction calculation
     point_source_correction = np.zeros(x_mm.shape)
     point_source_correction[ok] = 1./fiber_frac[ok]/isotropic_platescale[ok]**2
 
-    # normalize to one because this is a relative correction here
-    point_source_correction[ok] /= np.mean(point_source_correction[ok])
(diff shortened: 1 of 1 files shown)


julienguy (approved):
thanks!

rongpu:
`FLAT_TO_PSF_FLUX` (aka `point_source_correction`) came up in the sky spectra discussion today, and (if I heard correctly) Julien mentioned that `FLAT_TO_PSF_FLUX` does not only include the fiber positioning offset correction (`fiber_frac`) but also accounts for the fiber-to-fiber throughput variation. Looking at the code again, I see [in this line](https://github.com/desihub/desispec/blob/20732cccb35b1c8d634e2aeec1d17fb5d2dbda33/py/desispec/fiberfluxcorr.py#L76) that FLAT_TO_PSF_FLUX includes `fiber_frac` and `isotropic_platescale`, but not the throughput correction.

@julienguy Could you take another look and verify that setting `FLAT_TO_PSF_FLUX=1` for sky fibers is indeed the correct thing to do? Should we have keep the `isotropic_platescale` correction for the sky fibers?

PR #2457, opened 2025-03-27
log error when arc/flat set are rejected due to large time diff

sbailey:
This PR logs an error message with details for the case when an arc/flat set are rejected due to having too large a time difference.  I don't recall what night I was debugging when I added this to my local copy, but previously the code could reject all arc/flat sets, without anything in the log indicating why they were rejected.  The change is to add an else clause to this if statement instead of silently moving on:
```python
if np.abs(arcstart-flatend) < max_mjd_dt or np.abs(flatstart-arcend) < max_mjd_dt:
    ...
else:
    log.error(f'arc/flat time diff too large (>{max_mjd_dt}): {np.abs(arcstart-flatend)=}, {np.abs(flatstart-arcend)=}')
```

### DIFF
--- a/py/desispec/workflow/calibration_selection.py
+++ b/py/desispec/workflow/calibration_selection.py
@@ -549,7 +549,7 @@ def find_best_arc_flat_sets(exptable, ngoodarcthreshold=3, nflatlamps=4,
                 else:
                     log.info(f"Found an arc-flat set but with at least one issue.")
             else:
-                log.error(f'arc/flat time diff too large (>{max_mjd_dt}): {np.abs(arcstart-flatend)=}, {np.abs(flatstart-arcend)=}')
+                log.info(f'arc/flat time diff too large (>{max_mjd_dt}): {np.abs(arcstart-flatend)=}, {np.abs(flatstart-arcend)=}')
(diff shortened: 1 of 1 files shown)


akremin (commented):
This will raise an error even if a valid set is found, and could potentially throw several errors since it is inside the loop over arc sets and the loop over flat sets.

I don't think we want to raise errors on nights where we do have a valid combination, even if other combinations aren't valid. If you change this to "log.info", I would be comfortable with it. Otherwise I'd prefer to only raise this error if no combination has a dt below the threshold.

sbailey:
Depending upon what you mean by "raising an error", note that this update was logging an error but not raising an exception.  But point taken that this isn't an "error" per-se (something that isn't supposed to happen), so I changed it to an INFO-level log message, which also accomplishes the goals. The main thing I'm trying to avoid is the previous state where it would print that it found an arc/flat set, silently reject it, and then fail at the end with no arc/flat sets without the logs saying why it didn't use the initial set it had found.

akremin:
Yes I meant logging errors, thank you for changing it to an info level log. This looks good, merging.

PR #2459, opened 2025-03-27
fix read_frame_as_spectra

sbailey:
This PR fixes a bug found by @araichoor where `desispec.io.read_frame_as_spectra` was broken because it was creating a `Spectra` object with `meta` as a `astropy.io.fits.FITSHDR` instead of a `dict`.  This includes a unit test demonstrating the bug, and now it passes with this update.

Note: We do not use `read_frame_as_spectra` in the pipeline anymore thus we hadn't noticed that it was broken (we use `read_frame` and `read_spectra` instead).

### DIFF
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -492,7 +492,7 @@ def read_frame_as_spectra(filename, night=None, expid=None, band=None, single=Fa
         extra = {band : {"CHI2PIX" : fr.chi2pix}}
 
     spec = Spectra(bands, {band : fr.wave}, {band : fr.flux}, {band : fr.ivar},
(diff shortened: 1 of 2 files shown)


araichoor:
thanks.
I confirm that the command that was crashing before now successfully runs with this branch:
```
>>> from desispec.io import read_frame_as_spectra
>>> s = read_frame_as_spectra("/global/cfs/cdirs/desi/spectro/redux/loa/exposures/20240212/00225142/cframe-b0-00225142.fits.gz")
INFO:fibermap.py:339:read_fibermap: iotime 0.067 sec to read cframe-b0-00225142.fits.gz at 2025-03-27T13:57:09.283673
INFO:frame.py:226:read_frame: iotime 0.686 sec to read cframe-b0-00225142.fits.gz at 2025-03-27T13:57:09.296696
```

PR #2462, opened 2025-03-31
DESI_COMPRESSION

julienguy:
Add the possibility to choose the fits files compression mode with a DESI_COMPRESSION environment variable.

- default is 'GZ' (gzip compression), as in current pipeline (for instance if DESI_COMPRESSION is not set).
- possible values are DESI_COMPRESSION=GZ or DESI_COMPRESSION=NONE.

This feature is very useful for hardware tests because compression takes some time and this is sometimes inconvenient when doing real time tests.

### DIFF
--- a/py/desispec/io/meta.py
+++ b/py/desispec/io/meta.py
@@ -147,8 +147,7 @@ def findfile(filetype, night=None, expid=None, camera=None,
         rawdata_dir=None, specprod_dir=None, specprod=None,
         tiles_dir=None, outdir=None,
         download=False, return_exists=False,
-        readonly=False, logfile=False,
-        raise_error_if_alt_exists=False):
+        readonly=False, logfile=False):
     """Returns location where file should be
 
     Args:
@@ -179,7 +178,6 @@ def findfile(filetype, night=None, expid=None, camera=None,
         return_exists: if True, also return whether the file exists
         readonly: if True, return read-only version of path if possible
         logfile: if True, returns the pathname of the log instead of the data product itself
-        raise_error_if_alt_exists: If True, raises IOError if alternative file (.gz or not) exists.
 
     Returns filename, or (filename, exists) if return_exists=True
 
@@ -512,7 +510,7 @@ def findfile(filetype, night=None, expid=None, camera=None,
         filepath = download(filepath, single_thread=True)[0]
 
     try:
-        filepath = checkgzip(filepath,raise_error_if_alt_exists = raise_error_if_alt_exists)
+        filepath = checkgzip(filepath,readonly)
         exists = True
     except FileNotFoundError:
         exists = False
--- a/py/desispec/io/meta.py
+++ b/py/desispec/io/meta.py
@@ -147,7 +147,8 @@ def findfile(filetype, night=None, expid=None, camera=None,
         rawdata_dir=None, specprod_dir=None, specprod=None,
         tiles_dir=None, outdir=None,
         download=False, return_exists=False,
-        readonly=False, logfile=False):
+        readonly=False, logfile=False,
+        raise_error_if_alt_exists=False):
     """Returns location where file should be
 
     Args:
@@ -178,6 +179,7 @@ def findfile(filetype, night=None, expid=None, camera=None,
         return_exists: if True, also return whether the file exists
         readonly: if True, return read-only version of path if possible
         logfile: if True, returns the pathname of the log instead of the data product itself
+        raise_error_if_alt_exists: If True, raises IOError if alternative file (.gz or not) exists.
 
     Returns filename, or (filename, exists) if return_exists=True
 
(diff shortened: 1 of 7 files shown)


julienguy:
Note one feature: in case of a pre-existing output fits file, the data will be rewritten with the same preexisting file name, i.e., with the same compression level irrespectively of the DESI_COMPRESSION value. This behavior is due to the call to the desispec.io.util.checkgzip function in many places in the pipeline.

julienguy:
The code will now raise an IOError if the alternative compressed/uncompressed file exists unless the option readonly=True is set in checkgzip or findfile routines.

julienguy:
I am running `desi_proc_night -n 20250330 --tiles 10795` to see if this affects the pipeline run.

julienguy:
I was able to run a night of calibration + a tile, while changing the compression level half way through. This PR is ready for review.

py/desispec/io/meta.py
```diff
@@ -128,6 +128,18 @@ def get_findfile_argparser():
 
     return parser
 
+
+_valid_compression_types = ['NONE','GZ']
+
+def compression() :
```
sbailey, line 134: Please add docstring

py/desispec/io/meta.py
```diff
@@ -178,6 +190,18 @@ def findfile(filetype, night=None, expid=None, camera=None,
...
     log = get_logger()
+
+    comptype=compression()
+
+    if comptype=="GZ" :
+        compsuffix=".gz"
+    elif comptype=="NONE"  :
+        compsuffix=""
+    else :
+        raise KeyError(f"unknown compression type '{comptype}'")
```
sbailey, line 201: For consideration: how about making `compression` a longer more meaningful name and having it directly return the suffix to use rather than post-facto parsing the `comptype`, e.g. `compsuffix = get_compression_suffix()`?

py/desispec/io/meta.py
```diff
@@ -178,6 +190,18 @@ def findfile(filetype, night=None, expid=None, camera=None,
         exists; otherwise it returns the normal read/write path.
     """
     log = get_logger()
+
```
sbailey, line 193: The behavior of
> The code will now raise an IOError if the alternative compressed/uncompressed file exists unless the option
> readonly=True is set in checkgzip or findfile routines.

is new and potentially surprising, so please add a note about that in the `findfile` docstring, clarifying the relationship and consequences between the readonly option, $DESI_COMPRESSION, and the possible pre-existence of a file.  I think the chain of logic is
```
filepath = canonical location considering $DESI_COMPRESSION
if alternate (non)compressed file exists:
    if readonly:
        ok, use alternate path for filepath
    else:
        raise IOError  # due to pre-existing non-canonical file

return filepath
```

Is that correct and/or can we express it more simply?

py/desispec/preproc.py
```diff
@@ -597,7 +597,7 @@ def get_calibration_image(cfinder, keyword, entry, header=None):
             expid = header['EXPID']
             camera = header['CAMERA'].lower()
             if 'DESI_SPECTRO_REDUX' in os.environ and 'SPECPROD' in os.environ:
-                biasnight = findfile('biasnight', night, expid, camera)
+                biasnight = findfile('biasnight', night, expid, camera, readonly=True)
```
sbailey, line 600: Was this just because it was good to add readonly=True here, or was it actually necessary?

py/desispec/io/meta.py
julienguy: suggestion applied in last commit

py/desispec/io/meta.py
julienguy: suggestion applied in last commit

py/desispec/io/meta.py
julienguy: Yes it's correct. I have added a sentence in the Notes paragraph of findfile docstring,

py/desispec/preproc.py
julienguy: Without this option, the code was crashing because I had started to run the pipeline with DESI_COMPRESSION=NONE and then resumed running with DESI_COMPRESSION=GZ. It's a case where compressed or uncompressed filenames work when the option readonly=True is set. This error would not have occurred if I had not changed the compression level in the middle of the processing of a night!

sbailey (commented):
Functionality looks good.  Please add unit tests that will ensure that it continues to work in the future as we make other changes.  e.g.
* if $DESI_COMPRESSION is not set, default is .gz
* $DESI_COMPRESSION= 'gz', 'none', 'GZ', 'NONE', and 'None' all work as expected
  * Bonus: consider adding 'gzip' as a valid option
* The pre-existence of an alternate file has expected behavior
  * if readonly=True, alternate file is returned
  * if readonly=False, raises error

julienguy:
I added the unit tests

sbailey:
Thanks.  I updated the unit test setUp/tearDown to make sure that $DESI_COMPRESSION isn't set after the test is finished running, to avoid potential side-effects on other tests.  I'll merge after tests re-pass on github (they pass at NERSC).

PR #2464, opened 2025-04-03
Copy purged redux files to attic

rongpu:
This PR adds the option for `desi_purge_night` and `desi_purge_tilenight` to copy files to `$SPECPROD/attic/` before deletion.

Only the last purged version is preserved in attic: if a file already exist in the attic, the purge will overwrite it.

This option is enabled by default, and can be disabled with `--no_attic`.

### DIFF
--- a/py/desispec/scripts/purge_tilenight.py
+++ b/py/desispec/scripts/purge_tilenight.py
@@ -53,8 +53,11 @@ def remove_directory(dirname, dry_run=True, no_attic=False):
         else:
             log.info(f"Removing: {dirname}")
             if not no_attic:
-                attic_dir = dirname.replace(specprod_root(),
-                    os.path.join(specprod_root(), 'attic'))
+                if dirname.startswith('/'):  # absolute path (a la Anthony)
+                    attic_dir = dirname.replace(specprod_root(),
+                        os.path.join(specprod_root(), 'attic'))
+                else:  # relative path (a la Stephen)
+                    attic_dir = os.path.join('attic', dirname)
                 shutil.copytree(dirname, attic_dir, dirs_exist_ok=True)
             shutil.rmtree(dirname)
     else:
--- a/py/desispec/scripts/purge_tilenight.py
+++ b/py/desispec/scripts/purge_tilenight.py
@@ -51,14 +51,15 @@ def remove_directory(dirname, dry_run=True, no_attic=False):
         if dry_run:
             log.info(f"Dry_run set, so not performing any action.")
         else:
-            log.info(f"Removing: {dirname}")
             if not no_attic:
                 if dirname.startswith('/'):  # absolute path (a la Anthony)
                     attic_dir = dirname.replace(specprod_root(),
                         os.path.join(specprod_root(), 'attic'))
                 else:  # relative path (a la Stephen)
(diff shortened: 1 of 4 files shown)


py/desispec/scripts/purge_tilenight.py
```diff
@@ -48,12 +51,20 @@ def remove_directory(dirname, dry_run=True):
...
             log.info(f"Dry_run set, so not performing any action.")
         else:
+            if not no_attic:
+                if dirname.startswith('/'):  # absolute path
+                    attic_dir = dirname.replace(specprod_root(),
+                        os.path.join(specprod_root(), 'attic'))
+                else:  # relative path
+                    attic_dir = os.path.join('attic', dirname)
+                log.info(f"Copying {dirname} to attic")
+                shutil.copytree(dirname, attic_dir, dirs_exist_ok=True)
```
akremin, line 61: We use symlinks at a few points in the pipeline, particularly in the /exposures/ file. Looking at the `copytree` documentation it replaces symlinks with copies of the data by default. However, I don't think using `symlinks=True` will work because it will likely assign those links to the old filepaths that will then be removed with `shutil.rmtree`. The relative links might continue to work, but not for the `fiberflatexp-*` files that link to calibnight (which won't be purged in the case of `purge_tilenight`).

Because we're talking about a small amount of data, I think the code is fine as-is. I just wanted to point this out in case it is relevant in the future.

akremin (commented):
Thank you, Rongpu. The update looks good to me. I put one comment inline, but there is nothing you need to change, it's just documenting a subtlety for our future selves.

How much testing have you done with this? If you haven't done so, I'd like to see a mock prod created and 1-2 `purge_nights` run along with 3-4 `purge_tilenights`.

py/desispec/scripts/purge_tilenight.py
rongpu: Anthony and I talked in person, and we agree that to save disk space we will set `symlinks=True` and not make extra copies of the symlinked files. That would probably make the attic symlinks invalid, but we think that's fine.

rongpu: I have modified the code such that the attic symlinks will have the same absolute path as the original ones. So we should be able to run code in the attic without any issues.

akremin (commented):
Thanks, Rongpu! I think we can remove the new "if" statement you added and just include the new redefinition under the "else" that already existed above.

Once that's cleaned up we should be ready to merge.

py/desispec/scripts/purge_tilenight.py
```diff
@@ -48,12 +51,22 @@ def remove_directory(dirname, dry_run=True):
         if dry_run:
             log.info(f"Dry_run set, so not performing any action.")
         else:
+            if not no_attic:
+                if dirname.startswith('/'):  # absolute path
+                    attic_dir = dirname.replace(specprod_root(),
+                        os.path.join(specprod_root(), 'attic'))
+                else:  # relative path
+                    attic_dir = os.path.join('attic', dirname)
+                if not dirname.startswith('/'):  # use absolute path for symlinks
```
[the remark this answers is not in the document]
rongpu: Thanks! It's fixed.

akremin:
Thanks, merging now.

PR #2466, opened 2025-04-09
Added support to return redrock models in read_spectra

abhi0395:
### Summary

Since **kibo**, we have added support in **redrock** to write out best-fit model files (`rrmodel-?*.fits`) alongside the redshift fits. These files are now available on **NERSC** along with the Redrock redshift outputs.

### What's New

This pull request adds support to `desispec.io.spectra.read_spectra()` to optionally **return the Redrock model fits**, if the corresponding `rrmodel` file exists on disk for the given input spectra or directly from the `infile` if model exists in it. This adds a <1 s runtime to the existing code (if the model needs to be returned). Also added support to return **REDSHIFTS** table if user wants it either from `redrock-*.fits` file of from the `infile` if it exists

This simple addition will make it easier for users to directly access the **best-fit redrock models** without having to build it using templates and coefficients and **REDSHIFTS** table for those targets

### Notes

- No changes are required to existing workflows unless users want to take advantage of this feature.
- If `rrmodel` files are missing and `MODEL` HDU is not in `infile`, the function will fail citing `IOError`.
- Added tests in `test_spectra.test_slice` and `test_spectra.test_stack` for the model and redshift part.
- Also added optional model and redshift table write option in `desispec.io.spectra.write_spectra`

### Tests:

```
from desispec.io.spectra import read_spectra, write_spectra
coadd_file=f'/global/cfs/cdirs/desi/spectro/redux/loa/tiles/cumulative/8329/20240116/coadd-0-8329-thru20240116.fits'
spectra = read_spectra(coadd_file, return_models=True)
spectra.model
spectra = read_spectra(coadd_file, return_models=True, return_redshifts=True)
spectra.redshifts

## now write them in an output file:
write_spectra(output, spectra)
## if want to read this new output file:
new_spectra = read_spectra(output, return_models=True, return_redshifts=True)
```

It can also be used with `desispec.io.spectra.read_spectra_parallel()` as:

```
from desispec.io.spectra import read_spectra_parallel
tt = Table.read(f'{table_containing_some_target}')
tt = tt[np.argsort(tt["HEALPIX"])] # just to sort such that mpi doesn't have to work too much
spectra = read_spectra_parallel(tt, specprod='kibo', rdspec_kwargs={'return_models':True, 'return_redshifts':True})
spectra.model
spectra.redshifts
```

### Other important note:

- It also resolves `test_io_fibermap.py`, which was failing because `readonly=False` in the `io.fibermap.assemble_fibermap`:
```
rawfafile, exists = findfile('fiberassign', night=night, expid=expid,
tile=tileid, return_exists=True, readonly=True) # corrected script
```

### DIFF
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -221,31 +221,31 @@ def read_spectra(
     """
     Read Spectra object from FITS file.
 
-    This reads data written by the write_spectra function.  A new Spectra
+    This reads data written by the write_spectra function. A new Spectra
     object is instantiated and returned.
-
+    
     Args:
-        infile (str): path to read
-        single (bool): if True, keep spectra as single precision in memory.
-        targetids (list): Optional, list of targetids to read from file, if present.
-        rows (list): Optional, list of rows to read from file
-        skip_hdus (list): Optional, list/set/tuple of HDUs to skip
-        return_models (bool): Optional, if True will also read best-fit redrock models (default False)
-                        (loads directly from infile if _MODEL HDUs exist; otherwise, will look for corresponding rrmodel-*.fits file, otherwise raise IOError)
-        return_redshifts (bool): Optional, if True will also read redshift table for targets (default False)
-                        (loads directly from infile if REDSHIFTS HDU exists; otherwise, will look for corresponding redrock-*.fits file, otherwise raise IOError)
-        select_columns (dict): Optional, dictionary to select column names to be read. Default, all columns are read.
-
-    Returns (Spectra):
-        The object containing the data read from disk.
-
-    `skip_hdus` options are FIBERMAP, EXP_FIBERMAP, SCORES, EXTRA_CATALOG, MASK, RESOLUTION;
-    where MASK and RESOLUTION mean to skip those for all cameras.
-    Note that WAVE, FLUX, and IVAR are always required.
-
-    If a table HDU is not listed in `select_columns`, all of its columns will be read
-
-    User can optionally specify targetids OR rows, but not both
+        infile (str): Path to read.
+        single (bool): If True, keep spectra as single precision in memory.
+        targetids (list, optional): List of targetids to read from file, if present.
+        rows (list, optional): List of rows to read from file.
+        skip_hdus (list, optional): List/set/tuple of HDUs to skip.
+            Options: ``FIBERMAP``, ``EXP_FIBERMAP``, ``SCORES``, ``EXTRA_CATALOG``,
+            ``MASK``, ``RESOLUTION``. ``WAVE``, ``FLUX``, and ``IVAR`` are always required.
+        return_models (bool, optional): If True, also read best-fit redrock models.
+            Loads directly from ``infile`` if ``*_MODEL`` HDUs exist; otherwise,
+            looks for a corresponding ``rrmodel-*.fits`` file. Raises IOError if neither found.
+        return_redshifts (bool, optional): If True, also read redshift table for targets.
+            Loads directly from ``infile`` if ``REDSHIFTS`` HDU exists; otherwise,
+            looks for corresponding ``redrock-*.fits`` file. Raises IOError if neither found.
+        select_columns (dict, optional): Dictionary mapping HDU names to column names to read.
+            If a table HDU is not listed, all of its columns will be read.
+    
+    Returns:
+        Spectra: The object containing the data read from disk.
+    
+    Notes:
+        User can optionally specify either ``targetids`` or ``rows``, but not both.
     """
     log = get_logger()
     infile = checkgzip(infile)
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -196,7 +196,7 @@ def read_spectra(
     targetids=None,
     rows=None,
     skip_hdus=None,
-    rrmodel=False,
+    return_models=False,
     select_columns={
         "FIBERMAP": None,
         "EXP_FIBERMAP": None,
@@ -216,7 +216,7 @@ def read_spectra(
         targetids (list): Optional, list of targetids to read from file, if present.
         rows (list): Optional, list of rows to read from file
         skip_hdus (list): Optional, list/set/tuple of HDUs to skip
-        rrmodel (bool): Optional, if True will alo read best-fit redrock models (default False)
+        return_models (bool): Optional, if True will alo read best-fit redrock models (default False)
                         Also, it assumes that model is saved as rrmodel-??, in similar way ad coadd
         select_columns (dict): Optional, dictionary to select column names to be read. Default, all columns are read.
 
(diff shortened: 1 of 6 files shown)


py/desispec/io/spectra.py
```diff
@@ -215,6 +216,8 @@ def read_spectra(
         targetids (list): Optional, list of targetids to read from file, if present.
         rows (list): Optional, list of rows to read from file
         skip_hdus (list): Optional, list/set/tuple of HDUs to skip
+        return_models (bool): Optional, if True will alo read best-fit redrock models (default False)
+                        Also, it assumes that model is saved as rrmodel-??, in similar way ad coadd
```
sbailey, line 220: minor typos: alo -> also; ad -> as

py/desispec/io/spectra.py
```diff
@@ -389,16 +393,40 @@ def read_spectra(
     hdus.close()
     duration = time.time() - t0
     log.info(iotime.format("read", infile, duration))
-
+    
+    model_targetids = None # for the sanity checks in the end
+    if return_models:
+        rrmodel_file = infile.replace('coadd','rrmodel')
```
sbailey, line 399: Please use `desispec.io.util.replace_prefix(infile, 'coadd', 'rrmodel')`.  That will replace "coadd" only in the prefix of the filename part of infile, but not wherever else it might appear in the path (e.g. if $SPECPROD=testcoadd, the string "coadd" will appear twice in `infile`, but you only want to replace one of them).

py/desispec/io/spectra.py
```diff
@@ -389,16 +393,40 @@ def read_spectra(
...
+                if "MODEL" in hdu_name:
+                    ind_models.append(k)
+            for ind, band in zip(ind_models,bands):
+                log.debug('Reading %s_MODEL'%(band.upper()))
+                model[band] = _read_image(rrhdus, ind, np.float32, rows=rows)
+            rrhdus.close()
+            duration = time.time() - t0
+            log.info(iotime.format("read", rrmodel_file, duration))
+        else:
+            print(f'ERROR: {rrmodel_file} does not exist, so not returning redrock models')
```
sbailey, line 419: I suggest that this should be a fatal error: if the user requests `return_models=True` and there are no models to return, let's raise an exception rather than just printing an error message and not returning any models.  Anything printed here should use `log.critical` instead of `print`.

Also, if `return_models` is True, consider checking for the existence of the model file earlier in the code so that it would crash before reading everything else, rather than spending severals seconds reading the spectra and then crashing at the end due to a missing model.

py/desispec/io/spectra.py
```diff
@@ -416,20 +444,23 @@ def read_spectra(
...
+            rows = np.concatenate([np.where(input_targetids == tid)[0] for tid in found_targetids])
             log.debug("spec.fibermap['TARGETID'] = %s", np.asarray(spec.fibermap['TARGETID']))
             log.debug("rows for subselection=%s", rows)
             spec = spec[rows]
-
+        #consistency check between targetids (perhaps this is not necessary)
+        if model_targetids is not None and not np.all(input_targetids == model_targetids):
+            rows = np.concatenate([np.where(input_targetids == tid)[0] for tid in model_targetids])
+            spec = spec[rows]
+            log.debug("ordering corrected matching model targetids=%s", rows)
```
sbailey, line 463: This is fixing some other bug that you mentioned to me, but I forget exactly what case would trigger it.  Please document in this PR what input case was previously wrong but now right.  If reasonably possible, let's add a unit test that would have demonstrated the original problem and now passes, to make sure we don't break it again.

P.S. good use of `debug.log('message %s', value)` so that the string formatting is only calculated if the debug message would be printed.

py/desispec/spectra.py
```diff
@@ -45,6 +45,8 @@ class Spectra(object):
         Dictionary of arrays specifying the flux for each spectrum.
     ivar : :class:`dict`
         Dictionary of arrays specifying the inverse variance.
+    model : :class:`dict`, optional
+        Dictionary of arrays specifying the best-fit redrock model.
```
sbailey, line 49: Minor: Let's drop "redrock" from this description (and in another comment below), since in principle the model given to the Spectrum object could be from any other source, e.g. fastspecfit too.  In `read_spectra` it is specifically looking for the Redrock model, but here it  is more generic.

py/desispec/spectra.py
```diff
@@ -979,7 +1028,7 @@ def stack(speclist):
         _remove_tile_keywords(headers)
 
     sp = Spectra(bands, wave, flux, ivar,
-        mask=mask, resolution_data=rdat,
+        mask=mask, model=model, resolution_data=rdat,
```
sbailey, line 1031: I think you found all the cases where spectral slicing and stacking needs to manipulate the models just like it does for flux, ivar, etc., but please add checks for that to `desispec.test.test_spectra` in the `test_slice` and `test_stack` functions.

sbailey (changes requested):
Thanks!  This is definitely useful.  I put a few small suggestions inline, and one more substantive suggestion that if the user requests `return_models=True` but there are no models, then it should raise an exception rather than printing an error message and returning no models.  If you disagree, please explain why you think it is better to not crash.

sbailey:
And unit tests caught a case of "UnboundLocalError: local variable 'model' referenced before assignment".

py/desispec/io/spectra.py
abhi0395: Done!

py/desispec/io/spectra.py
abhi0395: Thanks, Done!

py/desispec/io/spectra.py
abhi0395: I moved the checking if file exists logic near the infile check in the beginning, and now raise the IOError, thanks for the suggestion! I think, I don't need log.critical here in this case.

py/desispec/io/spectra.py
abhi0395: This is actually just to check if the targetids of the models match the input targetids. This check only happened if targetids is not None, like if we want to read spectra for only user-provided targetids, in that case order of input and output targetids might be different.

py/desispec/spectra.py
abhi0395: Done!

py/desispec/io/spectra.py
abhi0395: The problem that I talked about, was happening in the `desispec.coaddition.coadd_cameras` which fails if there are duplicate targetids in the spectra.fibermap["TARGETID"]. In my view, we should just add the docstring explaining that, because otherwise, it's too much work to resolve that.

py/desispec/spectra.py
```diff
@@ -76,7 +78,7 @@ class Spectra(object):
     wavelength_unit = Unit('Angstrom')
     flux_density_unit = Unit('10-17 erg cm-2 s-1 AA-1')
 
-    def __init__(self, bands=[], wave={}, flux={}, ivar={}, mask=None,
+    def __init__(self, bands=[], wave={}, flux={}, ivar={}, model=None, mask=None,
             resolution_data=None, fibermap=None, exp_fibermap=None,
             meta=None, extra=None,
```
sbailey, line 83: Minor (maybe) : Please move the new `model` option to be after the current `extra` option, just in case there is any code that assumes that `mask` is the positional argument right after `ivar`.

abhi0395: thanks! done

sbailey (approved):
Thanks for the updates.  Looks good.

PR #2468, opened 2025-04-17
Have desi_compute_nightly_bias gracefully exit if default bias is missing

akremin:
Currently `desi_compute_nightly_bias` crashes with a KeyError if `calibfinder` can't find a valid BIAS calibration file for the current camera configuration. This updates logs an informative error message and skips that camera, allows MPI ranks to complete the remaining jobs, and returns an error code with the number of failed cameras.

I tested this with `desi_compute_nightly_bias -n 20250415 -c a01` which crashes in daily but exits with error code in this branch:

main:
```
ERROR:calibfinder.py:573:find_darks_in_desi_spectro_dark: Didn't find matching z1 calibration darks in $DESI_SPECTRO_DARK, falling back to $DESI_SPECTRO_CALIB
Traceback (most recent call last):
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/main/bin/desi_compute_nightly_bias", line 16, in <module>
    main()
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/main/py/desispec/scripts/nightly_bias.py", line 48, in main
    compute_nightly_bias(**args.__dict__, comm=comm)
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/main/py/desispec/ccdcalib.py", line 677, in compute_nightly_bias
    defaultbias = cf.findfile('BIAS')
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/main/py/desispec/calibfinder.py", line 398, in findfile
    return os.path.join(self.directory,self.value(key))
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/main/py/desispec/calibfinder.py", line 384, in value
    val=self.data[key]
KeyError: 'BIAS'
```

this branch:
```
INFO:ccdcalib.py:657:compute_nightly_bias: /global/cfs/cdirs/desi/spectro/redux/kremin/calibnight/20250415/biasnighttest-z1-20250415.fits.gz already exists; skipping
ERROR:calibfinder.py:573:find_darks_in_desi_spectro_dark: Didn't find matching z1 calibration darks in $DESI_SPECTRO_DARK, falling back to $DESI_SPECTRO_CALIB
ERROR:ccdcalib.py:681:compute_nightly_bias: rank=0, camera='z1' raised <class 'KeyError'> exception 'BIAS'
[...]

PR #2470, opened 2025-04-18
sys.exit in script for nightly_bias not in function

akremin:
Patch to PR #2468 . In that we have `sys.exit()` in the function instead of the script in `bin`. The wrapper `main()` script was being called by another function within python and causing that other function to crash. Now we only `sys.exit()` if we call the command line script directly.


sbailey (approved):
Good catch.

PR #2471, opened 2025-04-22
Numpy2 support, round 1

sbailey:
This PR makes some necessary changes for numpy 2.  It doesn't include everything needed for numpy 2 but it is a start, covering
* `np.in1d` -> `np.isin`
* `np.NaN` -> `np.nan`
* `np.linalg.linalg.LinAlgError` -> `np.linalg.LinAlgError`
* `np.dtype('bool8')` no longer exists
* difference in how to change from non-native endian to native endian

Other than possible typos, only the last one is somewhat subtle, but the implemented method follows the recommendation from the error message generated by the previous method, and the `test_native_endian` test in `desispec.test.test_io` covers that it works for both native and non-native endian cases.

Because desispec needs desihub/desiutil#218 (maskbit support) to be resolved to work with numpy, I have *not* added numpy 2.x to the test matrix, but the existing test matrix should help verify that all changes are backwards compatible with numpy 1.x.

### DIFF
--- a/py/desispec/scripts/trace_shifts.py
+++ b/py/desispec/scripts/trace_shifts.py
@@ -7,7 +7,7 @@ desispec.scripts.trace_shifts
 import os, sys
 import argparse
 import numpy as np
-from numpy.linalg.linalg import LinAlgError
+from numpy.linalg import LinAlgError
 import astropy.io.fits as pyfits
 from numpy.polynomial.legendre import legval,legfit
 from importlib import resources
--- a/bin/copyprod
+++ b/bin/copyprod
@@ -151,7 +151,7 @@ for night in np.unique(explist['NIGHT']):
     os.makedirs(os.path.dirname(outfile), exist_ok=True)
     if opts.filter_exptables:
         exptab = load_table(infile, tabletype='exptable')
-        ignore = np.in1d(exptab['EXPID'], explist['EXPID'], invert=True)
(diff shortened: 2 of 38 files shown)


py/desispec/scripts/trace_shifts.py
```diff
@@ -7,7 +7,7 @@
 import os, sys
 import argparse
 import numpy as np
-from numpy.linalg.linalg import LinAlgError
+from numpy.linalg import LinAlgError
```
weaverba137, line 10: Stylistic question here: in `fluxcalibration.py`, the exception catches `numpy.linalg.LinAlgError` and presumably in this file it's just `catch LinAlgError`. Both work, but do we want to prefer one style over the other?

weaverba137: See also `sky.py`.

weaverba137 (approved):
I think the code changes are fine. Just one comment about style when catching numpy exceptions.

py/desispec/scripts/trace_shifts.py
sbailey: For custom exceptions that aren't part of the Python standard namespace, I personally prefer using the longer fully qualified path `np.linalg.LinAlgError` at the point of catching the exception to make it more obvious where it came from without having to scroll back to the top to read the imports. But I don't mind the other form either. For the purposes of this PR, I tried to make only the minimal changes needed to support numpy 2 without delving into stylistic cleanup that wasn't strictly necessary.

PR #2477, opened 2025-04-28
Revert "Update nightly_bias.py to use 50 zeros by default"

akremin:
Reverts desihub/desispec#2476


akremin:
Upon discussion with Anand, we'll revert to using 25 zeros for a few more nights until we finish the cosmos observations. Since we used 25 zeros with the observations yesterday, this will give us a consistent dataset where all nights were processed with 25 zeros.
Then we can create a custom prod where we redo these tiles with 50 exposures. This will allow us to compare the two approaches.

PR #2490, opened 2025-05-19
Nightqa: handle bright1b/dark1b

araichoor:
This PR should address the small required updates to handle the bright1b/dark1b programs.
Modifications only are for:
- the `skyzfiber-NIGHT.png` plot;
- the `petalnz-NIGHT.pdf` plot.

My test bed is `20250506`, where we observed three dark1b tiles.

For the `skyzfiber-NIGHT.png` plot:
I was lazy and did a simple tweak to the code always "merge" the "1b" cases with the "normal" ones; if backup+backup1b, bright+bright1b, dark+dark1b (even if there will never be a backup1b program).
Here is the plot with current main branch:
skyzfiber-20250506
and here s with this PR branch (where the three dark1b tiles are added on the right panel):
skyzfiber-20250506-1b
Couple of remarks:
- the "ALL" case does not change: looking at the code, I think this "ALL" case is really all tiles, so ok;
- the transparency drops for the new version for the right panel: it s just that the code switches to higher transparency when there are more than 5000 fibers, which happens here when the three dark1b tiles are added.

For the `petalnz-NIGHT.pdf` plot:
I ve done the following:
- as for above, merge bright+bright1b and dark+dark1b tiles;
- added the `LGE` tracer for the dark case (so that also adds a row for the n(z) panels);
- newlya plot: extended the x-axis (previous tile coverage) from 6.5 to 8.5.
Note that there s an underlying "formula" giving frac_newlya = f(n_previous_tile_coverage), based on real data (https://github.com/desihub/desispec/pull/1780); but the coefficients are close to zero for passes=5 and 6; so I ve just took zero values for passes=7 and 8 (ie no code change); I *think* it is ok to assume that dark1b tile regions with more than 7 previous passes will have a negligible new lya counting.

Here s the plot with the current main branch:
[petalnz-20250506.pdf](https://github.com/user-attachments/files/20309321/petalnz-20250506.pdf)
and here s with this PR branch:
[petalnz-20250506-1b.pdf](https://github.com/user-attachments/files/20309357/petalnz-20250506-1b.pdf)

And, for what is worth, here s an example for `20250507`, where no dark1b tile has been observed (hence an empty row for the `LGE` n(z)):
[petalnz-20250507.pdf](https://github.com/user-attachments/files/20309482/petalnz-20250507.pdf)

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -1560,9 +1560,10 @@ def plot_newlya(
         * The plotted y-values are: (N_newlya_observed / N_newlya_expected) - 1.
         * The expected numbers are based on all main dark tiles (from daily) up to May 26th 2022.
         * The 1-2-3-sigma regions reflect the approximate scatter of those data.
+        * For the new Lya plot, we merge dark and dark1b tiles.
     """
     #
-    n_dark_passids = 7
+    n_dark_passids = 9
     if ntilecovs.shape[1] != n_dark_passids:
         msg = "ntilecovs.shape[1] = {} is different than n_dark_passids = {}".format(
             ntilecovs.shape[1], n_dark_passids,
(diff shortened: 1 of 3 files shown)


sbailey:
The changes in this PR look good.  The tests were broken due to an unrelated commit directly to main to fix the pipeline on Friday (new hardware functionality made a "this is an illegal config" test no longer correct).  I fixed the tests (at NERSC at least...).  Will wait to confirm before merging.

Issue #2494, opened 2025-05-27
Calibration errors are overestimated


julienguy:
- The calibration errors are over estimated because in the first version of the algorithm, I added a large error floor of 20% of the star flux for the outlier rejection stage (in order to avoid removing signal for bright stars).
- Now this extra variance is propagated (by mistake) to the final uncertainty reported in the inverse variance of the calibration vectors (hdu IVAR of the files fluxcalib-CAMERA-EXPID.fits.gz).
- This is results in uncertainties of about 0.2/sqrt(10) = 0.063 in the calibration vector error.
- This was not a problem before we fixed the propagation of the calibration error to the data inverse variance (see PR #2422).
- It results in an overestimated noise for bright objects; with an artificial saturation of the SNR to ~15.8 .
- This causes issues for the Lyman-alpha P1D analysis that relies on a precision measurement of the noise.

This is a bug that has to be fixed.

julienguy:
We need to keep the error floor when masking outliers but have to restore the original flux variance when estimating the final calibration uncertainties.

julienguy:
Fixed in PR #2496

PR #2496, opened 2025-05-29
Improve calibration uncertainty

julienguy:
This PR fixes issue #2494 .

An error floor was added to the uncertainties of the standard star spectra when computing the calibration vectors. This was to reduce the weight of the brightest stars (whose models are uncertain). It was also used to avoid masking pixels values with small uncertainties in the outlier rejection loop. The issue is that we left this error floor when computing the final calibration uncertainties, artificially inflating the calibration vector variance.
We fix this issue in this PR. After the fit, without altering the calibration vector values, we recompute the calibration vector variance using the original uncertainties in the spectra.

Because the calibration uncertainties are propagated to the uncertainties of the targets calibrated spectra, those were also overestimated. This mostly affect the brightest objects (stars). 
I checked that the rms of (data-model)/error is now close to 1 for the brightest star of exposure 20250526/00295628 camera b0.

### DIFF
--- a/py/desispec/fluxcalibration.py
+++ b/py/desispec/fluxcalibration.py
@@ -1295,7 +1295,24 @@ def compute_flux_calibration(frame, input_model_wave, input_model_flux,
     log.info("{} n flux values excluded = {}".format(camera, nout_tot))
 
     # solve once again to get deconvolved variance
-    #calibration,calibcovar=cholesky_solve_and_invert(A.todense(),B)
+    # by we now want to remove the large error floor that was set
+    # at the begining to remove outliers
+    # so we restart from the original ivar, but keep the information about the masked pixels
+    current_ivar=stdstars.ivar*(current_ivar>0)
+    sqrtw=np.sqrt(current_ivar)
+    for star in range(nstds) :
+        if badfiber[star] : continue
+        R = stdstars.R[star]
+        # diagonal sparse matrix with content = sqrt(ivar)*flat
+        D1.setdiag(sqrtw[star]*scale[star])
+        D2.setdiag(model_flux[star])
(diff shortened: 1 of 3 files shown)


sbailey:
Independent cross checks:

Comparing the chi2 of stdstar cframe flux and ivar to the fitted stdstar models, a significant contribution for some stars is due to broad-brand systematic offsets/colors rather than the small scale noise captured by the ivar, e.g. from night 20250526 expid 295628:
image

Since that broadband offset isn't the topic of this PR, I absorb that with a multiplicative offset+slope correction.  After doing that, the sigma of the (flux-model)/err distribution goes from 0.548 (daily, significantly overestimating the stdstar noise) to 1.026 (sjbailey test prod with this branch, much better error estimation).

Comparing the S/N for this PR vs. current main on night 20250526 expid 295628:
image
i.e. daily saturated at S/N ~15.  This saturation was known in issue #2420 / PR #2422 when we started propagating the calibration error, but at the time that was interpreted as a massive bug in the propagation of errors rather than a bug in the estimation of the original calibration errors.

Fortunately, that change was introduced after DR2/loa.  Comparing the S/N of this PR with Loa for night 20230914, the situation isn't quite so scary:
image
i.e. loa slightly underestimates the error, but not wildly.

Summarizing:
* DR2/loa: didn't propagate flux calibration errors, resulting in small underestimation of the true error for bright sources
* daily since 20241204: propagated flux calibration errors, but the input flux calibration error was significantly overestimated, causing the final S/N to incorrectly saturate at ~15.
* this PR: propagates more correct flux calibration errors.

Time's up for today; will plan to merge tomorrow.

sbailey:
Further notes: I'm not thrilled about the replicated code for the final fit, but the fits use enough intermediate variables that it would be somewhat messy to extract into a single fitting function that could be used for both the iterative loop and the final fit.  Instead, I just added some comments in each section noting the replicated code, in the hopes that if we have to update one of them in the future it will be a reminder to update the other too.  I'll merge after tests pass.

PR #2499, opened 2025-06-14
Allow 1200s dark for bad col selection

akremin:
#### Overview
This updates the pipeline to be able to handle the new 1200s "calib dark 20 min" taken each afternoon, while maintaining the ability to use the old 300s "calib dark 5 min" for bad column selection. Both the an `if` statement in in desi_proc needed to be updated. Since we are now allowing 300s or 1200s darks, I added an addition feature in the calibration selection that tries to identify calibration darks specifically, though if it fails it will pass on any 300s or 1200s dark. 

This also edits `/scripts/inspect_dark.py` to not scale the bad column selection threshold by exposure time, as I couldn't find any clear evidence that bright columns scaled with exposure time and testing was leading to fewer identified columns than the 300s dark on a given night. I did several studies to try and identify the correct scaling with exposure time, but the tests were inconclusive. There is no harm is keeping the reduced threshold in my tests, and we still maintain the same significance for keeping a detected column. 

The constant chosen is the same as what was used previously for the 300s dark, 0.005*300 = 1.5. 

#### Testing
Tests were performed in `/global/cfs/cdirs/desi/spectro/redux/test_badcol`.

I first ran the badcolumn finder on several nights using the afternoon 300s dark and the morning 1200s dark with varying threshold settings to determine if I could find a trend. What I found was that scaling with exposure time led to reduced column identification, which visually were present, so I removed the exposure time scaling.

I also ran badcolumn selection on multiple dark sequences taken on 20250602, which had multiple exposure times taken sequentially, hoping to identify the scaling of counts for bright columns with exposure time. The results were inconclusive, leading me to keep the uniform threshold.

I also tested the implementation to make sure that it ran on both old and new data. I tested 20201214, 20210514, 20250608, and 20250610, among others. They all worked as expected.

I will self merge so that this can be used over the weekend to rerun the nights from this week.

### DIFF
--- a/py/desispec/scripts/inspect_dark.py
+++ b/py/desispec/scripts/inspect_dark.py
@@ -27,7 +27,8 @@ def parse(options=None):
     parser.add_argument('--badfiber-table', type = str, default = None, required = False, help = 'output table with list of bad fibers (because of bad columns)')
     parser.add_argument('--badcol-table', type = str, default = None, required = False, help = 'output table with list of bad columns')
     parser.add_argument('--plot', action = 'store_true', help = 'plot the profiles')
-    parser.add_argument('--threshold', type = float , default = 0.005, required = False, help = 'threshold in electrons/sec to flag columns.')
+    parser.add_argument('--threshold', type = float , default = 0.005, required = False, help = 'threshold in electrons/sec to flag columns. '
+                                                                                                + 'Note not exptime dependent. Always multiplied by 300.')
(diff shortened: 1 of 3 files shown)


sbailey:
Note issue #2501 documenting the statement here that the bad column flux didn't vary with exposure time.  Apparently this is because the bias is varying rather than unmodeled dark current.  Let's continue any further discussion of that in the issue rather than here.

Issue #2498, opened 2025-06-10
Set correct PROGRAM/GOALTYPE value in exposures-daily.csv for DARK1B/BRIGHT1B exposures


araichoor:
we should decide what `PROGRAM` and `GOALTYPE` values the main dark1b (and probably bright1b) exposures have in `exposures-daily.csv`.

so far they have `PROGRAM=GOALTYPE="other"`:
```
>>> d = Table.read("/global/cfs/cdirs/desi/spectro/redux/daily/exposures-daily.csv")
>>> d = d[d["FAFLAVOR"]=="maindark1b"]
>>> d["FAFLAVOR","SURVEY","PROGRAM","GOALTYPE"]
<Table length=11>
 FAFLAVOR  SURVEY PROGRAM GOALTYPE
  str19     str7    str6    str6  
---------- ------ ------- --------
maindark1b   main   other    other
maindark1b   main   other    other
maindark1b   main   other    other
maindark1b   main   other    other
maindark1b   main   other    other
maindark1b   main   other    other
maindark1b   main   other    other
maindark1b   main   other    other
maindark1b   main   other    other
maindark1b   main   other    other
maindark1b   main   other    other
```

at least for the `PROGRAM`, I think this piece of code should be modified:
https://github.com/desihub/desispec/blob/7999ffe7fca907e80155172d1afb93fad398e99a/py/desispec/io/meta.py#L877-L914

araichoor:
I m happy to give a shot at a pr solving that (I m not that familiar with that script), but I d first need to know what do we want.

I would say we want, `PROGRAM = GOALTYPE = "dark"` for `dark1b` and `PROGRAM = GOALTYPE = "bright"` for `bright1b`.

(`GOALTYPE` is set in fiberassign that way: https://github.com/desihub/fiberassign/blob/8788ea5d98bd03b710988e92b87d4b90ad61813e/bin/fba_launch#L814-L816)

is that correct?

sbailey:
Thanks.  I think we want PROGRAM="dark1b" and GOALTYPE="dark" (and similarly for bright1b / bright).  This matches the FAPRGRM and GOALTYPE keywords of the original fiberassign files, and the PROGRAM keyword of the desi*.fits.gz raw data files (which gets its PROGRAM from NTS).

@akremin heads up: after we get this fixed for newly processed tiles, we'll have some surgery to do to fix the exposures and tiles files for past data.

sbailey:
[...]

PR #2511, opened 2025-06-30
Goaltype and program in exposures file for dark1b/bright1b tiles

araichoor:
This PR tries to address https://github.com/desihub/desispec/issues/2498.

Copying here the goal: "_we want PROGRAM="dark1b" and GOALTYPE="dark" (and similarly for bright1b / bright). This matches the FAPRGRM and GOALTYPE keywords of the original fiberassign files, and the PROGRAM keyword of the desi*.fits.gz raw data files (which gets its PROGRAM from NTS)._"

I am not familiar with all the possibly involved pieces of code here, and I don t know how to test that.
so it would be great if someone from the data team could take over.

the couple of commits I ve made should hopefully correctly set:
- the `PROGRAM` column value (in the `faflavor2program()` function);
- the `GOALTYPE` column value (in the `desi_tsnr_afterburner` script).

maybe that s all that is required? I admit I don t really know...
and I ve zero idea of how to test that!

### DIFF
--- a/bin/desi_tsnr_afterburner
+++ b/bin/desi_tsnr_afterburner
@@ -170,6 +170,9 @@ def update_targ_info(entry, targ_in):
     if entry['GOALTYPE'] in ('unknown', 'other'):
         entry['GOALTYPE'] = faflavor2program(entry['FAFLAVOR'])
 
+    if entry['GOALTYPE'] in ['dark1b', 'bright1b']:
+        entry['GOALTYPE'] = entry['GOALTYPE'].replace('1b', '')
+
     return entry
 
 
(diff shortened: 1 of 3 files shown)


sbailey:
Thanks.  I expanded our existing `faflavor2program` tests to explicitly check dark, dark1b, bright, bright1b which cause the problem that the PROGRAM column was dtype='U6', which truncated 'bright1b' back to 'bright'.  I fixed that.  It looks like the GOALTYPE code doesn't have that problem, but we don't have a clean unit test for that either.  I'm pretty confident that it will work, so I'll merge when the tests finish, then we can test with a real run of desi_tsnr_afterburner, and I'll take responsibility for anything that breaks if that was premature.

araichoor:
thanks @sbailey for taking over!
(btw I ve changed the pr title into something more meaningful than my branch name).

Issue #2047, opened 2023-05-22
Spectrophotometry of DESI standard stars


rongpu:
Hi @dkirkby @julienguy, there is a discrepancy in the spectrophotometry of DESI standard stars from speclite (which desispec uses) and from [Julien's magnitude.py script](https://github.com/desihub/desispec/blob/main/py/desispec/magnitude.py). The largest difference is ~12 mmag in g band, and ~2 mmag in r band. The comparison is done [in this notebook](https://github.com/rongpu/desi-examples/blob/master/misc/spectrophotometry/compare_speclite_with_julien_magnitude.ipynb).

To determine which calculation is correct, I did a simple experiment in which the filter curves and the spectrum are completely flat. In this simple case, the two magnitude calculations agree within 10 mmag. But when I set the redder half of the spectrum to zero (so the total flux sums up to 50% of the original flux), the magnitude.py script gives the expected 50% of the original flux, but speclite gives 47.2%. Similarly, if I set the bluer half to zero, magnitude.py gives 50% while speclite gives 52.8%. This experiment is done [in this notebook](https://github.com/rongpu/desi-examples/blob/master/misc/spectrophotometry/compare_speclite_with_julien_magnitude-experiment.ipynb). So unless I messed up somewhere, I think there is a bug in the speclite spectrophotometry calculation.

PR #2515, opened 2025-07-15
Update magnitude.py

rongpu:
This PR fixes a bug in the AB magnitude calculation where the multiplicative wavelength term is missing from the integral. It addresses https://github.com/desihub/desispec/issues/2047.

### DIFF
--- a/py/desispec/magnitude.py
+++ b/py/desispec/magnitude.py
@@ -7,20 +7,20 @@ Broadband flux and magnitudes.
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


rongpu:
I still think the `wave` term should be inside `compute_broadband_flux`. That is simply the definition of broadband flux in the AB system. When we compare the spectrophotometry with the fluxes in broadband photometry (e.g., in [Legacy Imaging](https://www.legacysurvey.org/dr9/files/#sweep-brickmin-brickmax-fits)), that is exactly what we would compute.

PR #2516, opened 2025-07-15
Limit input exposures for combined dark creation

jbanovetz:
Here is my addition to make sure we only limit the number of dark exposures that go into a combined dark (default is 50):

I added a flag '--max-exposures' to the compute_dark.py that will set an upper bound on the number of dark exposures used to create a combined dark. For this, I also changed the compute_dark.py a little including adding another for loop to retrieve the correct images. This was because there was an issue in only picking from files that can easily be grabbed. I also used a new function 'closest_exposures' which is described below. I tested this with a dry run and it was able to grab the correct files.

I added a function to util.py called 'closest_exposures' which just grabs an exposure table and uses the reference night to sort the exposure table by the reference night and then cut it to the first X number of entries (where X is the max number of exposures). This uses astropy.time.Time in order to do this and I added 1.5 days to the reference night in order to take into account the offset that DESI uses.

Finally, I added a test case for this 'test_closestexposures.py' which makes sure that 'closest exposures' can handle changes in months and changes in years. This is made by hard coding two example exposure tables and could be changed in the future. However, currently the exposure tables used in making the combined dark are not outputted.


sbailey (changes requested):
Thanks for this update, including putting the closest_exposures logic into a separate function with tests.  I included a number of comments inline; please don't be scared by the number of them on your first DESI PR!  Some of them are stylistic/pedagogical to help match our typical style.

Let's coordinate with @akremin in-person today before merging since this overlaps with his work incorporating the darks into the automated pipeline and there might be merge conflicts or conceptual overlap.

sbailey:
After further discussion with @akremin and @jbanovetz, we realized that this filtering needs to happen internal to the functions rather than at the very start, e.g. to avoid a case where we filter to the 50 nearest exposures but then discard half of them because we just made a hardware change.  @akremin has this implemented in a different branch, so we will close this one without merging.

PR #2521, opened 2025-07-25
Add redshift QA scripts

rongpu:
This PR adds the scripts that compute the per-fiber redshift performance stats and generate the per-fiber QA pages (see [this example](https://data.desi.lbl.gov/desi/users/rongpu/redshift_qa/new/kibo/html/fiber_21.html)).

It requires the [v2 zcatalog](https://github.com/desihub/desispec/pull/2396) as input.

### DIFF
--- a/etc/redshiftqa/desi_per_fiber_qa_html.py
+++ b/etc/redshiftqa/desi_per_fiber_qa_html.py
@@ -1,8 +1,11 @@
 #!/usr/bin/env python
 
+# This is step 3 of the redshift QA generation procedure
 # Create the redshift QA HTML files
+
 # Usage:
-# ./desi_per_fiber_qa_html.py --stats per_fiber_qa_stats.fits -o /global/cfs/cdirs/desicollab/users/rongpu/redshift_qa/new/loa
+# export QA_DIR=/global/cfs/cdirs/desicollab/users/rongpu/redshift_qa/new/loa
+# ./desi_per_fiber_qa_html.py --stats per_fiber_qa_stats.fits -o $QA_DIR
 
 from __future__ import division, print_function
 import sys, os, glob, time, warnings, gc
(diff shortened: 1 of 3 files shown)


py/desispec/redshiftqa/desi_per_fiber_qa_commands.sh
```diff
@@ -0,0 +1,14 @@
+# Compute the per-fiber redshift success rates and save them to disk
```
sbailey: bash scripts should go into bin/ or etc/, not under py/desispec/ .  bin/ is for scripts that are added to $PATH and could be run from the command line by any user from any directory; etc/ is for more specialized scripts or examples of what to run, but aren't included in what gets installed by default. I think this should go under etc/.

py/desispec/redshiftqa/desi_per_fiber_qa_html.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python
```
sbailey: This looks like a script that is intended to be run from the command line, in which case it should go in bin/ without the .py extension if you want it included in every install, or in etc/ .

There could be an argument for wrapping all of this in a function that someone could import and run from a jupyter notebook, and leave a mini `if __name__ == '__main__': ...` stub here so that it could also be called from the command line as a special case of non-installed code. But at minimum it shouldn't be left as-is in py/desispec/redshiftqa because e.g. if a document parser tries to import this module, it will end up running argument parsing, failing due to required options not being specified, etc.

In general:
* stuff under py/desispec : thing that you intend to "import desispec.blah" and run as a function
* stuff under bin/ : things that should be included in an installation and added to $PATH and be runable by anyone from anywhere
* stuff under etc/ : very specialized scripts or examples, but not intended as part of the core "this is what desispec provides for you to run"

py/desispec/redshiftqa/desi_per_fiber_qa_stats.py
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env python
```
sbailey: Same comments as for desi_per_fiber_qa_html.py

py/desispec/redshiftqa/desi_per_fiber_qa_plots.py
```diff
@@ -0,0 +1,257 @@
...
+    mask = cat['EFFTIME_SPEC']>min_depth
+else:
+    min_depth = 800.
+    mask = cat['EFFTIME_SPEC']>min_depth
+print('Min depth   ', np.sum(~mask), np.sum(mask), np.sum(~mask)/len(mask))
+cat = cat[mask]
+
+# Apply masks
+if tracer=='LRG':
+    lrgmask = Table(fitsio.read('/dvs_ro/cfs/cdirs/desi/users/rongpu/targets/dr9.0/1.1.1/resolve/dr9_lrg_1.1.1_lrgmask_v1.1_with_targetid.fits'))
```
sbailey: We shouldn't hardcode NERSC-specific or user-specific paths.  Is there a generic location under $DESI_ROOT that could be used instead?  Same for elgmask and other references to "cfs" and "rongpu" in other scripts.

py/desispec/redshiftqa/desi_per_fiber_qa_commands.sh
rongpu: All these scripts are moved to etc/.

py/desispec/redshiftqa/desi_per_fiber_qa_stats.py
rongpu: Same` if __name__`  restructuring.

py/desispec/redshiftqa/desi_per_fiber_qa_plots.py
rongpu: The scripts now point to the equivalent files in the LSS directory.

rongpu:
@sbailey I believe I have addressed all the issues that you raised. Please see the individual comments above.

sbailey (approved):

sbailey:
I had been hoping to wait until we had official Loa v2 catalogs before merging this, but the better is getting in the way of the good on that (mea culpa). Since this PR is isolated to 3 new files and doesn't change existing files, I will merge this now as-is.

Issue #2524, opened 2025-08-05
Write master bias filename in header of preprocessed exposures


julienguy:
I just noticed that we do not save the master bias name in the header contrary to other calibration data (dark,pixmask,flat ... etc).
We have to fix that.

PR #2525, opened 2025-08-05
save bias filename in header

julienguy:
Minor PR to fix issue #2524 .
The code now writes the bias filename in the header


jbanovetz:
It looks like something special was happening with biases in the `get_calibration_image` function where `depend.setdep` wasn't being set just for biases. It looks like this should now fix it.

jbanovetz (approved):

Issue #1512, opened 2021-12-06
robustness to DARKs that span noon NIGHT rollover


sbailey:
DARK exposure 20211205/00112406/desi-00112406.fits.fz is under NIGHT 20211205 but has header NIGHT=20211204.  Unluckily, it was also a 300s dark, which appears to have taken the spot as the first 300 sec DARK of the night even though the header keyword mismatch prevented it from getting used (initially no dark*.slurm job was created; I submitted the "correct" dark 112463 by hand).

Various remediations to be more robust:
  * only use DARK if PROGRAM starts with "CALIB Dark" (from the zeros+dark sequence) instead of "DARK" from the 90min dark sequences.
  * if an exposure is rejected, don't let it take the spot for the nightly badcolumn dark

akremin:
This is a useful discussion to have and come to a definitive answer on using `PROGRAM` for exposure filtering. Given that these PROGRAM values are pre-defined by the scripts used to run them they should be as robust as any other parameter, but we have had a collective hesitation because there is no precise reason for a given name and a well meaning person could change it to something else without realize the hard coded values in the code.

All of our cals have multiple subtypes. We currently use `EXPTIME` information to distinguish most of them and `SEQTOT` to distinguish between flat sequence types. We could do belt-and-suspender by adding PROGRAM checks, or alternatively move to a PROGRAM check rather than arbitrary exposure times.

Summary of types of cal follow:

#### zeros
1) "flush" zeros after turning VCCD on.
2) Actual bias frames:  b camera nightly biases and generating bias model

#### darks
1) 300s calibration dark: for bad column detection and dark model evolution modeling.
2) Dark sequence(s): for dark modeling

#### arcs
1) short arcs: for psf modeling
2) long arcs: for faint lines, not used currently

#### flats
1) short flat _sequence_: no longer used
2) long flat sequence: for flat fielding
3) short flat _exposure_: for cte identification.
There are three cases where we have multiple sub-types of obstypes: short vs long arcs, short flat sequence

sbailey:
This happened again with a dark 20241124/00264676/desi-00264676.fits.fz that spanned the noontime boundary and ended up with NIGHT=20241123 in the headers.  This led to pipeline bookkeeping confusion such that the 20241124 processing re-ran the 20241123 ccdcalib and then crashed because calibnight/20241124 files hadn't been created.

[...]

PR #2543, opened 2025-09-16
Dark rollover

jbanovetz:
To solve the problem from this ticket of darks rolling over (https://github.com/desihub/desispec/issues/1512), I added a small if section to two python files that make/update the exposure tables to make sure that exposures added match the with regards to the night of the table and the night from the header of the exposure. I tested this in a Jupyter notebook and found it works in adding `ignore` to the rollover exposures. We might want to change this in the future to move the exposure to the correct exposure table but this should stop any issues as described in the ticket.

### DIFF
--- a/py/desispec/scripts/exposuretable.py
+++ b/py/desispec/scripts/exposuretable.py
@@ -148,6 +148,13 @@ def create_exposure_tables(nights=None, night_range=None, path_to_data=None, exp
             if rowdict is not None and type(rowdict) is not str:
                 rowdict['BADCAMWORD'] = badcamword
                 rowdict['BADAMPS'] = badamps
+                
+                ## Check that the night from the header matches that given and if not, ignore the file
+                if rowdict['NIGHT']!=night:
+                    log.info(f"\n{exp} has the wrong NIGHT keyword, setting LASTSTEP to 'ignore'.")
+                    rowdict['LASTSTEP']='ignore'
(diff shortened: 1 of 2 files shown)


akremin (changes requested):
Hi John,

Both of these scripts call the same underlying function to create an exposure table row. I'd recommend for maintainability reasons that we put this logic in that function instead.

Relevant place in that function might be here: https://github.com/desihub/desispec/blob/6e36187d7dc9018b05d570b44e70a69005b644b9/py/desispec/workflow/exptable.py#L632

This is very slightly more complicated because we have multiple files open here that we're checking for information. There are nights where NIGHT wasn't defined in the raw data header due to ICS issues, for instance.

I think what we want here is to set the LASTSTEP to ignore if the OBSTYPE is dark and either the night isn't in the raw data file or differs from the expected night. I'm not sure if the former is a problem or not. Then you can append a comment string stating why it's ignored like you did before (emphasis on append since there might be other comments).

sbailey:
The unit tests are failing due to the pip/setuptools/desitarget incompatibility reported in desihub/desitarget#862.  A workaround by pinning the pip version was implemented in desispec PR #2557; we don't need to repeat that here.  The full suite of desispec tests pass at NERSC.  Let's merge on Monday (@akremin's review requested a change, but if I understand the comment thread correctly that has been addressed).

sbailey (approved):

Issue #2551, opened 2025-10-10
calibfinder requires darks to exist even if not asked for one


akremin:
When adding a new camera last week we ran into an issue where a new dark wasn't available for the camera in question. We now create darks as part of the pipeline, so this should not interfere with operations, but it did.

preproc raised the following error via calibfinder:
```
CRITICAL:calibfinder.py:569:find_darks_in_desi_spectro_dark: Didn't find matching z1 calibration darks in $DESI_SPECTRO_DARK, quitting
ERROR:preproc.py:278:preproc_file: Error while reading or preprocessing camera z1 in /global/cfs/cdirs/desi/spectro/data/20250929/00313643/desi-00313643.fits.fz
ERROR:preproc.py:279:preproc_file: Didn't find matching z1 calibration darks in $DESI_SPECTRO_DARK, quitting
ERROR:preproc.py:232:main: 1/1 cameras failed
CRITICAL:util.py:162:runcmd: FAILED rank 0 exception while running desispec.scripts.preproc.main(['-i', '/global/cfs/cdirs/desi/spectro/data/20250929/00313643/desi-00313643.fits.fz', '-o', '/global/cfs/cdirs/desi/spectro/redux/testdk/preproc/2025\
0929/00313643/preproc-z1-00313643.fits.gz', '--outdir', '/global/cfs/cdirs/desi/spectro/redux/testdk/preproc/20250929/00313643', '--cameras', 'z1']) called from line 450 in /global/homes/k/kremin/desi/python/desispec/py/desispec/scripts/proc.py
ERROR:util.py:170:runcmd: Traceback (most recent call last):
ERROR:util.py:170:runcmd: File "/global/homes/k/kremin/desi/python/desispec/py/desispec/util.py", line 148, in runcmd
    result = cmd(*args)
ERROR:util.py:170:runcmd: File "/global/homes/k/kremin/desi/python/desispec/py/desispec/scripts/preproc.py", line 238, in main
    sys.exit(int(num_failed))
ERROR:util.py:170:runcmd: SystemExit: 1
```

The reason is that we do need the calibfinder to read in the pixel mask and pixel flat. When that was loaded, it would raise an error because it couldn't find a viable dark for the new camera. 

We should change calibfinder to not raise an error unless explicitly asked for a dark that it doesn't have a location for.

PR #2552, opened 2025-10-10
Only Raise Error if CalibFinder Asked for Bias or Dark it can't Find

akremin:
This resolves issue #2551 in which preproc was crashing when a valid dark was available for a camera in `DESI_SPECTRO_CALIB` or `DESI_SPECTRO_DARK`. It is important to crash if we request the dark, but not if we are using calibfinder to find other files.

This works around that by setting class flags upon initialization that tell record the fact that we don't have a stored calibration dark (or bias). If the dark or bias location is requested and these flags are set, then we raise an error. If 'DARK' or 'BIAS' is never requested, then we don't raise an error. The current implementation keeps the DARK and BIAS connected, like it was previously, in the sense that if a bias is missing and we request a dark, it will still crash. But it is incrementally better than what was there and fixes the issue reported in #2551 without more elaborate surgery on `calibfinder.py`.

I've tested this by recreating the environment that triggered the failure (current trunk now has a dark, so it won't trigger the bug). I pinned DESI_SPECTRO_DARK to v2405, which is the version that crashes with current desispec/main.
Repo is here: `/global/cfs/cdirs/desi/spectro/redux/testdk`

### DIFF
--- a/py/desispec/calibfinder.py
+++ b/py/desispec/calibfinder.py
@@ -386,7 +386,7 @@ class CalibFinder() :
         if self.dark_or_bias_not_found and key in ['BIAS', 'DARK']:
             if self.crash_on_dark_or_bias_request:
                 log.critical(f"Didn't find matching {self.camera} calibration darks in $DESI_SPECTRO_DARK, quitting")
-                raise IOError(f"Didn't find matching {self.camera} calibration darks in $DESI_SPECTRO_DARK, quitting")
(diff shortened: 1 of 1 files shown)


sbailey (approved):
Functionality looks good, including updated tests, thanks. I put in one minor suggestion for exactly which error to raise. OK to leave as-is if you disagree.

Otherwise, @akremin and @jbanovetz please coordinate to get the missing mask for z3 in place before merging this.  When @jbanovetz has committed the mask, @akremin go ahead and merge when ready.

Issue #2533, opened 2025-09-10
desi_purge_night doesn't know about dark_preproc directory


akremin:
With the merging of the darknight PR, we now have a `dark_preproc` directory that is created on a given night. This should be removed if we are purging a night. 

I can't think of a reason why `desi_purge_tilenight` would care, since the science exposures don't save files there. So we just need to update `desi_purge_night`.

PR #2553, opened 2025-10-17
Purge dark_preproc in desi_purge_night

akremin:
Trivial change to resolve issue #2533. Now that we've introduced the `dark_preproc` directories, we need to purge them if we purge a night.


akremin:
Two failing tests are not related to this PR. I'm self merging.

PR #2554, opened 2025-10-17
Make compute_dark crashes more explicit

akremin:
One line change to `compute_dark` to make it more explicit that the code is exiting because of a missing raw data file. I also upgraded the `log.error` to `log.critical` in places that return 1, since those will crash.


akremin:
Failing tests are not related to this PR. Self merging.

PR #2555, opened 2025-10-20
additional logging of desi_compute_psf commands

sbailey:
This PR adds logging about what specex desi_compute_psf commands + options are run.  Previously it would only provide the command line if the command crashed, but not if it succeeded.  This makes the logging similar to other commands that are run via `runcmd`, which logs the command before running it and also afterwards.

Detail: desi_compute_psf = desispec.scripts.specex.main -> desispec.scripts.specex.run has a different structure to the MPI parallelism than other commands and thus doesn't fit with how runcmd works.  i.e. that's why we didn't get this pre/post logging for free via runcmd.

GitHub tests will likely fail due to unrelated numba/llvmlite installation version incompatibilities, to be solved separately. I tested this by running desi_proc on an arc exposure for which I was trying to get what exactly the command were that were being run.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -102,14 +102,15 @@ jobs:
         strategy:
             fail-fast: true
             matrix:
+                # matches python 3.10 environment above, approximately matching NERSC
                 os: [ubuntu-latest]
-                python-version: ['3.10'] # Approximately current NERSC setup
-                astropy-version: ['<6.1']  # Approximately current NERSC setup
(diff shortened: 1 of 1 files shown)


sbailey:
> GitHub tests will likely fail due to unrelated numba/llvmlite installation version incompatibilities, to be solved separately.

Actually, only the coverage tests failed this time but the normal unit tests worked, so I'm going to use this PR to see if I can debug the requisite version pinning. Apologies in advance for the failing test spam...

akremin:
Sorry for missing this tag yesterday. The logging changes look good.

PR #2557, opened 2025-10-24
Cosmic ray rejection: If statements for ITL devices on red cameras

jbanovetz:
Since installation, r7 has had high cosmic rates for amplifiers A and B. Looking into the data, it looks like some of the arc lines are getting masked as cosmic rays. Julien mentioned this is because we are using the wrong psf gradient parameters for the detector, parameters meant for LBNL detectors, not ITL. I have added some if statements to use the b camera psf parameters (since they are ITL devices) if the red cameras are ITLs using the CCDTYPE keyword, something unique to the newer detectors. Changing the parameters does help, though does not completely solve the problem of parts of arc lines being flagged as cosmic rays. Below shows comparisons between using the old (LBNL) psf parameters and the new (ITL) pdf parameters for r7. This just shows the cosmic ray mask plane in red overlaid with the corresponding arc (00316177 taken on 20251014).

image

image

This change should also help r1 which also has an ITL detector. Below are comparisons similar to those done with r7.

image

image

### DIFF
--- a/py/desispec/cosmics.py
+++ b/py/desispec/cosmics.py
@@ -382,7 +382,7 @@ def reject_cosmic_rays_ala_sdss(img,nsig=6.,cfudge=3.,c2fudge=0.5,niter=6,dilate
         # desi_compute_psf_gradients -i $DESI_SPECTRO_CALIB/spec/sp2/psf-b2-20190308.fit
         psf_gradients=np.array([0.758221,0.771945,0.570183,0.592199])
     elif band == 'r':
-        if 'CCDTYPE' in img.meta and ('ITL' in img.meta['CCDTYPE']):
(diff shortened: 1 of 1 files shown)


jbanovetz:
I have fine tuned some of the parameters to further reduce the number of arc lines that have pixels masked as cosmic. Comparing between the darks processed with old psf parameters, there is a 1000-1500 pixel discrepancy in the number of pixels masked. This seems to be that the new parameters can't fit the tails as well. Below is a comparison where the difference shows the pixels not being masked by the new parameters.

image

 I need to check if this is actually the old code overfitting the cosmic rays.

sbailey:
@jbanovetz it looks like you haven't pushed your updated coefficients to github, i.e. the ITL r coefficients are still the same as b. Please confirm.

PR #2562, opened 2025-11-13
remove pin on pip version

sbailey:
desihub/desitarget#864 fixed the underlying incompatibility that previously required us to pin pip==25.2 .  This PR removes that pinned version.  This may trigger similar issues in other packages for followup.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -83,6 +83,7 @@ jobs:
                 DESITARGET_VERSION: main
                 REDROCK_VERSION: 0.20.4
               run: |
+                # --no-build-isolation needed for older code tags with latest pip/setuputils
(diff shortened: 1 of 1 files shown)


sbailey:
This doesn't work with the pinned desimodel==0.19.3 which was before desimodel pyproject.toml and setup.cfg were added.  This highlights that we probably have a bunch of tags that can't be installed with pip 25.3.  To be continued...

weaverba137:
It looks like you didn't add `--no-build-isolation`.

sbailey:
Yes, let's address pyproject.toml in a separate PR.

Summarizing the state of installing other desi packages that were tagged before adding pyproject.toml + setup.cfg — either
* pin pip<25.3 (or perhaps pin setuptools, but pinning pip has the same effect); or
* pip install --no-build-isolation ...

Those workaround don't require any changes to desispec itself, but they impact how we install packages for testing.

Issue #2489, opened 2025-05-16
zall PHOTSYS incorrectly changed in loa for some targets


sbailey:
As discovered by @biprateep :  

in loa/zcatalog/v1/zall-pix-loa.fits, TARGETID=39627854849247294 incorrectly has PHOTSYS='N' for the SURVEY='main' PROGRAM='dark' row, when it should be PHOTSYS='' (blank) like it is in upstream files like zpix-main-dark.fits:
```
In [7]: zcat = fitsio.read('zall-pix-loa.fits', 1, columns=('TARGETID', 'PHOTSYS', 'SURVEY', 'PROGRAM'))

In [8]: Table(zcat[zcat['TARGETID']==39627854849247294])
Out[8]: 
<Table length=2>
     TARGETID     SURVEY PROGRAM PHOTSYS
      int64        str7    str6    str1 
----------------- ------ ------- -------
39627854849247294   main    dark       N
39627854849247294    sv3    dark       S

In [9]: zpix = fitsio.read('zpix-main-dark.fits', 1, columns=('TARGETID', 'PHOTSYS'))

In [10]: Table(zpix[zpix['TARGETID']==39627854849247294])
Out[10]: 
<Table length=1>
     TARGETID     PHOTSYS
      int64         str1 
----------------- -------
39627854849247294        
```

When observed in the main survey, this is a secondary target that should not have had a PHOTSYS associated with it.  It is correct in zall-pix-iron.fits.

Figure out what when wrong with assembling the zall file and document how many targets are affected.

sbailey:
Wow, thanks for tracking that down.  That astropy Table "feature" of treating blank strings as masked instead of just blank strings has bitten us multiple times.  Agreed about using `fitsio.read` instead.

FWIW, `desispec.io.read_table` is a wrapper around `t = Table(fitsio.read(filename))` that also propagates the header keywords into the `t.meta` dictionary like you get from `Table.read(filename)` but you don't by default get from `fitsio.read(filename)`. So if this is a case where we are also propagating header keywords, `desispec.io.read_table` could be useful.

rongpu:
The v2 zall catalogs with the bug fix are here:
`/global/cfs/cdirs/desicollab/users/rongpu/tmp/zcatalog/v0.4_bug_fix/zall`

The following columns (all strings) were affected by the bug (these are v2 zcatalog columns):
[...]

PR #2566, opened 2025-11-18
Fix zall blank string bug

rongpu:
This PR fixes the issue in https://github.com/desihub/desispec/issues/2489.

### DIFF
--- a/py/desispec/zcatalog.py
+++ b/py/desispec/zcatalog.py
@@ -45,7 +45,7 @@ import numpy as np
 import os
 from glob import glob
 from astropy.io import fits
-from astropy.table import Table, Column, vstack, join
+from astropy.table import Table, Column, MaskedColumn, vstack, join
 
 ## DESI related functions
 from desispec.io import specprod_root, read_table
@@ -425,6 +425,11 @@ def update_table_columns(table, specgroup):
     ## for rearranging the columns into a proper order
     sel = np.char.endswith(tab_cols, '_TARGET') & (tab_cols!='FA_TARGET')
     target_cols = list(tab_cols[sel])
+
+    ## Confirm that the only masked values are the _TARGET columns
+    for col in table.colnames:
+        if isinstance(table[col], MaskedColumn):
+            assert col in target_cols, f"Unexpected masked column {col}"
 
     ## Table with filled values
     table = table.filled(fill_value=0)


py/desispec/zcatalog.py
```diff
@@ -424,12 +426,8 @@ def update_table_columns(table, specgroup):
     sel = np.char.endswith(tab_cols, '_TARGET') & (tab_cols!='FA_TARGET')
     target_cols = list(tab_cols[sel])
 
-    for col in target_cols:
-        ## Fill the *TARGET columns that are masked with 0
-        table[col].fill_value = 0
-
     ## Table with filled values
-    table = table.filled()
+    table = table.filled(fill_value=0)
```
sbailey: Previously `fill_value` was set only for the `target_cols`, i.e. columns that end with "_TARGET" except for "FA_TARGET". This code will set `fill_value=0` for all columns, which I think will have the side effect of setting masked PHOTSYS and BRICKNAME to the string "0" instead of leaving them as blank strings.  Please check.

rongpu: And to clarify, with the bug fix, the only masked columns are the target bitmask ("_TARGET") columns, and they are filled with value 0. None of the string columns (`SUBTYPE, SPECTYPE_NEW, SUBTYPE_NEW, SUBTYPE_BEST, OBJTYPE, BRICKNAME, MORPHTYPE, REF_CAT, PHOTSYS`) are masked.

sbailey: This behavior surprised me, so I dug a bit deeper. The reason that `SUBTYPE`, `BRICKNAME`, etc. are not masked is that `read_table(filename)` purposefully does not generate masked columns; it just propagates forward whatever the file has. However, different input files have different sets of _TARGET columns, so when they get vstacked the output table creates the superset of _TARGET columns as masked columns.

i.e. `table = table.filled(fill_value=0)` works because the _TARGET columns are the only masked columns and 0 is the correct fill value in that case, but if other masked columns had appeared, 0 might not have been the choice we'd want (especially for strings/bytes). To guard against this, I added a safety check before doing the filling that the only masked columns are the _TARGET columns as expected.  If any other masked columns appear in the future, we can explicitly chose what to do with them instead of risking a string column getting filled with "0" instead of blank "".

sbailey (approved):
Thanks for doing the explicit check that the buggy column(s?) are the only ones impacted by the change. Approved for merging once GitHub tests finish.

Issue #2567, opened 2025-11-19
zall v2 incorrect SURVEY and PROGRAM keywords


sbailey:
After the zcat v2 update, `desispec.zcatalog.create_summary_catalog` generates warnings like
```
WARNING: MergeConflictWarning: Cannot merge meta key 'SURVEY' types <class 'str'> and <class 'str'>, choosing SURVEY='main' [astropy.utils.metadata.merge]
WARNING: MergeConflictWarning: Cannot merge meta key 'PROGRAM' types <class 'str'> and <class 'str'>, choosing PROGRAM='backup' [astropy.utils.metadata.merge]
WARNING: MergeConflictWarning: Cannot merge meta key 'PROGRAM' types <class 'str'> and <class 'str'>, choosing PROGRAM='bright' [astropy.utils.metadata.merge]
WARNING: MergeConflictWarning: Cannot merge meta key 'PROGRAM' types <class 'str'> and <class 'str'>, choosing PROGRAM='dark' [astropy.utils.metadata.merge]
WARNING: MergeConflictWarning: Cannot merge meta key 'SURVEY' types <class 'str'> and <class 'str'>, choosing SURVEY='special' [astropy.utils.metadata.merge]
WARNING: MergeConflictWarning: Cannot merge meta key 'SURVEY' types <class 'str'> and <class 'str'>, choosing SURVEY='sv1' [astropy.utils.metadata.merge]
WARNING: MergeConflictWarning: Cannot merge meta key 'PROGRAM' types <class 'str'> and <class 'str'>, choosing PROGRAM='other' [astropy.utils.metadata.merge]
WARNING: MergeConflictWarning: Cannot merge meta key 'SURVEY' types <class 'str'> and <class 'str'>, choosing SURVEY='sv2' [astropy.utils.metadata.merge]
WARNING: MergeConflictWarning: Cannot merge meta key 'SURVEY' types <class 'str'> and <class 'str'>, choosing SURVEY='sv3' [astropy.utils.metadata.merge]
```
And the output file has incorrect keywords
```
SURVEY  = 'sv3     '                                                            
PROGRAM = 'dark    '                                                            
```
i.e. these need to be stripped from the input catalogs before stacking the zall catalog which spans surveys and programs.

Example outputs in /pscratch/sd/s/sjbailey/desi/zcat/v2.1/zall, generated with
```
from desispec.zcatalog import create_summary_catalog
[...]

PR #2572, opened 2025-11-21
Fix bug in zall zcatlog PROGRAM and SURVEY header keywords

rongpu:
This PR address https://github.com/desihub/desispec/issues/2567. The bug was that the PROGRAM and SURVEY header keywords were removed from the primary (ZCATALOG) files and not the -imaging and -extra files. Now the two keywords are removed from all zall files.


sbailey (approved):
Looks good, thanks!

PR #2577, opened 2025-12-04
PR to survive missing calibration files if environment variable DESI_SPECTRO_ROBUST=True

julienguy:
- Use environment variable DESI_SPECTRO_ROBUST=True to complete the preprocessing of exposures even if there is no dark calibration file.
 - This is useful to preprocess images with new configurations, and for nightwatch to visualize data quality even in the absence of a full set of calibrations (that we cannot have with need hardware, or new hardware configuration)
 - If this specific environment variable DESI_SPECTRO_ROBUST is not set, the pipeline behavior is unaltered.
 - We can add later other cases where the pipeline would continue despite missing features when DESI_SPECTRO_ROBUST=True.
 -

### DIFF
--- a/py/desispec/calibfinder.py
+++ b/py/desispec/calibfinder.py
@@ -13,7 +13,7 @@ import numpy as np
 import yaml
 from astropy.table import Table
 
-from desispec.util import parse_int_args, header2night
+from desispec.util import parse_int_args, header2night, is_robust_mode
 from desiutil.log import get_logger
 
 def parse_date_obs(value):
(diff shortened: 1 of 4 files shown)


sbailey:
Thanks, this is good to have, and I appreciate that robust mode is opt-in not opt-out.  I put the robust-or-not code into a separate function `desispec.util.is_robust_mode`, inspired by this PR originally having two different definitions of which values were allowed to be considered robust :).  And added unit tests.  I also added text to various error messages, giving us a hint that setting $DESI_SPECTRO_ROBUST is how to work around the missing-calib errors if needed.

I tested this with `desi_preproc -n 20251204 -e 324997 --cameras b0 -o $SCRATCH/preproc-b0.fits` which does not yet have a config, with various combinations of $DESI_SPECTRO_ROBUST and $DESI_SPECTRO_DARK set or not.

Will merge when tests re-pass.

Issue #2560, opened 2025-11-06
zcat v2 definition of EFFTIME_SPEC


sbailey:
In zcat v2 from PR #2396, EFFTIME_SPEC is defined only in terms of TSNR2_LRG.
This matches Equation 22 of Guy et al. 2023 (arxiv:2209.14482), which is valid for the dark program, but it inconsistent with how EFFTIME_SPEC is tracked for the exposures and tiles of other programs.

Update https://github.com/desihub/desispec/blob/main/py/desispec/scripts/zcatalog.py#L602 to use `desispec.tsnr.tsnr2_to_efftime` (also used by bin/desi_tsnr2_afterburner).  Arguably that function should be updated / wrapped to standardize the choices based upon PROGRAM.  Currently for the tiles/exposures:

| PROGRAM | scaled from TSNR2 column |
| ---------- | ---------------------------- |
| dark | TSNR2_LRG |
| bright | TSNR2_BGS |
| backup | TSNR2_GPBBACKUP |
| other | I don't know; only used in cmx and sv1 |

rongpu:
Before we implement this change, @julienguy could you check that the program-dependent EFFTIME_SPEC definitions above are indeed reasonable? And if the answer is yes, should we just use TNSR2_LRG for PROGRAM== other?

PR #2583, opened 2026-01-07
Make EFFTIME_SPEC program-dependent

rongpu:
This PR addresses https://github.com/desihub/desispec/issues/2560. Rather than using TSNR2_LRG throughout, the TSNR2-to-EFFTIME_SPEC conversion now uses the program-dependent TSNR2 values as described by @sbailey in https://github.com/desihub/desispec/issues/2560. I have tested the new code on loa SV1 data and it works as expected.


rongpu (commented):
@sbailey I have made the changes that you requested.

Issue #2244, opened 2024-05-10
Combine PSFs even if individual bundles failed


sbailey:
Currently in the pipeline
  * `desi_proc` calls `desi_compute_psf` to calculate the PSF per exposure
  * `desi_compute_psf` (scripts.specex.main) calls `desi_fit_psf` to fit individual bundles, and then merges them back together with `desi_merge_psf`.
  * `desi_proc_joint_fit` then calls `scripts.specex.mean_psf` to combine PSFs across exposures into the nightly PSF

However, if `desi_fit_psf` fails for an individual bundle by having a bogus solution with overlapping fiber traces, this causes `desi_compute_psf` to fail and halt the entire pipeline.

It would be better if the bad bundle would be flagged as bad in the per-exposure PSF fit to prevent it from being used in `mean_psf`, but otherwise let the pipeline proceed.  It is relatively common for small updates to darks/masks to trigger new edge case failures leading to time consuming flagging, cleanup, and resubmissions in new prods.  Our current mitigation is to flag the entire camera as bad for that exposure and rerun, which looses the otherwise good bundles on that camera from the mean (ok, but not great).

Example failures of individual bundles in jura which had not been caught and flagged by earlier prods:
* night 20221209 exp 157081 camera z4 fibers 225-250
* night 20210327 exp 82431 camera r1 fibers 75-100
* night 20220506 exp 133476 camera r2 fibers 275-300

All of these would have been fine if it just logged an error, flagged the per-exposure bundle as bad, but then proceeded and dropped the bundle/exp from the psfnight merging.

julienguy:
None of the 3 examples above are failing either when using directly:
```
srun -n 20 desi_compute_psf_mpi --input-image preproc-r1-00082431.fits --input-psf  /global/homes/j/jguy/redux/daily/exposures/20210327/00082431/shifted-input-psf-r1-00082431.fits --output-psf psf-r1-bis.fits --broken-fibers 470
```
or with a custom script that reuses the logic in the pipeline (need to be ran with srun -n 21 because one rank is apparently used for some scheduling):
```
from mpi4py import MPI
import desispec.scripts.specex
comm=MPI.COMM_WORLD
[...]

PR #2584, opened 2026-01-08
Improved handling of the PSF fit failures of individual fiber bundles.

julienguy:
The PSF fit is performed independently per group (or bundle) of 25 fibers (because the bundles are well separated on the CCD). In this PR, we change the code behavior when the fit of a bundle fails. 
 - Instead of returning an error and throwing an exception for the full PSF fit, we do not merge the PSF parameters of the failed bundle in the output PSF file. In practice it means that the PSF parameters for the failed bundle are those of the input PSF. The fact that the fit failed for this bundle is recorded in the header of the 'PSF' HDU with a reduced chi2=0 and also in the 'STATUS' entry in the data table. 
 - When the parameters of several fitted PSF files are averaged, the reduced chi2 of the fit is read, and the PSF parameters of the  bundles with a reduced chi2=0 are excluded from the averaging.
 
This resolves issue #2244.

### DIFF
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -451,6 +451,9 @@ def merge_psf(inpsffile, inputs, output):
                      "B{:02d}NPAR".format(b) ]:
             if key in psf_hdulist["PSF"].header :
                 psf_hdulist["PSF"].header[key] = 0
+    # also reset the PSF STATUS
+    index = np.where(psf_hdulist["PSF"].data["PARAM"]=="STATUS")[0][0]
+    psf_hdulist["PSF"].data["COEFF"][index][:,0]=-1
 
     for input_filename in inputs :
         log.info("merging {} into {}".format(input_filename, inpsffile))


sbailey:
In `/pscratch/sd/s/sjbailey/temp/psf` I tried running
```
srun -n 20 desi_compute_psf_mpi   \
  --input-image /dvs_ro/cfs/cdirs/desi/spectro/redux/daily/attic/preproc/20251113/00321393/preproc-z0-00321393.fits.gz \
  --input-psf /dvs_ro/cfs/cdirs/desi/spectro/redux/daily/attic/exposures/20251113/00321393/shifted-input-psf-z0-00321393.fits \
  --output-psf fit-psf-z0-00321393.fits --broken-fibers 367
```
And it included
```
FATAL ERROR (other std) ERROR z0-00321393-09: problem with brent dchi2 = -3.06821e+23 (at line 1583 of file /global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/specex/main/src/specex_psf_fitter.cc)
...
WARNING:specex.py:305:main: The fit of the following bundles failed: [9]
```
But I don't see that reflected in the output file
```
psf = Table.read('fit-psf-z0-00321393.fits', 'PSF')
assert psf['PARAM'][57] == 'STATUS'  # confirm that row 57 has the STATUS coefficients
status = np.array(psf['COEFF'][57])
assert np.all(status == 0.0)
```
i.e. all STATUS values are 0, regardless of bundle.

Am I mis-interpreting the file and the problem with bundle 9 is stored separately, or is this a bug?

julienguy:
It is true that the STATUS was still 0 even for failed bundles (kept the value from the input PSF). I have changed that now. The averaging code was still working fine (i.e. ignoring failed bundles) because it is looking at the reduced chi2 values that are saved in the header to decide which data to include in the averaging (it discards chi2=0 or large chi2 values).

sbailey:
I confirmed that with PSFs generated with this update, `desi_average_psf` can survive 1 missing bundle but will fail if the same bundle failed on 2 or more exposures.  Good.

sbailey (approved):

Issue #2589, opened 2026-01-14
Refactor array copying in Spectra object to optimize file load speed and Spectra creation speed.


dylanagreen:
## The Problem
Some basic profiling of 10 raw data reads of the desispec.io.read_spectra function revealed that the load is spending most of its time copying the input spectra FLUX/IVAR/MASK/extra/etc. data to the Spectra object:
https://github.com/desihub/desispec/blob/3306287962a87d88974aaeb158bb2e566b4fc933/py/desispec/spectra.py#L209-L215

Profiling on `special/dark/272/27239/coadd-special-dark-27239.fits`:
```
Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       10    0.090    0.009   64.514    6.451 spectra.py:208(read_spectra)
      120    0.001    0.000   35.471    0.296 spectra.py:196(_read_image)
       10    0.253    0.025   27.726    2.773 spectra.py:83(__init__)
      150    0.001    0.000   25.173    0.168 image.py:246(read)
      150   25.163    0.168   25.163    0.168 {method 'read_image' of '_fitsio.FITS' objects}
      370   22.062    0.060   22.062    0.060 {method 'astype' of 'numpy.ndarray' objects}
   368430    0.667    0.000   13.373    0.000 resolution.py:82(__init__)
   368430    2.098    0.000   11.938    0.000 _dia.py:23(__init__)
   368430    0.429    0.000    2.783    0.000 _base.py:1532(_get_index_dtype)
   368430    0.368    0.000    2.780    0.000 _arraysetops_impl.py:145(unique)
   741330    2.770    0.000    2.770    0.000 {built-in method numpy.array}
   368430    0.962    0.000    2.232    0.000 _arraysetops_impl.py:348(_unique1d)
      120    0.000    0.000    2.207    0.018 _function_base_impl.py:916(copy)
   368430    1.426    0.000    2.139    0.000 _sputils.py:264(get_index_dtype)
   368430    0.747    0.000    1.512    0.000 _sputils.py:444(check_shape)
  2417306    0.457    0.000    0.775    0.000 {built-in method builtins.isinstance}
4455240/4454505    0.461    0.000    0.637    0.000 {built-in method builtins.len}
   368430    0.208    0.000    0.601    0.000 _sputils.py:372(isshape)
   736860    0.511    0.000    0.511    0.000 getlimits.py:698(__init__)
[...]

PR #2590, opened 2026-01-14
Add `copy` argument to Spectra object and only singly copy inputs when requested.

dylanagreen:
See https://github.com/desihub/desispec/issues/2589 for motivation. 

Originally spectral fluxes, ivars, and masks were doubly copied in initializing the Spectra object, because by default astype always copies arrays regardless of mismatches (or not) between input and requested dtypes. The primary purpose of this PR is to remove the spurious additional copies which were slowing down the Spectra object.

In addition, I added `copy_inputs` as an argument to the Spectra constructor, with a default of True (previous behaviour). This variable is passed directly to np.astype, and when set to False and the input and requested data types are the same, the np.astype will not copy. If the input and requested dtypes are different, then numpy has to incur a copy regardless.

Loading the spectra (in read_spectra) now requests no copies if possible through copy_inputs=False. Since the read in data is cast upon read, this prevents the Spectra object from doing _both_ copies, since now the input and requested dtypes will be the same. If read_spectra is read with default inputs (single=False, so no casting is necessary off disk), this saves ~30% of load time when reading spectra.

### DIFF
--- a/py/desispec/io/spectra.py
+++ b/py/desispec/io/spectra.py
@@ -515,7 +515,7 @@ def read_spectra(
         single=single,
         scores=scores,
         redshifts=redshifts,
-        copy_inputs=False
+        copy=False
     )
 
     # if needed, sort spectra to match order of targetids, which could be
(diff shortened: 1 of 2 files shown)


dylanagreen:
@sbailey This should be ready for review, although I think it's a pretty minor change.

Issue #2586, opened 2026-01-13
desi_preproc_darks does not use the CCD gains.


julienguy:
There is a bug in desi_preproc_darks somewhere because it does not read correctly the gains of the CCD that are saved in the calibration data:
```
grep -i gain ~/redux/daily/run/scripts/night/20260112/pdark-20260112-00331478-a0123456789-47729488.log 
ERROR:preproc.py:1123:preproc: Camera z0 missing keyword GAINA in header and nothing in calib data; using 1.0
```
(it's the same for all cameras and amps)

Because of this bug, the dark nights are in ADU and not in electrons.

julienguy:
It's because of this line in the preproc code:
```
need_cfinder = (mask is True) or (pixflat is True)
```

sbailey:
For the record, that `need_cfinder` line was added in PR #2552 while fixing issue #2551.  Whatever update we make for this, let's make sure we don't re-break what that was trying to fix.

akremin:
I hope that we can just replace this with `need_cfinder = (mask is True) or (dark is True) or (pixflat is True)`, which is what we used to use before I replaced it with the above. As I stated, I _think_ that was a belt-and-suspenders style change and that the other mitigations will be sufficient after we revert this.

Unfortunately I didn't have time today to commit or debug this. Hopefully someone else can commit and test this. 

The relevant line to change is: line 828 in preproc: https://github.com/desihub/desispec/blob/3306287962a87d88974aaeb158bb2e566b4fc933/py/desispec/preproc.py#L828

At least one test to perform is this one that Stephen ran in that last PR:
```
from desispec.calibfinder import CalibFinder
import fitsio
import os

os.environ['DESI_SPECTRO_DARK'] = '/global/cfs/cdirs/desi/spectro/redux/testdk/v2405'

rawfile = '/global/cfs/cdirs/desi/spectro/data/20250929/00313643/desi-00313643.fits.fz'
hdr0 = fitsio.read_header(rawfile, 0)
hdr1 = fitsio.read_header(rawfile, 'Z1')

cf = CalibFinder([hdr0, hdr1], fallback_on_dark_not_found=True)
cf.findfile('MASK')
cf.findfile('PIXFLAT')
cf.findfile('DARK')  # should fallback to desi_spectro_calib

cf = CalibFinder([hdr0, hdr1])  # crashes in main, not badcol_darknight
cf.findfile('MASK')
cf.findfile('PIXFLAT')
[...]

PR #2591, opened 2026-01-16
Apply gains to preprocessed darks

jbanovetz:
Following discussion in https://github.com/desihub/desispec/issues/2586 , I was able to preproc darks and have the gain applied by adding an extra 'or' statement to this line. This way anything that uses a bias will have gains applied which is what we want I believe.

### DIFF
--- a/py/desispec/preproc.py
+++ b/py/desispec/preproc.py
@@ -825,7 +825,7 @@ def preproc(rawimage, header, primary_header, bias=True, dark=True, pixflat=True
         if key in os.environ:
             depend.setdep(header, key, os.environ[key])
 
-    need_cfinder = (mask is True) or (pixflat is True)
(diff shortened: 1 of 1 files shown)


sbailey:
After sorting out my $DESI_LOGLEVEL settings, I confirm that this works:
```
srun -N 1 -n 30 -c 8 --cpu-bind=cores  desi_preproc_darks \
    -n 20260113 --expids=331536,331537 --camword a1 \
    --preproc-dark-dir $PSCRATCH --specprod daily --mpi
```

on main it incorrectly uses gain=1.0, and on this dark_gain branch it correctly finds and uses the gains.

Since there might be other calibfinder corner cases that we aren't considering, let's *not* merge this going into the long weekend, but merge+test for daily ops on Tuesday (recognizing that there will be a NERSC outage Weds/Thursday).

Issue #2594, opened 2026-01-20
Missing Data for August 23, 2023


jbanovetz:
I was trying to run a `desi_proc_night` in my own SPECPROD and got this error for August 23, 2023 on the `biaspdark` step:

`ERROR:ccdcalib.py:804:compute_nightly_bias:   OSError: Empty or corrupt FITS file`

This is for file `dvs_ro/cfs/cdirs/desi/spectro/data/20230823/00192200/desi-00192200.fits.fz` but possibly more

Looking at Slack, it looks like there is missing/corrupt data due to a full disk at the time. See https://desisurvey.slack.com/archives/C040Q62PD39/p1692924307846299.

I am not sure if we just need to flag some of the biases as bad or if the data exists elsewhere @weaverba137 ?

jbanovetz:
I started to flag them as bad and was curious why they weren't flagged before. Looking at the exposure table, the two exposures aren't actually in the table (192199 and 192200) in daily or in my table (which is just copied from daily). Looking into it further I think it's an issue of how the nightly bias is made. It looks like to find the biases, the exposure table isn't used to find the expids, but the raw data folder:

https://github.com/desihub/desispec/blob/5bd6b8bfc30177cc9abcece5e72f314fcedc6dd0/py/desispec/ccdcalib.py#L508-L511

Looking further into the process, it looks like the check for bad expids doesn't happen until later

https://github.com/desihub/desispec/blob/5bd6b8bfc30177cc9abcece5e72f314fcedc6dd0/py/desispec/ccdcalib.py#L548-L554


Maybe we want to add a check at the beginning that the expids are in the exposure table?

sbailey:
Thanks. Please proceed with a PR for that branch.

PR #2602, opened 2026-01-30
Only include biases that are in the exposure table for making a master bias to avoid corrupted data

jbanovetz:
Fix for https://github.com/desihub/desispec/issues/2594 so that only biases that are in the exposure tables are included to create the bias night. This way any data that is corrupted (such as two biases on 20230823) isn't included in the file list of biases to use that later crashes the pipeline.


jbanovetz:
Just to poke this again, I am not sure where it stands. It looks like the unit tests were cancelled and I don't know how to restart them for this.

Issue #2571, opened 2025-11-21
Add support for non-linear darks to dark_night


sbailey:
We now generate a nightly linear dark using `desi_compute_dark_night`.  However, at the start of DESI we needed to use non-linear darks generated by `desi_compute_dark_nonlinear`.  AFAIK neither the pipeline wrapper nor `desi_compute_dark_night` knows when or how to fallback to using a non-linear dark when needed.  To make matters worse, I don't think we have any automated way to know if a non-linear dark was needed, so we might just need to hardcode 30 dates before which non-linear darks are required.

To do:
  * determine those dates when we can start using linear darks
  * update the nightly dark logic in the pipeline so that future prods will know whether to make a non-linear or a linear dark night

This is a blocking factor for Matterhorn, and thus high priority.

@jbanovetz please take the lead on this, consulting with @akremin and @julienguy as needed.

jbanovetz:
After running into some issues with corner cases, I have confirmation that my change works! This is with branch `darknight_date_cutoff` and it adds in 2-3 lines of code that introduces a cutoff date of 20240509 as past this date, there is no more non-linearity in the darks with any of the cameras (see https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=9765 for more details). These changes are to only make `darknight` models after the cutoff date and to start preprocessing darks 20240408 (30 days before the cutoff). So non-linear darks will be used before the cutoff date and linear darks (via `darkinght`) will be used afterwards.

The tests that I ran:
Using dry runs to just produce the jobs
Date after 20240509 produce a darknight and preprocesses darks
Date before 20240509 and after 30 days previously doesn't produce a darknight and doesn't make preprocessed darks
Date before 20240509 but within 30 days doesn't produce a dark night but does make preprocessed darks

Using full runs:
Date before 20240509 and after 30 days before used the DESI_SPECTRO_DARK dark to process the images
Date after 20240509 used the darknight to process the images

If these tests are sufficient, I will make a PR with the changes.

akremin:
I think these tests are sufficient. Please proceed with a PR, and we can take a look.

PR #2603, opened 2026-01-30
Darknight date cutoff

jbanovetz:
Implementing changes discussed in https://github.com/desihub/desispec/issues/2571 . This will add a cutoff date where a non-linear dark will be used to process data before 20240509 and a linear dark (darknight) will be used to process data after 20240509. It also includes only beginning to preprocess darks for darknight 30 days before the cutoff date (20240408).


akremin (commented):
Thanks John, what you have here looks good, thank you for that and for the testing. 

Can you point me to your test repo? You report that everything looks good there, but we have some unit tests failing, which seem to indicate that there might be issues for the nights before the cuttoff.

The failing unit tests are all in test_proc_night, which tests using dry_run that the pipeline would submit all the expected jobs. In those tests the pipeline is exiting early saying it can't do calibrations. 

`ERROR:proc_night.py:606:proc_night: Exiting because not all calibration files accounted for with links or submissions and require_cals is True.
`

jbanovetz:
Just gave it a look and it looks good to me!

PR #2610, opened 2026-02-12
Dark restart

jbanovetz:
Following what was discussed here: https://github.com/issues/assigned?issue=desihub%7Cdesispec%7C2607, I created a new way for us to reset what darks are used with darknight. It does use a new keyword so the relevant DESI_SPECTRO_CALIB folder used for testing is /global/cfs/cdirs/desi/users/banovetz/desi_spectro_calib_new_dark_keyword. 

Essentially what it does is loads a reference config from the yaml file and looks for a new keyword DARK_RESET and if true, applies it (right now I just have it set up that if the keyword is present to do something and doesn't actually look at the value). Then when running through the list of darks to use, it keeps a running track of when the last time a DARK_RESET was set (dark_reset_begin) and when it was set in the future (dark_reset_end) and just uses the files in between those dates.

A couple of changes is that it does invoke the CalibFinder and so it might add to the run time a little bit and there is now a lengthy if/then section that might need some clean-up (though it does run as expected!).

### DIFF
--- a/py/desispec/workflow/submission.py
+++ b/py/desispec/workflow/submission.py
@@ -261,6 +261,7 @@ def submit_biasnight_and_preproc_darks(night, dark_expids, proc_obstypes,
 
     linked_bias = 'biasnight' in files_to_link
     dobias = (not linked_bias) and ('biaspdark' not in ptable['JOBDESC']) and 'zero' in proc_obstypes and len(zero_expids) > 0
+    
    # Only submit pdark if it is after 30 days before 20240509 (see desispec issue #2571)
     dark_date=night>20240408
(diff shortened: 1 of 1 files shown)


jbanovetz:
I just pushed my latest version of the DARK_RESET logic. I found a bug in the logic running some tests that is now fixed and there is both log and comments which should help guide through the logic/report when something is skipped. I haven't made it into a function and unit test yet but for m1 it should work. We would have to implement the changes to DESI_SPECTRO_CALIB as well though and make new configs for z7 and z0 with the new keyword. It would take about an hour or 2 I think so we could push m1 before that gets done. I did check and everything still runs fine even without the change in keywords to the yaml files in DESI_SPECTRO_CALIB.

Issue #2592, opened 2026-01-16
desi_preproc_darks ragged nested sequences deprecation


sbailey:
From `/global/cfs/cdirs/desi/spectro/redux/daily/run/scripts/night/20260112/pdark-20260112-00331478-a0123456789-47729488.log`:

```
/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/desispec/main/py/desispec/scripts/preproc_darks_mpi.py:193: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  all_data_header_cams = np.array(all_data_header_cams)
```

At best this is a deprecated feature that will stop working in a future version of numpy.  At worst this isn't doing what we intend it to do and it is a bug.

Investigate what `all_data_header_cams` is prior to wrapping it with `np.array(...)` and what the intended result is meant to be, and update.

PR #2619, opened 2026-02-24
Fix preproc_darks_mpi.py ragged array and cleanup some MPI therein

akremin:
This fixes issue #2592 by removing the need for a numpy ragged array and instead using a simple list of lists. I also cleanup up a few typos and improved the MPI logic. 

There was reports of the job hanging due to the ragged array issue. This is a fundamental problem with having different ranks doing different tasks, if one fails the others will wait at the next blocking operation forever. The solution I have found based on a quick search is to wrap the entire code block in a try/except to catch the exceptions and then call `comm.Abort(1)` after raising the error. I have not implemented that, as I don't think we do that elsewhere in the code. But it would be a way of avoiding job hangs and having our codes exit more quickly.

Finally, the code now works in the current `main` environment. It fails in the `test-main` environment for `m1` due to a seemingly unrelated error with MPI and cuda:

```
  File "/global/homes/k/kremin/desi/python/desispec/bin//desi_proc", line 13, in <module>
    sys.exit(proc.main(args))
             ~~~~~~~~~^^^^^^
  File "/global/homes/k/kremin/desi/python/desispec/py/desispec/scripts/proc.py", line 111, in main
    comm, rank, size = assign_mpi(do_mpi=args.mpi, do_batch=args.batch, log=log)
                       ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/global/homes/k/kremin/desi/python/desispec/py/desispec/workflow/desi_proc_funcs.py", line 170, in assign_mpi
    from mpi4py import MPI
ImportError: libcudart.so.12: cannot open shared object file: No such file or directory
```

Because the error is unrelated and more about packages than this code, I would like to merge this and resolve this new issue independently.

### DIFF
--- a/py/desispec/scripts/preproc_darks_mpi.py
+++ b/py/desispec/scripts/preproc_darks_mpi.py
@@ -34,7 +34,7 @@ def preproc_darks_parser():
     parser.add_argument('-n', '--nights', type=str, default = None, required=False,
                         help='Comma separated list of YEARMMDD nights where we find the darks to run through preproc')
     parser.add_argument('--reference-night', type=int, default = None, required=False,
-                        help='YEARMMDD reference night defining the hardware state for this dark frame (default is most recent, option cannot be set at the same time as reference-expid)')
+                        help='YEARMMDD reference night defining the hardware state for this dark frame (default is most recent)')
     parser.add_argument('-c','--camword', type=str, required = True,
                         help = 'cameras to process, e.g. a0123456789')
     parser.add_argument('--bias', type = str, default = None, required=False,


sbailey (approved):
The core issue of `all_data_header_cams` being a ragged array is fixed and looks good.

I did not review all the interactions with copilot, but the final result also looks good.

Let's save the MPI robustness for a separate PR.  In general we do not use `comm.Abort(1)` because that is a really hard MPI exit. That might be fine for a single script that really should crash, but it prevents that script from being wrapped in a larger parallel job in the future.  Elsewhere we wrap things in try/except and increment an error counter that gets summed across ranks, and then the script returns the total number of errors encountered.  Outer wrappers can decide what to do with that.

I have not been able to reproduce the test-main + desi_proc mpi4py crash. Please post the exact command run.  I agree that if this works with main and the crash is specific to test-main, we could merge this anyway and debug that separately.

akremin:
I can't reproduce the error now using test-main with the updates from this PR, nor can I reproduce it using a smaller toy example. I was doing some environment manipulations under-the-hood in my PR testing, so it is possible that there were old and new desiconda packages competing and causing incompatibilities. Since I can't reproduce the error, I will not create a spinoff ticket. Here's to hoping that it was an issue with my environment during testing.

PR #2636, opened 2026-02-27
Replace binary_closing with closing.

### LINKED ISSUE #2620: skimage.morphology.binary_closing renamed to closing
While testing desispec with skimage/0.26.0:
```
  /global/common/software/desi/users/sjbailey/desispec/py/desispec/joincosmics.py:158: FutureWarning: `binary_closing` is deprecated since version 0.26 and will be removed in version 0.28. Use `skimage.morphology.closing` instead.
    bc = bc | binary_closing(bmask, footprint=se.se)
```

It looks like a simple rename, but double check that they didn't also change the API, and also check for backwards compatibility with skimage/0.22.0 (our current default version).

dylanagreen:
This simple PR replaces binary_closing with closing everywhere to fix #2620. To maintain compatibility with the "redefined" interface for when skimage is not in the environment, I also renamed `binary_closing` there to `closing`.

### DIFF
--- a/py/desispec/joincosmics.py
+++ b/py/desispec/joincosmics.py
@@ -17,7 +17,7 @@ from desispec.maskbits import specmask
 
 try:
     # Note: scikit-image is not part of desiconda.
-    from skimage.morphology import binary_closing
+    from skimage.morphology import closing
     from skimage import __version__ as _skimage_version
 except ImportError as e:
     _skimage_version = '0.0.0'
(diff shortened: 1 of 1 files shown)


sbailey (approved):
I was curious so I also confirmed that `closing` and `binary_closing` are equivalently fast.  Originally `binary_closing` was optimized for binary input, but `closing` does just as well now.

PR #2641, opened 2026-03-02
Bias runtime

jbanovetz:
This updates the `batch` and `batch_writer` script so that bias jobs have a shorter wall time (less than 30 minutes so that they can be run in the debug queue). This changes some of the parameters used to calculate the time and also fixes an issue that only one node was being asked when 2 were needed for biases.

Testing on my own prod, I get 2 nodes for all the jobs (both `biasnight` and `biaspdark`) and wall times of 15-20 minutes depending on if there were `pdarks` that needed to happen.

### DIFF
--- a/py/desispec/workflow/batch_writer.py
+++ b/py/desispec/workflow/batch_writer.py
@@ -511,7 +511,8 @@ def create_ccdcalib_batch_script(night, expids, camword='a0123456789',
         cte_expstr = ','.join(np.array(cte_expids).astype(str))
         cmd = f"desi_fit_cte_night -n {night} -c {camword} -e {cte_expstr}"
         script_body += wrap_command_for_script(cmd, nodes, ntasks=ntasks, threads_per_task=threads_per_task, stepname='ctecorr')
-
+    if nodes==0:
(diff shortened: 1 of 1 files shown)


py/desispec/workflow/batch_writer.py
```diff
@@ -505,7 +511,8 @@ def create_ccdcalib_batch_script(night, expids, camword='a0123456789',
         cte_expstr = ','.join(np.array(cte_expids).astype(str))
         cmd = f"desi_fit_cte_night -n {night} -c {camword} -e {cte_expstr}"
         script_body += wrap_command_for_script(cmd, nodes, ntasks=ntasks, threads_per_task=threads_per_task, stepname='ctecorr')
-
+    if nodes==0:
+        log.critical('No nodes allocated for job, exiting without writing script')
```
sbailey: `log.critical` just prints the log message, but it doesn't actually raise an exception or change the code flow.  Note that if we made the above suggestion of starting with nodes=1 and then using `max(nodes, ...)` this should be changed to check the runtime as the belt-and-suspenders measure.

PR #2642, opened 2026-03-02
Adjust the Resolution Matrix for Flux Calibration (v2)

dylanagreen:
In this PR I address the points brought up by @segasai in #2419, superseding a previous effort to do so in #2423.

## Mathematical Particulars (for the curious of heart)
The key different in this PR versus that one is that in this one we correctly propagate and use the "deconvolved" flux calibration vector calculated during the calibration step when calibrating the resolution matrix. To summarize:

The flux calibration in DESI is a per spectrum flux calibration vector, $C_i$. Using a hat to denote a vector in flux units, where no hat represents a vector in electron counts, the resolution matrix is derived in electron counts,

$$ f_i = R_i f $$

but has been historically (incorrectly) applied on flux vectors (for example, a redrock model of the true flux):

 $$ \hat{f}_i = R_i \hat{f} $$

so we would like to "calibrate" the resolution matrix to $\hat{R}_i$, such that  $\hat{f}_i = \hat{R}_i \hat{f}$ is true.

The flux calibration vector in DESI, $C_i$ is derived assuming that there is a global (i.e. spectrum independent) calibration vector encompassing all effects including throughputs,

 $$ {f}_i = R_i C \hat{f}_{true} $$

such that we return the calibrated flux

 $$\hat{f}_i = C_i^{-1} R_i C \hat{f}_{true} $$

$C_i$ is then calculated as

$$ C_i \approx \frac{R_i C \hat{f}_{true} }{R_i \hat{f}_{true} } $$

where the approximation is due to the fact that in the calculation we use a flat $\hat{f}_{true}$. Given these specifics, the correct, *calibrated* resolution matrix is $\hat{R}_i = C_i^{-1} R_i C$. The key different from Sergey's original PR is in the use of the deconvolved, estimated, global $C$ on the right hand side, rather than the estimate $C_i$. In other words, Sergey's original PR computed $\hat{R}_i = C_i^{-1} R_i C_i$, which we believe to be an inaccurate estimate of $\hat{R}_i = C_i^{-1} R_i C$.

## What this PR Does
This PR calibrates all resolution matrices as above, with the following changes to the fluxcalibration code:
- The `fluxcablib-*.fits.gz` files now store an additional header: `DECONVOLVED_CALIB`, which stores (as expected) the calculated deconvovled calibration vector $C$ as float64. The header has correctly set fits units which are the same as the `FLUXCALIB` units.
  - This header is only saved when $C$ is actually calculated. If the median SNR of the standard stars is < 2, $C$ is never calculated, a simple scaling calibration is applied to calculate $C_i$ instead.
  - Since $C$ itself is not calculated,  there's nothing to save.
- The `FLUXCALIB` header has been promoted to float64 (from float32) when saving. 
  - This choice, and the choice to save the `DECONVOLVED_CALIB` vector both arise from the fact that the flux calibration calculation (generating the `fluxcablib-*.fits.gz` files) is an independent step from applying the flux calibration (generating the `cframe-*.fits.gz` files), where the former from disk is loaded to perform the latter. To preserve the full resolution of the resolution matrix calibration, we prefer to do these computations with the full float64 precision.
 - The resolution matrix in the `cframe*.fits.gz` files is calibrated as above to $\hat{R}_i = C_i^{-1} R_i C$ when $C$ is available. If the loaded fluxcalib file has no `DECONVOLVED_CALIB` header, then no $C$ is loaded. In these rare cases, the resolution matrix is instead left as $\hat{R}_i = R_i$. This is not *strictly* correct, but the calculation of $C_i$ in these cases is also not *strictly* correct either, and this choice is **more** correct than any alternative.
 - The docstring for `compute_flux_calibration` has been updated to make the user aware of the face that the $C_i$ returned by that function already includes the `FLAT_TO_PSF_CORRECTION` value applied. When computing the calibrated resolution matrix I correspondingly re-divide this correction out of the calibration vector to ensure that the calibration of the resolution does not include this effect.

## Downstream Effects

To test the effects of this PR I ran a test night through the full desispec stack both with `main` (in this case, the state of desispec at the commit I branched off for this PR, since desispec has changed quite a bit in the two weeks since I started drafting this work) and the work in this branch. 

The night I used to test for this PR is **20260115**. This was a night that observed all of our currently active surveys including some tertiary49 tiles (Observed 1x BACKUP, 1x partial BRIGHT1B, 3x BRIGHT1B, 3x 83580 2x 83581, 13x DARK1B, 2x partial DARK, 10x DARK.) with a total of 180000 spectra so provides a good "all around" test.

Generally, the calibration of the resolution matrix in this way leaves most but not all of the resolution matrix unaffected. The general sum of the resolution matrix terms per pixel remains similar if not identical to the previously uncalibrated values except for a small number of pixels. The median (and also maximum) number of pixels affected in a single spectrum on the test night is **30 pixels.** 

The downstream redrock results on these updated spectra are

```
For 180000 spectra for ZWARN == 0 for both fits...
Changed Spectype: 0.000812410872477421
All redshift changes:
	 Changed z by more than dv=10: 0.01621364677412385
	 Changed z by more than dv=50: 0.005341169353096236
	 Changed z by more than dv=100: 0.0032064301456289704
	 Changed z by more than dv=500: 0.0004839894559439955
	 Changed z by more than dv=1000: 0.0002938507411088544
Changed ZWARN !=0 to ZWARN == 0: 0.004955555555555556
Changed ZWARN == 0 to ZWARN != 0: 0.005694444444444445
```
All spectra which were previously ZWARN == 0 and are now ZWARN != 0 have ZWARN == 4 (low delta chi^2).

Here is an extremely rare example of something that changed both spectype and redshift: 
spectype(6)

This is probably the clearest case that we *might* have gone to the wrong classification, but the spectrum is so low signal to noise I can't tell:
spectype(7)

Here's the histogram of the median SNR of spectra who changed spectype, most of them are quite low in SNR (there are ~5 spectra > 10), where here median SNR was calculated as median(flux * sqrt(ivar))
snr_changes

Based on my spot checks these changes in redrock are very minimal, to a very small amount of spectra. Many of the larger velocity differences (which is again a very small fraction) are broad line quasars where the fit has shifted one or two pixels. 

The results used to generate these statistics and figures is on the CFS, at `$DESI_SPECTRO_REDUX/dylang/branch_c2`

### DIFF
--- a/py/desispec/test/test_flux_calibration.py
+++ b/py/desispec/test/test_flux_calibration.py
@@ -199,6 +199,60 @@ class TestFluxCalibration(unittest.TestCase):
         with self.assertRaises(SystemExit):  #should be ValueError instead?
             apply_flux_calibration(frame,fc)
 
+    def test_apply_fluxcalibration_resolution(self):
+        """Test that apply_flux_calibration updates resolution_data as C_i^-1 * R * C"""
+        wave = np.arange(5000, 5010, dtype=float)
+        nwave = len(wave)
+        nspec = 2
+
+        flux = np.ones((nspec, nwave))
+        ivar = np.ones((nspec, nwave))
+
+        # Tridiagonal resolution matrix (width=3): identity (diagonal=1, off-diags=0)
+        width = 3
+        resolution_data = np.zeros((nspec, width, nwave))
+        resolution_data[:, 1, :] = 1.0  # set only the diagonal
+
+        frame = Frame(wave, flux.copy(), ivar.copy(),
+                      resolution_data=resolution_data.copy(), spectrograph=0)
+
+        # Known constant calibration values
+        c_val = 2.0        # convolved calib C
+        c_d_val = 3.0      # deconvolved calib C_d
+        f_val = 0.5        # FLAT_TO_PSF_FLUX correction per fiber
+
+        calib = np.full((nspec, nwave), c_val)
+        fcivar = np.ones((nspec, nwave))
+        mask = np.zeros((nspec, nwave), dtype=np.uint32)
+        deconvolved_calib = np.full(nwave, c_d_val)
+        fibercorr = {"FLAT_TO_PSF_FLUX": np.full(nspec, f_val)}
+
+        fc = FluxCalib(wave, calib, fcivar, mask,
+                       deconvolved_calib=deconvolved_calib,
+                       fibercorr=fibercorr)
+
+        apply_flux_calibration(frame, fc)
+
+        # Expected scaling: resolution_data[spec, diag_offset, wave] is scaled by
+        # C_d[wave] / (C[spec, wave] * f[spec]), where diag_offset=1 is the diagonal.
+        # With constant values and identity R (diagonal=1, off-diagonals=0):
+        #   diagonal (offset=1) -> 1 * c_d_val / (c_val * f_val)
+        #   off-diagonals (offset=0, 2) -> 0 * ... = 0 (unchanged)
+        expected_diag = c_d_val / (c_val * f_val)
+        self.assertTrue(np.allclose(frame.resolution_data[:, 1, :], expected_diag),
+                        msg='Diagonal elements do not match expected C_i^-1 * R * C scaling')
+        self.assertTrue(np.allclose(frame.resolution_data[:, 0, :], 0.0),
+                        msg='Upper off-diagonal should remain 0')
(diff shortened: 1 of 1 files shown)


sbailey (approved):

segasai:
Sorry @sbailey, I think I got confused. I can see now. 
Also just this
` R = Resolution(mean_res_data)`
just before the loop that convolves the spectra with the resolution matrix led me astray. 

So, yes, all is good then. Sorry for false alarm.

Issue #2622, opened 2026-02-25
biasnight jobs requesting way more time than needed


sbailey:
m1/run/scripts/night/20210216/biasnight-20210216-00076136-a0123456789.slurm requests almost an hour and a half of runtime:
```
#SBATCH --time=01:28:00
```

However, running that script by hand in the interactive queue, it finished in a minute and a half:
```
desi@nid004272:/global/cfs/cdirs/desi/spectro/redux/m1/run/scripts/night/20210216> source biasnight-20210216-00076136-a0123456789.slurm |& tee biasnight-20210216-00076136-a0123456789-$SLURM_JOBID.log
Starting at Wed Feb 25 09:07:39 AM PST 2026
Running srun -N 2 -n 15 -c 32 --cpu-bind=cores desi_proc --cameras a0123456789 -n 20210216 --nightlybias --mpi
INFO:proc.py:117:main: rank 0 started /global/common/software/desi/perlmutter/desiconda/20260220-2.3.0/code/desispec/main/py/desispec/scripts at 2026-02-25T09:07:46.614591
...
INFO:proc.py:336:main: All done at Wed Feb 25 09:09:26 2026; duration 1m31s
biasnight succeeded at Wed Feb 25 09:09:28 AM PST 2026
All done at Wed Feb 25 09:09:28 AM PST 2026
```

Track why the request time is so long, and hopefully reduce it to a few minutes so that the jobs get through the queue faster.

jbanovetz:
So looking at the `biaspdark` logs from 2025-2026, 15 minutes looks like a good number. Here are the runtimes from the logs:

<img width="543" height="453" alt="Image" src="https://github.com/user-attachments/assets/ff7786e6-d09d-46d8-bb22-eca4ead1c9f8" />

While most are around 3-4 minutes, there are some that go for 15~ish minutes. I will investigate the dates later to see why these times are so long.

akremin:
Completed with PR #2641 . The new distribution for `biaspdark` jobs is between 8 minutes and ~25 minutes. The bottom is explicitly set after discussion about python startup times and the top is set by nights with lots of darks. All `biasnight` jobs now have 8 minute runtimes.

PR #2643, opened 2026-03-03
Update bias and pdark resource logic

akremin:
This solves the underlying problem leading to issue #2622 . The problem was lines 186 and 187 in current main:
https://github.com/desihub/desispec/blob/0f1c09175e11230d68300bfee2f9be0ed36c2308/py/desispec/workflow/batch.py#L186C1-L187C83

Essentially it is trying to compute the number of loops through the data we need and adding runtime associated with each loop. The number of total tasks is `ncameras*nexps`, and the total number of tasks we can do in one loop is the number of cores `nodes*cores_per_node`. So the correct scaling should be `np.ceil( ncameras*nexps / nodes*cores_per_node )`. 

I have implemented that, and also simplified things such that if we have a biaspdark job, one call to `determine_resources()` returns the total runtime, rather than having the script writer do computation related to that. 

I ran through a loop of calling determine_resources() with a number of different ncameras, nexps, and jobdescs. Below are the plots. We see the scaling of runtime with exposures and cameras was completely wrong before. We see that it is now much better and does what is expected. Note that below 15 cameras, only one node is requested for bias-type jobs so there are discontinuities at that number because we suddenly double the available nodes, cores, and halve the number of loops required.

Current main:
resource_heatmaps_main

This branch:
resource_heatmaps_biastiming

### DIFF
--- a/py/desispec/workflow/batch_writer.py
+++ b/py/desispec/workflow/batch_writer.py
@@ -276,7 +276,8 @@ def create_biaspdark_batch_script(night, expids,
 
     Args:
         night (str or int): The night in which the biaspdark script will be run.
-        expids (list of int or np.array): The exposure id(s) for the data.
+        expids (list of int or np.array): The exposure id(s) for the data. These are the
+            dark expids if pdark or biaspdark is being run. Otherwise it is a zero expid.
         jobdesc (str, optional): Description of the job to be performed. If None, will
             default to 'biaspdark' or 'pdark' depending on do_biasnight and do_pdark.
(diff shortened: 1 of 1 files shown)


akremin:
My end-to-end test completed with 0 failures. All jobs requested and used 2 nodes (they were all bias[night,pdark] jobs, so that is correct). The runtimes were also much more reasonable, possibly even too short for the biasnight jobs at 5 minutes. But I think John's sanity checks this should be fine as long as Perlmutter stays happy:

image

Issue #2182, opened 2024-02-21
Heliocentric corrections variation with the field


segasai:
Hi, 

Currently DESI applies heliocentric corrections based on the ra,dec of the center of the field, as opposed to star by star. 
The difference between those can be computed easily: 
![image](https://github.com/desihub/desispec/assets/109513/5957f224-ca3a-4227-931b-d9262b12e3d7)
It can reach .8 km/s and  has a stddev of 0.35 km/s

This effect is detectable, by looking at the delta(barycentric correction) (i.e. barycentric correction computed for the center of the field  minus barycentric correction for a given star) vs delta RV in external surveys (SoS in this case).
The plot is made using z-band only RVSpecfit measurements in iron single exposures.
![image](https://github.com/desihub/desispec/assets/109513/9b64f022-e42e-4a71-831e-26e643573829)

Given that I was able for the first time detect this, one can try to start thinking about how to actually fix that.

And the easiest way of doing that I think is through trace_shifts.py where one can simply add a term for barycentric correction of the fiber minus barycentric correction of the center of the frame.

Thoughts, comments ?

sazabi4:
I do want to point out that the shift is significant for a good fraction of stars and therefore will affect our science results using these stars. So it would be good to fix this in Year 3 data release.

djschlegel:
Adding my old notes on this:
Thinking about the barycentric corrections again, the worst-case
scenario for errors by using the tile center instead of the fiber position is
   (cos(45+1.6 deg) - cos(45 deg)) * 30 km/s = 0.6 km/s
Across a tile, that’s 1.2 km/s peak-to-peak error.
(I don't see how we get the larger 0.8 km/s shown above.)

Sky-subtraction needs to be done on the as-observed wavelength
scale, meaning either before the barycentric correction or with the
same done per fiber.

But a way to fix this would be to make small offsets to the final resolution
matrix to account for the residual per-fiber errors.  The central fibers
have no change, nor do fibers offset in the perpendicular direction to
the earth’s motion vector.  Fibers parallel to that motion would have
their resolution matrices shifted by ~0.01 - 0.02 pix.

This approach would give us data products that are unchanged, with
the same wavelength scale for all fibers.  But for the detailed analysis,
like projecting stellar models to the as-observed spectra, this would
formally have the correct velocity information.

segasai:
[...]

PR #2648, opened 2026-03-06
Applying heliocentric correction offsets to the resolution matrices in cframes

segasai:
This is a test patch for adjusting spectra for incorrect barycentric shifts before doing the coadd (to address #2182 )

This applies the idea outlined in https://github.com/desihub/desispec/issues/2182#issuecomment-4007096973
where the resolution matrix is adjusted for the true_helicor != applied_heliocor

It also writes the HELIOCOR variable into spectra files. 

I haven not yet tested this whole patch, but only the resolution matrix shifting code.

### DIFF
--- a/py/desispec/heliocentric.py
+++ b/py/desispec/heliocentric.py
@@ -24,8 +24,6 @@ try:
 except TypeError:
     kpno = None
 
-log = get_logger()
-
 
 def heliocentric_velocity_corr_kms(ra, dec, mjd) :
     """Heliocentric velocity correction routine.
@@ -135,6 +133,7 @@ def heliocentric_shift_res_data(fibermap, resolution_data, wave,
         heliocor_offset: (nspec) array of applied heliocor corrections
     """
     
+    log = get_logger()
     nspec, ndiag, nwave = resolution_data.shape
     shifted_res_data = np.zeros_like(resolution_data)
 


sbailey:
Quick clarification note: I don't think we should apply this at the time of extraction, since that could negatively impact sky subtraction as @julienguy noted.  I think we should consider applying it at the time of creating the cframe files, i.e. after we've solved for the sky model and flux calibration and are calculating flux = (photons - sky)/calibration.  Or perhaps when regrouping cframe -> spectra.  I agree that it is hard to do the equivalent shifts after the coadd is made, while in principle a user could do this themselves on the pre-coadded data, so the question is how much of this do we want to enforce / pre-calculate for the end-user.

sbailey:
OK, yes, let's move this logic into py/desispec/scripts/procexp.py so that the resolution matrices in the cframe (and thus spectra) are already offset, and the coadd doesn't have to do anything new.

 > Also, applying the correction at extraction requires I think being more explicit, what the resolution matrix columns K_ij really are -- discrete samples of a smooth extraction kernel in wavelength corresponding for every extracted pixel...

I'm not sure I understand this concern. It seems the two paths are equivalent for the purposes of the coadd and redshift fitting:
* current PR: save unshifted R, read unshifted R, shift R in memory, coadd
* proposed update: shift R, save shifted R, read shifted R, coadd

with the only difference being whether the uncoadded cframe and spectra files have the shifted or the unshifted version of R, but the actual shift logic is the same and the impact on the coadd and redshifts is identical.

If I'm missing something, let's resolve that before doing the refactor.

Where to save the shifts: if we move the shift logic into the generation of the cframes, this has the further advantage of being able to keep the existing HELIOCOR header keyword and the extraction-time shift since it applies to all fibers in the entire cframe, and a new fibermap column HELIOCOR_OFFSET could then store per-fiber offsets.  That would be for debugging / information only; it would not be needed by the downsteam coadd or analysis code. Do you agree?

Issue #2654, opened 2026-03-07
m2 20250531 and 20250401 ccdcalib failed due to dark reset logic


sbailey:
From [m2/run/scripts/night/20250331/ccdcalib-20250331-00286417-a0123456789-49750667.log](https://data.desi.lbl.gov/desi/spectro/redux/m2/run/scripts/night/20250331/ccdcalib-20250331-00286417-a0123456789-49750667.log)

```
...
called from line 111 in /global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desispec/main/py/desispec/scripts/compute_dark_night.py
ERROR:util.py:182:runcmd: Traceback (most recent call last):
ERROR:util.py:182:runcmd: File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desispec/main/py/desispec/util.py", line 160, in runcmd
    result = cmd(*args)
ERROR:util.py:182:runcmd: File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desispec/main/py/desispec/scripts/compute_dark.py", line 337, in main
    compute_dark_file(args.images, args.outfile, camera=args.camera, bias=args.bias,
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                      nocosmic=args.nocosmic,
                      ^^^^^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
                      min_dark_exposures=args.min_dark_exposures,
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                      max_dark_exposures=args.max_dark_exposures)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR:util.py:182:runcmd: File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desispec/main/py/desispec/ccdcalib.py", line 147, in compute_dark_file
    elif calib.data['DATE-OBS-BEGIN']>dark_reset_end:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR:util.py:182:runcmd: TypeError: '>' not supported between instances of 'str' and 'int'
[...]

PR #2659, opened 2026-03-10
Cast yaml dates as int for dark reset logic

jbanovetz:
This is designed to fix https://github.com/desihub/desispec/issues/2654 by making sure that all of the dates are cast as `int`. In the yaml files, sometimes the dates (e.g. 20250331) are `int` types and others are of type `str`.  This PR should fix this by casting all the dates as type `int` in that logic chain.


sbailey:
It looks like this change would work, but I'm going to delay us just a bit further: I meant to suggest doing the int-casting once in the `desispec.calibfinder.CalibFinder` constructor instead of in `desispec.ccdcalib.compute_dark_file`.  Then all uses of a `CalibFinder` would automatically get `DATE-OBS-BEGIN` as an int and `compute_dark_file` wouldn't need any changes.

sbailey (approved):

Issue #2674, opened 2026-03-17
Override files for Matterhorn


akremin:
In debugging issue #2656 , I noticed that that night is only trying to link the badcol files from a previous night. In looking, however, there are 1200s darks available on the night which are not flagged. Historically we required a 300s dark for the purpose of bad column identification, but changed that when we moved to darknights. Now we only take 1200s darks on a night and use those for bad column identification. Since we use these now successfully, we should be able to use the historic 1200s darks. This will eliminate the need to link to other nights for three historic cases:

Night **20230309**
```
EXPID  OBSTYPE  EXPTIME    CAMWORD   BADCAMWORD BADAMPS
------ ------- --------- ----------- ---------- -------
170910    dark 1200.0545 a0123456789
170911    dark 1200.0596 a0123456789
```

Night **20231220**
```
EXPID  OBSTYPE  EXPTIME    CAMWORD   BADCAMWORD BADAMPS
------ ------- --------- ----------- ---------- -------
211365    dark 1200.0368 a0123456789
211366    dark 1200.0605 a0123456789
```

Night **20241123**
```
EXPID  OBSTYPE  EXPTIME    CAMWORD   BADCAMWORD BADAMPS
------ ------- --------- ----------- ---------- -------
264645    dark 1200.0365 a0123456789
264646    dark 1200.0444 a0123456789
```

There are still 5 historic nights where we do not have any valid darks and will need to link, but this reduces the number if we can use the 1200s darks. I have asked @jbanovetz to assess the darks to make sure there are no glaring issues why we can't use them. Since some cals on the night are flagged, it is worth being cautious.

PR #2679, opened 2026-03-20
fix BADCOLUMN vs. BADFIBER typo; add tests that would have caught it

sbailey:
While checking #2674 I found that inspect_dark was using `fibermask.BADCOLUMN` which doesn't exist at the fibermask level (it does at the specmask level).  I have updated this to `fibermask.BADFIBER` which is less descriptive but does exist.

With the help of the LBL cborg coder, I also added unit tests that would have caught this in the first place and adds coverage for inspect_dark which previously was not covered by any unit test.


sbailey:
Curious.  Tests pass on my laptop with a 184-sigma significant bad column 10, but failed for at least one combination here.  Investigating.

Issue #2585, opened 2026-01-12
Problematic emline result


rongpu:
@araichoor I found (from looking at the emline results from https://github.com/desihub/desispec/issues/2456) this bizarre case where the original (Loa) OII_FLUX and OII_FLUX_IVAR are normal but a slightly different redshift produces unreasonably large OII_FLUX and unreasonably small OII_FLUX_IVAR.

```
TARGETID            Z                   OII_FLUX    OII_FLUX_IVAR
39627618835760858   0.0987038947889492  12.493738   0.042262517
39627618835760858   0.0986977191929144  20570.734   4.2293852E-11
```

https://inspector.desi.lbl.gov/dr2/spectra/tiles/39627618835760858?format=html&plotnoise=1
<img width="835" height="409" alt="Image" src="https://github.com/user-attachments/assets/9256f3fe-41b7-4f2f-a4ad-814afb5f8ab5" />

The relevant files are:
`/pscratch/sd/r/rongpu/tmp/emline/emline-main-dark-25529-39627618835760858.fits` (the original emline)
`/pscratch/sd/r/rongpu/tmp/emline/emline-main-dark-25529_dz1.5-39627618835760858.fits` (the problematic result)

The emline PDF files are below. The second PDF is the problematic result, but it does not show anything wrong (except for the very low OII SNR).
[redrock-main-dark-25529-39627618835760858.pdf](https://github.com/user-attachments/files/24576006/redrock-main-dark-25529-39627618835760858.pdf)
[redrock-main-dark-25529_dz1.5-39627618835760858.pdf](https://github.com/user-attachments/files/24576007/redrock-main-dark-25529_dz1.5-39627618835760858.pdf)

sbailey:
Fixed by PR #2685. Closing.

PR #2685, opened 2026-03-24
Fix emission line fitting

rongpu:
This PR fixes the emline afterburner issue described in https://github.com/desihub/desispec/issues/2585. It includes the following changes:
1. Changing the curve_fit bounds to use physically motivated limits: 
  a. minimum FWHM of 0.8 Angstrom
  b. 0.35 < I(3729) / I(3726) < 1.5
2. For the model flux, integrate over the wavelength pixel (which is given by the error function for the Gaussian profile)
3. A bug fix for the CHI2 calculation (which should be proportional to 1 * ivars, not 1 / ivars**2); I don't think CHI2 value is used anywhere downstream.

Here are the before and after plots for a previously problematic emline fit:

TARGETID = 39628465707680538
image
image

And as a sanity check, here is a good emline fit from before, and as expected the fit changed very little:

TARGETID = 39627730592995420
image
image

I can run the new emline fitter on a few tiles and report back.

### DIFF
--- a/py/desispec/test/test_emlinefit.py
+++ b/py/desispec/test/test_emlinefit.py
@@ -7,7 +7,8 @@ import unittest
 import tempfile
 
 import numpy as np
-from desispec.emlinefit import get_emlines
+from scipy.special import erf
+from desispec.emlinefit import emlines_gaussfit, get_emlines
 
 #- some tests require data only available at NERSC
 _everest = '/global/cfs/cdirs/desi/spectro/redux/everest'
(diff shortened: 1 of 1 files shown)


sbailey:
@rongpu thanks for those additional checks and plots.  It is nice to see the progress of "most of the 0.4% lost are problematic OII fits and should have been rejected" and a little disconcerting that we've been living with those bad fits for so long.  This is a nice algorithmic update for Matterhorn.  I'll merge as soon as tests finish running.

Issue #2701, opened 2026-04-03
b8 arc bundles failing on 20221121


akremin:
Bundles 10 and 11 are failing for all 5 exposures of b8 on 20221121. `b8B` is flagged as a badamp, but that was flagged in Kibo as well and it still produced outputs. 

logs:
https://data.desi.lbl.gov/desi/spectro/redux/m3/run/scripts/night/20221121/psfnight-20221121-00154099-a02346789-50549430.log
https://data.desi.lbl.gov/desi/spectro/redux/m3/run/scripts/night/20221121/arc-20221121-00154099-a02346789-50337448.log  
https://data.desi.lbl.gov/desi/spectro/redux/m3/run/scripts/night/20221121/arc-20221121-00154101-a02346789-50337450.log  
https://data.desi.lbl.gov/desi/spectro/redux/m3/run/scripts/night/20221121/arc-20221121-00154103-a02346789-50337453.log
https://data.desi.lbl.gov/desi/spectro/redux/m3/run/scripts/night/20221121/arc-20221121-00154100-a02346789-50337449.log  
https://data.desi.lbl.gov/desi/spectro/redux/m3/run/scripts/night/20221121/arc-20221121-00154102-a02346789-50337452.log

PR #2714, opened 2026-04-08
Changed to look at Bundle Status to determine if something was missing or failed

jbanovetz:
The goal of this PR is to determine if bundles were missing due to a bad amplifier. This will be implemented at the same time as a PR in specex to change the values of different outcomes such that missing bundles return a STATUS of -1 and failed bundles return a STATUS of 4.

This code change only should affect the `psfnight` job when averaging PSFs. It will record the bundles that are flagged as missing and then not include them in the final averaging if they appear in all 5 exposures. This should fix the issue seen here: https://github.com/desihub/desispec/issues/2701

### DIFF
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -534,7 +534,7 @@ def mean_psf(inputs, output):
     bundle_rchi2=[]
     nbundles=None
     nfibers_per_bundle=None
-
+    missing_bundles=[]
     for input in inputs :
         log.info("Adding {}".format(input))
         if not os.path.isfile(input) :
@@ -574,6 +574,15 @@ def mean_psf(inputs, output):
         nbundles=rchi2.size
         bundle_rchi2.append(rchi2)
 
+        #Check for missing bundles
+        t=psf["PSF"].data
+        i=np.where(t["PARAM"]=="BUNDLE")[0][0]
+        i_status=np.where(t["PARAM"]=="STATUS")[0][0]
+        bundles=t["COEFF"][i][:,0].astype(int)
+        status=t["COEFF"][i_status][:,0].astype(int)
+        missing_mask=(status<0)
+        missing_bundles.append(bundles[missing_mask])
+
     npsf=len(tables)
     bundle_rchi2=np.array(bundle_rchi2)
     log.debug("bundle_rchi2= {}".format(str(bundle_rchi2)))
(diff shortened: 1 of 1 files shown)


jbanovetz:
It seems like it worked as intended as the `psfnight` job was successfully ran! (see here: `/global/cfs/cdirs/desi/spectro/redux/banovetz/run/scripts/night/20221121/psfnight-20221121-00154099-a02346789-51244679.log`

I will run it on a different night that doesn't have a bad amp and test it there as well.

jbanovetz:
I implemented Copilots changes and initialize the `output_coeff` to -1 to get past silently accepting and changed it so that it only counts a bundle as missing if all of the fibers are missing. I tested this with `20221121` and `b8` and it works. The only difference is before bundle 10 was included as missing and now isn't. Investigating, I think that is because bundle 10 is on the amp boundary and has some fibers missing because of the bad amp. Bundles 11-19 were successfully skipped though, as expected.

akremin:
Thank you, please also test on a night with all good data to make sure this never triggers the skipping of a bundle. Then we want a third test where 2-3 exposures have a bundle issue but the others don't. We want to run that and verify that the code doesn't skip those bundles since they aren't skipped on all petals. Finally we want to flag too many badamp that we can't create the bundles and we still get a crash (expected).

I've setup a test prod for this:
`/global/cfs/cdirs/desi/spectro/redux/fiber_bundle_bad_amp`

Night 1 (1 arcs have a badamp): 20260320 (expect no skipped bundles and no crash)
Night 2 (all 5 arcs have a badamp): 20260321 (expect skipped bundles and no crash)
Night 3 (no arcs have a badamp): 20260322 (expect no skipped bundles and no crash)
Night 4 (4 arcs have a badamp): 20260323 (expect a crash since can't take median of 1 but not all 5 are flagged)

akremin:
I've linked all the relevant input cals, so you should be able to just run the psf's and psfnight commands in an interactive node. Please post results showing we get the expected behavior in each of the 4 cases.

jbanovetz:
I ran these tests last night and got these results for the different nights:

Night 1 (1 arcs have a badamp): 20260320 (expect no skipped bundles and no crash)
Result: No skipped bundles and no crash

Night 2 (all 5 arcs have a badamp): 20260321 (expect skipped bundles and no crash)
Result: skipped bundles and CRASH This is because bundle 10 is included where some of the fibers are in amp B and some are not. This was fine before when I checked because I had it set up that if ANY of the fibers were missing, then skip the bundle.  Now it is set up that only if ALL of the fibers in a bundle are missing to skip it.

Night 3 (no arcs have a badamp): 20260322 (expect no skipped bundles and no crash)
No skipped bundles no crash.

Night 4 (4 arcs have a badamp): 20260323 (expect a crash since can't take median of 1 but not all 5 are flagged)
No skipped bundles and crash

jbanovetz:
Rerunning just the `psfnight` job for 20260321 did succeed and successfully skipped the missing bundles. I am investigating but It might have just been an issue of running not with my branch.

jbanovetz:
Looks like it was just a case of running with the wrong branch! All the tests have past and the corresponding logs are in the corresponding `/global/cfs/cdirs/desi/spectro/redux/fiber_bundle_bad_amp/run/scripts/night/` folders with `mydesispec_psfnight_*`. I did also confirm with 20260321 that it works with `desi_proc_night` using the command: `desi_proc_night --proc-obstypes=bias,dark,arc -q regular -n $NIGHT --dont-require-cals`

akremin:
Thanks for the overnight testing, John. Happy to hear we have at least 3/4 with correct behavior. Let us know what you find for Night 3. I thought the initial crash was because we didn't have any fibers in a bundle, but it is possible that the threshold is more than 1 valid fiber. Maybe we need to tune that to this case where a badamp removes some but not all fibers. Or maybe it's not an issue and re-testing will show that we're fine.

akremin:
> Looks like it was just a case of running with the wrong branch! All the tests have past and the corresponding logs are in the corresponding `/global/cfs/cdirs/desi/spectro/redux/fiber_bundle_bad_amp/run/scripts/night/` folders with `mydesispec_psfnight_*`. I did also confirm with 20260321 that it works with `desi_proc_night` using the command: `desi_proc_night --proc-obstypes=bias,dark,arc -q regular -n $NIGHT --dont-require-cals`

Great news!

julienguy:
- I checked that the 500 fibers of b8 with the bad amp have STATUS=-1 and all other coefficients are the values of the input psf (not zero), which I think is OK.
- One minor comment would to be add to the message "Bundle XXX is missing in all input PSFs, likely due to a bad amp. This bundle will be ignored in the merging." the name of camera, otherwise it's impossible to know which camera is affected when reading the lok.
- Except for that I think this PR is ready to be merged.

jbanovetz:
Changed to now include the camera. The output is:

`WARNING:specex.py:613:mean_psf: Bundle 19 is missing in all input PSFs for camera 'b8      ', likely due to a bad amp. This bundle will be ignored in the merging.`

akremin (approved):
Thanks these changes look good. A very productive feedback loop with copilot here to catch initialization concerns. I think it does what we want now and my independent testing shows the correct behavior.

PR #2722, opened 2026-04-17
use_reservation add biasnight,biaspdark; handle DependencyNeverSatisfied

sbailey:
This PR updates desi_use_reservation to match what we are doing for Matterhorn:
1. don't move "DependencyNeverSatisfied" jobs into the reservation (!)
2. prioritize psfnight and arc CPU jobs to unblock downstream GPU jobs to be able to run

(2) is what we want for the case where there aren't enough eligible GPU jobs and we need to get CPU jobs through the queue to unblock GPU jobs.  There is another extreme where we run out of CPU jobs but are so backlogged on GPU jobs that we can't submit more nights.  It is unclear to me if we'd want to do anything different in that case since the core problem is that we need to finish GPU jobs as fast as possible, regardless of order.

### DIFF
--- a/bin/desi_use_reservation
+++ b/bin/desi_use_reservation
@@ -79,7 +79,7 @@ def use_reservation(name=None, resinfo=None, extra_nodes=0, dry_run=False):
     ressize = resinfo['node_count']
 
     #- job types for CPU and GPU, in the order they should be considered for the reservation
-    cpujobs = ['linkcal', 'ccdcalib', 'psfnight', 'arc']
+    cpujobs = ['psfnight', 'arc', 'ccdcalib', 'biasnight', 'biaspdark', 'linkcal']
     gpujobs = ['nightlyflat', 'flat', 'ztile', 'tilenight', 'zpix']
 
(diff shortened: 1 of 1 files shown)


akremin (commented):
Thanks, the current updates look good, but I will leave this open over the weekend to make sure that we don't want to make any other changes. I don't think we have any pressing need to merge this at this time.

One question to monitor is whether the change of CPU priorities will continue to be optimal if we enter a GPU-bottlenecked queue state. During the testing, this has worked well, but we have been in a CPU-bottlenecked state where those are the limiting factor and getting arcs+psfnights done enabled more GPU jobs to be freed up.

sbailey:
Over the past week we've had reservations with a 1:4 CPU:GPU node ratio, and using this code to move jobs from the regular queue into the reservation.  This seemed to work well to free up both CPU and GPU jobs to run.  We started the first reservation in an unbalanced state (not enough GPU jobs eligible, blocked by pending CPU jobs), and this prioritization order plus the 1:4 reservations got us out of it and stayed balanced for the duration of the runs.

Issue #2723, opened 2026-04-20
Psfnight not working with nights where some of the amps are flagged


jbanovetz:
With Matterhorn, there are two nights (20211028 and 20211106) where 2 of the arcs have a bad amp flagged. This causes a problem with psfnight that now throws an error whenever 2 or more bundles of fibers have missing PSF measurements. This is also not caught by this PR: https://github.com/desihub/desispec/pull/2714 since it is only 2 and not all of the exposures.

I think a fix that we could implement is to only count fit failures towards counting total errors instead of including both fit failures and missing bundles. With the PR mentioned above, it should be easy to distinguish the two and I will look into making this change.

jbanovetz:
I have started to work on this and have a solution though I think there is a better one (in branch `missingamp_some_exposures`).

The change is to include one line before the new failure check


```
for bundle in fibers_in_bundle.keys() :

            ok=np.where((bundle_rchi2[:,bundle]>0)&(bundle_rchi2[:,bundle]<rchi2_threshold))[0]

            if entry==0 :
                log.info("for fiber bundle {}, {} valid PSFs".format(bundle,
                    ok.size))
            
            if bundle not in all_missing_bundles.flatten(): #New line
                nfailed = np.sum(bundle_rchi2[:,bundle]==0)
                if nfailed > 1 :
                    message=f"{nfailed} fit failures for bundle {bundle} indicate potential issue with unmasked CCD features or with the input PSF."
                    log.critical(message)
                    raise RuntimeError(message)
```

In this case, `all_missing_bundles` are bundles identified as missing due to a bad amp in any of the exposures. So essentially we skip over this fit failure check if ANY of the exposures have a missing bundle. I did confirm with 20211028 that this does work and produces `psfnight` for r8.

HOWEVER, after talking with Anthony, we discovered a possible corner case. This change will skip over bundles if ANY of the exposures have the bundle as missing and skip over the check for ALL exposures. So there could be a case where the bundle is missing in one exposure and ends up failing in another two exposures, there won't be a failure reported and the script would continue.

I think the better solution would be to do something more bookkeeping and mask out exposures and bundles that are missing. But this error only affects two nights (or have thus far) and so we could just check if there are failed fits on the other non-flagged amp exposures. That said, there are maybe other corner cases not thought of yet that we could potentially be silently letting through with the current method that I present. I think the extra bookkeeping could help reduce the likeness of an unexpected corner case.

Issue #91, opened 2015-11-03
desi_fit_stdstars.py should be flexible about what filters it needs


sbailey:
Issue #89 with flux calibration crashing was due to desi_fit_stdstars.py not finding SDSS_R as a magnitude in the input standard stars.  PR #90 fixes this for now, but it raises several to do items:

* if desi_fit_stdstars.py can't find a filter that it needs, it should raise an error and stop instead of writing a bogus file
* we need unit tests on the underlying code
* eventually the real data will have a mix of DECam, Bok, and Mosaic fluxes, and desi_fit_stdstars should work flexibly with what it can find.  Exactly how those filters will be named will be determined by the Legacy Survey team, but this week at Argonne is a good chance to talk to them about their future plans for including these filters.

sbailey:
Fixed in PR #90

PR #2732, opened 2026-04-23
Bad amp in only some of the input arc exposures

jbanovetz:
This pull request is to fix an issue we were running into with matterhorn where some of the input arcs (>2) had bad amps which triggered an the error which triggers when more than 2 arcs have `rchi==0`, usually corresponding to a fit failure. Because it was not all of the arcs, this skips over a previous check. More information can be found here: https://github.com/desihub/desispec/issues/2723

My solution is to mask known exposures with missing bundles from the check for fit failures. I tested this with 20211028 and this successfully ran and produced the expected logs with the warning. See `/global/cfs/cdirs/desi/spectro/redux/banovetz/run/scripts/night/20211028/psfnight-20211028-00106397-a0123456789-51950571.log` for the log.

### DIFF
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -668,12 +668,16 @@ def mean_psf(inputs, output):
             
             # We use all_missing_bundles since we want to avoid flagging bundles are flagged as 
             # bad for only some of the input exposures
-            if bundle not in all_missing_bundles.flatten():
-                nfailed = np.sum(bundle_rchi2[:,bundle]==0)
-                if nfailed > 1 :
-                    message=f"{nfailed} fit failures for bundle {bundle} indicate potential issue with unmasked CCD features or with the input PSF."
-                    log.critical(message)
-                    raise RuntimeError(message)
+            # if bundle not in all_missing_bundles.flatten():
+            mask=[bundle in missing for missing in missing_bundles]
+            masked_bundle_rchi2 = bundle_rchi2.copy()[~np.array(mask)]
+            if len(masked_bundle_rchi2)!= len(bundle_rchi2) :
+                log.warning(f"Bundle {bundle} is missing in some but not all input PSFs for camera {refhead['CAMERA']}. Only {len(masked_bundle_rchi2)} PSFs will be used.")
+            nfailed = np.sum(masked_bundle_rchi2[:,bundle]==0)
+            if nfailed > 1 :
+                message=f"{nfailed} fit failures for bundle {bundle} indicate potential issue with unmasked CCD features or with the input PSF."
+                log.critical(message)
+                raise RuntimeError(message)
 
             # We finally resorted to use a mean instead of a median here for two reasons.
             # First, there is already a vetting of PSF bundles with good chi2 above
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -663,13 +663,17 @@ def mean_psf(inputs, output):
             if entry==0 :
                 log.info("for fiber bundle {}, {} valid PSFs".format(bundle,
                     ok.size))
-
-
-            nfailed = np.sum(bundle_rchi2[:,bundle]==0)
-            if nfailed > 1 :
-                message=f"{nfailed} fit failures for bundle {bundle} indicate potential issue with unmasked CCD features or with the input PSF."
-                log.critical(message)
-                raise RuntimeError(message)
+            
+            # This checks and makes sure that we are only looking for bundles with fit failures
+            
+            # We use all_missing_bundles since we want to avoid flagging bundles are flagged as 
+            # bad for only some of the input exposures
+            if bundle not in all_missing_bundles.flatten():
+                nfailed = np.sum(bundle_rchi2[:,bundle]==0)
+                if nfailed > 1 :
+                    message=f"{nfailed} fit failures for bundle {bundle} indicate potential issue with unmasked CCD features or with the input PSF."
+                    log.critical(message)
+                    raise RuntimeError(message)
 
             # We finally resorted to use a mean instead of a median here for two reasons.
             # First, there is already a vetting of PSF bundles with good chi2 above
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -664,11 +664,9 @@ def mean_psf(inputs, output):
                 log.info("for fiber bundle {}, {} valid PSFs".format(bundle,
                     ok.size))
             
-            # This checks and makes sure that we are only looking for bundles with fit failures
-            
-            # We use all_missing_bundles since we want to avoid flagging bundles are flagged as 
-            # bad for only some of the input exposures
-            # if bundle not in all_missing_bundles.flatten():
+            # Only count fit failures for exposures where this bundle is present.
+            # Exposures listed in missing_bundles are excluded so that bundles
+            # missing from only some input PSFs are not treated as failed fits.
             mask=[bundle in missing for missing in missing_bundles]
             masked_bundle_rchi2 = bundle_rchi2.copy()[~np.array(mask)]
             if len(masked_bundle_rchi2)!= len(bundle_rchi2) :
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -664,18 +664,18 @@ def mean_psf(inputs, output):
                 log.info("for fiber bundle {}, {} valid PSFs".format(bundle,
                     ok.size))
             
-            # Only count fit failures for exposures where this bundle is present.
-            # Exposures listed in missing_bundles are excluded so that bundles
-            # missing from only some input PSFs are not treated as failed fits.
-            mask=[bundle in missing for missing in missing_bundles]
-            masked_bundle_rchi2 = bundle_rchi2.copy()[~np.array(mask)]
-            if len(masked_bundle_rchi2)!= len(bundle_rchi2) :
-                log.warning(f"Bundle {bundle} is missing in some but not all input PSFs for camera {refhead['CAMERA']}. Only {len(masked_bundle_rchi2)} PSFs will be used.")
-            nfailed = np.sum(masked_bundle_rchi2[:,bundle]==0)
-            if nfailed > 1 :
-                message=f"{nfailed} fit failures for bundle {bundle} indicate potential issue with unmasked CCD features or with the input PSF."
-                log.critical(message)
-                raise RuntimeError(message)
+                # Only count fit failures for exposures where this bundle is present.
+                # Exposures listed in missing_bundles are excluded so that bundles
+                # missing from only some input PSFs are not treated as failed fits.
+                mask=[bundle in missing for missing in missing_bundles]
+                masked_bundle_rchi2 = bundle_rchi2.copy()[~np.array(mask)]
+                if len(masked_bundle_rchi2)!= len(bundle_rchi2) :
+                    log.warning(f"Bundle {bundle} is missing in some but not all input PSFs for camera {refhead['CAMERA']}. Only {len(masked_bundle_rchi2)} PSFs will be used.")
+                nfailed = np.sum(masked_bundle_rchi2[:,bundle]==0)
+                if nfailed > 1 :
+                    message=f"{nfailed} fit failures for bundle {bundle} indicate potential issue with unmasked CCD features or with the input PSF for camera {refhead['CAMERA']}."
+                    log.critical(message)
+                    raise RuntimeError(message)
 
             # We finally resorted to use a mean instead of a median here for two reasons.
             # First, there is already a vetting of PSF bundles with good chi2 above
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -244,6 +244,15 @@ def main(args=None, comm=None):
                           "value {} running {}".format(rank, retval, comstr))
                 failcount += 1
                 failed_bundles.append(b)
+                # f =  "{}_{:02d}.fits".format(outroot, b)
+                # other_psf_hdulist=fits.open(f)
+
+                # # look at what fibers where actually fit
+                # i=np.where(other_psf_hdulist["PSF"].data["PARAM"]=="STATUS")[0][0]
+                # status_of_fibers = \
+                #     other_psf_hdulist["PSF"].data["COEFF"][i][:,0].astype(int)
+                # log.info(f'status_of_fibers: {status_of_fibers}')
+                bundles_to_merge.append(b)
         else:
             log.info(f"proc {rank} succeeded generating {outbundlefits}")
             bundles_to_merge.append(b)
@@ -305,6 +314,13 @@ def main(args=None, comm=None):
             log.warning(f"The fit of the following bundles failed: {failed_bundles}")
             for x in failed_bundles :
                 f =  "{}_{:02d}.fits".format(outroot, x)
+                # other_psf_hdulist=fits.open(f)
+
+                # # look at what fibers where actually fit
+                # i=np.where(other_psf_hdulist["PSF"].data["PARAM"]=="STATUS")[0][0]
+                # status_of_fibers = \
+                #     other_psf_hdulist["PSF"].data["COEFF"][i][:,0].astype(int)
+                # log.info(f'status_of_fibers: {status_of_fibers}')
                 if  os.path.isfile(f):
                     os.remove(f)
 
@@ -457,15 +473,20 @@ def merge_psf(inpsffile, inputs, output):
 
     for input_filename in inputs :
         log.info("merging {} into {}".format(input_filename, inpsffile))
-        other_psf_hdulist=fits.open(input_filename)
+        other_psf_hdulist=fits.open(input_filename,memmap=False)
 
         # look at what fibers where actually fit
         i=np.where(other_psf_hdulist["PSF"].data["PARAM"]=="STATUS")[0][0]
         status_of_fibers = \
             other_psf_hdulist["PSF"].data["COEFF"][i][:,0].astype(int)
+        # log.info("status of fibers in PSF {} = {}".format(input_filename,
+        #     status_of_fibers))
         selected_fibers = np.where(status_of_fibers==0)[0]
+        failed_fibers=np.where(status_of_fibers>0)[0]
         log.info("fitted fibers in PSF {} = {}".format(input_filename,
             selected_fibers))
+        log.info("failed fibers in PSF {} = {}".format(input_filename,
+            failed_fibers))
         if selected_fibers.size == 0 :
             log.warning("no fiber with status=0 found in {}".format(
                 input_filename))
@@ -485,6 +506,9 @@ def merge_psf(inpsffile, inputs, output):
             i1=np.where(other_psf_hdulist["PSF"].data["PARAM"]==param)[0][0]
             psf_hdulist["PSF"].data["COEFF"][i0][selected_fibers] = \
                 other_psf_hdulist["PSF"].data["COEFF"][i1][selected_fibers]
+            if param=="STATUS" :
+                psf_hdulist["PSF"].data["COEFF"][i0][failed_fibers] = \
+                    other_psf_hdulist["PSF"].data["COEFF"][i1][failed_fibers]
 
         # copy bundle chi2
         i = np.where(other_psf_hdulist["PSF"].data["PARAM"]=="BUNDLE")[0][0]
@@ -535,6 +559,7 @@ def mean_psf(inputs, output):
     nbundles=None
     nfibers_per_bundle=None
     missing_bundles=[]
+    failed_bundles=[]
     for input in inputs :
         log.info("Adding {}".format(input))
         if not os.path.isfile(input) :
@@ -585,7 +610,11 @@ def mean_psf(inputs, output):
         missing_mask = np.array(
             [np.all(status[bundles == bundle] < 0) for bundle in unique_bundles],
             dtype=bool)
+        failed_mask=np.array(
+            [np.any(status[bundles == bundle] > 0) for bundle in unique_bundles],
+            dtype=bool)
         missing_bundles.append(unique_bundles[missing_mask])
+        failed_bundles.append(unique_bundles[failed_mask])
 
     npsf=len(tables)
     bundle_rchi2=np.array(bundle_rchi2)
@@ -599,6 +628,8 @@ def mean_psf(inputs, output):
     WAVEMAX=refhead["WAVEMAX"]
     FIBERMIN=int(refhead["FIBERMIN"])
     FIBERMAX=int(refhead["FIBERMAX"])
+    log.info(f"Input {input} has failed bundles: {failed_bundles}")
+    failed_bundles=np.array([x for sublist in failed_bundles for x in sublist])
 
 
     fibers_in_bundle={}
@@ -672,6 +703,7 @@ def mean_psf(inputs, output):
                 if len(masked_bundle_rchi2)!= len(bundle_rchi2) :
                     log.warning(f"Bundle {bundle} is missing in some but not all input PSFs for camera {refhead['CAMERA']}. Only {len(masked_bundle_rchi2)} PSFs will be used.")
                 nfailed = np.sum(masked_bundle_rchi2[:,bundle]==0)
+                nfailed += len(failed_bundles[failed_bundles==bundle])
                 if nfailed > 1 :
                     message=f"{nfailed} fit failures for bundle {bundle} indicate potential issue with unmasked CCD features or with the input PSF for camera {refhead['CAMERA']}."
                     log.critical(message)
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -473,6 +473,9 @@ def merge_psf(inpsffile, inputs, output):
 
     for input_filename in inputs :
         log.info("merging {} into {}".format(input_filename, inpsffile))
+
+        # JB Keep memmap=False to avoid issues with specex qa and memory mapping
+        # See desispec PR 2732 for more details
         other_psf_hdulist=fits.open(input_filename,memmap=False)
 
         # look at what fibers where actually fit
@@ -483,10 +486,16 @@ def merge_psf(inpsffile, inputs, output):
         #     status_of_fibers))
         selected_fibers = np.where(status_of_fibers==0)[0]
         failed_fibers=np.where(status_of_fibers>0)[0]
+
+        if failed_fibers.size > 0 :
+            log.warning("failed fibers in PSF {} = {}".format(input_filename,
+                failed_fibers))
+            i_p=np.where(psf_hdulist["PSF"].data["PARAM"]=='STATUS')[0][0]
+            psf_hdulist["PSF"].data["COEFF"][i_p][failed_fibers] = \
+                                other_psf_hdulist["PSF"].data["COEFF"][i][failed_fibers]
+
         log.info("fitted fibers in PSF {} = {}".format(input_filename,
             selected_fibers))
-        log.info("failed fibers in PSF {} = {}".format(input_filename,
-            failed_fibers))
         if selected_fibers.size == 0 :
             log.warning("no fiber with status=0 found in {}".format(
                 input_filename))
@@ -506,9 +515,6 @@ def merge_psf(inpsffile, inputs, output):
             i1=np.where(other_psf_hdulist["PSF"].data["PARAM"]==param)[0][0]
             psf_hdulist["PSF"].data["COEFF"][i0][selected_fibers] = \
                 other_psf_hdulist["PSF"].data["COEFF"][i1][selected_fibers]
-            if param=="STATUS" :
-                psf_hdulist["PSF"].data["COEFF"][i0][failed_fibers] = \
-                    other_psf_hdulist["PSF"].data["COEFF"][i1][failed_fibers]
 
         # copy bundle chi2
         i = np.where(other_psf_hdulist["PSF"].data["PARAM"]=="BUNDLE")[0][0]
--- a/py/desispec/scripts/specex.py
+++ b/py/desispec/scripts/specex.py
@@ -244,15 +244,10 @@ def main(args=None, comm=None):
                           "value {} running {}".format(rank, retval, comstr))
                 failcount += 1
                 failed_bundles.append(b)
-                # f =  "{}_{:02d}.fits".format(outroot, b)
-                # other_psf_hdulist=fits.open(f)
-
-                # # look at what fibers where actually fit
-                # i=np.where(other_psf_hdulist["PSF"].data["PARAM"]=="STATUS")[0][0]
-                # status_of_fibers = \
-                #     other_psf_hdulist["PSF"].data["COEFF"][i][:,0].astype(int)
-                # log.info(f'status_of_fibers: {status_of_fibers}')
-                bundles_to_merge.append(b)
+                if os.path.isfile(outbundlefits):
+                    bundles_to_merge.append(b)
+                else:
+                    log.error(f"no output file {outbundlefits} from failed bundle {b}; not merging it")
         else:
             log.info(f"proc {rank} succeeded generating {outbundlefits}")
             bundles_to_merge.append(b)
@@ -309,20 +304,7 @@ def main(args=None, comm=None):
                 if os.path.isfile(f):
                     os.remove(f)
 
-        if len(failed_bundles) > 0 :
-            failed_bundles = np.hstack(failed_bundles).astype(int)
             log.warning(f"The fit of the following bundles failed: {failed_bundles}")
-            for x in failed_bundles :
-                f =  "{}_{:02d}.fits".format(outroot, x)
-                # other_psf_hdulist=fits.open(f)
-
-                # # look at what fibers where actually fit
-                # i=np.where(other_psf_hdulist["PSF"].data["PARAM"]=="STATUS")[0][0]
-                # status_of_fibers = \
-                #     other_psf_hdulist["PSF"].data["COEFF"][i][:,0].astype(int)
-                # log.info(f'status_of_fibers: {status_of_fibers}')
-                if  os.path.isfile(f):
-                    os.remove(f)
 
     return
 
@@ -565,7 +547,7 @@ def mean_psf(inputs, output):
     nbundles=None
     nfibers_per_bundle=None
     missing_bundles=[]
-    failed_bundles=[]
+    failed_bundles_per_psf=[]
     for input in inputs :
         log.info("Adding {}".format(input))
         if not os.path.isfile(input) :
@@ -616,11 +598,12 @@ def mean_psf(inputs, output):
         missing_mask = np.array(
             [np.all(status[bundles == bundle] < 0) for bundle in unique_bundles],
             dtype=bool)
+        # Only works with changes made in desihub/specex#91
         failed_mask=np.array(
             [np.any(status[bundles == bundle] > 0) for bundle in unique_bundles],
             dtype=bool)
         missing_bundles.append(unique_bundles[missing_mask])
-        failed_bundles.append(unique_bundles[failed_mask])
+        failed_bundles_per_psf.append(unique_bundles[failed_mask])
 
     npsf=len(tables)
     bundle_rchi2=np.array(bundle_rchi2)
@@ -634,9 +617,6 @@ def mean_psf(inputs, output):
     WAVEMAX=refhead["WAVEMAX"]
     FIBERMIN=int(refhead["FIBERMIN"])
     FIBERMAX=int(refhead["FIBERMAX"])
-    log.info(f"Input {input} has failed bundles: {failed_bundles}")
-    failed_bundles=np.array([x for sublist in failed_bundles for x in sublist])
-
 
     fibers_in_bundle={}
     i=np.where(tables[0]["PARAM"]=="BUNDLE")[0][0]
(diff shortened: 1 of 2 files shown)


julienguy (approved):
Looks good

sbailey:
Moving to WIP / draft so that we don't accidentally merge this before converging on the right long-term solution.

jbanovetz:
I added some additional code as before, `psf`s with failed fibers were being passed. This additional code goes with changes to `specex` and identifies failed fiber fits (most likely due to fibers crossing) and counts them as failures and causes the pipeline to crash (as intended). These failed fibers will have a `STATUS` of 4 and this is what the code uses to identify the bad fibers.

py/desispec/scripts/specex.py
```diff
@@ -663,13 +694,20 @@ def mean_psf(inputs, output):
...
+            
+                # Only count fit failures for exposures where this bundle is present.
+                # Exposures listed in missing_bundles are excluded so that bundles
+                # missing from only some input PSFs are not treated as failed fits.
+                mask=[bundle in missing for missing in missing_bundles]
+                masked_bundle_rchi2 = bundle_rchi2.copy()[~np.array(mask)]
+                if len(masked_bundle_rchi2)!= len(bundle_rchi2) :
+                    log.warning(f"Bundle {bundle} is missing in some but not all input PSFs for camera {refhead['CAMERA']}. Only {len(masked_bundle_rchi2)} PSFs will be used.")
+                nfailed = np.sum(masked_bundle_rchi2[:,bundle]==0)
+                nfailed += len(failed_bundles[failed_bundles==bundle])
```
akremin: Can a fiber be both RCHI2=0 and STATUS>0? If so, I think you count that bundle twice and then throw out the bundle in the subsequent lines when really there is only one actual failure.

py/desispec/scripts/specex.py
```diff
@@ -244,6 +244,15 @@ def main(args=None, comm=None):
...
                 failed_bundles.append(b)
+                # f =  "{}_{:02d}.fits".format(outroot, b)
+                # other_psf_hdulist=fits.open(f)
+
+                # # look at what fibers where actually fit
+                # i=np.where(other_psf_hdulist["PSF"].data["PARAM"]=="STATUS")[0][0]
+                # status_of_fibers = \
+                #     other_psf_hdulist["PSF"].data["COEFF"][i][:,0].astype(int)
+                # log.info(f'status_of_fibers: {status_of_fibers}')
+                bundles_to_merge.append(b)
```
akremin: Are you confident that the later additions capture all possible corner cases where this could come back to bite us? I worry about including things that failed and then try to exclude bad things later, versus only keeping good things and then explicitly adding back in the failures that we deem acceptable. The former risks corner cases where the latter doesn't. I have a very superficial understanding of this so please let me know if I'm being overly nervous.

akremin (commented):
Thank you for your work on this! Overall these changes look to do what we want but I do have two important questions I'd like clarification on before I accept the PR.

py/desispec/scripts/specex.py
jbanovetz: I don't think so as RCHI2=0 is only true for missing fibers (STATUS<0). When merging the individual bundle files into the primary file, the RCHI2 is set to 0 and is populated with the correct value unless that bundle is missing  (https://github.com/desihub/desispec/blob/000242d19ed0a149d3bd2952b763cf38c28022f5/py/desispec/scripts/specex.py#L449-L456). So I don't think there is ever a case where this should happen as I also checked and the example of a failed fit does have a RCHI2>0.

py/desispec/scripts/specex.py
jbanovetz: I think this captures all known corner cases. I tested this with 4 different cases:

1. Missing amplifiers in all arc exposures (passes with warning in logs)
2. Missing amplifier in only a few arc exposures (passes with warning in logs)
3. Arcs with bad fiber fits (fails as expected/wanted)
4. Normal set of arcs (passes)

I can't think of another case where there would be a silent failure but I could add more warnings if that makes debugging down the road easier.

akremin:
Thanks, @jbanovetz . Can you provide the nights and exposure id's you used to tests the 4 different cases you mention above? And ideally the locations at NERSC where your tests reside. Thanks!

jbanovetz:
Sorry for the delay, here are the dates and then they can all be found at `/global/cfs/cdirs/desi/spectro/redux/banovetz`. I did rerun a few times so I would look at the most recent logs.

Some arcs with missing amps: 20211028
106399- r8A flagged
106400 - r8A flagged
(This night also only uses 4 arcs since 106396 was ignored for failing psf fitting)

All arcs with missing amps: 20221121
154099-103 b8B is missing

Normal night: 20230829

Fit failures: 20250822
307722 fibers 250 and 251 in bundle 10 cross
3077225 fibers 250 and 251 in bundle 10 cross

jbanovetz:
Adding some more documentation and context to this PR:

TLDR I changed some of the bookkeeping done on the `desispec` side and `specex` side so that we can separate out between bundle/fiber fit failures and flagged, missing bundles in order for the pipeline to crash as we want to (with the fit failures and not with the missing bundles).

This began when running Matterhorn and new additions to the `psfnight` caused the pipeline to crash when there were flagged missing amps (see [here](https://github.com/desihub/desispec/issues/2701)). Essentially the pipeline is set to crash when two or more bundles have a `rchi2==0` which would happen for cases of **missing bundles or failed fiber fits**. Not counting those missing bundles if present in **all** input exposures as failed fits was addressed in PR https://github.com/desihub/desispec/pull/2714

This also corresponded to a change in `specex` where I tried to change the C++ code with the help of Claude to initialize the PSF STATUS of bundles as -1 (so missing bundles should have a status of -1) and then when a fit failure happens to change it to 4. This resulted in this PR (https://github.com/desihub/specex/pull/86).

This then created a problem where if two input exposures had flagged amplifiers, then the pipeline would crash as not all the exposures had the same flagged amplifiers/bundles and there were two exposures with `rchi2==0`. This created this issue: https://github.com/desihub/desispec/issues/2723 which spawned this PR.

Upon looking into this, I found that the C++ wasn't actually doing what I expected and missing bundles and bundles with fit failures still had a PSF Status of -1. This prompted a PR to change the C++ code back to what it was and to make the PSF STATUS change further downstream with the Python QA part of `specex`. See https://github.com/desihub/specex/pull/91 for more details.

This branch should now be able to split missing and fit failure bundles based on PSF STATUS and this was shown with the tests labeled above. **This however does rely on https://github.com/desihub/specex/pull/91 to be pushed in order to work.**

jbanovetz:
One last summary post with a direct comparison between `matterhorn` and my changes (`SPECPROD=banovetz`):

### All arcs with missing amps: 20221121
154099-103 b8B is missing
`matterhorn`: Bundles 11-19 missing and skipped
`banovetz`:  Successfully skips Bundles 11-19: `/global/cfs/cdirs/desi/spectro/redux/banovetz/run/scripts/night/20221121/psfnight-20221121-00154099-a02346789-51299656.log`

### Normal night: 20230829
`matterhorn`: successfully completes
`banovetz`: successfully completes: `/global/cfs/cdirs/desi/spectro/redux/banovetz/run/scripts/night/20230829/psfnight-20230829-00193225-a0123456789-53965464.log`

### Some arcs with missing amps: 20211028, **what the PR tries to fix**
106399- r8A flagged
106400 - r8A flagged
(This night also only uses 4 arcs since 106396 was ignored for failing psf fitting)

`matterhorn`: First attempt failed:
`CRITICAL:specex.py:671:mean_psf: 2 fit failures for bundle 0 indicate potential issue with unmasked CCD features or with the input PSF.`
Second attempt succeeded after flagging one of the exposures as bad (106399)

`banovetz`: Unflagging 106399 and flagging r8A as originally done, it successfully runs with the changes and a new warning:
`WARNING:specex.py:704:mean_psf: Bundle 0 is missing in some but not all input PSFs for camera 'r8      '. Only 2 PSFs will be used.` 
`/global/cfs/cdirs/desi/spectro/redux/banovetz/run/scripts/night/20211028/psfnight-20211028-00106397-a0123456789-54019366.log`

### Fit failures: 20250822, **a test case that should fail**
307722 fibers 250 and 251 in bundle 10 cross in z7
307725 fibers 250 and 251 in bundle 10 cross in z7

`matterhorn`: First attempt failed as expected:
`CRITICAL:specex.py:671:mean_psf: 2 fit failures for bundle 10 indicate potential issue with unmasked CCD features or with the input PSF.`
Second attempt succeeded after flagging one of the exposures as bad (307722)

`banovetz`: Unflagging 307722, it successfully failed as expected:
`CRITICAL:specex.py:709:mean_psf: 2 fit failures for bundle 10 indicate potential issue with unmasked CCD features or with the input PSF for camera 'z7      '.`
`/global/cfs/cdirs/desi/spectro/redux/banovetz/run/scripts/night/20250822/psfnight-20250822-00307722-a0123456789-56472263.log`

py/desispec/scripts/specex.py
```diff
@@ -244,6 +244,15 @@ def main(args=None, comm=None):
...
                 failcount += 1
                 failed_bundles.append(b)
+                # f =  "{}_{:02d}.fits".format(outroot, b)
+                # other_psf_hdulist=fits.open(f)
+
+                # # look at what fibers where actually fit
+                # i=np.where(other_psf_hdulist["PSF"].data["PARAM"]=="STATUS")[0][0]
+                # status_of_fibers = \
+                #     other_psf_hdulist["PSF"].data["COEFF"][i][:,0].astype(int)
+                # log.info(f'status_of_fibers: {status_of_fibers}')
```
akremin: Please delete this commented-out debug block rather than leaving it in. Same for the copies at L317-323 and L482-483.

py/desispec/scripts/specex.py
akremin: **This is the one I'd most like changed before merging.** Adding failed bundles to `bundles_to_merge` unconditionally means `merge_psf` will `fits.open()` a file that may not exist.

It usually does exist — `run_specex` calls `write_psf()` unconditionally after `fit_psf()`, which is why your 20250822 test worked. But `fit_psf` returns `EXIT_FAILURE` out of a `catch (std::exception&)` that skips `pyps.psf = fitter.psf`, and `read_psf` / `read_preproc` / `write_psf` can throw on their own. If the file is missing or truncated, `merge_psf` raises, the exception is caught and swallowed at L301-304, and **no merged PSF is written at all** for that camera. Before this change the failed bundle was simply excluded and we still produced a PSF from the remaining 19 bundles — so this is a robustness regression on the failure path.

Suggest gating on the file actually existing, checked on the rank that wrote it:

```suggestion
                if os.path.isfile(outbundlefits):
                    bundles_to_merge.append(b)
                else:
                    log.error(f"no output file {outbundlefits} from failed bundle {b}; not merging it")
```

As belt-and-braces it would also be worth wrapping the per-input body of the loop in `merge_psf` in a try/except so that one unreadable bundle file can't take out the whole merge.

py/desispec/scripts/specex.py
```diff
@@ -305,6 +314,13 @@ def main(args=None, comm=None):
...
                 f =  "{}_{:02d}.fits".format(outroot, x)
+                # other_psf_hdulist=fits.open(f)
+
+                # # look at what fibers where actually fit
+                # i=np.where(other_psf_hdulist["PSF"].data["PARAM"]=="STATUS")[0][0]
+                # status_of_fibers = \
+                #     other_psf_hdulist["PSF"].data["COEFF"][i][:,0].astype(int)
+                # log.info(f'status_of_fibers: {status_of_fibers}')
                 if  os.path.isfile(f):
                     os.remove(f)
```
akremin: Two things here.

First, please remove the commented-out debug block. This one isn't hypothetical — the live version of it caused a spurious job failure in your own 20250822 test run. The merge succeeded and the PSF was written, the bundle files were correctly deleted at L308-310, and then this `fits.open(f)` re-opened the deleted file and failed the entire z7 arc job:

```
INFO:specex.py:531:merge_psf: Wrote PSF .../fit-psf-z7-00307722.fits
INFO:specex.py:306:main: done merging
WARNING:specex.py:314:main: The fit of the following bundles failed: [10]
ERROR:specex.py:418:fitframe: FAILED: MPI ranks 41-60 on camera z7
ERROR:specex.py:421:fitframe: [Errno 2] No such file or directory: '.../fit-psf-z7-00307722_10.fits'
```

Second, now that failed bundles are included in `bundles_to_merge`, their files are already removed by the loop at L308-310, so this cleanup loop is dead code. I'd reduce the whole block to just the `log.warning` on L314.

py/desispec/scripts/specex.py
```diff
@@ -457,15 +473,20 @@ def merge_psf(inpsffile, inputs, output):
 
     for input_filename in inputs :
         log.info("merging {} into {}".format(input_filename, inpsffile))
-        other_psf_hdulist=fits.open(input_filename)
+        other_psf_hdulist=fits.open(input_filename,memmap=False)
```
akremin, line 461: Could you add a one-line comment explaining why `memmap=False` is needed here? I assume it's working around something real (the bundle files get unlinked shortly afterwards), but as an unexplained change it's likely to get "tidied up" back to the default by someone later.

py/desispec/scripts/specex.py
```diff
@@ -457,15 +473,20 @@ def merge_psf(inpsffile, inputs, output):
...
             other_psf_hdulist["PSF"].data["COEFF"][i][:,0].astype(int)
+        # log.info("status of fibers in PSF {} = {}".format(input_filename,
+        #     status_of_fibers))
         selected_fibers = np.where(status_of_fibers==0)[0]
+        failed_fibers=np.where(status_of_fibers>0)[0]
         log.info("fitted fibers in PSF {} = {}".format(input_filename,
             selected_fibers))
+        log.info("failed fibers in PSF {} = {}".format(input_filename,
+            failed_fibers))
         if selected_fibers.size == 0 :
```
akremin, line 481: **I think this is a real hole, and it's the other one I'd block on.**

When *every* fiber in the bundle failed, `selected_fibers` is empty and we `continue` here — before the new `failed_fibers` STATUS propagation at L509-511 ever runs. So the merged PSF keeps `STATUS = -1` for those fibers (from the reset at L472) and `B{bb}RCHI2 = 0` (never set, since the header copy at L519-523 is also below this `continue`).

`mean_psf` then sees `np.all(status < 0)` for that bundle, classifies it as **missing**, masks it out of `nfailed`, and no error is raised.

This isn't an exotic case: specex sets `fit_status` to 1 (cholesky error) or 2 (no convergence) for the bundle and writes it to every fiber (`specex_psf_proc.cc`), so a whole-bundle fit failure lands here. That's precisely the case 0dfcafb2 ("average PSF return error if there is more than 1 fit error in a bundle") was added to catch — before this PR the bundle wasn't merged, `RCHI2` stayed 0, and `nfailed` counted it and raised.

Fix should be small: move the STATUS write for `failed_fibers` above this early `continue` (write the STATUS column for the failed fibers first, then `continue` on the empty-`selected_fibers` case).

This would also be a good thing to pin down in a unit test.

py/desispec/scripts/specex.py
```diff
@@ -485,6 +506,9 @@ def merge_psf(inpsffile, inputs, output):
             i1=np.where(other_psf_hdulist["PSF"].data["PARAM"]==param)[0][0]
             psf_hdulist["PSF"].data["COEFF"][i0][selected_fibers] = \
                 other_psf_hdulist["PSF"].data["COEFF"][i1][selected_fibers]
+            if param=="STATUS" :
+                psf_hdulist["PSF"].data["COEFF"][i0][failed_fibers] = \
+                    other_psf_hdulist["PSF"].data["COEFF"][i1][failed_fibers]
```
akremin: Two notes.

Minor: this `if` is inside the `for param in parameters` loop, so it's evaluated once per PARAM in order to act once. Cleaner to hoist it out of the loop entirely.

More substantive, and I don't think it's discussed in the PR: for failed fibers we copy only the `STATUS`, not `XTRACE`/`YTRACE`/coefficients — so those fibers keep the *input/shifted reference* PSF values. But the bundle now gets a valid `RCHI2 > 0`, so in `mean_psf` that exposure passes the `ok` selection and is averaged in. Your 20250822 log shows this (`for fiber bundle 10, 4 valid PSFs` with two exposures flagged on fibers 250/251).

With >= 2 failures we crash, so no harm done. But with exactly **one** failure we don't crash, and the psfnight trace for those fibers becomes `mean(N-1 real fits, 1 stale reference trace)` — silently degraded. Under the old behaviour that exposure's whole bundle was excluded from the average.

Could we either exclude the failed fibers from the per-fiber average, or at minimum emit a warning when `nfailed == 1` so it's visible in the logs?

py/desispec/scripts/specex.py
```diff
@@ -585,7 +610,11 @@ def mean_psf(inputs, output):
         missing_mask = np.array(
             [np.all(status[bundles == bundle] < 0) for bundle in unique_bundles],
             dtype=bool)
+        failed_mask=np.array(
+            [np.any(status[bundles == bundle] > 0) for bundle in unique_bundles],
+            dtype=bool)
         missing_bundles.append(unique_bundles[missing_mask])
+        failed_bundles.append(unique_bundles[failed_mask])
```
akremin: Suggest keeping this per-input-PSF rather than flattening it later at L632 — see my comment on the `nfailed` computation. Something like a `failed_bundles_per_psf` list parallel to `missing_bundles` lets the failure count be expressed as a per-exposure boolean OR instead of a sum of two counts.

Also worth noting in the PR body / changelog that this whole mechanism is a no-op without desihub/specex#91, since nothing sets `STATUS > 0` before that merges.

py/desispec/scripts/specex.py
```diff
@@ -599,6 +628,8 @@ def mean_psf(inputs, output):
     WAVEMAX=refhead["WAVEMAX"]
     FIBERMIN=int(refhead["FIBERMIN"])
     FIBERMAX=int(refhead["FIBERMAX"])
+    log.info(f"Input {input} has failed bundles: {failed_bundles}")
+    failed_bundles=np.array([x for sublist in failed_bundles for x in sublist])
```
akremin: This log line looks like leftover debugging, and as written it's misleading: it's outside the `for input in inputs` loop, so `input` is the *last* input file while `failed_bundles` covers all of them. From your 20250822 log:

```
INFO:specex.py:631:mean_psf: Input .../fit-psf-z7-00307726.fits has failed bundles: [array([10]), array([], dtype=int64), array([], dtype=int64), array([10]), array([], dtype=int64)]
```

307726 isn't one of the two failing exposures. Either move it inside the loop so the filename matches, or drop the filename and just log the per-exposure lists.

On L632: I'd rather not flatten here — see the `nfailed` comment below.

py/desispec/scripts/specex.py
```diff
@@ -663,13 +694,20 @@ def mean_psf(inputs, output):
...
-            if nfailed > 1 :
-                message=f"{nfailed} fit failures for bundle {bundle} indicate potential issue with unmasked CCD features or with the input PSF."
-                log.critical(message)
-                raise RuntimeError(message)
+            
+                # Only count fit failures for exposures where this bundle is present.
+                # Exposures listed in missing_bundles are excluded so that bundles
+                # missing from only some input PSFs are not treated as failed fits.
+                mask=[bundle in missing for missing in missing_bundles]
+                masked_bundle_rchi2 = bundle_rchi2.copy()[~np.array(mask)]
```
akremin: Copilot's suggestion on this line hasn't been applied and I think it's worth taking, with one addition: `bundle_rchi2.copy()[~np.array(mask)]` allocates a full copy and then slices it, inside the bundle loop. `bundle_rchi2[~np.array(mask, dtype=bool)]` is equivalent.

The explicit `dtype=bool` matters beyond style: if `missing_bundles` is ever empty, `np.array([])` is float64 and `~` raises `TypeError`.

Also, L697 has trailing whitespace.

py/desispec/scripts/specex.py
```diff
@@ -663,13 +694,20 @@ def mean_psf(inputs, output):
...
+                if len(masked_bundle_rchi2)!= len(bundle_rchi2) :
+                    log.warning(f"Bundle {bundle} is missing in some but not all input PSFs for camera {refhead['CAMERA']}. Only {len(masked_bundle_rchi2)} PSFs will be used.")
```
akremin: The wording here is misleading. The mask only affects the *failure counting*; the averaging selection at L692 still uses the full, unmasked `bundle_rchi2`, so the number of PSFs actually used for this bundle is `ok.size`, which is logged separately on L695.

They happened to agree in your 20211028 run (2 of 4 exposures missing r8A, and `ok.size == 2`), but they're different quantities — `ok` also applies the chi2 threshold. Suggest rewording to something like "Bundle {bundle} is present in only {n} of {npsf} input PSFs for camera {camera}" and leaving the "will be used" statement to the existing L695 message.

py/desispec/scripts/specex.py
akremin: Following up on my earlier question about double-counting: your answer is right for the code as it stands today — a partially-failed bundle does get a real `RCHI2 > 0` from L519-523, so the two terms are disjoint. But that's an invariant holding by coincidence rather than by construction, and it's exactly the kind of thing that breaks quietly if specex ever writes `RCHI2 = 0` for a bundle that also has `STATUS > 0` fibers. One exposure would then count as two failures and we'd get a spurious `RuntimeError`.

Making it structurally impossible is about the same amount of code — count *exposures*, with a boolean OR:

```python
present = ~np.array([bundle in m for m in missing_bundles], dtype=bool)
failed  = np.array([bundle in f for f in failed_bundles_per_psf], dtype=bool)
nfailed = np.sum(present & ((bundle_rchi2[:, bundle] == 0) | failed))
```

This needs `failed_bundles` kept per-PSF instead of flattened at L632, and it has the nice side effect that `nfailed` is unambiguously "number of exposures in which this bundle failed", which is what the error message claims.

py/desispec/scripts/specex.py
```diff
@@ -663,13 +694,20 @@ def mean_psf(inputs, output):
...
+                nfailed = np.sum(masked_bundle_rchi2[:,bundle]==0)
+                nfailed += len(failed_bundles[failed_bundles==bundle])
+                if nfailed > 1 :
+                    message=f"{nfailed} fit failures for bundle {bundle} indicate potential issue with unmasked CCD features or with the input PSF for camera {refhead['CAMERA']}."
+                    log.critical(message)
+                    raise RuntimeError(message)
 
             # We finally resorted to use a mean instead of a median here for two reasons.
             # First, there is already a vetting of PSF bundles with good chi2 above
```
akremin, line 701: This comment is about the `else` fallback further down at L733-738, which isn't in the diff:

```python
else : # we have a problem here, take the smallest rchi2
    i=np.argmin(bundle_rchi2[:,bundle])
```

Missing exposures have `rchi2 == 0`, which is always the minimum — so when `ok` is empty this copies coefficients from an exposure in which the bundle was **never fit**, silently, and writes `B{bb}RCHI2 = 0` to the output header.

This is pre-existing, but I think this PR makes it newly reachable. Fully-missing bundles are dropped from `fibers_in_bundle` at L642-645, so it couldn't be hit for those; *partially*-missing bundles are the new case this PR introduces, and they do reach it. Concrete scenario: 3 arcs, bundle missing in 2 of them, and the one exposure that did fit it lands above `rchi2_threshold` — `ok` is empty, and we take the bundle from one of the two exposures where it was masked out.

Since you're already touching this block, could the argmin be restricted to `bundle_rchi2[:,bundle] > 0`, with a loud warning (or an error) if that set is empty?

akremin (commented):
Thanks for the additional work on this, and for the very helpful summary of the four test cases. I asked Claude to provide a thorough review and it certainly did so. I vetted them for substance, but I admittedly have not lived with this code long enough to know what is and isn't a true issue. I apologize in advance if some of these are non-issues.

Review from Claude:
The overall design is right: separating "bundle masked/missing because of a bad amp" from "bundle genuinely failed to fit" is the correct fix for #2723, and the 20211028 and 20250822 logs show it doing the intended thing in those two cases. I dug into the surrounding code and the specex side (desihub/specex#91) and found a couple of things I think we need to resolve before merging, plus a handful of smaller cleanups. Details are in the inline comments; summarizing the two I'd consider blocking:

1. **A bundle in which *every* fiber fails is now silently reclassified as "missing" rather than "failed"** (comment on `merge_psf`, L485-490). `merge_psf` `continue`s before the new STATUS propagation when `selected_fibers` is empty, so the merged PSF keeps `STATUS=-1` and `RCHI2=0`, and `mean_psf` then masks it out of `nfailed`. specex sets `fit_status` to 1 (cholesky error) or 2 (no convergence) on *every* fiber of the bundle, so this is the classic whole-bundle fit failure — i.e. exactly the case the check in 0dfcafb2 was added to catch. Before this PR it raised; now it passes silently.

2. **`merge_psf` can now abort entirely for a camera** (comment on L255). Failed bundles go into `bundles_to_merge` unconditionally. `run_specex` does call `write_psf()` even when `fit_psf()` fails, so the file usually exists — but `fit_psf` returns `EXIT_FAILURE` from a `catch(std::exception&)` that skips `pyps.psf = fitter.psf`, and `read_psf`/`read_preproc`/`write_psf` can throw on their own. If the file is missing or truncated, `merge_psf` raises, the exception is swallowed at L301-304, and **no merged PSF is written at all** for that camera. Previously the bad bundle was just excluded and we still got a PSF from the remaining 19 bundles.

This is the concrete version of the worry I raised earlier about "include everything then exclude the bad later" — both of these are that pattern biting us, and both have small local fixes.

Two more general points:

* **Please remove the commented-out debug blocks** (L247-254, L317-323, L482-483). Not just style: the live version of the L317 block caused a real spurious job failure in your 20250822 run — the merge succeeded and wrote the PSF, then the debug `fits.open(f)` re-opened the already-deleted bundle file and marked the whole z7 arc job FAILED:
  INFO:specex.py:531:merge_psf: Wrote PSF .../fit-psf-z7-00307722.fits
  INFO:specex.py:306:main: done merging
  WARNING:specex.py:314:main: The fit of the following bundles failed: [10]
  ERROR:specex.py:418:fitframe: FAILED: MPI ranks 41-60 on camera z7
  ERROR:specex.py:421:fitframe: [Errno 2] No such file or directory: '.../fit-psf-z7-00307722_10.fits'

* **Please note explicitly in the PR body which specex version this requires. Relatedly, it would be really valuable to add a small unit test — there's no `test_specex.py` today, and the missing-vs-failed classification in `mean_psf` is pure array logic that could be exercised with synthetic `status`/`rchi2` arrays. Given this logic has now been revised across #2714, this PR (5 commits), and two specex PRs, a test that pins down all four of your cases would save us the next round of this.

One cross-repo note: desihub/specex#91 changes the `PSF_Params` ctor initializer list from `chi2(0), chi2_in_core(0), ndata(0), ndata_in_core(0), nparams(0), fit_status(4), ...` to `chi2(0), ndata(0), fit_status(-1), ...`, which drops the initializers for `chi2_in_core`, `ndata_in_core` and `nparams`. Those become indeterminate, and `specex_pyfitting.cc` divides by `ndata_in_core` when logging. Worth restoring them there alongside the `fit_status(-1)` change.

py/desispec/scripts/specex.py
jbanovetz: I will add a comment when I do the rest of the clean up but this was added for two reasons, if I remember correctly. The first was that sometimes I would hit a memory limit when running and go very slowly and cause timeouts (this was actually more with the `specex/qa.py` side of things). 

The second reason was that this actually keeps the first file you load in memory and will sometimes load that one and not do a fresh reload even when you made changes to the file (see here: https://docs.astropy.org/en/latest/io/fits/index.html#working-with-large-files). What was happening was that the `specex/qa.py` part would load the fits file and change the PSF STATUS based on if there was any QA issues. Then when `desispec` would load the file here to merge, it would actually be loading the pre-changed version of the fits file and merge that. 

Thank you for pointing out that I should comment on this as it took some real time to figure out this nuance!

py/desispec/scripts/specex.py
jbanovetz: I agree and moved up the failed_fiber part to right before the `continue`. I am writing a unit test with the help of Claude to check this as well.

py/desispec/scripts/specex.py
jbanovetz: I removed the if and see what you mean about the degraded PSF. I added in a warning for the `nfailed==1` case as I think the individual fiber bookkeeping is a bigger effort.

py/desispec/scripts/specex.py
jbanovetz: I added the change and also made sure that the case of all coefficients are 0 to have it be 0 (which shouldn't happen so maybe could be removed) but would like to know @julienguy 's thoughts about the change. 

In your example from a QA side (and more ease of doing QA), I might prefer that the bundle `rchi2=0` to the large `rchi2 > rchi2_threshold` as if I don't know what the threshold is, I might not know if that bundle failed or not (though to be fair, in my experience, the fit usually fails catastrophically it is usually pretty easy to see what is outside of family from the other bundles). Keeping the `rchi2=0` does help in identifying the problem bundles much easier and programmatically is easier. It also might be somewhere else in the pipeline that then discards bundles based off of `rchi2=0` (One example could be `FIBERSTATUS` mask is set).

I think at the very least we should add something to each of the `ok.size` statements here outside of `log.debug`, or at the level of when `ok` is first instantiated. Something along the lines of a `log.warning` maybe of just `ok.size < number of input exps: found that X exps have bundle Y missing and Z exps have a rchi2 above threshold for bundle Y`

py/desispec/scripts/specex.py
jbanovetz: I agree with the changes to `failed_bundles_per_psf`, it is a lot cleaner than flattening and counting the flatten array and avoids potential double counting. I will implement the changes with the other comments.

Already mentioned in the PR body as a comment and added as a comment to the code.

py/desispec/scripts/specex.py
jbanovetz: Removed the debugging code and removed the flattening part as a part of the larger effort to change the `nfailed` calculation.

py/desispec/scripts/specex.py
jbanovetz: Changed and good to know! It also looks cleaner.

py/desispec/scripts/specex.py
jbanovetz: Changed the warning to better reflect what is being counted

py/desispec/scripts/specex.py
jbanovetz: I agree with these changes and have made them throughout the code following the comments suggestions.

The only change I made was:
`present = ~np.array(mask, dtype=bool)`
since mask is already set a few lines above as:
`mask=[bundle in missing for missing in missing_bundles]`

py/desispec/scripts/specex.py
jbanovetz: I agree that this is a good check, especially if something changes in the future. right now the script will continue, is `log.error` good enough or should I also throw an error to stop the script?

jbanovetz:
I just ran `desi_proc_night` on the same 4 cases above and they all ran as expected! Here are the logs:

20211028:
`/global/cfs/cdirs/desi/spectro/redux/banovetz/run/scripts/night/20211028/psfnight-20211028-00106397-a0123456789-58130508.log`

20221121:
`/global/cfs/cdirs/desi/spectro/redux/banovetz/run/scripts/night/20221121/psfnight-20221121-00154099-a02346789-58130479.log`

20230829:
`/global/cfs/cdirs/desi/spectro/redux/banovetz/run/scripts/night/20230829/psfnight-20230829-00193225-a0123456789-58130431.log`

20250822 (expected to fail on one job with crossed fiber trace fits):
`/global/cfs/cdirs/desi/spectro/redux/banovetz/run/scripts/night/20250822/psfnight-20250822-00307722-a0123456789-58130336.log`

As a reminder: **THIS ONLY WORKS WITH PR https://github.com/desihub/specex/pull/91**

I got burned by that during this latest round of testing so we should merge that one before merging this PR.

jbanovetz:
One thing to note is that I haven't tested the new unittest that @akremin suggested. If that is blocking Nevis, I could remove it from this PR and add it in with a new PR.

py/desispec/scripts/specex.py
akremin: The motivation of the comment was to prevent the script from crashing when a file doesn't exist, so let's NOT throw an error to stop the script. `log.error` noting the problem and then moving on smoothly is the behavior we want, similar to the suggested solution above.

akremin:
This is what my commit changed:
  - Restore the `if len(failed_bundles) > 0` guard dropped in 15e390ee,
    which made every clean arc job log "The fit of the following bundles
    failed: []".
  - Raise if no merged PSF was written. A merge_psf exception was caught
    and counted into failcount, which main() then discarded, so the camera
    was logged SUCCESS with no PSF on disk. Raise before deleting the
    per-bundle files so they survive for debugging.
  - Move the fit-failure check out of the per-PARAM averaging loop. It does
    not depend on the entry being averaged, so it now runs once, before any
    averaging work, rather than being gated on `entry == 0`.
  - When falling back to the smallest non-zero rchi2, only consider inputs
    that contain the bundle. The previous `i = 0` could select an exposure
    where the bundle was masked out entirely.
  - Pass bundlefiles[0] as an input as well as the template under
    --dont-merge-with-input, so a fit failure in the first bundle is not
    erased by the template STATUS reset.
  - Strip the quotes specex writes into the CAMERA header value, so 
    messages read "camera z7" rather than "camera 'z7      '".
  - Add tests for mean_psf: the four production cases from the PR
    discussion, both fallback branches, and PARAM row ordering. One test
    pins the current XTRACE/YTRACE averaging (#2819) so that changing it is
    deliberate.

py/desispec/scripts/specex.py
```diff
@@ -457,13 +468,25 @@ def merge_psf(inpsffile, inputs, output):
...
             other_psf_hdulist["PSF"].data["COEFF"][i][:,0].astype(int)
         selected_fibers = np.where(status_of_fibers==0)[0]
+        failed_fibers=np.where(status_of_fibers>0)[0]
+
+        if failed_fibers.size > 0 :
+            log.warning("failed fibers in PSF {} = {}".format(input_filename,
+                failed_fibers))
+            i_p=np.where(psf_hdulist["PSF"].data["PARAM"]=='STATUS')[0][0]
+            psf_hdulist["PSF"].data["COEFF"][i_p][failed_fibers] = \
+                                other_psf_hdulist["PSF"].data["COEFF"][i][failed_fibers]
```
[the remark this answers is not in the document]
akremin: The premise that they are assigned to -1 is incorrect and likely misled by a unit test that does that. The real code doesn't. Updated the unit test instead

py/desispec/scripts/specex.py
```diff
@@ -244,6 +244,10 @@ def main(args=None, comm=None):
                           "value {} running {}".format(rank, retval, comstr))
                 failcount += 1
                 failed_bundles.append(b)
+                if os.path.isfile(outbundlefits):
+                    bundles_to_merge.append(b)
+                else:
+                    log.error(f"no output file {outbundlefits} from failed bundle {b}; not merging it")
```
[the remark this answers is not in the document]
akremin: Addressed in latest commit

py/desispec/scripts/specex.py
```diff
@@ -296,17 +303,21 @@ def main(args=None, comm=None):
...
             log.info('done merging')
 
+            #- a merge_psf exception is caught above, but that must not be
+            #- reported as success: without this check the camera is logged
+            #- SUCCESS with no merged PSF on disk.  Raise before removing the
+            #- per-bundle files so that they survive for debugging.
+            if not os.path.isfile(outfits) :
+                message = f"merging failed: {outfits} was not written"
+                log.critical(message)
+                raise RuntimeError(message)
```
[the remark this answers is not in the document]
akremin: Addressed in latest commit

akremin:
I'm out of copilot tokens, so I asked GPT-6 Astra to weigh in. I used Claude in development so this is a ~independent review:

Verdict
I would approve the current PR for its intended scope. I found no additional merge-blocking bug. It retains a known numerical limitation, so I wouldn’t describe the remaining concerns as purely cosmetic.
Revision matters: GitHub is at 63a6dfc; your local checkout is at the older e19975ac, which lacks several subsequent fixes. This assessment applies to the current PR #2732.
What looks correct
The current implementation correctly handles the central cases:
- Bundles missing from some exposures no longer count as fit failures.
- Missing exposures do not conceal two genuine failures in the remaining exposures.
- An exposure with both RCHI2 == 0 and positive STATUS counts only once.
- Whole-bundle failures retain their failure flags, even when no fiber fitted successfully.
- The fallback avoids selecting missing bundles merely because their chi-square is zero.
- Retries remove stale bundle files; merge exceptions propagate and preserve diagnostic files.
- --dont-merge-with-input preserves failure flags from the first bundle.
These address the substantive problems raised in earlier reviews.
Remaining concerns
1. One partially failed exposure still contributes to the average
When exactly one exposure contains failed fibers but has acceptable positive chi-square, the code warns and still includes that exposure in the average. Failed fibers retain reference coefficients from merge_psf, so those reference values contribute alongside successful fits. STATUS is also averaged, potentially producing fractional values.
I reproduced this: coefficients of 100 from the failed/reference input and 20, 60 from successful inputs produce 60, whereas excluding the failed input produces 40.
This is a real numerical compromise. However, it was explicitly discussed and accepted with a warning. I would track improved selection of valid fits as follow-up work rather than reopen it as a newly discovered blocker.
2. Trace averaging still ignores bundle selection
XTRACE and YTRACE are averaged across all inputs, including those excluded from coefficient averaging. This is pre-existing behavior, explicitly documented by the new tests and already identified for separate work. It remains a limitation, but I would not attribute it to this PR.
3. The specex dependency must be deployed
The required specex #91 has merged. The processing environment must actually have those changes installed; merging the companion PR alone does not update deployed installations.
Validation
- 12 PR tests passed, plus two subtests.
- 10 additional temporary checks passed, covering retry handling, merge exceptions, first-bundle failures, missing output, and mixed missing/failed exposures.
- All reported GitHub CI checks passed.
- I inspected the companion specex writer/QA and downstream callers.
The additional checks used real FITS files with a mocked fitter. I did not rerun compiled specex, MPI, or the NERSC production cases.
Recommendation: merge the current revision with the required specex version deployed, retaining the averaging limitations as explicit follow-up work. Your checkout was left unchanged.

PR #2746, opened 2026-05-05
Fix program-to-efftime in zcatalog

rongpu:
The TSNR2-to-EFFTIME_SPEC conversion is program-specific. Previously, we read the `PROGRAM` value from the redrock header, but it can be blank or take arbitrary values in some tiles (that were entered by hand, e.g., "fibertest", "rotation test", "bgs tile 80660") rather than the standard program values (dark, bright, backup, etc.).

This PR implements a workaround that was suggested by @sbailey: instead of reading the `PROGRAM` keyword, we now read the `FAFLAVOR` keyword (from the `EXP_FIBERMAP` extension) and translate it to `program` with `faflavor2program`.

This PR also adds `bright1b` and `dark1b` to `program_to_tsnr2_colname_dict` (that chooses the TSNR2 value to convert to EFFTIME_SPEC).

### DIFF
--- a/py/desispec/scripts/zcatalog.py
+++ b/py/desispec/scripts/zcatalog.py
@@ -126,7 +126,7 @@ def read_redrock(rrfile, group=None, pertile=False, counter=None):
         # PROGRAM is needed for TSNR2 -> EFFTIME_SPEC conversion
         # cannot use the "PROGRAM" header keyword because some files have arbitrary values
         hdr1 = fx['EXP_FIBERMAP'].read_header()
-        program = faflavor2program(hdr1['PROGRAM'])
+        program = faflavor2program(hdr1['FAFLAVOR'])
 
         hdr = fx[0].read_header()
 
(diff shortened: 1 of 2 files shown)


rongpu:
@sbailey noted that unfortunately the `FAFLAVOR` keyword is present in matterhorn, but not in loa redrock files. We decided to instead use `--program` argument as the keyword. This makes `--program` a required argument (which should be fine since we always need to specify the program when producing the zcatalog).

I am doing a test run with the update code and the outputs are in `/pscratch/sd/r/rongpu/tmp/matterhorn/zcatalog/v2_new`

sbailey:
Looks good, thanks. Let's wait to merge until you are finished generating the matterhorn tile-based zcatalog files, including zall, to make sure there aren't other problems that need to be addressed for matterhorn.

sbailey (approved):

sbailey:
@rongpu generated draft matterhorn/zcatalog/v2 tile-based redshift catalog files in /pscratch/sd/r/rongpu/tmp/matterhorn/zcatalog/v2_new/zall (Thanks!).  I'm moving these into place in matterhorn now, with the expectation that we'll rerun again after some additional QA and cleanup.

PR #2751, opened 2026-05-08
Add CCD history script

rongpu:
This PR adds a script that retrieves the change history of the DESI CCDs (from the spectrograph YAML files), prints it and (optionally) saves it to a YAML file (so it can be easily ingested by another program).


rongpu:
The CCD history from running the script can be found here:
https://desi.lbl.gov/trac/wiki/InstOpsPages/CCDHistory

sbailey (approved):

PR #2755, opened 2026-05-15
desi_healpix_redshifts use adaptive healpix (uniqpix)

sbailey:
This PR updates `desi_healpix_redshifts` and `desi_zproc` to use adaptive-nside uniqpix instead of fixed-nside healpix.  This allows dense pixels to be recursively split into smaller pixels (higher nside) to achieve a maximum number of targets per pixel, which helps with memory and runtime uniformity.

See the TODO section at the end for items that need attention before merging, including why some tests are broken.

## New file organization

Now cross-tile files are kept under `spectra/SURVEY/PROGRAM/PIXGROUP/UNIQPIX/` instead of `healpix/SURVEY/PROGRAM/PIXGROUP/HEALPIX/`.  Structurally these are identical except for "healpix" -> "spectra" and the numbers now mean something different.

`spectra/SURVEY/PROGRAM/hpix2upix-SURVEY-PROGRAM.[fits,json]` have the mapping of healpix at the maximum nside to which uniqpix they ended up in.  e.g.
```python
hpix2upix = Table.read('hpix2upix-special-dark.fits')
nside = hpix2upix.meta['NSIDE']  # 256
hpix = desimodel.footprint.radec2pix(nside, ra, dec)  # or healpy.ang2pix(nside, ra, dec, lonlat=True, nest=True)
upix = hpix2upix['UNIQPIX'][hpix]
```

## Code

The core updates are

  * `pixgroup.get_exp2uniqpix_map` replaces `pixgroup.get_exp2healpix_map`. The new function needs the ztile catalog for the requested SURVEY and PROGRAM so that it can calculate the uniqpix hierarchy, then it cross matches with the FRAMES HDU of the production exposures-SPECPROD.fits file to get which exposures exists for each TILE, PETAL_LOC.
  * `io.findfile` updated to support uniqpix by default, while still supporting groupname='healpix' for backwards compatibility with looking up files in previous productions.  The logic is a bit of a mess with all of the special cases, so the options are perhaps easiest to understand by looking at the updates to test.test_io.
  * `pixgroup.uniqpix_to_map` calculates the mapping from healpix at a high nside to which uniqpix contains them (used to generate the `hpix2upix-*.* files)
  * Other changes are mostly about renaming "healpix" to "uniqpix" in arguments / variables.

## Tests

I created an "m5" prod with links to various matterhorn inputs, then ran
```
desi_healpix_redshifts --survey special --program dark --bundle-pix 10 --nosubmit
```
This generated the scripts under `run/scripts/uniqpix/special/dark/` and pre-populated the `spectra/special/dark/*/*/pixexp-*.csv` files that have the map of which exposure petals cover each uniqpix.

I then ran `run/scripts/uniqpix/special/dark/6996/zpix-special-dark-699663-699671.slurm` in an interactive node, which processed pixels 699663-699671 (only 9 because it was the last batch out of 259 input uniqpix to process).

## TODO

  * [x] double check bookkeeping — are all frames correctly represented in the correct uniqpix?
  * [x] as currently implemented, the hpix2upix map for each survey/program uses the largest nside that was needed for that survey/program, which can be different than other survey/programs. Although that is kept as a header keyword, this smells like a hassle.  Maybe better to standardize on nside_max being the same for all survey/programs within a production, e.g. nside_max=256 or 512.
  * [x] naming - uniqpix_to_map, get_exp2uniqpix_map isn't particularly clear.  The script directory is `run/scripts/uniqpix/SURVEY/PROGRAM` while the output directory is `spectra/SURVEY/PROGRAM`. etc.
  * [x] This broke the ability to run desi_zproc from a production exposures file instead of pre-generating the per-pixel mapping of exposure-petals -> pixels. That isn't needed for productions, but it broke tests and maybe is useful for custom runs.


sbailey:
@copilot please review, even though this is still a draft.

py/desispec/io/meta.py
```diff
@@ -322,9 +328,11 @@ def findfile(filetype, night=None, expid=None, camera=None,
         # z catalogs
         #
         zcatalog='{specprod_dir}/zcatalog-{specprod}.fits',  # deprecated
-        zcat_hp = '{specprod_dir}/zcatalog/{version}/zpix-{survey}-{faprogram}.fits',
+        zcat_hp = '{specprod_dir}/zcatalog/{version}/zpix-{survey}-{faprogram}.fits', # deprecated
```
akremin: Not critically important, but on line 291-302 you use *_hp , while here you deprecate *_hp and add *_pix naming for what is otherwise an identical string. Should we also rename the files in 291-302? And why do need to keep both labels around for zcatalog? I would have thought we use the zcat label is far fewer places than the spectra and redrock labels.

py/desispec/io/meta.py
```diff
@@ -383,12 +403,24 @@ def findfile(filetype, night=None, expid=None, camera=None,
     if tile is not None and healpix is not None:
         raise ValueError(f'Set healpix or tile but not both ({healpix}, {tile})')
 
-    #- Setting healpix implies groupname='healpix'
-    if healpix is not None and groupname is None:
```
akremin, line 387: It appears that we lost this old check for groupname None when healpix/uniqpix is defined. Perhaps you could add it to the current if statement under line 362?

py/desispec/io/meta.py
```diff
@@ -383,12 +403,24 @@ def findfile(filetype, night=None, expid=None, camera=None,
...
-        groupname = 'healpix'
+    #- uniqpix or healpix but not both
+    if uniqpix is not None and healpix is not None:
+        raise ValueError(f'Set uniqpix or healpix but not both ({uniqpix}, {healpix})')
+
+    pixbase = pix = None
+    if groupname == 'healpix':
+        pixbase = 'healpix'
+        pix = healpix
+    elif groupname in ('uniqpix', None):
```
akremin: Maybe this resolves my previous comment, but I am surprised that there isn't a filetype check here. This sets the variables if the groupname is `None` for any reason. I would have guessed we only do that for pix-based filetypes.

py/desispec/io/util.py
```diff
@@ -378,23 +378,18 @@ def _dict2ndarray(data, columns=None):
     return xdata
 
 
-def healpix_subdirectory(nside, pixel):
+def healpix_subdirectory(pixel):
```
akremin: Should we rename this `pix_subdirectory`?

py/desispec/scripts/healpix_redshifts.py
```diff
@@ -73,59 +76,87 @@ def main(args):
...
+        allpixels = [int(p) for p in args.uniqpix.split(',')]
     else:
-        allpixels = np.unique(np.asarray(exppix['HEALPIX']))
+        allpixels = np.unique(np.asarray(exppix['UNIQPIX']))
+
+    #- Save mapping of healpix to uniqpix as the maximum nside in uniqpix
+    # TODO: should this be based upon all the pixels in the exppix table, or just the ones we're going to run on?
+    uniqpix_for_map = np.unique(exppix['UNIQPIX'])
+    hpix2upix_map, nside_max = uniqpix_to_map(uniqpix_for_map)
+    outdir = f'{reduxdir}/spectra/{args.survey}/{args.program}'
```
akremin: I think this should use `findfile`

py/desispec/workflow/redshifts.py
```diff
@@ -106,35 +106,35 @@ def get_ztile_script_suffix(tileid, group, night=None, expid=None, subgroup=None
         log.warning(f'Non-standard tile group={group}; writing outputs to {suffix}.*')
     return suffix
 
-def get_zpix_redshift_script_pathname(healpix, survey, program):
-    """Return healpix-based coadd+redshift+afterburner script pathname
+def get_zpix_redshift_script_pathname(uniqpix, survey, program):
```
akremin: Again consider a rename to `get_pix_redshift_script_pathname`

py/desispec/pixgroup.py
```diff
@@ -47,80 +49,125 @@ def fibermap2tilepix(fibermap, nside=64):
...
+    log.info('Trimming zcat to one row per TILEID, PETAL_LOC, UNIQPIX')
+    columns = ['TILEID', 'PETAL_LOC', 'UNIQPIX']
+    upix_tiles = zcat.groupby(columns).size().reset_index(name='NTARGETS')
+
+    #- Add PETAL_LOC to frames based on CAMERA
+    log.info('Converting frames CAMERA to PETAL_LOC')
+    frames['PETAL_LOC'] = frames['CAMERA'].str[-1].astype(int)  # extract last character of CAMERA and convert to int
+
+    #- Drop camera; just use PETAL_LOC
+    frames = frames[['NIGHT', 'EXPID', 'TILEID', 'PETAL_LOC']].drop_duplicates(subset=['EXPID', 'PETAL_LOC'])
```
akremin, line 124: What is the benefit of creating pandas dataframes here and converting back at the end? Is it much more efficient?
Table joins and grouping by columns can be done with astropy tables, so I think this line and line 111 below are the only things that can't be done natively. However, they can be pretty easily done with a `np.unique` command, I think.

akremin (commented):
Thanks, overall this looks very nice. I have minor in-line comments mostly related to naming schemes. 

We discussed the TODO items in person, which I'll summarize here: 

- I think we should use a fixed NSIDE for the lookup table for all surveys and programs in a production, but we should let users assume that maximum is the same across different productions. 
- We should keep the scripts and data directory naming consistent
- We can reduce the lookup table from a `HPIX,UNIQPIX` table to just `UNIQPIX` since the healpix is just the index of the array.
- We should strive to support uniqpix even if the mapping files don't exist, but we don't need those for Matterhorn and therefore don't have to have that for this PR. It is okay to remove those tests and create an issue to address in the future.

py/desispec/io/util.py
```diff
@@ -379,22 +379,22 @@ def _dict2ndarray(data, columns=None):
 
 
 def healpix_subdirectory(nside, pixel):
+    log = get_logger()
+    log.warning('DEPRECATED: healpix_subdirectory is deprecated. Use pix_subdirectory instead.')
+    return pix_subdirectory(pixel)
+
```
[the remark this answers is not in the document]
sbailey: @copilot apply changes based on [this feedback](https://github.com/desihub/desispec/pull/2755#discussion_r3251565143)

py/desispec/pixgroup.py
```diff
@@ -47,80 +49,128 @@ def fibermap2tilepix(fibermap, nside=64):
...
+    nside, ipix = desiutil.healpix.upix2hpix(uniqpix)
+    order = np.log2(nside).astype(int)
+
+    if nside_max is None:
+        nside_max = int(np.max(nside))
+    elif nside_max < np.max(nside):
+        raise ValueError(f"{nside_max=} is too small for the maximum nside {np.max(nside)} in uniqpix")
+
+    order_max = np.log2(nside_max).astype(int)
+    npix_max = 12 * nside_max ** 2
```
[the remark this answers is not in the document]
sbailey: @copilot apply changes based on [this feedback](https://github.com/desihub/desispec/pull/2755#discussion_r3251565160)

py/desispec/scripts/zcatalog_wrap.py
```diff
@@ -98,7 +98,7 @@ def main(args=None):
 
     ## Define filetype based on healpix vs not healpix
     if args.group == 'healpix':
-        ftype = 'zcat_hp'
+        ftype = 'zcat_pix'
     else:
         ftype = 'zcat_tile'
```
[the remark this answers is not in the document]
sbailey: fixed in commit b219474f (to be pushed in a batch, not yet on github)

py/desispec/scripts/uniqpix_redshifts.py
```diff
@@ -73,59 +78,94 @@ def main(args):
...
+    hpix2upix_map, nside_max = get_hpix2upix_map(uniqpix_for_map, args.nside_max)
+    outdir = io.findfile('spectra_base', survey=args.survey, faprogram=args.program)
+    header = dict(
+            NSIDE = nside_max,
+            HPXNSIDE = nside_max, # same as NSIDE, but consistent with other files
+            NPXNEST = True,
+            SURVEY = args.survey,
+            PROGRAM = args.program,
+            SPECPROD = os.getenv('SPECPROD', 'unknown'),
+            )
```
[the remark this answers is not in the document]
sbailey: fixed in commit 7361b118 (to be pushed to github in a batch)

py/desispec/util.py
```diff
@@ -845,4 +845,29 @@ def wrap_long_logs(text, width=120):
...
+    if isinstance(data, Table):
+        return data[columns].to_pandas()
+    elif isinstance(data, np.ndarray):
+        data = data[columns]
+        data = data.astype(data.dtype.newbyteorder('='))  # ensure native byte order for pandas
+        return pd.DataFrame(data)
+    elif isinstance(data, pd.DataFrame):
+        return data[columns]
+    else:
+        raise ValueError(f'Unexpected {type(data)=}, should be astropy Table, numpy structured array, or pandas DataFrame')
```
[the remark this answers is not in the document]
sbailey: This is a specialized function not intended for most end users, so I don't want to bring in pandas as a required dependency for all users via setup.cfg (it is part of our core desiconda installation).  I think the exceptions are clear enough as is if pandas is missing or the users passes an invalid array. Not taking action on this comment.

py/desispec/io/meta.py
```diff
@@ -283,17 +288,19 @@ def findfile(filetype, night=None, expid=None, camera=None,
...
-        hpixexp='{specprod_dir}/healpix/{survey}/{faprogram}/{hpixdir}/hpixexp-{survey}-{faprogram}-{healpix}.csv',
+        spectra_base  = '{specprod_dir}/spectra/{survey}/{faprogram}',
+        coadd_pix     = '{specprod_dir}/{pixbase}/{survey}/{faprogram}/{pixdir}/coadd-{survey}-{faprogram}-{pix}.fits',
+        rrdetails_pix = '{specprod_dir}/{pixbase}/{survey}/{faprogram}/{pixdir}/rrdetails-{survey}-{faprogram}-{pix}.h5',
+        rrmodel_pix   = '{specprod_dir}/{pixbase}/{survey}/{faprogram}/{pixdir}/rrmodel-{survey}-{faprogram}-{pix}.fits',
+        spectra_pix   = '{specprod_dir}/{pixbase}/{survey}/{faprogram}/{pixdir}/spectra-{survey}-{faprogram}-{pix}.fits{compsuffix}',
+        redrock_pix   = '{specprod_dir}/{pixbase}/{survey}/{faprogram}/{pixdir}/redrock-{survey}-{faprogram}-{pix}.fits',
+        qso_mgii_pix  = '{specprod_dir}/{pixbase}/{survey}/{faprogram}/{pixdir}/qso_mgii-{survey}-{faprogram}-{pix}.fits',
+        qso_qn_pix    = '{specprod_dir}/{pixbase}/{survey}/{faprogram}/{pixdir}/qso_qn-{survey}-{faprogram}-{pix}.fits',
+        emline_pix    = '{specprod_dir}/{pixbase}/{survey}/{faprogram}/{pixdir}/emline-{survey}-{faprogram}-{pix}.fits',
```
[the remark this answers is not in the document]
sbailey: implemented in commit f8ca1d71, along with healpix_redshifts.py cleanup to use it.

py/desispec/workflow/redshifts.py
```diff
@@ -106,35 +107,35 @@ def get_ztile_script_suffix(tileid, group, night=None, expid=None, subgroup=None
...
+        scriptname = f'zpix-{survey}-{program}-{uniqpix[0]}.slurm'
     else:
-        scriptname = f'zpix-{survey}-{program}-{hpixmin}-{hpixmax}.slurm'
+        scriptname = f'zpix-{survey}-{program}-{upixmin}-{upixmax}.slurm'
 
     reduxdir = desispec.io.specprod_root()
-    return os.path.join(reduxdir, 'run', 'scripts', 'healpix',
-                        survey, program, str(hpixmin//100), scriptname)
+    return os.path.join(reduxdir, 'run', 'scripts', 'spectra',
+                        survey, program, str(pix_subdirectory(upixmin)), scriptname)
```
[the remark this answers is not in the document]
sbailey: There is opportunity for confusion here, but we felt that "spectra" as a directory name on disk was less cryptic for the end-user than "uniqpix", while "uniqpix" in option names for generating felt more descriptive than e.g. `--spectra 1234`.  No action yet, and we'll see how it goes.

py/desispec/workflow/redshifts.py
```diff
@@ -183,9 +184,9 @@ def create_desi_zproc_batch_script(group,
...
         night = thrunight
-    elif group == 'healpix':
-        if (healpix is None or survey is None or program is None):
-            msg = f"group='healpix' must define healpix,survey,program (got {healpix},{survey},{program})"
+    elif group == 'uniqpix':
+        if (uniqpix is None or survey is None or program is None):
+            msg = f"group='uniqpix' must define uniqpix,survey,program (got {uniqpix},{survey},{program})"
             log.error(msg)
             raise ValueError(msg)
     else:
```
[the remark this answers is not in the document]
sbailey: Good catch. This problem also exists in main, causing healpix jobs to use "customztile" instead of "healpix" (confirmed in kibo logs)..  Updated to fully support uniqpix in commit e656d5e8.

py/desispec/scripts/zproc.py
```diff
@@ -79,15 +79,15 @@ def parse(options=None):
...
-            help="healpix pixels (nested nside=64")
-    healpix_options.add_argument("--survey", help="survey name, e.g. main,sv1,sv3")
-    healpix_options.add_argument("--program", help="program name, e.g. dark,bright,backup,other")
-    healpix_options.add_argument("--expfiles", nargs='*',
-                        help="csv files with NIGHT,EXPID,SPECTRO,HEALPIX")
-    healpix_options.add_argument("--prodexpfile", help="production summary exposure file (using pre-generated --expfiles is more efficient)")
+    #- Options for uniqpix-based redshifts
+    uniqpix_options = parser.add_argument_group("uniqpix-based options (--groupname uniqpix)")
+    uniqpix_options.add_argument("-p", "--uniqpix", type=int, nargs='*', default=None,
+            help="uniqpix pixels (healpix + 4*nside^2")
```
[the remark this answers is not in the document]
sbailey: Odd miss.  While it is indeed a typo in a help string, it is just instead a string that doesn't impact whether the code can run. Fixed typo.

py/desispec/io/meta.py
sbailey: This is indeed handled by the later logic, but the overall flow is hardly clear and we're largely relying upon the unit tests to check that the various cases are covered.  I added your suggestion for clarity and robustness.

py/desispec/io/meta.py
sbailey: The tile-based cases are handled previously, defaulting to groupname='cumulative'. If the code gets to this point with groupname==None, it is uniqpix-based. Admittedly not super-clear / straightforward. I added a comment to at least help a bit.

py/desispec/io/util.py
sbailey: good idea; I also left `healpix_subdirectory(nside, pixel)` behind with a deprecation warning in case anyone was using it (unlikely).  The desispec code is updated to use `pix_subdirectory`.

py/desispec/workflow/redshifts.py
sbailey: I went the other direction on this, updating it to `get_zpix_script_pathname` for symmetry with the `get_ztile_script_pathname` function and because the script itself is called "zpix-*.slurm".

py/desispec/scripts/healpix_redshifts.py
sbailey: good idea, done.

py/desispec/pixgroup.py
sbailey: I added an implementation note comment about this.  The core reason is that pandas.DataFrame is ~10x faster than astropy Table for the merge/join operations on 10s of millions of rows from the zcatalog, reducing the runtime from minutes to seconds, specifically adding UNIQPIX to every row in the zcatalog, after it had been calculated for each unique TARGETID in a trimmed "targets" catalog:
```python
zcat = zcat.merge(targets[['TARGETID', 'UNIQPIX']], on='TARGETID', how='inner')
```

I went through multiple implementation options, and this pandas-based solution was the fastest I found (kudos to Claude for suggesting trying pandas when astropy.table.join and other by-hand merge options seemed too slow).

py/desispec/io/meta.py
sbailey: Cleaned up to use just _pix and _tile for zcat and zall.

sbailey:
I believe I have addressed all comments from copilot and @akremin .  I'm proceeding with a final round of testing before merging and tagging.

akremin:
I apologize because I think it was my comment/request that led to this new bug, but I don't think the findfile logic works for healpix anymore, (however uniqpix does work as expected):
```
In [14]: findfile('qso_qn_pix', healpix=10000, survey='main', faprogram='dark')
Out[14]: '/global/cfs/cdirs/desi/spectro/redux/kremin/spectra/main/dark/263/26384/qso_qn-main-dark-26384.fits'

In [15]: findfile('qso_qn_pix', uniqpix=10000, survey='main', faprogram='dark')
Out[15]: '/global/cfs/cdirs/desi/spectro/redux/kremin/spectra/main/dark/100/10000/qso_qn-main-dark-10000.fits'
```

Note that (1) it lists healpix as under `/spectra/` instead of `/healpix/` and (2) that it seems to be changing the number from what the user specified.

sbailey:
> In [14]: findfile('qso_qn_pix', healpix=10000, survey='main', faprogram='dark')
> Out[14]: '/global/cfs/cdirs/desi/spectro/redux/kremin/spectra/main/dark/263/26384/qso_qn-main-dark-26384.fits'

That behavior was purposeful, but admittedly surprising. The idea was that we should default to the new uniqpix/spectra scheme unless explicitly specifying groupname='healpix'.  After discussion with @akremin, we agreed to special case that: if you specify only healpix, you mean groupname=healpix, but if you specify both nside and healpix, it should default to the new uniqpix=healpix+4*nside^2 scheme.  The full table of options:
```
  groupname   nside   healpix uniqpix Result    Comment
  --          --      --      --      tiles     not healpix or uniqpix
* --          --      --      1234    uniqpix
* --          --      1234    --      healpix   with nside=64
  --          --      1234    1234    Error     don't mix healpix and uniqpix
  
  --          64      --      --      Error     must specify healpix or uniqpix
  --          64      --      1234    Error     don't mix nside and uniqpix
! --          64      1234    --      uniqpix   from nside,healpix -> uniqpix
  --          64      1234    1234    uniqpix/Error if uniqpix != healpix + 4*nside^2
  
  healpix     --      --      --      Error     must specify healpix
  healpix     --      --      1234    Error     don't mix healpix and uniqpix
* healpix     --      1234    --      healpix   assume nside=64
  healpix     --      1234    1234    Error     don't mix healpix and uniqpix
  
  healpix     64      --      --      Error     must specify healpix
  healpix     64      --      1234    Error     don't mix nside and uniqpix
  healpix     64      1234    --      healpix
  healpix     64      1234    1234    healpix/Error if uniqpix != healpix + 4*nside^2
  
  uniqpix     --      --      --      Error     must specify uniqpix
* uniqpix     --      --      1234    uniqpix
  uniqpix     --      1234    --      Error     must specify nside too
  uniqpix     --      1234    1234    Error     don't mix healpix and uniqpix
  
  uniqpix     64      --      --      Error     must specify uniqpix, not nside
  uniqpix     64      --      1234    Error     don't mix nside and uniqpix
  uniqpix     64      1234    --      uniqpix   Use uniqpix = healpix + 4*nside^2
  uniqpix     64      1234    1234    uniqpix/Error if uniqpix != healpix + 4*nside^2
```

"*" highlights intended standard usage, and "!" highlights the potentially surprising case that will emit a warning but still default to groupname='uniqpix', e.g.
```
In [2]: findfile('qso_qn_pix', healpix=10000, survey='main', faprogram='dark')
Out[2]: '/global/cfs/cdirs/desi/spectro/redux/m5/healpix/main/dark/100/10000/qso_qn-main-dark-10000.fits'

In [3]: findfile('qso_qn_pix', nside=64, healpix=10000, survey='main', faprogram='dark')
WARNING:meta.py:409:findfile: Using uniqpix=healpix+4*nside^2; if you really want healpix-based files, set groupname="healpix"
Out[3]: '/global/cfs/cdirs/desi/spectro/redux/m5/spectra/main/dark/263/26384/qso_qn-main-dark-26384.fits'

In [4]: findfile('qso_qn_pix', nside=64, healpix=10000, survey='main', faprogram='dark', groupname='healpix')
Out[4]: '/global/cfs/cdirs/desi/spectro/redux/m5/healpix/main/dark/100/10000/qso_qn-main-dark-10000.fits'
```

Tests now explicitly cover all of those situations in the above table, and the findfile logic has been "simplified" to group all of that if/then logic into a continuous place in the code.

sbailey:
For the record: test runs highlighted cases like special bright uniqpix=123 with 5000 targets but 85000 input spectra (!).  If that pixel gets combined with other pixels in the same zproc job, it runs out of memory.  Updated logic for desi_uniqpix_redshifts:
* partition the sky by uniqpix with a default maximum of 5000 targets before splitting into subpixels
* submit jobs with a max of 16 uniqpix, but use fewer uniqpix if needed to have a maximum of 50,000 input spectra.
  * special case: if a uniqpix has >50,000 input spectra, it runs in a single job by itself.
* job names are determined by a hash of the input pixels that they process.  `run/scripts/spectra/SURVEY/PROGRAM/pixjobs-SURVEY-PROGRAM.csv` track which which pixels got processed by which jobs.

I have tested this with various extreme cases, all of which ran in <1 hour without blowing memory:

```
                JOBHASH          NPIX NTARGETS NSPECTRA UNIQPIX
                ---------------- ---- -------- -------- -------
special/bright  d2732c294ac87abf    1     5000    85000     123
special/bright  749035f9d6d5b367   16    23506    46309   25957,28244,28247...
main/dark       166b6381f1e8066f   16    41392    49487   27550,44732,27784...
```

sbailey:
Copilot review number 2 suggestions have been implemented.

PR #2767, opened 2026-06-12
Only try make science based nightqa PDFs if the data exists.

dylanagreen:
@abrodze encountered the first night with no science data since the daily environment changeover on May 29. Turns out that matplotlib 3.10 removed the ability to create empty pdfs, something which nightqa depended on in the old environment for nights where there were no data. The removal of this ability meant that on nights with no data, `sframesky` (and incidentally `tileqa` and `petalnz`) would crash, crashing the script shortly after the dark steps and without generating an html page.

`skyzbifer` would succeed, but only because it makes a PNG instead of a PDF. 

In this PR, we intercept at the start of nightqa when looking for science data. If no science data is found, we no longer try to do any of the aforementioned steps that require science data.

Note that the `html` step is robust to missing data, but does require the values of expids, tileids and surveys, so we consider it for triggering the search for those values on a given night, but do not skip that step if we didn't find any data.

### DIFF
--- a/bin/desi_night_qa
+++ b/bin/desi_night_qa
@@ -159,8 +159,11 @@ def main():
     # DG - the following steps require science data, so if any of them
     # are requested, check and find expids, tileids and surveys observed on the night.
     # fF we find no science data that night, then remove these steps from steps_tbd.
-    steps_need_science = ["sframesky", "tileqa", "skyzfiber", "petalnz"]#, "html"]
-    if np.isin(steps_need_science, steps_tbd).sum() > 0:
+    steps_need_science = ["sframesky", "tileqa", "skyzfiber", "petalnz"]
+    # DG - HTML doesn't need science data to exist, but needs the values in expids
+    # tileids, surveys for generation, so we consider for triggering looking
+    # for those values, but don't remove it if there's no science data.
+    if np.isin(steps_need_science + ["html"], steps_tbd).sum() > 0:
         expids, tileids, surveys = get_surveys_night_expids(args.night)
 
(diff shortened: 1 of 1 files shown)


bin/desi_night_qa
```diff
@@ -156,10 +156,21 @@ def main():
...
+    # are requested, check and find expids, tileids and surveys observed on the night.
+    # fF we find no science data that night, then remove these steps from steps_tbd.
+    steps_need_science = ["sframesky", "tileqa", "skyzfiber", "petalnz"]
+    # DG - HTML doesn't need science data to exist, but needs the values in expids
+    # tileids, surveys for generation, so we consider for triggering looking
+    # for those values, but don't remove it if there's no science data.
+    if np.isin(steps_need_science + ["html"], steps_tbd).sum() > 0:
         expids, tileids, surveys = get_surveys_night_expids(args.night)
 
+        if len(expids) == 0:
```
akremin, line 169: Two efficiency notes that are irrelevant for this case since we're only searching a list of a few strings, but rather than two list comprehensions this would likely be more efficient with set operations. The second list comprehension also would only need to check in the subset of `skipped`, since the first list comprehension already assures the other steps in `steps_need_science` are not present.
Again, neither of these substantively matter for this case. I'll approve this as-is.

akremin (approved):
Thank you for the quick fix. As discussed in person this is clean both for future maintainability of the code and readability of the logs.
I put one comment inline about best practices, but it is not important for this particular case and I'll merge as-is.

Issue #2771, opened 2026-06-15
Use the DR11 layer in tile QA images


sybenzvi:
As noted by @djschlegel [on Slack](https://desisurvey.slack.com/archives/C01HNN87Y7J/p1781366223839559), the tile QA plots should start to use the DR11 layer (currently `ls-dr11-early-v2`) when fetching cutouts from [legacysurvey.org](https://www.legacysurvey.org/viewer-desi).

I believe changing the default layer in `get_viewer_cutout` in `tile_qa_plot.py` (lines [270](https://github.com/desihub/desispec/blob/9a44952611b29b53befcab1ee7bc788b461a47d6/py/desispec/tile_qa_plot.py#L270) and [390](https://github.com/desihub/desispec/blob/9a44952611b29b53befcab1ee7bc788b461a47d6/py/desispec/tile_qa_plot.py#L390)) is sufficient.

PR #2772, opened 2026-06-15
Switch to DR11 layer in legacysurvey URLs.

sybenzvi:
PR to address #2771.

Note that the DR9 pixweights are still being used in [plot_mw_skymap](https://github.com/desihub/desispec/blob/9a44952611b29b53befcab1ee7bc788b461a47d6/py/desispec/tile_qa_plot.py#L759). Those may need to be updated, too.


akremin (approved):
Thanks @sybenzvi the changes look good and minimal. Thank you for also updating the docstrings.
It does not appear that pixweights exist yet within the dr11 directory, so we will have to update that at a later date. Perhaps when we go from `dr11-early-v2` to `dr11`.

PR #2774, opened 2026-07-06
night QA petalnz fix when there are no tiles to plot

sbailey:
This PR fixes a nightqa bug that occurred on 20260703 and 20260704.  Basically if there were no tiles observed to make n(z) plots, then there would be no petalnz_pdf made but it would still try to rename the non-existent temporary file to the final file, e.g.
```
WARNING:night_qa.py:1766:create_petalnz_pdf: no /dvs_ro/cfs/cdirs/desi/spectro/redux/daily/tiles/cumulative/45991/20260703/tile-qa-45991-thru20260703.fits file, proceeding to next tile
  File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desispec/main/bin/desi_night_qa", line 247, in <module>
    main()
    ~~~~^^
  File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desispec/main/bin/desi_night_qa", line 228, in main
    create_petalnz_pdf(outfns["petalnz"], args.night, args.prod, unq_tileids,
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                       unq_surveys, dchi2_threshold=25, group=args.group)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desispec/main/py/desispec/night_qa.py", line 2120, in create_petalnz_pdf
    os.rename(tmp_outpdf, outpdf)
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/global/cfs/cdirs/desi/spectro/redux/daily/nightqa/20260703/petalnz-20260703_tmp1922671.pdf' -> '/global/cfs/cdirs/desi/spectro/redux/daily/nightqa/20260703/petalnz-20260703.pdf'
```

This fix adds safety checks to not crash in this situation.  I have tested it by using this branch to generate the 20260703 and 20260704 nightqa pages.

I'm a little surprised that we didn't encounter this before, but I haven't dug into exactly why not. It is possible that this is a change of behavior in the latest version of PdfPages and this is the first time we've had an exclusively backup night since then.

### DIFF
--- a/py/desispec/night_qa.py
+++ b/py/desispec/night_qa.py
@@ -1916,9 +1916,9 @@ def create_petalnz_pdf(
         "ELG" : "b",
         "QSO" : "orange",
     }
-    with PdfPages(tmp_outpdf) as pdf:
-        # AR we need some tiles to plot!
-        if ntiles["bright"] + ntiles["dark"] > 0:
+    # AR we need some tiles to plot!
+    if ntiles["bright"] + ntiles["dark"] > 0:
+        with PdfPages(tmp_outpdf) as pdf:
             for survey in np.unique(surveys):
                 ntiles_surv = {
                     faprgrm : np.unique(
@@ -1927,6 +1927,7 @@ def create_petalnz_pdf(
                 }
                 # AR plotting only if some tiles
                 if np.sum([ntiles_surv[faprgrm] for faprgrm in faprgrms]) == 0:
+                    log.info(f"no SURVEY={survey} tiles with FAPRGRM in {faprgrms}, skipping plots")
                     continue
                 # AR three plots:
                 # AR - fraction of VALID fibers, bright+dark together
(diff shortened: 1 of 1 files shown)


akremin (approved):
Thanks this looks good. One thing to note is that this removes any existing outpdf even if the function didn't produce a file, which is a change of behavior from what it used to do (it would crash and leave behind an old file). This new behavior is what we want, however, so it's a good change.

sbailey:
converting to a draft to avoid merging while investigating a test crash, which might be unrelated. TBD.

sbailey:
For the record, the crash I was seeing was a side effect of having SPECPROD=sjbailey instead of SPECPROD=daily while testing.  I think we're good to go for merging after tests re-pass.

PR #2776, opened 2026-07-08
add good_z_lya

abrodze:
Added the quality flags for non qso targets to define GOOD_Z_LYA. The corresponding Z_QSO values will need updated still and a decision of Z_CONF level


abrodze:
The selection cuts reproduce /global/cfs/cdirs/desicollab/users/martini/QSO-Catalogs/QSO_cat_matterhorn_main_dark_healpix_v2.fits if you also apply the COADD_FIBERSTATUS variable cut. The WG applies this cut because of potential forest correlations and not because the redshift is bad, so I chose not to here

From in-person discussions with Rongpu, the lingering decision is whether the new Z_QSO values become Z_BEST and whether they get Z_CONF 2 or 3.

PR #2780, opened 2026-07-13
add good_z_lya criteria

abrodze:
(1) adds good_z_lya column to redshift catalog following mkQSO_2.py in LSS
(2) sets Z_BEST = Z_QSO if good_z_lya is set 
(3) sets zconf=3 for good_z_lya objects
(4) does not use BGS cuts in GOOD_Z_LYA. This has poor accuracy and corresponds to <<1% of objects z>1.77 so it is irrelevant for Lya anyway

### DIFF
--- a/py/desispec/validredshifts.py
+++ b/py/desispec/validredshifts.py
@@ -169,7 +169,8 @@ def validate(redrock_path, fiberstatus_cut=True, return_target_columns=False, ex
     return cat
 
 
-def actually_validate(cat, fiberstatus_cut=True, ignore_emline=False, ignore_qso=False, ignore_lya=False):
+def actually_validate(cat, fiberstatus_cut=True, ignore_emline=False, ignore_qso=False, ignore_lya=False,
+                      populate_empty_columns=False):
     '''
     Apply redshift quality criteria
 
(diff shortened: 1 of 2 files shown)


rongpu:
@abrodze Thanks. I have moved the GOOD_Z_LYA checks to validredshifts.py (as we had discussed) and fixed some minor bugs. Please take a look.

I am restricting the GOOD_Z_LYA assessment to the main survey because I have only validated (VI'ed) the main-survey redshifts (but feel free to expand it to other surveys if you are confident about their LyA redshifts).

py/desispec/validredshifts.py
abrodze: This and other uses of get_good_fiberstatus are redundant with the zcatalog.py good_spec condition. I think this is fine so that validredshifts.py works in isolation, but noting.

We should apply &= get_good_fiberstatus(cat) to GOOD_Z_LYA, the only one missing this condition currently.

abrodze: Noting that the results would be unchanged for zcatalog.py script, but does change this script's results

rongpu: @abrodze Thanks for catching this. I have fixed it and reorganized the `fiberstatus_cut` instances in `validredshifts.py` for better readability.

Indeed we should keep `validredshifts.py` as an independent script. In `zcatalog.py` we do apply the `get_good_fiberstatus` requirement twice (first when running `validredshifts.actually_validate` and then when requiring `GOOD_SPEC==True`), but I think it's fine.

Issue #2785, opened 2026-08-08
validredshifts now requires SCND_TARGET column; and also add WISE_VAR_QSO column?


araichoor:
if I m correct, https://github.com/desihub/desispec/pull/2780 introduced the following issue:
because of this line https://github.com/desihub/desispec/blob/4ed24c7851348d007c9c6df1dbdfa5459efab6bc/py/desispec/validredshifts.py#L238, it nows uses `SCND_TARGET`, which is not present currently:
```
>>> d = validate("/global/cfs/cdirs/desi/spectro/redux/matterhorn/tiles/cumulative/1000/20210517/redrock-0-1000-thru20210517.fits")
[...]
 File "/global/common/software/desi/perlmutter/desiconda/20260227-2.3.1/code/desispec/main/py/desispec/validredshifts.py", line 238, in actually_validate
    is_wise_qso = (cat['SCND_TARGET'] & 2**35) != 0  # scnd_mask.WISE_VAR_QSO
                   ~~~^^^^^^^^^^^^^^^
[...]
KeyError: 'SCND_TARGET'
```
I think the fix is:
- adds `scnd=True` in the `main_cmx_or_sv()` call here https://github.com/desihub/desispec/blob/4ed24c7851348d007c9c6df1dbdfa5459efab6bc/py/desispec/validredshifts.py#L93, ie:
```
surv_target, surv_mask, surv = main_cmx_or_sv(tmp_fibermap, scnd=True)
```
- and then adapts these two lines https://github.com/desihub/desispec/blob/4ed24c7851348d007c9c6df1dbdfa5459efab6bc/py/desispec/validredshifts.py#L97-L98 with:
```
    desi_target_col, bgs_target_col, _, scnd_target_col = surv_target
    desi_mask, bgs_mask, _, _ = surv_mask
```

Loosely related, I m wondering if it would make sense to add a `WISE_VAR_QSO` boolean column here https://github.com/desihub/desispec/blob/4ed24c7851348d007c9c6df1dbdfa5459efab6bc/py/desispec/validredshifts.py#L136-L144...

sbailey:
@abrodze please take a look. Let's also include unit tests that would have caught this error, plus covering the case from PR #2780 that did not trigger this crash to make sure we don't break that while fixing this.  I'm not quite sure what all the special cases are (e.g. what happens if the input file doesn't have SCND_TARGET at all?)

abrodze:
[...]

PR #2786, opened 2026-08-11
add wise_var_qso to validredshifts function

abrodze:
This PR updates the validate function in validredshifts to accommodate the new GOOD_Z_LYA column (PR #2780), addressing issue #2785.

### DIFF
--- a/py/desispec/test/test_validredshifts.py
+++ b/py/desispec/test/test_validredshifts.py
@@ -1,0 +1,617 @@
+"""
+test desispec.validredshifts
+"""
+
+import os
+import shutil
+import tempfile
+import unittest
+
+import numpy as np
+from astropy.table import Table
+import fitsio
+
+from desispec.maskbits import fibermask
+from desispec.validredshifts import (get_good_fiberstatus, actually_validate,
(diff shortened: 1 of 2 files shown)


rongpu:
Since DR1/iron is our only major public data release, I would prefer that anyone using the public DR can still run this script on redrock files without crashing. So I made the DELTACHI2_NEW column optional: if it does not exists in the QN afterburner files, it automatically switches to DELTACHI2. Otherwise this PR looks good.

The only place where DELTACHI2_NEW is used is the ["low-z" QSO failure mode cut](https://github.com/desihub/desispec/blob/3eb39778534ef67a54d76702fd7e7b4b1dbd2645/py/desispec/validredshifts.py#L265-L266). I have checked this and found that almost all (98%) of the previously identified low-z failures are still identified using DELTACHI2 (and few extra failures are introduced). So I think this is the best we can do given the limitations. And in fact this is exactly what I did for the v2 zcatalog for Iron.

I have not read the unit test. LLMs tell me that it looks good.

abrodze:
I think this looks good, happy to merge after tests finish running

Issue #2371, opened 2024-09-12
consolidate psf and flat jobs


sbailey:
One of the challenges of running kibo was the sheer number of jobs, ~50k.  This required extra queue management to avoid submit limits, made rare startup problems common to encounter, and added overhead to reservations which sometimes had ~10% of the nodes idle while turning over jobs.  To reduce the number of jobs,
* bundle arc and psfnight jobs together for a 6 -> 1 reduction for those steps.  Consider also bundling the ccdcalib job into this bunch.
* bundle flat and nightlyflat jobs together for a 13 -> 1 reduction of those steps

Although there are still more science jobs than there are arc/psf/flat/nightlyflat jobs, these calib jobs are tightly coupled and blocking factors for getting to the more genuinely independent tilenight and ztile jobs, so bundling them would streamline the workflow.

akremin:
Resolved with PR #2799

Issue #2784, opened 2026-08-04
Reconsider psf arc job size


sbailey:
We currently run arc / psf / specex jobs on 3 CPU nodes with e.g.
```
export NIGHT=20250914
export EXPID=311138
time srun -N 3 -n 301 -c 2 --cpu-bind=cores desi_proc -n $NIGHT -e $EXPID --mpi
```
That uses 300 ranks + 1 coordination rank to process 600 bundles on 3 nodes.  That takes ~7 minutes, i.e. ~21 node-minutes.

I just tried
```
time srun -N 1 -n 101 -c 2 --cpu-bind=cores desi_proc -n $NIGHT -e $EXPID --mpi
```
i.e. 100 worker ranks on a single node.  That took ~11 minutes, i.e. 60% longer wallclock time but 1.9x cheaper in node-minute time.  Both the 3 node and 1 node jobs were reproducible over 2 runs each.

Revisiting this optimization may be especially advantageous if we bundle all 5 arc exposures into a single job, potentially needing 5 nodes for 11 minutes instead of 5 jobs * 3 nodes/job = 15 nodes for 7 minutes.

Heads up @akremin

akremin:
This was done as part of the updated arc job bundling in PR #2799

PR #2799, opened 2026-08-14
Bundle arcs into a single psfnight job and do similar for flats and cteflats

akremin:
#### Overview
This resolves issues #2371 and #2784 by consolidating all arcs and psfnight jobs into one psfnight job that first runs the individual arc commands and then the joint fit job. A similar thing is done for the flats and nightlyflat, and for the CTE flats.
This is a 6->1, 13->1, and 3->1 reduction in jobs, respectively.

Important choices made:
- I've reduce the number of nodes per arc job to 1 so that all 5 can run in parallel in the Perlmutter realtime queue for daily operations. This is actually more compute-efficient anyway, as noted by Stephen in issue #2784 .
- The twelve flats each need a node and cannot run in parallel within our realtime allocation, so I have throttled them down to 4 nodes. GNU parallel takes care of managing the job execution, which I've tested and works well. This takes three times longer in wall-clock compared to running all in parallel, but is roughly the same in compute time and greatly reduces the job-footprint by a factor of 13.
- I have invented a new cteflat job that bundles the 0-3 CTE flats taken on a night into a single job. This requests one node and each flat is run in serial. There is no joint fit job at the end. If no CTE flats exist on a night, no job is submitted and the code proceeds normally. 

I used coding assistants to help with this PR. I admittedly have not yet reviewed all of the new unit tests written yet, but have reviewed all of the other code.

Another thing that could use some more stress-testing is that all camword corner cases work, but there were multiple dry_run nights that included all exposures with less than 30 cameras and those at least propagate correctly.

Finally, I did not test the ability of the script to pickup if several exposures already exist on disk. The update does check for that and submits only the subset that needs processing, but I didn't explicitly test that functionality as it is unlikely to happen often in practice with calibrations.

#### Tests
My tests are at NERSC: `/global/cfs/cdirs/desi/spectro/redux/jobbundle`

I ran 3 sets of tests:

1. I ran 20260319 in it's entirety from biaspdark through redshifts to verify that the new changes didn't impact other job types. This included 45 biaspdark jobs, which I ran interactively and then restarted the pipeline after some processing table hacking on the ccdcalib job dependencies only. I resubmitted all jobs using `desi_resubmit_queue_failures`, which worked as expected with the new bundled jobs and all downstream jobs. All bundle scripts have the correct number of nodes and parallelism, and the dependencies look right. And the expected output files exist.
2. I ran 20210514 in it's entirety from biaspdark through redshifts to verify that the new changes didn't impact other job types. I only ran with `--n-nights-darks=3` to reduce the number of nights that needed biaspdark jobs. This early night didn't have CTE flats and the pipeline happily submitted the whole night without submitting a cteflat job since there was nothing to do. Again all the scripts look good and outputs exist. As of writing the psfnight job ran and produced the expected outputs, the nightlyflat job is in the queue.
3. In dry_run_level=1 mode, I ran the 16th of every month from 2021-2026 if the exposure_table existed. This produced processing tables and scripts but didn't submit them. All of these produced scripts with the correct number of nodes and  parallelism. 

#### AI disclosure
I used Claude Opus 5 and ChatGPT Sol 5.6 in planning and executing this PR.

### DIFF
--- a/py/desispec/scripts/proc_night.py
+++ b/py/desispec/scripts/proc_night.py
@@ -34,8 +34,8 @@ from desispec.workflow.processing import define_and_assign_dependency, \
     create_and_submit, \
     submit_tilenight_and_redshifts, \
     generate_calibration_dict, \
-    night_to_starting_iid, make_joint_prow, \
-    set_calibrator_flag, make_exposure_prow, \
+    night_to_starting_iid, make_exposure_prow, \
+    make_calibration_bundle_prow, \
     all_calibs_submitted, \
     update_and_recursively_submit, update_accounted_for_with_linking, \
     submit_redshifts, check_darknight_deps_and_update_prow
@@ -586,12 +586,27 @@ def proc_night(night=None, proc_obstypes=None, z_submit_types=None,
     calibjobs = generate_calibration_dict(ptable, files_to_link)
 
     ## Determine the appropriate set of calibrations
-    ## Only run if we haven't already linked or done fiberflatnight's
+    ## Only run if we haven't already linked or done fiberflatnight's, or if
+    ## cte flats might still be missing. The cte flat bundle is independent of
+    ## the standard calibrations: a night can have every standard calibration
+    ## accounted for, including a linked fiberflatnight, and still need its own
+    ## cte flats processed for the nightly detector QA.
+    standard_calibs_needed = not all_calibs_submitted(calibjobs['accounted_for'],
+                                                      do_cte_flats, do_darknight)
+    ## cheap pre-check so nights that took no cte flats behave exactly as before
+    night_has_cte_flats = len(split_normal_and_cte_flats(etable)[1]) > 0
     cal_etable = etable[[]]
-    if not all_calibs_submitted(calibjobs['accounted_for'], do_cte_flats, do_darknight):
+    if standard_calibs_needed or (do_cte_flats and night_has_cte_flats):
         cal_etable = determine_calibrations_to_proc(etable,
                                                     do_cte_flats=do_cte_flats,
                                                     still_acquiring=still_acquiring)
+
+    cte_flats_needed = False
+    if do_cte_flats and len(cal_etable) > 0:
+        selected_ctes = split_normal_and_cte_flats(cal_etable)[1]
+        processed_cte_expids = get_processed_cte_expids(ptable)
+        cte_flats_needed = bool(np.any(~np.isin(selected_ctes['EXPID'],
+                                                processed_cte_expids)))
 
     ## Determine the appropriate science exposures
     sci_etable, tiles_to_proc = determine_science_to_proc(
@@ -632,8 +647,9 @@ def proc_night(night=None, proc_obstypes=None, z_submit_types=None,
         return prow, proctable
 
     ## Actually process the calibrations
-    ## Only run if we haven't already linked or done fiberflatnight's
-    if not all_calibs_submitted(calibjobs['accounted_for'], do_cte_flats, do_darknight):
+    ## Only run if we haven't already linked or done fiberflatnight's, or if
+    ## there are cte flats that no cteflat bundle covers yet
+    if standard_calibs_needed or cte_flats_needed:
         ptable, calibjobs, int_id = submit_calibrations(cal_etable, ptable,
                                                 cal_override, calibjobs,
                                                 int_id, night, files_to_link,
@@ -820,32 +836,19 @@ def submit_calibrations(cal_etable, ptable, cal_override, calibjobs, int_id,
     if len(cal_etable) == 0:
         return ptable, calibjobs, int_id
 
-    if len(ptable) > 0:
-        ## we use this to check for individual jobs rather than combination
-        ## jobs, so only check for scalar jobs where JOBDESC == OBSTYPE
-        ## ex. dark, zero, arc, and flat
-        explists = ptable['EXPID'][ptable['JOBDESC']==ptable['OBSTYPE']]
-        if len(explists) == 0:
-            processed_cal_expids = np.array([]).astype(int)
-        elif len(explists) == 1:
-            processed_cal_expids = np.unique(explists[0]).astype(int)
-        else:
-            processed_cal_expids = np.unique(np.concatenate(explists).astype(int))
-    else:
-        processed_cal_expids = np.array([]).astype(int)
+    ## CTE flats are submitted independently of the other calibrations, so
+    ## track which of them a cteflat bundle already covers. Arcs and normal
+    ## flats are covered by the psfnight/nightlyflat accounted_for flags.
+    processed_cte_expids = get_processed_cte_expids(ptable)
 
     ## Otherwise proceed with submitting the calibrations
     ## Define objects to process
-    darks, flats, ctes = list(), list(), list()
+    darks = list()
     zeros = cal_etable[cal_etable['OBSTYPE']=='zero']
     arcs = cal_etable[cal_etable['OBSTYPE']=='arc']
     if 'dark' in cal_etable['OBSTYPE']:
         darks = cal_etable[cal_etable['OBSTYPE']=='dark']
-    if 'flat' in cal_etable['OBSTYPE']:
-        allflats = cal_etable[cal_etable['OBSTYPE']=='flat']
-        is_cte = np.array(['cte' in prog.lower() for prog in allflats['PROGRAM']])
-        flats = allflats[~is_cte]
-        ctes = allflats[is_cte]
+    flats, ctes = split_normal_and_cte_flats(cal_etable)
 
     do_darknight = do_darknight and not calibjobs['accounted_for']['darknight']
     do_badcol = len(darks) > 0 and not calibjobs['accounted_for']['badcolumns']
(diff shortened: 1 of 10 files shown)


py/desispec/workflow/batch_writer.py
```diff
@@ -549,6 +590,424 @@ def create_ccdcalib_batch_script(night, expids, camword='a0123456789',
...
+    if system_name == 'perlmutter-gpu':
+        mps_wrapper = 'desi_mps_wrapper'
+        step_gpu_opt = '--gpus-per-node={} '.format(batch_config['gpus_per_node'])
+
+    ## OMP_NUM_THREADS is exported once for the whole exposure block
+    if step_class == 'ARC':
+        step_threads = max([res[2] for res in step_resources], default=1)
+    else:
+        step_threads = 1
+
```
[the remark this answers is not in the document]
sbailey: I agree with the concern but the proposed fix is incomplete.  I update the code myself to set OMP_NUM_THREADS per arc step.

py/desispec/workflow/batch.py
```diff
@@ -125,6 +125,29 @@ def parse_reservation(reservation, jobdesc):
...
+        int: the largest node count that should be requested for this job type.
+
+    Note:
+        PSFNIGHT is allowed more nodes than everything else because a bundled
+        arc job runs every selected arc exposure concurrently, one node each.
+        Everything else is throttled so that two jobs can run together in the
+        10-node realtime queue.
+    """
+    if jobdesc is not None and jobdesc.upper() == 'PSFNIGHT':
+        return 15
```
sbailey, line 146: This appears to be a leftover from when we thought we'd need 5*3=15 nodes to process arcs in parallel. This branch now uses 1 node per arc exposure for a total of 5 nodes and I don't think this cap of 15 is ever exercised.

py/desispec/workflow/batch_writer.py
```diff
@@ -549,6 +590,421 @@ def create_ccdcalib_batch_script(night, expids, camword='a0123456789',
...
+        queue (str, optional): the Slurm queue that will be used.
+        system_name (str, optional): name of batch system, e.g. perlmutter-cpu.
+
+    Returns:
+        tuple: A tuple containing:
+
+        * nodes: int, number of nodes for this step, always 1.
+        * ntasks: int, number of MPI ranks for this step.
+        * threads_per_task: int, number of threads per rank.
+        * runtime: float, estimated runtime of this step in minutes.
```
sbailey: Returning these as a tuple results in other code being concise but cryptic with magic numbers to index the tuple, e.g.
```
    step_runtimes = [res[3] for res in step_resources]
```
I suggest returning these as a dictionary so that they can be accessed with e.g. `res['runtime']` instead of `res[3]`.

sbailey (commented):
Overall, looks really good including the testing. I put two questions / suggestions inline in the context of long-term maintainability; they don't impact current functionality.

py/desispec/test/test_workflow_batch_writer.py
```diff
@@ -0,0 +1,468 @@
...
+ARC_NIGHT = 20260706
+ARC_EXPIDS = [360185, 360186, 360187, 360188, 360189]
+FLAT_NIGHT = 20260806
+FLAT_EXPIDS = [364647, 364648, 364649, 364652, 364653, 364654,
+               364657, 364658, 364659, 364662, 364663, 364664]
+CTE_NIGHT = 20260806
+CTE_EXPIDS = [364665, 364666, 364667]
+CTE_CAMWORDS = ['a0123456789', 'a012345678', 'a0123456789']
+FULL_CAMWORD = 'a0123456789'
+REFERENCE_REDUXDIR = '/global/cfs/cdirs/desi/spectro/redux/jobbundle'
```
sbailey: Claude-flavored code leftover: Let's not depend on a hardcoded NERSC path to a test production. Or perhaps this is a string in an evals comparison script that I haven't found, so maybe this just needs an explanatory comment.

py/desispec/test/test_workflow_batch_writer.py
```diff
@@ -0,0 +1,468 @@
...
+
+_evals_dir = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(
+        os.path.dirname(os.path.abspath(__file__))))), 'evals')
```
sbailey: Where is this `evals` dir?  It looks like it is in the top-level checkout of your desispec clone, but never committed?  It would be better to put this under py/desispec/tests/data with a more descriptive name than "evals", and then use `importlib.resources.files` to find it rather than a bunch of chained `dirname` calls.

sbailey (changes requested):
Review part 2, about test infrastructure.

py/desispec/workflow/batch.py
akremin: It is leftover from the original planning of 3 nodes x 5 arc jobs, but I wanted to keep it in case we ever changed our mind in the future. This piece of code is very subtle and has always silently reduced the resources from what was requested based on upper limits we've wanted to enforce. For that reason I wanted the upper limit to be 15 such that we could change to 3x5 nodes in the future if we wanted without extra code changes, while still keeping the sanity check in place so that we don't accidentally code up a 100 node job.

I'm happy to remove the 'psfnight' exception and limit everything to 5 nodes, remove this check entirely and trust the determine_resources() and the people who update it, or leave it as-is with 15 nodes. Please let me know what you'd prefer.

py/desispec/test/test_workflow_batch_writer.py
akremin: Yes this was a remnant of my development on my local machine. `evals` was where I put all of the planning and reviewing markdown files and it though those would be committed and kept in the repo, which is not my intension. Sorry for missing this. I'll get this cleaned up

py/desispec/workflow/batch.py
akremin: Speaking to Stephen he is indifferent to the three options listed and left it to me. I will leave it as-is at 15 node maximum. That way if we decide in the future to move back to 3 nodes per arc for speed reasons, we can do so without making this additional change.

py/desispec/workflow/batch_writer.py
akremin: That's a great suggestion, and I've implemented it in the latest commit.

py/desispec/test/test_workflow_batch_writer.py
akremin: This was a remnant of Claude wanting to use my example templates as an end-to-end comparison in unit tests rather than as a design schematic to build toward. I have removed this reference to a file path and the use of templates.

py/desispec/test/test_workflow_batch_writer.py
akremin: I have removed the use of templates from the tests and increased the coverage through bespoke tests instead.

py/desispec/scripts/proc_night.py
```diff
@@ -586,13 +586,28 @@ def proc_night(night=None, proc_obstypes=None, z_submit_types=None,
...
+    ## the standard calibrations: a night can have every standard calibration
+    ## accounted for, including a linked fiberflatnight, and still need its own
+    ## cte flats processed for the nightly detector QA.
+    standard_calibs_needed = not all_calibs_submitted(calibjobs['accounted_for'],
+                                                      do_cte_flats, do_darknight)
+    ## cheap pre-check so nights that took no cte flats behave exactly as before
+    night_has_cte_flats = len(split_normal_and_cte_flats(etable)[1]) > 0
     cal_etable = etable[[]]
-    if not all_calibs_submitted(calibjobs['accounted_for'], do_cte_flats, do_darknight):
+    if standard_calibs_needed or (do_cte_flats and night_has_cte_flats):
```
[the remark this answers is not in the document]
akremin: This is fixed in the latest push. It isn't the most important bug since we don't care much about cte flat processing and we care even less about them on nights with no science data, but it was still a bug and worth fixing.

py/desispec/workflow/processing.py
```diff
@@ -407,14 +526,21 @@ def desi_proc_joint_fit_command(prow, queue=None):
...
+    if direct_mode:
+        cmd += ' --mpi'
     return cmd
 
 
 def create_batch_script(prow, queue='realtime', dry_run=0, joint=False,
-                        system_name=None, use_specter=False, extra_job_args=None):
+                        system_name=None, use_specter=False,
+                        check_for_outputs=True, resubmit_partial_complete=True,
+                        extra_job_args=None):
```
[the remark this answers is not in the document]
akremin: This is not a concern as these are only used within the desispec pipeline infrastructure and are always called by name.

py/desispec/scripts/proc_night.py
```diff
@@ -920,70 +938,92 @@ def submit_calibrations(cal_etable, ptable, cal_override, calibjobs, int_id,
...
+    new_ctes = [cte_erow for cte_erow in ctes
+                if cte_erow['EXPID'] not in processed_cte_expids]
+    if len(new_ctes) > 0:
+        bundle_prow, step_prows, int_id = make_calibration_bundle_prow(
+                new_ctes, descriptor='cteflat', internal_id=int_id,
+                calibjobs=calibjobs)
+        extra_job_args = {'calibration_bundle_steps': step_prows,
+                          'bundle_full_camword': bundle_prow['PROCCAMWORD']}
+        bundle_prow, ptable = create_submit_add_and_save(bundle_prow, ptable,
+                                                         extra_job_args=extra_job_args)
```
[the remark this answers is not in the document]
akremin: Good catch. That was a known limitation that I failed to mention in the PR. Rather than leave it as an undocumented limitation that would require a future update, I've implemented a fix in the latest commit.

akremin:
Tests all pass and all AI and human comments have been resolved. Stephen gave offline approval to merge when ready, so I will proceed with merging

Issue #2792, opened 2026-08-12
20220605 b4 biasnightest


jbanovetz:
In creating the nightly bias on 20220605, b4 ran into an issue and only created a biasnighttest file. This is documenting the investigation into why.

jbanovetz:
I am seeing this in the `matterhorn` log:

```
INFO:ccdcalib.py:900:compute_nightly_bias: Nightly bias b4: maxabsdiff 1.41, stddev 0.39
INFO:ccdcalib.py:901:compute_nightly_bias: Default bias b4: maxabsdiff 0.43, stddev 0.13
WARNING:ccdcalib.py:907:compute_nightly_bias: Nightly bias worse than default; leaving /global/cfs/cdirs/desi/spectro/redux/matterhorn/calibnight/20220605/biasnighttest-b4-20220605.fits.gz for inspection
```

Looking at nightwatch, I don't see anything obvious besides an oscillating signal in one of the amps. Will run preproc on the images to check.

jbanovetz:
It looks like something happened in the afternoon that might require some flagging

<img width="592" height="453" alt="Image" src="https://github.com/user-attachments/assets/038a9cee-a363-43c8-8cb4-a1b83ba4a3bd" />

jbanovetz:
Made a override file for this date and added it to the commit mentioned here: https://github.com/desihub/desispec/issues/2777

Issue #2790, opened 2026-08-12
20220531 b4 biasnightest


jbanovetz:
In creating the nightly bias on 20220531, b4 ran into an issue and only created a `biasnighttest` file. This is documenting the investigation into why.

jbanovetz:
Looking at the log, I found this:

```
INFO:ccdcalib.py:900:compute_nightly_bias: Nightly bias b4: maxabsdiff 1.35, stddev 0.35
INFO:ccdcalib.py:901:compute_nightly_bias: Default bias b4: maxabsdiff 0.19, stddev 0.06
WARNING:ccdcalib.py:907:compute_nightly_bias: Nightly bias worse than default; leaving /global/cfs/cdirs/desi/spectro/redux/matterhorn/calibnight/20220531/biasnighttest-b4-20220531.fits.gz for inspection
```

I don't see anything obvious in the nightwatch images, I will have to preproc the individual biases in order to figure out if there are some to flag.

jbanovetz:
It looks like something happened in the afternoon that maybe requires some flagging

<img width="592" height="453" alt="Image" src="https://github.com/user-attachments/assets/949d8f9e-4cfb-4128-9e4d-257f5bda9a40" />

jbanovetz:
Made an override file for this date and camera and was a part of the svn commit mentioned here: https://github.com/desihub/desispec/issues/2777

Issue #2735, opened 2026-04-28
biasnight-r8-20211109 failing (and z6 too)


sbailey:
From matterhorn/run/scripts/night/20211109/biasnight-20211109-00108054-a0123456789-52191355.log
```
INFO:ccdcalib.py:423:compute_bias_file: reading /dvs_ro/cfs/cdirs/desi/spectro/data/20211109/00108083/desi-00108083.fits.fz r8
ERROR:preproc.py:184:calc_overscan: error in sigma clipping for overscan measurement, return result without clipping
ERROR:preproc.py:184:calc_overscan: error in sigma clipping for overscan measurement, return result without clipping
ERROR:preproc.py:184:calc_overscan: error in sigma clipping for overscan measurement, return result without clipping
ERROR:preproc.py:184:calc_overscan: error in sigma clipping for overscan measurement, return result without clipping
...
INFO:ccdcalib.py:900:compute_nightly_bias: Nightly bias r8: maxabsdiff nan, stddev nan
INFO:ccdcalib.py:901:compute_nightly_bias: Default bias r8: maxabsdiff 0.28, stddev 0.10
WARNING:ccdcalib.py:907:compute_nightly_bias: Nightly bias worse than default; leaving /global/cfs/cdirs/desi/spectro/redux/matterhorn/calibnight/20211109/biasnighttest-r8-20211109.fits.gz for inspection
```
(and similar messages for z6)

These CCDs were going back and forth between 2-amp and 4-amp mode (see #2733) and the data might be genuinely bad, but it is somewhat suspicious that they did succeed on 20211108 and 20211115 despite 2/4-amp testing then too.  Check if this is recoverable.

The same thing happened in Kibo, so for now I am letting Matterhorn proceed for 20211109 but flagging for QA.

jbanovetz:
After some investigation, I am thinking it might be a VCCD issue. It looks like both r8 and z6 were reset with VCCDSEC times of ~1200 seconds. I am not sure how that related to the overscan regions and will test that as the line:

`INFO:ccdcalib.py:900:compute_nightly_bias: Nightly bias r8: maxabsdiff nan, stddev nan`

is a little concerning and unlike other `biasnightest` cases I see.

jbanovetz:
I am seeing that for r8 more than 50% of the detector has nans for the `biasnightest` shown in red in the figure below:

[...]

PR #2801, opened 2026-08-20
Override On a Per Camera Basis - Biases

jbanovetz:
There are a handful of cases in Matterhorn where something is happening to the biases in the afternoon and they behave differently than in the morning or that flagging the bad biases will result in there being less than 18 viable ones for a detector. See these as examples:

https://github.com/desihub/desispec/issues/2735
https://github.com/desihub/desispec/issues/2790
https://github.com/desihub/desispec/issues/2792

Talking with Stephen and Julien, we want to be able to flag individual cameras for this case instead of flagging all 30 cameras for a night, which is currently what happens. 

Here I found a way to do this by changing a handful of files. Luckily, it looks like there was already some infrastructure on the processing side to take individual cameras from an `override.yaml` file using the keyword `camword`. The first change I made was in the `create_override` script to include this keyword if given a new argument `-c`. I also changed the corresponding `bin` wrapper for it.

The second was changes to `workflow/submission.py`. I noticed that while on the processing side it accepts the keyword, there is still an assumption that the override covers **all** cameras. Thus a `linkcal` job would get made but just for the flagged cameras and `biasnight` would never run for the remaining, unflagged cameras. To fix this, I included some logic to check if there is a `camword` keyword in the override file and if not, treat it like before. If there is, run `linkcal` on the flagged cameras and then run the usual `desi_compute_nightly_bias` on all the other cameras. It will also split biases and darks if this is the case as from what I can tell, the darks are generally fine and/or we can flag individual ones.

I have ran this with 3 separate situations and they were all able to get through the bias stage with no issues:
1. A night where two cameras need to have an override (20211109)
2. A night where one camera needs to have an override (20220531)
3. A night where no cameras need to have an override (20220601)

For these tests, I just ran the bias step: e.g. `desi_proc_night --proc-obstypes=zero -q regular -n 20220601 --dont-require-cal`. I plan on testing this through all calibration steps just to make sure nothing breaks down the line. I could also run on a different night with an `override.yaml` file to make sure that it still runs.

Some things I think could be improvements but outside the scope/might not be ready in time for Nevis:
1. `desi_compute_nightly_bias` doesn't know about the `override.yaml` file and will break on the test nights if ran outside of processing. I could try and fix that too or at least a log message just in case.
2. The new keyword for `create_override` is not something that can be changed with the questions like the others. I could try and include a camera questionnaire as a part of this function.
3. This only works for **biases/zeros**! I haven't touched darks since these are created from either `darknight` (combining multiple nights) or `DESI_SPECTRO_DARK` (from premade darks). I also haven't worried about arcs and flats as I figure those should still be all cameras most likely as I think that all exposures are flagged. I could try and split them up in the event in the future that there is one camera that needs a `linkcal` to a `biasnight` and arcs need a `linkcal` that involves a different set/all cameras.

### DIFF
--- a/py/desispec/scripts/createoverride.py
+++ b/py/desispec/scripts/createoverride.py
@@ -27,7 +27,7 @@ def valid_night(night):
     isvalid = night.isnumeric() \
         and np.abs(int(night[:4]) - 2023) < 10 \
         and np.abs(int(night[4:6]) - 6.5) < 6 \
-        and np.abs(int(night[6:]) - 16) < 15
+        and np.abs(int(night[6:]) - 16) < 16
     if not isvalid:
         print(f"--> Received invalid response: '{night}'. Must be YEARMMDD.")
     return isvalid
--- a/py/desispec/scripts/createoverride.py
+++ b/py/desispec/scripts/createoverride.py
@@ -48,7 +48,8 @@ def valid_cameras(cameras):
     except:
         isvalid = False
     if not isvalid:
-        print(f"--> Received invalid response: '{cameras}'. Must be in DESI format (i.e. a1, b2, r3, z4, r8z6, etc.).")
+        print(f"--> Received invalid response: '{cameras}'. Must be a list "
+              + "of cameras or camwords, i.e. a1, b2, r3, z4, r8z6, etc..")
     return(isvalid)
 
 def valid_yes_no(string):
@@ -101,7 +102,9 @@ def get_camera(prompt):
     Prompt the user for a camera string and return the camera string as a string
     """
     cameras = get_response(prompt, valid_cameras)
-    return(cameras)
+    if cameras == '':
+        cameras = 'a0123456789'
+    return parse_cameras(cameras, loglevel='ERROR')
 
 def is_yes(prompt):
     """
@@ -129,14 +132,14 @@ def create_override_file(args):
     if linkcal:
         refnight = get_night("What is the reference night? ")
         good_bias = is_yes("Are there valid zeros for biases? ")
-        good_cte = is_yes("Are there valid falts for cte corections? ")
+        good_cte = is_yes("Are there valid flats for cte corections? ")
         good_badcol = is_yes("Is there a valid dark for badcolumn detection? ")
         good_psf = is_yes("Are there valid arcs for psf generation? ")
         good_flats = is_yes("Are there valid flats for "
                             + "fiberflatnight generation? ")
         if not good_bias:
             biaslink_camword = get_camera("What cameras need a biasnight link? " +
-                                          "If all, input 'a0123456789'")
+                                          "Leave blank if all. ")
             if biaslink_camword == 'a0123456789':
                 print("Since all cameras need a biasnight link, " +
                       "'biaslink_camword' will NOT be set")
--- a/py/desispec/scripts/createoverride.py
+++ b/py/desispec/scripts/createoverride.py
@@ -9,7 +9,8 @@ import time
 import numpy as np
 
 from desispec.io.meta import findfile
-from desispec.io.util import decode_camword, create_camword
+# from desispec.io.util import decode_camword, create_camword
+from desispec.io.util import parse_cameras
 
 def valid_night(night):
     """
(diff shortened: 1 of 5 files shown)


akremin:
Some thoughts and comments:

1. There are only two calibration files that we can feasibly do per-camera specification for: biasnight and badcol. For now we don't need the ability to specify badcol, so we can leave it at just biasnight, as was done here. However, we want to be explicit in our override file variable name since we could in theory be e.g linking a subset of bias cameras but all arcs and flats. I suggest `biaslink_camword` rather than just `camword`. And I would have it as a question to the user rather than a command line argument to `desi_create_override_file`. Let's maintain the `camword` argument in its old capacity as a global camword for all steps. That doesn't need its own command line argument, however, let's just leave it as an undocumented feature that currently remains unused.
2. We need to update lincal to support the specified linking for the biases but not for the other linked cals.
3. As copilot also identified above, I think the cleanest way to handle such nights is to have the code submit separate bias and pdark jobs since they'll have different camwords. That is more straightforward than adding the complicated logic to propagate multiple camwords to the batch writer and submission functions. Assuming this only happens for a handful of nights I think the simplicity warrants the few additional jobs that we would have to submit.

akremin:
I think you could also leverage `parse_cameras` rather than making the new function `valid_cameras`. Either by putting it in a try-except to catch a raised error or having the error be what the user sees. I think your current `valid_cameras` would also just raise an error if the decoder or encoder failed, so I think `parse_cameras` is what you want.

akremin:
I have made the additional updates to propagate the relevant information to the script generator and to the script generator itself. If multiple linkings are requested and biasnight has a different specified camword then the job is split into two linking commands that are executed in serial inside the one linkcal script. If they share the same camword or there is only biases requested for linking, then only one command is written in the script.

I have run 4 tests that can be found in `/global/cfs/cdirs/desi/spectro/redux/override_percam`

1. 20250102: A regular night with no override file. This worked as expected.
2. 20250106: Link biasnight, psfnight, fiberflatnight for all cameras. This worked as expected.
3. 20250116: Link just biasnight for a subset of cameras. This new functionality worked as desired.
4. 20250126: Link all cameras for psfnight and fiberflatnight, but only a subset of cameras for biasnight. This new functionality worked as desired.

All of the tests passed and I believe it sufficiently tests all corner cases that could arise in the pipeline. Assuming copilot doesn't identify any clear bugs I think we are ready to merge this.

akremin (approved):
John's updates looked good except for one minor bug that I fixed while adding some functionality on top of it.

py/desispec/workflow/submission.py
jbanovetz, line 321: Sorry if I missed something above, do we need `, biascmd=None` added to the end of this?

py/desispec/scripts/createoverride.py
```diff
@@ -31,6 +32,26 @@ def valid_night(night):
...
+        camera(s), str, follows DESI camera input (a,b,r,z)(0-9)
+
+    Returns:
+        bool, True if valid camera sequence
+    """
+    try:
+        cams = parse_cameras(cameras)
+        isvalid =  isinstance(cams, str)
+    except:
+        isvalid = False
```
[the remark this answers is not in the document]
jbanovetz: I am good with this change, I checked and 'x1' does return 'ax1' so we probably want to avoid that for any new user. Thoughts @akremin ?

jbanovetz:
One other thing we can maybe change is this line and 15->16:

https://github.com/desihub/desispec/blob/02bc7e6634ec6cf1fc04725a93594bb879f4cdd7/py/desispec/scripts/createoverride.py#L26-L29

I am trying to create an override file for 20240131 and it is not allowing me to do that.

py/desispec/scripts/createoverride.py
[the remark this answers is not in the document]
akremin: This should really be dealt with in parse_cameras rather than here and that's outside the scope of this PR. I'm not worried about this. Let's leave it resolved for now.

akremin:
Good point, that's an embarrassing typo. I'll make that change.

py/desispec/workflow/submission.py
jbanovetz: Or sorry, do we need to add `biascmd=biaslinkcamword` at the end. `biaslinkcamword` is initialized as `None` so I think it's fine to put in.

akremin: No worries! That function doesn't have biascmd as an argument and recreates the biaslink_camword from the file once again. There is some duplicate code and I/O inefficiency involved with that, but the file was already being loaded for other reasons within the function, so it was easier to re-derive the camword for independence-sake.
We actually never pass the camword anywhere, the `biascmd` that is an argument a few places is the entire executable string to run the bias linking.

Issue #2803, opened 2026-08-25
Add missing LSS columns to zcat v2


sbailey:
Most of what the LSS catalogs need are already in zcat v2, except for 3 columns:
  * `ZWARN` from the zmtl files (has additional QA bits set compared to `ZWARN` from the redrock files)
  * `OIII_FLUX` and `OIII_FLUX_IVAR` from emlinefit files (in addition to OII columns already propagated)

I propose that we add zmtl `ZWARN` as `ZWARN_QA` (or `ZWARN_MTL`?) into the zcat -extra.fits file, but *not* change `ZWARN_BEST` in the primary zcat file.  That lets LSS use it, but doesn't suddenly force a change on other people's code who may have been using cut of ZWARN_BEST=0 or ZWARN_BEST=0 or 4.

healpix/uniqpix-based redshifts do not have zmtl files, so catalogs based on those would not have a `ZWARN_QA` column.

`OIII_FLUX` and `OIII_FLUX_IVAR` should be straighforward since zcat v2 already opens the emline files which exist for both tile-based and pix-based redshifts.

For the record: I considered suggesting adding a `GOOD_SPEC_LSS` column with the more restrictive LSS hardware cuts as a compliment to the `GOOD_SPEC` column. But that smells like an opportunity to get out of sync with the actual final LSS cuts and is perhaps better done as documentation, i.e. defining how a theoretical `GOOD_SPEC_LSS = GOOD_SPEC & (other cuts based on zcat v2 columns...)`

@rongpu could you add these?

rongpu:
The [`BAD_SPECQA` and `BAD_PETALQA` bits](https://desidatamodel.readthedocs.io/en/latest/bitmasks.html#zwarn) in zmtl ZWARN are compressions of `QAFIBERSTATUS` in the `tile-qa-*.fits` (in the `FIBERQA` HDU). See the definition below.

In addition to `ZWARN_MTL`, should we also add `QAFIBERSTATUS` to the zcat `-extra.fits`, so that we keep the provenance of the `BAD_*QA` bits and make it possible/easy to change the `BAD_*QA` definition in the future? @sbailey

The `QAFIBERSTATUS` -> `BAD_*QA` compression (summarized by Codex):
```
  BAD_SPECQA combines any of:
  - UNASSIGNED
  - BROKENFIBER
  - MISSINGPOSITION
  - BADPOSITION
  - POORPOSITION
  - BADFIBER
  - BADTRACE
  - BADFLAT
[...]

PR #2806, opened 2026-08-28
Add OIII and QAFIBERSTATUS columns

rongpu:
This PR addresses https://github.com/desihub/desispec/issues/2803. It adds QAFIBERSTATUS (ztile only), OIII_FLUX, and OIII_FLUX_IVAR columns to the v2 zcat extra file.

I tested the new script by running it on matterhorn SV1 data, and it ran successfully.


sbailey (approved):
Looks good; thanks for checking for special cases of tile-qa contents being out of sync with redrock files.

PR #2809, opened 2026-09-03
update determine_science_to_proc

sbailey:
## Primary purpose of this PR ##

This PR updates `determine_science_to_proc` with several end-user simplifications for addressing the case of "which tiles will be processed by this production?":
  * if `etable` is None, load and stack all exposure tables in the production
    * enabled by new function `read_all_exptables` which parallelizes the loading of all exposure tables
  * every option now has a default, instead of having to pass multiple `None` values just to get default behavior
  * adds `startnight` option to remove exposures before a given date (default 20201214, start of SV).

This snippet now predicts what tiles will be included in a standard production, e.g. Nevis:
```
sciexp, tiles = determine_science_to_proc(laststeps='all', all_tiles=False, complete_tiles_thrunight=20260414)
```
The `laststeps` and `all_tiles` options are needed to override the defaults for daily operations, and the `complete_tiles_thrunight` provides an optional production cutoff.

Caveat: current surveyops/ops/tiles-specstatus.ecsv includes 22 backup tiles that we continued observing after Matterhorn. These were included in Matterhorn but now would fail the complete_tiles_thrunight cut.

## Other changes ##

Adding parallelism for `read_all_exptables` highlighted that `default_nproc` wasn't considering NERSC login node usage, and several places were allowing up to 128 cores on a login node.  This PR updates those to standardize the calculation and throttle to 8 processes when running on a login node.

Note: this does do minor calculations upon import of desispec.parallel. Normally that is discouraged, but I think this case is acceptable to maintain the current usage of `default_nproc` as a variable instead of turning that into a function (which might break other external code).

Text matrix was updated to include python 3.14 and astropy 8 in prep for Nevis.

### DIFF
--- a/py/desispec/workflow/exptable.py
+++ b/py/desispec/workflow/exptable.py
@@ -5,6 +5,7 @@ desispec.workflow.exptable
 """
 import os
 import glob
+import multiprocessing as mp
 import numpy as np
 from astropy.table import Table, vstack
 from astropy.io import fits
@@ -17,6 +18,7 @@ from desiutil.log import get_logger
 from desispec.util import header2night
 from desispec.io.util import create_camword, parse_badamps, checkgzip
 from desispec.io.meta import findfile
+from desispec.parallel import default_nproc
 
 #############################################
 ##### Exposure Table Column Definitions #####
@@ -1027,6 +1029,49 @@ def read_minimal_science_exptab_cols(nights=None, tileids=None,
         outtable = outtable[np.isin(outtable['TILEID'], tileids)]
 
     return outtable
+
+
+def _read_one_exptable(etab_file):
+    """
+    Helper function for read_all_exptables() that reads a single exposure
+    table file. Defined at module level so that it can be pickled for use
+    with multiprocessing.
+    """
+    from desispec.workflow.tableio import load_table
+    return load_table(tablename=etab_file, tabletype='exptable', suppress_logging=True)
+
+
+def read_all_exptables(nproc=default_nproc, specprod=None):
+    """
+    Reads and stacks all exposure tables for a production.
+
+    Args:
+        nproc (int, optional): Number of multiprocessing processes to use
+            when reading the exposure table files. Default is
+            desispec.parallel.default_nproc.
+        specprod (str, optional): Production name, or full path to a
+            production, passed to findfile(). Default is None, i.e. use
+            $DESI_SPECTRO_REDUX/$SPECPROD.
+
+    Returns:
+        astropy.table.Table: The exposure table entries from every exposure
+        table file in the production, stacked into a single table.
+    """
+    log = get_logger()
+    etab_path = findfile('exptable', night='99999999', readonly=True, specprod=specprod)
+    glob_path = etab_path.replace('99999999', '202?????').replace('999999', '202???')
+    etab_files = sorted(glob.glob(glob_path))
+    log.info(f"Found {len(etab_files)} exposure tables to read")
+
+    if nproc is not None and nproc > 1 and len(etab_files) > 1:
+        with mp.Pool(nproc) as pool:
+            etables = pool.map(_read_one_exptable, etab_files)
+    else:
+        etables = [_read_one_exptable(etab_file) for etab_file in etab_files]
+
+    etables = [t for t in etables if t is not None and len(t) > 0]
+
+    return vstack(etables)
 
 
 def _select_sciences_from_etab(etab):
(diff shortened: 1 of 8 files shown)


py/desispec/workflow/exptable.py
```diff
@@ -1029,6 +1031,49 @@ def read_minimal_science_exptab_cols(nights=None, tileids=None,
...
+
+    Returns:
+        astropy.table.Table: The exposure table entries from every exposure
+        table file in the production, stacked into a single table.
+    """
+    log = get_logger()
+    etab_path = findfile('exptable', night='99999999', readonly=True, specprod=specprod)
+    glob_path = etab_path.replace('99999999', '202?????').replace('999999', '202???')
+    etab_files = sorted(glob.glob(glob_path))
+    log.info(f"Found {len(etab_files)} exposure tables to read")
```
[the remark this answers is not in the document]
sbailey: Good point; updated, albeit with a slightly different solution.

sbailey:
Note: tests are failing for python 3.14 for unrelated multi-processing reasons. I will address that in a separate PR instead of mixing up a py3.14 fix with this PR which is already mixing topics.

py/desispec/workflow/exptable.py
```diff
@@ -1029,6 +1031,54 @@ def read_minimal_science_exptab_cols(nights=None, tileids=None,
     return outtable
 
 
+def _read_one_exptable(etab_file):
+    """
+    Helper function for read_all_exptables() that reads a single exposure
+    table file. Defined at module level so that it can be pickled for use
+    with multiprocessing.
+    """
+    from desispec.workflow.tableio import load_table
```
akremin, line 1044: Is there a circular import issue with having this imported globally? I think python caches imported functions, but that would be once we worker for your parallel load case.

py/desispec/workflow/science_selection.py
```diff
@@ -70,6 +77,11 @@ def determine_science_to_proc(etable, tiles, surveys, laststeps,
             if None or all_tiles is True.
         specstatus_path (str, optional): Location of the surveyops specstatus
             table.Default is $DESI_SURVEYOPS/ops/tiles-specstatus.ecsv.
+        startnight (int, optional): Default is 20201214. Exposure table
```
akremin, line 80: I'm okay with hard coding this, but you have a check for `startnight is None` below so just verifying you want it to be hardcoded to the start of SV? The `None` is in case someone wants to opt out of it explicitly?

py/desispec/workflow/exptable.py
```diff
@@ -1029,6 +1031,54 @@ def read_minimal_science_exptab_cols(nights=None, tileids=None,
...
+    return load_table(tablename=etab_file, tabletype='exptable', suppress_logging=True)
+
+
+def read_all_exptables(nproc=default_nproc, specprod=None):
```
akremin, line 1048: This is unfortunately close to `desispec.workflow.exptable.read_minimal_science_exptab_cols` but does have the key distinctions of not sub-selecting LASTSTEP or down-selecting columns, and loads in parallel which is nice. This is mostly just a note to your future selves, the referenced function wouldn't work for your needs without modification that would then require changes elsewhere in the pipeline where we already call `read_minimal_science_exptab_cols`.

akremin (approved):
I made three comments but none of them are blocking factors. Two are genuine questions, but assuming they were intensional choices then I am happy to resolve them unchanged. The third is more of a comment that a change request and has no action associated with it.

py/desispec/workflow/exptable.py
sbailey: Good question. I double checked this, and because the import is inside the function rather than at the module level it is ok. Circular imports are only a problem if both modules import each other, but if they contain functions that internally load the other module, that is fine.

That being said, this is an indication that the import relationship between tableio.py, exptable.py, and proctable.py isn't great, but that goes deeper than what is added by this PR.  Let's defer that cleanup since it might have knock-on effects.

py/desispec/workflow/science_selection.py
sbailey: Yes, the None check is for safety and a mechanism for not filtering at all (0 would also work for that). I'll update the docstring.

py/desispec/workflow/exptable.py
sbailey: Yes, I first tried read_minimal_science_exptab_cols but it didn't meet the "ready everything" needs.  I'll add comments to the docstrings that reference each other to help mitigate this similarity confusion.

Issue #2813, opened 2026-09-09
Table strings whitespace change in astropy>8


sbailey:
Claude identified an issue for an upcoming version of astropy (still ok for current astropy 8) — numpy is deprecating `numpy.char.chararray` so astropy Tables will move from using `numpy.char.chararray` to `numpy.ndarray` for string columns.  The problem is that `chararray` supported auto-stripping trailing whitespace, while `ndarray` does not.  This risks having columns like SPECTYPE suddenly have extra padding (e.g. `"GALAXY"` and `"QSO   " instead of "QSO"`). I'm not entirely sure that this will be an issue in practice, but at minimum we should watch out for it and explicitly check when moving to astropy/9.x.

## References ##

- astropy PR [#19267](https://github.com/astropy/astropy/pull/19267) "Handle chararray deprecation" - the implementation, and where the **v9.0 target (spring 2027)** is stated:
  "The plan is to remove the compatibility code and use of chararray for v9.0 (don't think we can
  wait longer since at some point Numpy will remove chararray)." Note the published astropy docs
  only say "in a future version"; v9.0 is from the PR review discussion.
- astropy issue [#3862](https://github.com/astropy/astropy/issues/3862) "Deprecate use of chararray in io.fits" - the original issue, which is explicit that the behavior at stake is chararray's "ability to automatically `rstrip()` individual strings returned from the array".
- astropy issue [#19216](https://github.com/astropy/astropy/issues/19216) - the numpy chararray deprecation surfacing as an astropy bug.
- Upstream driver: [numpy 2.5.0 release notes](https://numpy.org/devdocs/release/2.5.0-notes.html) - `numpy.char.chararray` and `numpy.char.[as]array` deprecated, use plain arrays plus the `np.strings` namespace; the `numpy.chararray` re-export was removed in 2.5.

PR #2815, opened 2026-09-09
Add tests for python/3.14 and astropy/8.x

sbailey:
Adds python/3.14 and astropy/8.x to the github test matrix.  Adds new tests that cover astropy usage patterns in case those change in the future (e.g. #2813) so that we find out incompatibilities before running the full pipeline. No code changes, just updated tests.


sbailey:
Tests are intermittently getting stuck on py3.10 in test_bitmask (both in this PR and #2815).  Seems unrelated to current updates. I'm going to merge as-is and deal with that later.

PR #2822, opened 2026-09-15
splrep takes 1/sigma as input

segasai:
the input splrep is incorrect, it should be 1/sigma, not 1/sigma^2


sbailey:
Double checked and confirmed that splrep weights are sqrt(ivar) not ivar itself.  Oops.

PR #2823, opened 2026-09-15
incorrect variance in stdstars

segasai:
when computing the combined flux as 
$sum(w_i I_i f_i)/sum(w_i I_i)$
the variance is $sum(w_i^2 I_i^2 *(1/I_i) )/(sum(w_i I_i))^2 = sum(w_i^2 I_i )/(sum(w_i I_i))^2$, not
$1/(sum(w_i I_i))$


julienguy:
good catch.

PR #2824, opened 2026-09-15
fix incorrect equation for sigma_coadd_norm

segasai:
The equation for sigma_coadd_norm is not correct it seems. 

if chi2 =  (w (f_c - b * f_i)^2) as defined in the code
$$b=sum(w*f_c*f_i)/sum(w*f_i^2)$$
$$Var[b] = sum(w^2 f_c^2 Var_i)/sum(w*f_i^2)^2$$
The code give eqn for $a=1/b$ , then $$Var[a]=Var[b]/b^4$$ which will give the equation in the patch 
(the derivation also assumes w_tot>> w_i)


julienguy:
another good catch

PR #2826, opened 2026-09-17
correctly compute the profile by summing along one of the axes

segasai:
From the same audit as other PRs. 

It seems compute_x_offset_from_central_band_cross_dispersion_profile incorrectly computes the fiber profile by doing the sum over 2d array rather than one dimension.


julienguy:
Wow. This is a bug fix in a routine I wrote at some moment to try and detect large shifts in fiber flat calibration images. The algorithm could not have worked properly with this bug. I checked this routine has not been used in production (neither for the afternoon calibrations or science observations at night).

julienguy:
I checked that the new version works. It turns out the version before the bug fix also worked by pure chance, because of the  term 1/ivar was resulting in a profile good enough for cross-correlation (!).
I am merging this bug fix.

PR #2828, opened 2026-09-18
Only use good fits for XTRACE and YTRACE means

### LINKED ISSUE #2819: `XTRACE` and `YTRACE` include average over bad exposures
`mean_psf` averages the `XTRACE` and `YTRACE` HDUs over every input PSF, with none of the bundle selection that is applied to the PSF coefficients. Found while reviewing #2732; **not** a regression from that PR, and deliberately left out of it.
     
The PSF coefficients are merged per bundle, using `rchi2` as a selection score and skipping inputs where the bundle is missing or failed:
     
     ```python
     for bundle in fibers_in_bundle.keys() :
         ok=np.where((bundle_rchi2[:,bundle]>0)&(bundle_rchi2[:,bundle]<rchi2_threshold))[0]
         ...
         if ok.size>=2 : # use mean
             for f in fibers_in_bundle[bundle]  :
                 output_coeff[f]=np.mean(coeff[ok,f],axis=0)
     ```
     
The traces are then averaged with a flat `np.mean` over all inputs, with no `ok`, no `bundle_present`, and no exclusion of bundles dropped because they were missing from every input:
     
     ```python
     if len(xtrace)>0 :
         xtrace=np.array(xtrace)
         ytrace=np.array(ytrace)
         ...
         hdulist["xtrace"].data = np.mean(xtrace,axis=0)
         hdulist["ytrace"].data = np.mean(ytrace,axis=0)
     ```
     
## Consequence
     
For a bundle masked out by a bad amp in some of the input arcs, the output PSF coefficients correctly come only from the exposures that have the bundle, but the output trace is pulled towards the exposures that do not. Those inputs carry whatever the reference PSF had for the masked fibers rather than a fit, so the averaged trace is a blend of fitted and unfitted values. The same applies to a bundle that is missing from *all* inputs and is therefore dropped from the merge entirely (`Bundle N is missing in all input PSFs ... will be ignored in the merging`) — its coefficients keep the reference values, but its trace is still averaged.
     
This is the case #2732 is about, so it is most visible on nights like 20211028 (r8A flagged in some arcs) and 20221121 (b8B missing in all arcs).
     
     ## Notes
     
     - Whether this matters in practice depends on how much the unfitted reference trace differs from the fitted ones; it may well be small. Worth quantifying before deciding how to fix it.
     - The current behavior is pinned by `test_traces_are_averaged_over_all_inputs` in `py/desispec/test/test_specex.py` (added in #2732), so a change here will fail that test on purpose. Update the test alongside any fix.
     - A fix would need to decide what a trace should be for a bundle that no input successfully fit — probably keep the reference trace, matching what the coefficients already do.
     
     Raised during the review of #2732.

akremin:
Fixes #2819.

## Problem

`mean_psf` applies two different rules to the two halves of the file it writes.

The **PSF table coefficients** are merged per fiber bundle: inputs are kept only
where `0 < rchi2 < threshold`, bundles missing from every input are dropped from
`fibers_in_bundle` and keep the reference PSF's values, and there is a
smallest-non-zero-rchi2 fallback when nothing passes.

The **`XTRACE`/`YTRACE` HDUs** were merged with a bare `np.mean(xtrace, axis=0)`
over every input, with no bundle logic at all.

So a bundle's shape and its position could come from different sets of exposures.
Where a bundle was not fit, the traces averaged in values that were never
measured from that arc: `desi_psf_fit` is never invoked for a bundle skipped on a
bad amp (`specex.py:220-222`), and `merge_psf` starts from the input PSF
(`specex.py:473`), resets STATUS to -1 (`:483`), and copies XTRACE/YTRACE only
for fibers with STATUS==0 (`:496`, `:515-518`). Any bundle with no successfully
fit fiber therefore still holds the input PSF's traces verbatim, whether it was
masked out or attempted and failed.

## Changes

* `_select_bundle_inputs()` now holds the per-bundle input selection. It is
  called once per bundle instead of once per bundle per PARAM row, since the
  selection never depended on the row being averaged; previously it was
  recomputed ~50 times per camera, including its warnings.
* `XTRACE`/`YTRACE` use that selection. A bundle with nothing to average keeps
  the reference PSF's traces, matching what the coefficients already did.
* Only inputs that actually fit a bundle can contribute to its traces. Presence
  is not enough: a bundle that failed outright has STATUS>0 rather than <0, so
  it is "present" while still carrying the input PSF's traces.
* `xtrace`/`ytrace` are kept aligned with the input list, with a `None`
  placeholder for a missing HDU, so an input lacking trace HDUs is dropped from
  each bundle's selection rather than disabling selection for the whole camera.
* Fixed a latent bug in the trace Legendre refit, which iterated
  `range(icoeff.shape[0])` using a variable left over from the coefficient loop
  and never assigned in that block. The same block also relied on the
  coefficient loop having executed its function-local `legval`/`legfit` import,
  which only worked because both branches share the `wavemins[p] != WAVEMIN`
  condition; it now imports them itself.

Nominal nights are unaffected: when every input passes the cut, the selection
keeps them all and the result is the previous plain mean.

The non-fallback path needs no separate "was it fitted" test, which is worth
stating because it is not obvious: `merge_psf` zeroes every `B{bb}RCHI2` up
front (`:476-480`) and copies one back only for a bundle with a STATUS==0 fiber,
so `rchi2 > 0` already implies a fitted trace exists.

## Why the rchi2 cut applies to the traces, not just the unfitted case

Issue #2819 proposes the narrower fix, retaining the reference trace where
nothing was fit. This PR reuses the whole selection, rchi2 cut included, because
specex fits the traces as free parameters of the same least-squares that produces
the rchi2:

* `specex/src/specex_psf_fitter.cc:2554-2561` runs a stage with
  `fit_trace=true` ("FitSeveralSpots TRACE"), and `:2562-2578` a "FLUX+TRACE"
  loop that iterates until the traces move less than 0.5 pix.
* `FitEverything` then recomputes the weight image and sets
  `psf_params->chi2 = ParallelizedComputeChi2AB(false)` over the whole bundle
  model with those traces in place (`:2845-2857`), which
  `specex/py/specex/io.py:204` writes as `B{bb}RCHI2`.

So `rchi2` is a joint trace-plus-shape goodness of fit, and an input rejected for
a bundle's coefficients is also evidence against that input's trace for it.

Downstream correction is low order and cannot repair an arbitrary per-bundle
error, though it is not merely a constant: the science, flat and sky path runs
`desi_compute_trace_shifts` with `--degxx 2 --degxy 0` and `--degyx 2 --degyy 0`
(`proc.py:556-601`). The degree-zero call is the arc pre-specex step that
produces the `shifted-input-psf-*` files (`proc.py:609-636`).

## One deliberate divergence from the coefficients

When no input passes the rchi2 cut, the traces do not follow the coefficients'
single-input fallback. They average the inputs that fit the bundle, compared
against a threshold built from that bundle alone.

`rchi2_threshold` is computed once per camera from the median of every positive
bundle rchi2 (`specex.py:702-708`), so a bundle can fall entirely above it
without any of its inputs being an outlier among the others. b8 bundle 10 of
20221121 is `4.95, 4.62, 4.81, 5.03, 4.61`: the coefficients must choose one
input and take the 4.61 arc, but preferring it for the traces over the 5.03 arc
would trade the averaging of five exposures for an 8% rchi2 difference that
carries no information.

Averaging is only defensible where the inputs really are comparable, and the
camera-wide threshold does not establish that. The fallback therefore applies the
same median+1 rule to the bundle's own values, so `4.6, 5.0, 100` drops the 100
instead of blending it in at equal weight. This is an empirical policy with an
explicit criterion, not a property of the camera-wide threshold. Its weakness is
that `+1` is calibrated for rchi2 near 1, so it is a wide tolerance at 1.25 and a
tight one at 5; a multiplicative form is closer in spirit but needs an arbitrary
constant.

The two cases differ physically, which is what motivates treating them
differently:

| | what the traces contain | information |
|---|---|---|
| masked, or failed outright | the input PSF's traces, inherited verbatim | none of its own |
| fitted but above the cut | genuine per-bundle fit output | real, just noisier |

## Production impact

Measured by running the new `mean_psf` over real arc PSFs and comparing the
traces against what the old code would have written, in CCD pixels.

On clean recent nights (daily, 20260805-09, 600 camera-bundles each), **5-14
bundles per night change**, roughly 1-2%, all through the partial-rchi2 route.
**0 partly-masked bundles**, so the case the issue is named for does not arise
on those nights. Cameras with no affected bundle change by exactly zero.

Per-bundle change on 20260807, over the fiber and wavelength grid:

| bundle | median | p90 | max |
| --- | ---: | ---: | ---: |
| r7 b17 | 0.0017 | 0.0139 | 0.1780 |
| r7 b19 | 0.0018 | 0.0136 | 0.1177 |
| r6 b16 | 0.0008 | 0.0088 | 0.1048 |
| b2 b16 | 0.0014 | 0.0045 | 0.0554 |

The maxima are not the effect size. For r7 b17 the maximum is at fiber 449 at
7838 A, the bundle-edge fiber at the extreme red end of the range where Legendre
differences naturally diverge, and only 1.8% of the grid exceeds 0.05 pix. The
typical change is ~0.002 pix.

For scale, the unfitted-trace case is the larger one: a reference trace sits
about 0.03 pix (median) from a fitted one, roughly twenty times the
rchi2-filtering effect.

### Arc-residual validation: a null result

r7 bundle 17 on 20260807, rchi2 `[2.08, 3.06, 2.91, 2.63, 2.88]` against a 2.72
threshold. 69 arc lines over fibers 425-449, ~4500 line-fiber centroids across
the five preproc images, measured by iterating the flux-weighted centroid to
convergence so the result cannot depend on which trace seeded the window.

    rms residual, pixels          dx                      dy
      all exposures    old=0.0398 new=0.0403    old=0.0519 new=0.0525
      kept arcs        old=0.0416 new=0.0418    old=0.0547 new=0.0550
      dropped arcs     old=0.0385 new=0.0393    old=0.0499 new=0.0506

No detectable difference. The harness is sensitive enough for that to mean
something: injecting a known +0.20 pix into XTRACE moved rms dx from 0.040 to
0.1992, and +0.10 pix into YTRACE moved rms dy from 0.052 to 0.106. The
exposure's own fitted trace is also no better on its own exposure than a
five-arc mean is, because every trace variant agrees to far better than the
centroid measurement noise where arc lines actually fall.

**This is a consistency fix of small measured magnitude, not a demonstrated
improvement in extraction accuracy.** It removes a real inconsistency, and the
rchi2 filtering of traces is retained because the traces and coefficients then
describe the same exposures and the per-bundle fallback cut guards a
pathological case that typical-case measurement cannot rule out. Arc residuals
do not show a benefit, and no extraction comparison was attempted.

## Tests

`py/desispec/test/test_specex.py`, 22 tests, of which 8 fail against the base
implementation. The pre-existing coefficient tests pass untouched.

`py/desispec/test/test_specex.py`, 22 tests, of which 8 fail against the base
implementation. The pre-existing coefficient tests pass untouched.

* the four trace-selection cases: bundle missing in some inputs, missing in all,
  failed in one, and rejected by the rchi2 cut
* the fallback cases: averaging when nothing passes the cut, excluding an input
  that never fit the bundle, and dropping an outlier among the fitted inputs
* an input lacking trace HDUs, asserting it is dropped per bundle rather than
  disabling selection for the camera
* a nominal-night regression guard, asserting the traces are unchanged
* `test_inputs_on_another_wavelength_range_are_refit_before_averaging`, covering
  the Legendre refit. `compatible()` does not compare WAVEMIN/WAVEMAX, so that
  branch is live code that had no coverage, which is how the stale `icoeff`
  reference survived in it. It passes against base, so it documents existing
  behavior rather than demonstrating that defect. It uses non-constant traces on
  purpose, since a constant refits to itself and would pass even with the refit
  deleted, and checks the merged trace evaluated at a wavelength against the
  mean of the inputs evaluated at the same wavelength rather than
  reimplementing the refit's algebra. Verified to fail when the refit is
  disabled.

`test_traces_are_averaged_over_all_inputs`, which pinned the old behavior
deliberately, is replaced.

Separately, a differential check over 52 randomized cases covering masked,
failed, partially masked, high-chi-square and mixed-wavelength-range inputs
confirms the PSF coefficient arrays and output bundle rchi2 are **bit-identical**
to the base implementation. Only the traces change.

## For reviewers

* **The rchi2 cut is often marginal.** b2 bundle 16 of 20260807 is
  `2.27, 2.38, 2.45, 2.28, 2.29` against a 2.39 threshold: it drops 2.45 and
  keeps 2.38, a 3% difference. The "a relative cut does not always discriminate"
  argument behind the fallback divergence applies here too. This PR does not act
  on it, since the coefficients already use exactly this cut and the traces
  following it is at least self-consistent, but it is worth a second opinion.
* **Real fitted data is being dropped**, not only contamination removed. In the
  partial-rchi2 case the excluded arcs have STATUS=0 and genuine traces. The
  justification is that rchi2 says the model fits those exposures worse, which is
  a chi2 argument rather than a provenance one, and the arc residuals above do
  not show that the result is better.
* **No clean production example of the masked-bundle path** was found to test
  against. A 40-night scan of recent daily found masked bundles on only two
  nights, both with reprocessing provenance that could not be resolved, and the
  20211028/20221121 arcs in loa have no masked bundles at all (that masking needs
  the specex#91 changes, which postdate that production). The path is covered by
  unit tests and is unambiguous in `merge_psf`, but it is not exercised here on
  real data.
* **Untouched, pre-existing:** `merge_psf` keeps the reference trace for
  individually failed fibers (STATUS>0) while `mean_psf` selects only per bundle,
  so a bundle passing selection can still leave such a fiber on its reference
  trace; and if the *first* input lacks XTRACE/YTRACE, writing the output still
  raises `KeyError`, before and after this change.


akremin:
GPT-6 Astra review:

**Review of desispec PR #2828: Only use good fits for XTRACE and YTRACE means**

Reviewed on 2026-09-18. This assessment covers [PR #2828](https://github.com/desihub/desispec/pull/2828), addressing [issue #2819](https://github.com/desihub/desispec/issues/2819).

- Reviewed head: `8e6b18a079022269cdcb3d46813fcd6b56e814e5` (`tracemeanfix`).
- Compared against base: `3574633fef45232d79db2c81431cb79d65589aa2`.
- Scope: all three changed files, the surrounding PSF merge logic, relevant callers, and the trace-shift configuration.
- No implementation or test files were changed during the review. The additional checks described below were run separately.

**Assessment.** The PR fixes the main masked-bundle problem for normal specex inputs containing both trace HDUs. The coefficient refactor also appears to preserve existing behavior. However, two fallback paths still average unfitted reference traces into the result. I recommend addressing those paths before describing #2819 as fully resolved.

Both findings below reproduce behavior that existed before this PR. They are incomplete fixes, not newly introduced regressions. I found no coefficient regression or new crash in the paths tested.

1. **[P2] Exclude completely failed fits from the no-passing-input trace fallback.**

   Location: [specex.py, lines 785-791](https://github.com/desihub/desispec/blob/8e6b18a079022269cdcb3d46813fcd6b56e814e5/py/desispec/scripts/specex.py#L785-L791).

   When no input passes the chi-square cut, the trace fallback selects every input where the bundle is present. Presence only establishes that the bundle was not entirely masked out. An attempted but completely failed fit is also present, and its trace still contains reference values because `merge_psf` did not copy a fitted solution.

   I reproduced this using the test fixtures, with the other bundles at nominal `rchi2=1.25`, giving a selection threshold of 2.25:

   | Input | Result for the affected bundle | rchi2 | Synthetic trace coefficient value |
   | --- | --- | ---: | ---: |
   | A | Completely failed; STATUS=2 for every fiber | 0 | 10 |
   | B | Successfully fitted, above the cut | 5.0 | 20 |
   | C | Successfully fitted, above the cut | 3.7 | 60 |

   Both the base and PR produce a trace coefficient of **30**, averaging all three inputs. Under the PR's policy of averaging successfully fitted inputs when none passes the cut, the result should be **40**, excluding A. Following the coefficient fallback instead would select C and give 60. The current value includes an input known not to have a fitted trace.

   The coefficients come from C, and the output STATUS is zero. The output therefore does not retain a failure status that would expose this trace contamination.

   Suggested change: distinguish successfully fitted inputs from merely present inputs when constructing the trace fallback. Preserve averaging of high-chi-square fits if desired, but exclude outright failures and define a separate fallback when no successful fit exists. Add a regression test combining one complete failure with otherwise successful fits that all exceed the cut.

2. **[P2] Preserve bundle filtering when an input lacks a trace HDU.**

   Location: [specex.py, lines 877-884](https://github.com/desihub/desispec/blob/8e6b18a079022269cdcb3d46813fcd6b56e814e5/py/desispec/scripts/specex.py#L877-L884).

   If any input lacks either trace HDU, the code disables selection for the whole camera and averages all available paired traces. This retains the original issue even for bundles whose correct trace selection is fully known.

   The PR's own [missing-HDU test](https://github.com/desihub/desispec/blob/8e6b18a079022269cdcb3d46813fcd6b56e814e5/py/desispec/test/test_specex.py#L829-L853) demonstrates the problem: only input 0 fitted bundle 0; input 1 lacks trace HDUs; input 2 has an unfitted reference trace for that bundle. The test expects the mean of inputs 0 and 2, **35**, while the valid fitted trace and selected coefficients are **10**.

   The comment's explanation that selection cannot reliably identify exposures is no longer applicable. The new `None` placeholders already keep trace arrays aligned with the input tables and chi-square values.

   Suggested change: intersect each bundle's selected indices with the inputs that have usable trace HDUs. Handle an empty intersection explicitly, without disabling filtering for other bundles. Bundles missing from every input should still retain the reference trace. Update the test to check that the masked input is excluded.

   This is a compatibility-path concern: normal specex outputs contain both trace HDUs. It is nevertheless a supported path explicitly implemented and tested by the PR.

**What the implementation gets right.** For inputs containing both trace HDUs, the normal selection path correctly excludes bundles masked out of some exposures and retains the first input's reference traces for bundles missing from every exposure. It also excludes zero-chi-square failed fits and high-chi-square inputs when another input passes the cut. When every input passes, the result remains the ordinary mean.

Computing the selection once per bundle and reusing it for each parameter row is appropriate. The selection does not depend on the row being averaged. The changes also preserve the coefficient fallback, output chi-square values, and existing multiple-failure checks in the cases tested. Keeping the trace lists aligned with input indices is necessary and correctly implemented.

**The fallback's statistical explanation needs revision.** The threshold is calculated from the median of all positive bundle/input chi-square values for the camera, plus one. It is not a median calculated separately for each bundle. See [specex.py, lines 702-708](https://github.com/desihub/desispec/blob/8e6b18a079022269cdcb3d46813fcd6b56e814e5/py/desispec/scripts/specex.py#L702-L708).

Consequently, no input passing the cut does not imply that the inputs are similarly bad or that there is no outlier. I verified that a bundle with values `4.6, 5.0, 100`, against a threshold of 2.25, takes the averaging fallback and includes all three traces. The claim in the [fallback comment](https://github.com/desihub/desispec/blob/8e6b18a079022269cdcb3d46813fcd6b56e814e5/py/desispec/scripts/specex.py#L776-L784) is therefore not guaranteed by the code.

Preserving averaging may be a defensible compatibility choice. The near-equal values cited in the PR support considering that choice, but do not justify it for every case where no input passes. The explanation should distinguish an empirically motivated policy from a property of the threshold. If near-equality is essential to the policy, it needs an explicit criterion rather than an assumption.

Applying the existing coefficient cut to traces is reasonable because the fitted model includes both trace and shape parameters. However, measured trace movements demonstrate a change, not necessarily an improvement in extraction accuracy. Arc residuals or comparisons of extracted spectra would provide stronger validation of the broader high-chi-square filtering policy. I did not independently reproduce the production-night measurements in the PR.

**Two factual corrections to the PR writeup.**

- The command with `--degxx 0 --degxy 0 --degyx 0 --degyy 0` is in the arc preprocessing path, before specex fitting: [proc.py, lines 624-627](https://github.com/desihub/desispec/blob/8e6b18a079022269cdcb3d46813fcd6b56e814e5/py/desispec/scripts/proc.py#L624-L627). The science path uses `--degxx 2 --degxy 0` and `--degyx 2 --degyy 0`: [proc.py, lines 573-583](https://github.com/desihub/desispec/blob/8e6b18a079022269cdcb3d46813fcd6b56e814e5/py/desispec/scripts/proc.py#L573-L583). These smooth corrections still cannot generally repair arbitrary errors in individual bundles, but the claim that science processing applies only a constant displacement is inaccurate.
- Running the current 19 `TestMeanPSF` tests against the base implementation produced **six failures**, not seven: five trace-result assertions and the missing-HDU warning assertion. The wavelength-refit test passes against the base implementation. It is useful coverage of existing behavior, but does not demonstrate the claimed latent defect. Replacing the stale `icoeff` reference with the trace array's own shape improves clarity; a claim that this fixes a specific failure needs a test that triggers that failure.

**Existing limitations remain distinct from these findings.** The acknowledged per-fiber limitation remains: a bundle with an individually failed fiber can pass selection when its bundle chi-square is acceptable. Restricting the traces to that input can leave the failed fiber using only its reference trace. I reproduced that selection, but it follows the pre-existing coefficient policy and should not be presented as newly introduced failure detection. The first-input-missing-trace-HDU `KeyError` is also pre-existing and acknowledged in the PR.

**Validation performed.**

- `test_specex.py`: **20 tests passed**, plus **2 subtests**.
- `test_xytraceset.py` and `test_trace_shift.py`: **4 tests passed**.
- Thirty generated differential cases exercised missing bundles, failures, high chi-square values, and mixed wavelength ranges. The complete PSF coefficient arrays and output bundle chi-square headers were **exactly identical** between the base and PR implementations in all 30 cases.
- The two fallback problems above were reproduced against both implementations. Additional checks exercised widely separated above-threshold chi-square values and selection of an input containing an individually failed fiber.
- The current `TestMeanPSF` suite was run against the base implementation to identify which tests actually distinguish the change.
- `git diff --check origin/main...HEAD` passed.

The repository test commands were:

```bash
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=py .venv/bin/python -m pytest \
    -q -p no:cacheprovider py/desispec/test/test_specex.py

PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=py .venv/bin/python -m pytest \
    -q -p no:cacheprovider py/desispec/test/test_xytraceset.py \
    py/desispec/test/test_trace_shift.py
```

The generated differential checks were separate review checks, not additions to the repository test suite. They establish preservation of coefficient behavior for the exercised cases; they do not replace production-data or end-to-end extraction validation.

akremin:
GPT-6 Astra:

I don’t see any further blocking correctness issues in the updated code. I reviewed bafa214c against the current PR base and the previously reviewed version. Both earlier findings are resolved.
The revised implementation correctly:
- Excludes completely failed fits from the trace fallback.
- Drops inputs lacking trace HDUs per bundle, preserving selection for other bundles.
- Retains reference traces, with warnings, when no usable trace remains.
- Rejects the rchi2=100 example using the new bundle-specific cutoff.
Verification completed:
- 26 tests passed, plus 2 subtests.
- Nine additional checks passed, covering empty selections, entirely unfitted bundles, missing only one trace HDU, and missing HDUs combined with wavelength refitting. Input files remained unchanged.
- 30 differential cases produced coefficient arrays and chi-square headers exactly identical to the base implementation.
- Confirmed that eight current tests fail against the base, matching the revised PR writeup.
- git diff --check passed.
One minor documentation issue remains: the test docstring at [test_specex.py (line 729)](/Users/kremin/Projects/desispec/py/desispec/test/test_specex.py:729) still says that nothing passing the camera-wide cut means nothing is an outlier. The implementation now correctly handles that distinction; the docstring should describe the additional bundle-specific cut.
The PR writeup also still calls the Legendre cleanup a “latent bug” fix near the top, despite later correctly explaining that its test passes against the base. That wording is worth reconciling, but neither documentation point warrants blocking the code.
The previously acknowledged per-fiber limitation remains, and I did not independently reproduce the new production-data measurements. Within those limits, I would approve the updated implementation. No repository files were changed.

akremin:
Merging as Claude+ChatGPT signed off in addition to me. This will enable further testing over the weekend.

Issue #2802, opened 2026-08-22
biasnight and biaspdark jobs not in processing dashboard


akremin:
biasnight and biaspdark jobs are not currently in processing dashboard. Update the code to include them.

sbailey:
Related: #2677 which also mentions linkcal and ccdcalib as not covered by the dashboard.

akremin:
Resolved with PR #2829

Issue #2677, opened 2026-03-20
Improve dashboard visibility of calib failures


sbailey:
The dashboard tracking starts with psfs and fiberflats, but doesn't track earlier calibration steps like linkcal, ccdcalib, biaspdark, etc.  As a result, those jobs can fail but still have the dashboard show all blue (good) for those nights.

e.g. in the current state of the m3 prod
* linkcal 20260130 failed but it shows blue (good)
* linkcal 20201217 failed but shows gray (in progress, likely because downstream jobs are in "DependencyNeverSatisfied" rather than removed from the queue yet).

I don't want to add N>>1 additional columns, but perhaps we could have one generic column for any calib jobs prior to PSFs, or a catchall column for flagging that any jobs from the night are in a bad state.

akremin:
Resolved with PR #2829

PR #2829, opened 2026-09-18
Update dashboards to show one year at a time and add calibration rows

akremin:
This addresses https://github.com/desihub/desispec/issues/2802 and https://github.com/desihub/desispec/issues/2677, adding linkcal, biasnight, biaspdark, pdark, and ccdcalib jobs to the processing dashboard.

It also adds the feature that it now only displays one year at a time with a year selector at the top of the page. It uses iframes to render the years, so it reduces the browser memory footprint for large productions.


akremin:
This was reviewed by an independent Claude Opus 5 session and a GPT-6 Astra session, in addition to my oversight and review. The dashboard is not used for creating any data products, so I will merge it based on this "review."

PR #2831, opened 2026-09-20
ivar of rebinned spectra was incorrect

segasai:
The original code computed the ivar of (sum(x_i)) as sum(ivar_i), while it should be 1/sum(1/ivar_i)


julienguy:
checked the algorithmic change and ran a test. merging

PR #2832, opened 2026-09-20
fix uncorrect variance calculation when convolving

segasai:
Similar issue to #2831


julienguy:
Algorithmic change is good. I also tested the branch on one exposure just as a double cross-check. Merging.

PR #2835, opened 2026-09-20
scattered light calculates the median over data  combined with ivar i…

segasai:
scattered light calculates the median over data combined with ivar, as median_numba returns two arrays


julienguy:
this is another bug that should have resulted in a non-functional algorithm.
I have to see why it was kind of working despite this mistake.

julienguy:
This bug only affects the evaluation of a threshold to the mask gaps between fiber blocks affected by light from bright stars. It did not affect the background estimation on a test dark time exposure I looked at.
This has however the potential of affecting (i.e. improving) the background subtraction of some science exposures.

